PR #39, opened 2015-01-14
Add Origin to ALTSVC HTTP/2 frame.

bencebeky:
A possible implementation of http://lists.w3.org/Archives/Public/ietf-http-wg/2015JanMar/0062.html.


reschke:
Thanks for the proposal; will use a variant using 16 bits.

PR #47, opened 2015-02-18
Proposal for header purpose.

hruellan:


mnot:
Herve, could you please open an issue for this, so that we can assure that we track the underlying issue? It's OK to merge the proposal if the editor feels that it's helpful, but we need to make sure the WG gets consensus here.

Thanks,

hruellan:
Linked to issue #49.

PR #54, opened 2015-03-17
Falsification

martinthomson:
This adds a section regarding falsification of Tunnel-Protocol.  I didn't expand this to note that WebRTC won't suffer from this problem because the application won't be able to set the header field.  That seemed like too much detail.

### DIFF
--- a/draft-ietf-httpbis-tunnel-protocol.xml
+++ b/draft-ietf-httpbis-tunnel-protocol.xml
@@ -266,6 +266,14 @@ Tunnel-Protocol: h2, http%2F1.1
         from the case where either client or proxy does not support the
         Tunnel-Protocol header field.
       </t>
+      <t>
+        The value of the Tunnel-Protocol header field could be falsified by a
+        client.  The use of protocols like <xref target="RFC5246">TLS</xref>
+        make it very difficult or impossible to verify that the header field is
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-tunnel-protocol.xml
```diff
@@ -266,6 +266,14 @@ Tunnel-Protocol: h2, http%2F1.1
         from the case where either client or proxy does not support the
         Tunnel-Protocol header field.
       </t>
+      <t>
+        The value of the Tunnel-Protocol header field could be falsified by a
+        client.  The use of protocols like <xref target="RFC5246">TLS</xref>
+        make it very difficult or impossible to verify that the header field is
+        correct even after a connection is established.  A proxy therefore
```
ekr: "makes it very difficult"

Perhaps rewrite
"If the data being sent through the tunnel is encrypted (e.g., with <xref target="RFC5246">TLS</xref>), then the proxy will not be able to directly
inspect it to verify that the claimed protocol is the one which is actually
being used, although it may be able to perform traffic analysis. [0]"

[0] http://alfredo.pironti.eu/research/publications/full/identifying-website-users-tls-traffic-analysis-new-attacks-and-effective-counterme

PR #56, opened 2015-03-24
Precise signification of http/1.1 ALPN token.

hruellan:
The `http/1.1` ALPN token identifies HTTP/1.1 over TLS.


draft-ietf-httpbis-alt-svc.xml
```diff
@@ -240,6 +240,11 @@
    (<xref target="frame"/>).
 </t>
 <t>
+   Note that for the purpose of identifying an alternative service, the
+   "http/1.1" token defined by <xref target="RFC7301"/> identifies HTTP/1.1 over
+   TLS.
```
martinthomson: I think that you might also note: Neither RFC 7301, nor this document defines a token that identifies HTTP/1.1 over unsecured TCP.

PR #66, opened 2015-04-26
Avoid ambiguity with "may not" and a "MAY" requirement

aprescott:
RFC 2119 defines "MAY" as a keyword, which may cause confusion with "may not", even though the latter is not uppercased.

This hopefully better clarifies that the server in question may or may not be an origin server, as opposed to implying that origin servers are prevented from returning a 451 response status.

### DIFF
--- a/draft-ietf-httpbis-legally-restricted-status.xml
+++ b/draft-ietf-httpbis-legally-restricted-status.xml
@@ -103,9 +103,9 @@
       <t>This status code indicates that the server is denying access
       to the resource as a consequence of a legal demand.</t> 
       
-      <t>The server in question may not be an origin server.  This type of
-      legal demand typically most directly affects the operations of ISPs and
-      search engines.</t>
+      <t>The server in question is not necessarily an origin server.  This type
+      of legal demand typically most directly affects the operations of ISPs
+      and search engines.</t>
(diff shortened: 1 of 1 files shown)


aprescott:
I re-wrapped the content to 80 characters after making this change, although I'm happy to undo that if the diff noise is unwelcome.

I also would be glad to go with a different choice of wording if there's something more preferable.

reschke:
Sounds good to me.

aprescott:
Out of curiosity, why this was closed?

mnot:
because the RFC has been published.

reschke:
Actually, it has been approved but not published, and this is a change that could be done during AUTH48 (a minimal change would be "may" -> "might").

mnot:
@timbray, thoughts? I think this is editors' discretion (although historically we do try to avoid using RFC2119 keywords in this manner).

aprescott:
Happy to resubmit a PR with the change, unless @timbray or someone else would just make that small modification.

reschke:
@timbray I see the spec is now in AUTH48; are you going to make this change?

reschke:
I think this can be considered to be resolved due to the changes in f36fb4af.

PR #77, opened 2015-05-28
Going down for maintenance?

MikeBishop:
Spec commends the use of Alt-Svc when a server is "going down for maintenance."  But new clients will always be trying to contact the origin server, and existing clients may always fall back to it for various reasons.  Alt-Svc isn't a redundancy strategy for the origin, and I don't think we should present it as one here.  Servers will still need to cluster, anycast, etc. or shed load by changing DNS.

### DIFF
--- a/draft-ietf-httpbis-alt-svc.xml
+++ b/draft-ietf-httpbis-alt-svc.xml
@@ -109,9 +109,8 @@
 <t>For example:
   <list style="symbols">
     <t>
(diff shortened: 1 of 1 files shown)


martinthomson:
Yep.

reschke:
See https://github.com/httpwg/http-extensions/commit/0b847c05be2dca2280007c84f87ed6fb2ec1a5d1

PR #78, opened 2015-05-29
Alt-Svc Editorial Nits

MikeBishop:
Assorted:
  - Alt-Svc doesn't introduce 421, just points to RFC 7540
  - Missing "the" in overview
  - Requests != connections in HTTP/2; presuming you mean connections throughout
  - Non-normative "can" mismatch with normative SHOULD later in the document
  - Clients who wish to avoid tracking are also permitted to just drop Alt-Used; turned to normative MAY to match earlier text
  - Word order in scheme confusion text

### DIFF
--- a/draft-ietf-httpbis-alt-svc.xml
+++ b/draft-ietf-httpbis-alt-svc.xml
@@ -136,8 +136,8 @@
    service was used (<xref target="indicator"/>).
 </t>
 <t>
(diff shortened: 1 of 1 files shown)


reschke:
(see code comments; cherry-picked everything else in https://github.com/httpwg/http-extensions/commit/65ffed008a9eb9d06762217cfd6e03401fef640f)

Issue #16, opened 2014-08-18
Alt-Svc alternative cache invalidation


mnot:
In 3 The Alt-Svc HTTP Header Field, there's:

> When an Alt-Svc response header field is received from an origin, its value invalidates and replaces all cached alternative services for that origin.

However, in several other places, we now say that multiple alternatives can co-exist (with the client figuring out which to use). 

Is this still our intent -- i.e., that the header field has a special cache invalidation semantic -- or is it just left over from our previous approach?

mnot:
Discussed on-list. Cache invalidation is to be scoped to a specific discovery mechanism; e.g., the alternatives you discover via the response header will be invalidated when you see a new response header, while those that were discovered via the frame will be invalidated only when a new frame is received.

This means each mechanism needs to define its own exact invalidation semantics, and probably needs to be capable of carrying multiple alternatives.

reschke:
For the header field we already say: "When an Alt-Svc response header field is received from an origin, its value invalidates and replaces all cached alternative services for that origin."

Just duplicate that in the description of the frame?

mnot:
Almost. That existing text needs to be qualified to say that the response header invalidates previous services discovered via the response header, and then the section on the frame needs to say that it invalidates previous alt services discovered via a frame.

reschke:
That sounds a bit like the client might have to maintain two sets; is that really true?

mnot:
That's what we discussed on list.

reschke:
I looked at the email thread, and it seems the use case mentioned was 1.1 --> op-sec --> 2 and then alt-svc being used for load balancing. In this particular case we'd indeed have on set received by header field and one set received by frame. But is relying on exactly these two sets sufficient? In particular, tying it just to the header field/frame distinction seems funky...

martinthomson:
It's not the case that there are exactly two sets.   It's the case that one source cannot override information acquired from another source, but it must override information from the same source.

mnot:
Discussed in Prague; sense of room was to keep a single global scope for invalidation, to keep things simple. Need to adjust ABNF to allow an empty set. Add advice that server preference is expressed by lexical ordering.

reschke:
I'm having second thoughts abound the "empty list" solution. In the past I've seen broken support for empty header field values (software components not being able to distinguish between "empty" and "not present").

mnot:
[...]

PR #90, opened 2015-07-21
Allow empty list of alternative services.

bencebeky:
Allow empty list of alternative services both in header field value and in
frame.  See
https://github.com/httpwg/http-extensions/issues/16#issuecomment-123273055.


reschke:
Addressed by introducing the "clear" keyword (see https://github.com/httpwg/http-extensions/commit/68970e3555ba77cd56418c036e601af520b17711)

Issue #76, opened 2015-05-28
Alt-Svc and Cert Pinning


MikeBishop:
If http://www.example.com sends Alt-Svc pointing to evil.haxor.bg showing a cert from a hacked CA for the right origin, we don’t want to trust it.  Cert pinning would mitigate this for https:// origins, and we could go a step further for https:// origins without pinning and validate that the cert on the alternate resembles the cert we last saw on the origin (identical cert, same issuer, same issuer country, etc.).

However, for http:// origins, we're pretty vulnerable.  That scenario looks a lot like 9.2 again for http:// origins – if you’re MITM’d once, you can be MITM’d for as long as you trust the fishy cert.

martinthomson:
The alternative needs to present credentials that are valid.  That means that if the original requires pinning, then the alternative needs to present a certificate that is valid against the pin set.

Requiring that an alternative present an identical certificate would make this considerably less attractive.

mnot:
Mike, is this really an issue, or just a question? Do you have proposed text?

MikeBishop:
This is a concern that came up in our internal discussions.  Reasonable text could range anywhere from general guidance that implementations might use caution in which certificates they accept to specific requirements on the cert, depending on the consensus of the rest of the group.

More generally, I think this is something a cautious implementation might do anyway, and if it's at least mentioned in the spec, origin operators will be more prepared for it than if it's strictly an implementation "quirk."

mnot:
Discussed in Prague; need to add security considerations text explaining the potential attack and possible mitigations.

mnot:
@MikeBishop, perhaps you could suggest some text?

PR #98, opened 2015-10-06
Security for spoofed certs

MikeBishop:
Text giving recommendations for guarding against spoofed certs.


martinthomson:
The text on pinning is fine.  If the intent is expand on the authentication requirements for alternatives, the first sentence would suffice.

I'm concerned that the "additional requirements" you describe here might lead to interoperability problems.  If these aren't specified, we are left to guess what the validation requirements are for any given client.  I recognize that failure to authenticate an alternative doesn't necessarily translate into a fatal error - the alternative is just not used.  However, I'm wondering what the motivation for the change is.

MikeBishop:
This is from #76, which I should have tagged.  You argued against the spec requiring an identical cert, and I agree that synchronization and security of the cert could certainly be an issue.  However, the client might already have some data about the certs used by the origin and could reasonably expect some similarity between them.

As I said in the issue, I think cautious implementations will probably do *something* here, and I'd prefer the possibility to be mentioned in the spec so it's something operators will be aware of.

reschke:
@martinthomson are you ok with this proposal?

martinthomson:
Yes, it is what we discussed in the meeting.

PR #101, opened 2015-10-06
Reconciling MAY/can vs. SHOULD

MikeBishop:
Client MAY (2.2) vs. SHOULD (2.4) use alternatives they're aware of; clients "can" (2.4) vs. SHOULD (5) include the Alt-Used header. Reconciling these both to SHOULDs.

### DIFF
--- a/draft-ietf-httpbis-alt-svc.xml
+++ b/draft-ietf-httpbis-alt-svc.xml
@@ -293,7 +293,7 @@ uri-host      = &lt;uri-host, see <xref target="RFC7230" x:rel="#uri"/>&gt;
    parameter.
 </t>
 <t>
-   Clients &MAY; choose to use an alternative service instead of the origin at any
+   Clients &SHOULD; choose to use an alternative service instead of the origin at any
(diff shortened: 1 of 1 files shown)


martinthomson:
LGTM

mnot:
Discussed in Yokohama; seems reasonable.

reschke:
The "seems reasonable" applies to my comment, right?

reschke:
There are unfortunately 2 change request here, one of which I partly agree with, the other I disagree with.

The first one is caused by having normative language spread over 2.2 and 2.4; I tried to address this in https://github.com/httpwg/http-extensions/commit/acc3ae3c4290323069501d55ea8cdb5bacdbc6e8.

The second one is about adding a duplicate requirement; I continue to believe we shouldn't do that. We can mention requirements multiple times, but only one place should have the magic keyword (otherwise we risk ending up with exactly the kind of inconsistencies this pull request is about).

@MikeBishop are you ok with the change?

mnot:
I think this is resolved, but if you disagree, please open a new issue ASAP.

Issue #135, opened 2016-01-28
encryption-encoding: use two octets to indicate record padding length


beverloo:
Each record currently includes between 1 and 256 octets of padding, while the default record size is set to 4096 octets. This is in many cases not sufficient, for example in the pad-to-next-power-of-two case.

Extending the padding length to two octets will allow up to 65537 octets (65KB) of padding.

I raised this on the ietf-http-wg list here:
    https://lists.w3.org/Archives/Public/ietf-http-wg/2016JanMar/0171.html

Martin [raises a good point](https://lists.w3.org/Archives/Public/ietf-http-wg/2016JanMar/0175.html) about an upgrade path for existing users. I'll reply on the mailing list to not split the discussion.

martinthomson:
This was fixed in #136 and #137.

PR #136, opened 2016-01-28
Use two octets to indicate record padding length in httpbis-encryption-encoding

beverloo:
This is a proposal to address issue #135

/cc @martinthomson


beverloo:
I've updated the graph in section 2 to note the new range. PTAL.

PR #137, opened 2016-01-28
Rename aesgcm128 to aesgcm to provide an update path

beverloo:
In order to allow existing applications to update from one octet of
padding per record to two (PR #136, Issue #135), this PR renames
"aesgcm128" to "aesgcm" as the value for the Content-Encoding header,
as well as the parameter name in the Crypto-Key header.

This also affects the context used for deriving the nonce and the
content encryption key. Examples have been updated.

What I did not update is the JWE mapping (currently "A128GCM"). Do
we wish to change that to "AGCM", which may be less descriptive?

### DIFF
--- a/draft-ietf-httpbis-encryption-encoding.md
+++ b/draft-ietf-httpbis-encryption-encoding.md
@@ -134,9 +134,9 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 interpreted as described in [RFC2119].
 
 
-# The "aesgcm128" HTTP Content Encoding {#aesgcm128}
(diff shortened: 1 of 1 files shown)


beverloo:
@martinthomson - for your consideration

I updated the hashes, but would like you to verify those when you've got a moment. I wasn't able to reproduce the output of the explicit example using the current spec.

martinthomson:
Awesome.  I'm still awaiting the return of my implementation (I'm annoyed that I put it on a boat), but I will verify when able.  It's entirely possible that one (or more) of the examples are in error.

martinthomson:
I still need to double-check the examples, but I won't be able to do that for a while now.  Stupid travel.  Might as well make progress at least.

Thanks for putting this all together @beverloo.

PR #151, opened 2016-03-08
Adding `commit` member to JSON

martinthomson:


draft-ietf-httpbis-http2-encryption.md
```diff
@@ -203,81 +200,54 @@ form of protection against active attacks can be provided for clients on subsequ
...
-~~~
 
-When it appears in a HTTP response from a strongly authenticated alternative service, this header
-field indicates that the availability of the origin through TLS-protected alternative services is
-"sticky", and that the client MUST NOT fall back to cleartext protocols while this information is
-considered fresh.
+## Opportunistic Commitment
 
-For example:
+A alternative service can commit to providing a secured alternative by including a `commit` member
```
mnot: Shouldn't that be "An origin can commit..."?

draft-ietf-httpbis-http2-encryption.md
```diff
@@ -203,81 +200,54 @@ form of protection against active attacks can be provided for clients on subsequ
...
-~~~ example
-    HTTP/1.1 200 OK
-    Content-Type: text/html
-    Cache-Control: max-age=600
-    Age: 30
-    Date: Thu, 1 May 2014 16:20:09 GMT
-    HTTP-TLS: ma=3600
+{
+  "origins": ["http://example.com:80", "http://www.example.com/:81"],
+  "commit": "2016-03-12T12:27:00Z"
```
mnot: So, why a timestamp instead of a delta? Using a TS means the server will have to regenerate the file or serve it dynamically. Also brings clock sync issues.

draft-ietf-httpbis-http2-encryption.md
```diff
@@ -203,81 +200,54 @@ form of protection against active attacks can be provided for clients on subsequ
...
 
-When it appears in a HTTP response from a strongly authenticated alternative service, this header
-field indicates that the availability of the origin through TLS-protected alternative services is
-"sticky", and that the client MUST NOT fall back to cleartext protocols while this information is
-considered fresh.
+## Opportunistic Commitment
 
-For example:
+A alternative service can commit to providing a secured alternative by including a `commit` member
+in the http-opportunistic well-known resource.  This field includes an RFC3339 [RFC3339] timestamp.
```
mnot: "commit" makes me think "commit message." How about "available-until" or something similar?

draft-ietf-httpbis-http2-encryption.md
```diff
@@ -203,81 +200,54 @@ form of protection against active attacks can be provided for clients on subsequ
...
-
-Authentication for HTTP over TLS is described in Section 3.1 of {{RFC2818}}, noting the additional
-requirements in Section 2.1 of {{I-D.ietf-httpbis-alt-svc}}. The header field MUST be ignored if
-strong authentication fails; otherwise, an attacker could create a persistent denial of service by
-falsifying a commitment.
+The value of the `commit` member MUST be ignored unless the alternative service can be strongly
+authenticated.  Minimum authentication requirements for HTTP over TLS are described in Section 2.1
+of {{I-D.ietf-httpbis-alt-svc}} and Section 3.1 of {{RFC2818}}.  As noted in
+{{I-D.ietf-httpbis-alt-svc}}, clients can impose other checks in addition to this minimum set.  For
+instance, a client might choose to apply key pinning {{RFC7469}}.
```
mnot: So, this is kind of weird, because the .well-known is only for the case when the alternative *isn't* strongly authenticated. Restructuring the spec might help.

draft-ietf-httpbis-http2-encryption.md
martinthomson: Yes, sync issues.  But certificates are similarly challenged, and I want to be able to lift one of these out of a cache.  That's a reasonable thing to do right?

draft-ietf-httpbis-http2-encryption.md
martinthomson: This is the essential weirdness of this mechanism.  You don't *have* to authenticate, but you can.

draft-ietf-httpbis-http2-encryption.md
martinthomson: Done.

draft-ietf-httpbis-http2-encryption.md
martinthomson: 🚲🏚

draft-ietf-httpbis-http2-encryption.md
mnot: I was assuming that a typical case would be just wanting to try it out by assigning a relatively short commitment period.

martinthomson:
And lo, relativism became ascendant.

PR #158, opened 2016-03-14
Updated encryption examples

martinthomson:
Thanks to @beverloo for identifying the bug in my code that caused this to take so long.

### DIFF
--- a/draft-ietf-httpbis-encryption-encoding.md
+++ b/draft-ietf-httpbis-encryption-encoding.md
@@ -558,20 +558,21 @@ Encryption: keyid="mailto:me@example.com";
 
 ## Encryption with Explicit Key {#explicit}
 
+This example shows the UTF-8 encoded string "I am the walrus" encrypted using an
+directly provided value for the input keying material.  The content body
+contains a single record only and is shown here encoded in URL-safe base64 for
+presentation reasons only.
+
 ~~~ example
 HTTP/1.1 200 OK
-Content-Length: 32
+Content-Length: 33
 Content-Encoding: aesgcm
 Encryption: keyid="a1"; salt="vr0o6Uq3w_KDWeatc27mUg"
 Crypto-Key: keyid="a1"; aesgcm="csPJEXBYA5U-Tal9EdJi-w"
 
-OwW_-ChdUuV44mRDjlg1lF4pviF1AsfHv_9wbnSHLoxj
+VDeU0XxaJkOJDAxPl7h9JD5V8N43RorP7PfpPdZZQuwF
 ~~~
 
-This example shows the string "I am the walrus" encrypted using an directly
-provided value for the input keying material.  The content body contains a
-single record only and is shown here encoded in URL-safe base64 for presentation
-reasons only.
 
 
 ## Diffie-Hellman Encryption
--- a/draft-ietf-httpbis-encryption-encoding.md
+++ b/draft-ietf-httpbis-encryption-encoding.md
@@ -516,12 +516,17 @@ material that is identified by a URI.
 Note that the media type has been changed to "application/octet-stream" to avoid
 exposing information about the content.
 
+
 ## Encryption and Compression
+
+In this example, a response is first compressed, then encrypted.  Note that this
+particular encoding might compromise confidentiality if the contents of the
+response could be influenced by an attacker.
 
 ~~~ example
 HTTP/1.1 200 OK
 Content-Type: text/html
-Content-Encoding: aesgcm, gzip
+Content-Encoding: gzip, aesgcm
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-encryption-encoding.md
```diff
@@ -598,6 +629,27 @@ added for presentation purposes only.
...
+~~~
+HTTP/1.1 200 OK
+Content-Length: 33
+Content-Encoding: aesgcm
+Encryption: keyid="dhkey"; salt="lngarbyKfMoi9Z75xYXmkg"
+Crypto-Key: keyid="dhkey";
+                dh="BNoRDbb84JGm8g5Z5CFxurSqsXWJ11ItfXEWYVLE85Y7
+                    CYkDjXsIEc4aqxYaQ1G8BqkXCJ6DPpDrWtdWj_mugHU"
+
+QPsWCEH-8WqiXl7Ez23A6u4rwH2hFBifCybQQz-SxsjF
```
beverloo: I am unable to reproduce this in either of my implementations, using the following information:

**Sender public**: BNoRDbb84JGm8g5Z5CFxurSqsXWJ11ItfXEWYVLE85Y7CYkDjXsIEc4aqxYaQ1G8BqkXCJ6DPpDrWtdWj_mugHU
**Sender private**: nCScek-QpEjmOOlT-rQ38nZzvdPlqa00Zy0i6m2OJvY
**Receiver public**: BCEkBjzL8Z3C-oi2Q7oE5t2Np-p7osjGLg93qUP0wvqRT21EEWyf0cQDQcakQMqz4hQKYOQ3il2nNZct4HgAUQU
**Salt**: lngarbyKfMoi9Z75xYXmkg
**Auth secret**: R29vIGdvbyBnJyBqb29iIQ
**Padding**: 0 bytes
**Plaintext**: I am the walrus

I get the following:

**Shared secret**: RNjC-NVW4BGJbxWPW7G2mowsLeDa53LYKYm4--NOQ6Y
**PRK**: EhpZec37Ptm4IRD5-jtZ0q6r1iK5vYmY1tZwtN8fbZY
**CEK**: AN2-xhvFWeYh5z0fcDu0Ww
**Nonce**: JY1Okw5rw1Drkg9J
**Ciphertext**: 6nqAQUME8hNqw5J3kl8cpVVJylXKYqZOeseZG8UueKpA

Decryption fails as well, so I can't see whether you used a different `plaintext`.

draft-ietf-httpbis-encryption-encoding.md
```diff
@@ -359,13 +359,13 @@ header field.
 
 aesgcm:
 
-: The "aesgcm" parameter contains the URL-safe base64 [RFC4648] octets of the
-input keying material.
+: The "aesgcm" parameter contains the URL and filename safe base64 [RFC4648]
+  octets of the input keying material with padding removed.
```
beverloo: Do you want to be strict in regards to omitting padding? Right now our implementation is lenient and just ignores it.

draft-ietf-httpbis-encryption-encoding.md
martinthomson: Ahh, I get the same answer that you do.  Strange.  I must have I copied the wrong value in somehow.  Thanks for checking!  I'll change the ciphertext.

draft-ietf-httpbis-encryption-encoding.md
martinthomson: Well, it turns out that '=' is an invalid character unless you quote the string.  I think that we should be strict here.  I believe that Firefox is, but I could be wrong about that.

beverloo: I'll change Chrome to play ball then.

draft-ietf-httpbis-encryption-encoding.md
```diff
@@ -598,6 +629,28 @@ added for presentation purposes only.
...
+~~~
+HTTP/1.1 200 OK
+Content-Length: 33
+Content-Encoding: aesgcm
+Encryption: keyid="dhkey"; salt="lngarbyKfMoi9Z75xYXmkg"
+Crypto-Key: keyid="dhkey";
+                dh="BNoRDbb84JGm8g5Z5CFxurSqsXWJ11ItfXEWYVLE85Y7
+                    CYkDjXsIEc4aqxYaQ1G8BqkXCJ6DPpDrWtdWj_mugHU"
+
+6nqAQUME8hNqw5J3kl8cpVVJylXKYqZOeseZG8UueKpA
```
beverloo: Confirmed! Thanks.

martinthomson:
Merged manually

PR #163, opened 2016-03-30
WIP: addressing CH feedback

igrigorik:
https://lists.w3.org/Archives/Public/ietf-http-wg/2016JanMar/0519.html

Outstanding issues:
- [x] reference User-Agent header field definition
- [x] reference cookie spec
- [x] make "proactive content negotiation [RFC7231]" reference more specific
- [ ] most of the "this document uses..." references do not seem to be used; cleanup.
- [x]  drop "common syntax" leftover from previous drafts.
- [x] make "Vary" reference more specific
- [x] reference HTML spec
- [x] include a reference to "CSS px"
- [x] include a reference to "layout viewport"
- [ ] clarify whether this is for user agents or all clients

### DIFF
--- a/draft-ietf-httpbis-client-hints.md
+++ b/draft-ietf-httpbis-client-hints.md
@@ -91,7 +91,7 @@ code and issues list for this draft can be found at <https://github.com/httpwg/h
 
 There are thousands of different devices accessing the web, each with different device capabilities and preference information. These device capabilities include hardware and software characteristics, as well as dynamic user and client preferences.
 
-One way to infer some of these capabilities is through User-Agent (UA; section 5.5.3 of {{RFC7231}}) detection against an established database of client signatures. However, this technique requires acquiring such a database, integrating it into the serving path, and keeping it up to date. However, even once this infrastructure is deployed, UA sniffing has numerous limitations:
+One way to infer some of these capabilities is through User-Agent (UA; Section 5.5.3 of {{RFC7231}}) detection against an established database of client signatures. However, this technique requires acquiring such a database, integrating it into the serving path, and keeping it up to date. However, even once this infrastructure is deployed, UA sniffing has numerous limitations:
 
   - UA detection cannot reliably identify all static variables
   - UA detection cannot infer any dynamic client preferences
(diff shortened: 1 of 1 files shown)


igrigorik:
@reschke @mnot most of the remaining issues are references. Any tips or examples for how to reference W3C docs? Similarly, for Vary, how do I deeplink into the spec?

reschke:
W3C: try something like:

 informative:
+  W3C.REC-html5-20141028:

For deep linking, I believe kramdown can't help. Just say "Section xyz of [RFC....]".

draft-ietf-httpbis-client-hints.md
```diff
@@ -50,6 +50,7 @@ normative:
       organization: Google
   W3C.REC-html5-20141028:
   W3C.CR-css-values-3-20150611:
+#  W3C.REC-CSS2-20110607:
```
igrigorik: @reschke @mnot I get a crash when I try to build with this. A bug in the fetched ref?

```
Error at xsl:value-of on line 2278 of rfc2629.xslt:
  SERE0014: Illegal HTML character: decimal 135
  at xsl:apply-templates (file:/Users/igrigorik/git/specs/http-extensions/lib/rfc2629.xslt#2519)
     processing /rfc/back[1]/references[1]/reference[10]
  at xsl:apply-templates (file:/Users/igrigorik/git/specs/http-extensions/lib/rfc2629.xslt#1958)
     processing /rfc/back[1]/references[1]
  at xsl:apply-templates (file:/Users/igrigorik/git/specs/http-extensions/lib/rfcbootstrap.xslt#150)
     processing /rfc/middle[1]
  at xsl:call-template name="body" (file:/Users/igrigorik/git/specs/http-extensions/lib/rfc2629.xslt#2667)
  at xsl:apply-templates (file:/Users/igrigorik/git/specs/http-extensions/lib/rfc2629.xslt#3700)
     processing /rfc
Illegal HTML character: decimal 135
Makefile:81: recipe for target 'draft-ietf-httpbis-client-hints.html' failed
```

igrigorik:
@reschke thanks, please review. I think I got most of the things fixed.

draft-ietf-httpbis-client-hints.md
reschke: Yes, it seems the reference is borked. I believe the only way to fix this right now is to inline the full text...

igrigorik:
@mnot @reschke squashed. Please review and feel free to merge if it looks good.

reschke:
Looked at it; I think it improves many things so we can merge it (should I?); but there maybe a few more things to do (but we can do that as separate changes).

igrigorik:
+1 to merge and iterate, as it unblocks other work on this end.

PR #175, opened 2016-05-10
Fix formatting

martinthomson:
Putting the meat of the document in Section 1.2 was a bit odd.  Also, the fields descriptions were glommed together.


mnot:
Weird; didn't happen on local copies.

PR #193, opened 2016-06-03
A commitment, once made, requires both reasonable assurances and stro…

### LINKED ISSUE #191: Strong Authentication and commit
Erik raises in post-WGLC:

The text is still a little unclear on whether commit then requires subsequent connections to use "strong authentication" or just "authentication".  In particular, Section 3 seems to add some confusion about whether "reasonable assurances" and "server authentication" are the same thing or not but then 5.2 sometimes uses "authenticated without "strongly authenticated".  For example, this paragraph is not particularly crisp on what the requirements are (strong authentication or something else):

    A commitment is not bound to a particular alternative service.
    Clients are able to use alternative services that they become aware
    of.  However, once a valid and authenticated commitment has been
    received, clients SHOULD NOT use an unauthenticated alternative
    service.  Where there is an active commitment, clients SHOULD ignore
    advertisements for unsecured alternative services.  A client MAY send
    requests to an unauthenticated origin in an attempt to discover
    potential alternative services, but these requests SHOULD be entirely
    generic and avoid including credentials.

martinthomson:
…ng authentication

Closes #191 if Erik is cool with it.

### DIFF
--- a/draft-ietf-httpbis-http2-encryption.md
+++ b/draft-ietf-httpbis-http2-encryption.md
@@ -254,10 +254,12 @@ services for the duration of the commitment interval, except to discover new pot
 
 A commitment is not bound to a particular alternative service. Clients are able to use alternative
 services that they become aware of. However, once a valid and authenticated commitment has been
-received, clients SHOULD NOT use an unauthenticated alternative service. Where there is an active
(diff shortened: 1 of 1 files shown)


mnot:
@enygren?

mnot:
ETIMEDOUT

@enygren, if you object please reopen.

PR #205, opened 2016-06-30
Provide rules for handling duplicate keys

martinthomson:
Raised on list by Kari Hurtta.  Option 6.

### DIFF
--- a/draft-ietf-httpbis-http2-encryption.md
+++ b/draft-ietf-httpbis-http2-encryption.md
@@ -171,6 +171,12 @@ Connection: close
 }
 ~~~
 
+The JSON format for the http-opportunistic resource makes it possible to include multiple values
+for the same origin.  Variations in case allow for duplicate keys that a JSON parser will treat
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2-encryption.md
```diff
@@ -171,6 +171,12 @@ Connection: close
 }
 ~~~
 
+The JSON format for the http-opportunistic resource makes it possible to include multiple values
+for the same origin.  Variations in case allow for duplicate keys that a JSON parser will treat
+as different.  Duplicated keys might not be detected by a client (see Section 4 of {{RFC7159}}).
+A client SHOULD NOT use origin objects if multiple keys match, though it MAY use any origin
+object with a matching key.
+
```
reschke: The SHOULD NOT might not be implementable if the parser doesn't report the fact...

martinthomson: That's why it's a SHOULD NOT and not MUST NOT.

martinthomson:
This has been OBE.

PR #207, opened 2016-06-30
Making tls-commit and tls-ports mutually exclusive

martinthomson:
This probably isn't the end of this issue though.  We require that an opportunistic upgrade uses the same hostname.  The text I added here lifts the requirement regarding port numbers, but should it also say that the hostname needs to remain the same?

I've left this open here, but I think that we probably should leave the hostname requirement in place.  That way, we still retain some of the aspects of http:// authority.

### DIFF
--- a/draft-ietf-httpbis-http2-encryption.md
+++ b/draft-ietf-httpbis-http2-encryption.md
@@ -232,7 +232,6 @@ origin makes such a commitment for the duration of the origin object lifetime.
 ~~~ example
 {
   "http://www.example.com": {
-    "tls-ports": [443,8080],
     "tls-commit": true,
     "lifetime": 3600
   }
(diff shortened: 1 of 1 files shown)


MikeBishop:
I disagree.  As soon as you're fully authenticating the TLS certificate, you're doing vanilla Alt-Svc and this draft could be totally out of the picture.  The only thing you're gaining from the http-opportunistic resource is a commitment that the TLS Alt-Svc will remain available and a suggestion that the client refuse to fall back to the plain-text origin.  I don't see that same-host plays into that.

Alternately phrased -- what about the alternative being on a different host implies the origin should *not* be able to make a commitment that it will remain available?

martinthomson:
Sorry about being so slow, but @MikeBishop, if I understand what you are implying, you are suggesting that we should remove the host restriction if the certificate fits.  I was thinking about a client using the process we outline in the draft to arrive at such a host, which isn't really possible.  

However, I guess we have to accept that a client might learn of this resource from an HTTPS connection to some other host (maybe because of connection coalescing).  In which case, then I find that I agree.  Some more careful words needed, I guess.

martinthomson:
I think that this has been OBE.

mnot:
Ah, sorry - forgot to look at the pulls.

PR #210, opened 2016-07-05
feat: Add an example of a composite Crypto-Key header

jrconlin:
Crypto-Key is a composite data field. Curretly both draft ECE and draft
VAPID use Crypto-Key to store information. This has lead to some
confusion regarding how data should be stored and accessed.

### DIFF
--- a/draft-ietf-httpbis-encryption-encoding.md
+++ b/draft-ietf-httpbis-encryption-encoding.md
@@ -366,13 +366,16 @@ avoided).
 # Crypto-Key Header Field {#crypto-key}
 
 A Crypto-Key header field can be used to describe the input keying material
-used in the Encryption header field.
+used in the Encryption header field. Crypto-Key is a composite field,
+where multiple protocols may store information. Each parameter set
+shall be separated by a comma and should be identified by a "keyid"
+parameter.
 
 The Crypto-Key header field uses the extended ABNF syntax defined in Section 1.2
 of [RFC7230] and the `parameter` and `OWS` rules from [RFC7231].
 
 ~~~ abnf7230
-  Crypto-Key = #crypto_key_params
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-encryption-encoding.md
```diff
@@ -366,13 +366,16 @@ avoided).
...
+where multiple protocols may store information. Each parameter set
+shall be separated by a comma and should be identified by a "keyid"
+parameter.
 
 The Crypto-Key header field uses the extended ABNF syntax defined in Section 1.2
 of [RFC7230] and the `parameter` and `OWS` rules from [RFC7231].
 
 ~~~ abnf7230
-  Crypto-Key = #crypto_key_params
+  Crypto-Key = crypto_key_params *(COMMA crypto_key_params)
```
martinthomson: This change isn't correct, or even needed.  The `#` has the same effect as what you are doing here, but it also allows for whitespace in the usual ways.

draft-ietf-httpbis-encryption-encoding.md
```diff
@@ -410,6 +413,18 @@ encryption key is the necessary size.
...
 
+An example of a composite Crypto-Key field is
+
+~~~ example
+   CryptoKey: keyid=p256dh;dh=BN8VZLdyi…,p256ecdsa=BFoWLF…
+~~~
+
+The proceeding example contained two parameter sets. The first
+contains "keyid" and "dh" used by this protocol (the values have been
+truncated with a "…" for readablity). The second contains "p256ecdsa"
```
martinthomson: This ellipsis won't render in our wonderful ASCII format.

martinthomson:
Redundant with the removal of said header.

PR #218, opened 2016-08-04
Remove DH

martinthomson:
This is only really needed by webpush and that spec can take the necessary pieces.  This is a LOT simpler as a result of this change.

[Preview the changes](http://httpwg.org/http-extensions/simple_base/draft-ietf-httpbis-encryption-encoding.html).


martinthomson:
Manually rebased and merged.

PR #221, opened 2016-08-09
Explicitly note case-sensitivity of prefix

ericlaw1979:
As cookie attributes are case-insensitive, I think it makes sense to explicitly call out case-sensitivity of prefixes.


mikewest:
I pulled this into https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-01#section-5.3 (steps 13 and 14). Thanks!

PR #239, opened 2016-10-07
update reference to CR-css-values-3

reschke:


igrigorik:
If memory serves, we froze older version due to build errors.. and it appears that this is still the case based on Travis results.

igrigorik:
@reschke I'm inclined to close this as wontfix.. since this breaks the build and I'm not sure how to work around it.

PR #250, opened 2016-10-19
clarify Accept-CH

igrigorik:
See
https://lists.w3.org/Archives/Public/ietf-http-wg/2016OctDec/0146.html

@reschke ptal.


reschke:
Sounds good to me.

igrigorik:
Thanks!

PR #252, opened 2016-10-29
Remove Encryption header field

martinthomson:
This moves the values that were in the Encryption header field to
the start of the payload.  This provides a meagre efficiency gain
as well as removing the need to have two correlated header fields.

This should go most of the way to address WGLC comments.

### DIFF
--- a/draft-ietf-httpbis-encryption-encoding.md
+++ b/draft-ietf-httpbis-encryption-encoding.md
@@ -35,12 +35,16 @@ informative:
     author:
       - ins: D. Eastlake
       - ins: J. Reagle
+      - ins: F. Hirsch
+      - ins: T. Roessler
       - ins: T. Imamura
       - ins: B. Dillaway
       - ins: E. Simon
-    date: 2002-12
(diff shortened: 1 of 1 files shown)


reschke:
Looks good to me.

draft-ietf-httpbis-encryption-encoding.md
```diff
@@ -324,14 +290,20 @@ Thus, the final nonce for each record is a 12 octet value:
...
+permits truncation of the tail of the sequence (see {{aes128gcm}} for how this
+is avoided).
 
 
 # Crypto-Key Header Field {#crypto-key}
 
-A Crypto-Key header field can be used to describe the input keying material
-used in the Encryption header field.
+A Crypto-Key header field can be used to describe the input keying material used
+in the Encryption header field.
```
reschke: "header field" -> "header block"?

martinthomson:
@reschke happy to correct any parsing weirdness, but need to know precisely what it is first.

PR #257, opened 2016-11-01
Try to explain and justify the use of the absolute form in HTTP/1.1

martinthomson:
Based on the discussion on-list, the requirement to use the absolute form is a little hard.  This tries to justify it on the basis that the secure server is acting as a proxy.  Maybe that's dishonest, but I think that it's accurate.  I would rather finagle this than have an experimental document update a standards-track one.

### DIFF
--- a/draft-ietf-httpbis-http2-encryption.md
+++ b/draft-ietf-httpbis-http2-encryption.md
@@ -102,8 +102,14 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 An origin server that supports the resolution of `http` URIs can indicate support for this
 specification by providing an alternative service advertisement {{RFC7838}} for a protocol
-identifier that uses TLS, such as `h2` {{RFC7540}}, or `http/1.1` {{?RFC7301}}.  Note that HTTP/1.1
-requests MUST use the absolute form (see Section 5.3.2 of {{RFC7230}}).
+identifier that uses TLS, such as `h2` {{RFC7540}}, or `http/1.1` {{?RFC7301}}.
+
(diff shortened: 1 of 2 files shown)


draft-ietf-httpbis-http2-encryption.md
```diff
@@ -102,8 +102,14 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
-identifier that uses TLS, such as `h2` {{RFC7540}}, or `http/1.1` {{?RFC7301}}.  Note that HTTP/1.1
-requests MUST use the absolute form (see Section 5.3.2 of {{RFC7230}}).
+identifier that uses TLS, such as `h2` {{RFC7540}}, or `http/1.1` {{?RFC7301}}.
+
+Requests for `http` resources using TLS MUST include the scheme in the request.  HTTP/2 requires
+this and provides the `:scheme` pseudo-header field for this purpose.  HTTP/1.1 provides no such
+field, and even prohibits the use of the absolute form for requests that are sent directly to an
+origin server (see Section 5.3.1 of {{RFC7230}}).  This document overrides that requirement for
+requests with the `http` scheme that are sent over TLS; these requests MUST be made using the
+absolute form as though the server were a proxy.
```
mnot: I don't think this is going to pass the sniff test unless we update 7230. 

How will this transit a proxy -- aren't you requiring reverse proxies et al to change their behaviour? 

It's also not at all clear that this information will be made available to applications.

mcmanus: I really do think its fine to abandon h1 for this. The core difference between h1 and h2 is at the transport level and this is a transport level feature.

PR #258, opened 2016-11-01
remove bogus "!"

reschke:
it leads to kramdown warnings ("overwriting RFC....").


reschke:
might not be a warning after all...

martinthomson:
@reschke, there are two ways to add references in the kramdown variant of markdown.  Adding "x" to the appropriate references block and then using `{{x}}`, OR using `{{?x}}` (informative) or `{{!x}}` (normative).

The warning comes about because the references block already contains "x", or because the reference is being used inconsistently (mixing styles is bad).

PR #259, opened 2016-11-02
Fix examples

martinthomson:
Somehow I managed to mess up the examples royally.  They were kinda valid, but the rs and id fields didn't match the rest of the example.


reschke:
Works for me (see https://gist.github.com/reschke/46659c912b426dffeac41d9a21421c95)

martinthomson:
Superceded by other changes.

PR #262, opened 2016-11-03
Exclude HTTP/1.1

martinthomson:
I tried to make this work, but @mcmanus is right: this is entirely a transport decision and given that our best option was to violate a MUST in RFC 7230, that's too hard.  This just says that HTTP/1.1 is no good because we can't include the scheme (because RFC 7230 says we can't).  I almost made this h2-specific, but then remembered that there might be another protocol that has favourable properties.


mnot:
It's official - MT says "HTTP/1.1 is no good."

PR #270, opened 2016-11-21
Don't allow clients to send http and https requests coalesced.

### LINKED ISSUE #269: mixed schemes
Can we coalesce http and https?  The implication from everything we've said is "yes", but Erik remains reluctant.

martinthomson:
This is kinda silly, but it seems like adding support for coalescing is
something we could fix later.

Closes #269 with the nuclear option.


mcmanus:
the pull request is fine by me. I do wonder if we shouldn't make the json a tad more complicated to allow for a reasonable mixed definition in the future. But an unreasonable one is obviously possible :)

PR #272, opened 2016-11-21
Advice on padding strategy

### LINKED ISSUE #260: padding is hard
We should provide some advice, along the lines of:

1. padding is hard
2. padding doesn't stop side channels from processing data
3. should pad the entire message
4. might want to distribute padding evenly

martinthomson:
Closes #260

### DIFF
--- a/draft-ietf-httpbis-encryption-encoding.md
+++ b/draft-ietf-httpbis-encryption-encoding.md
@@ -535,6 +535,14 @@ provides.  Alternatively, splitting up content into segments and storing the
 separately might reduce exposure. HTTP/2 {{?RFC7540}} combined with TLS
 {{?RFC5246}} might be used to hide the size of individual messages.
 
+Developing a padding strategy is difficult.  A good padding strategy can depend
+on context.  Common strategies include padding to a small set of fixed lengths,
+padding to multiples of a values, or padding to powers of 2.  Even a good
+strategy can still cause size information to leak if processing activity of a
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-encryption-encoding.md
```diff
@@ -535,6 +535,14 @@ provides.  Alternatively, splitting up content into segments and storing the
 separately might reduce exposure. HTTP/2 {{?RFC7540}} combined with TLS
 {{?RFC5246}} might be used to hide the size of individual messages.
 
+Developing a padding strategy is difficult.  A good padding strategy can depend
+on context.  Common strategies include padding to a small set of fixed lengths,
+padding to multiples of a values, or padding to powers of 2.  Even a good
+strategy can still cause size information to leak if processing activity of a
+recipient can be observed.  This is especially good if the trailing records of
+a message contain only padding.  To minimize this information, non-padding data
```
ekr: I am guessing "good " is not the word you want here. Perhaps "true"?

draft-ietf-httpbis-encryption-encoding.md
```diff
@@ -535,6 +535,14 @@ provides.  Alternatively, splitting up content into segments and storing the
...
+can be evenly distributed between records.
```
ekr: Perhaps something slightly stronger like "should" or "MAY WISH TO"

martinthomson:
Merged manually.

PR #277, opened 2016-12-09
make title consistent with WG terminology

reschke:


bsdphk:
I have struggled with the naming from the start, and am open to any suggestions on that, with the single caveat that making the name too long is probably not a good idea without at good acronym.

mnot:
OBE

PR #280, opened 2017-01-03
Clarify paragraph on requirements for alt-svc

martinthomson:
This was getting unwieldy.  Change it so that there are two musts and two supporting statements.


martinthomson:
Merged manually.

PR #283, opened 2017-01-30
Move to using padding that is more like TLS

martinthomson:
Following the discussion on the list.


manger:
Need to change the language about the size of the last record: from MUST be shorter, to CAN be shorter (in text and diagram).

PR #285, opened 2017-02-01
DNS relaxation, smaller initial Origin Set

mnot:
- Relax DNS requirements when ORIGIN is received
- Change the initial origin set to just SNI

### DIFF
--- a/draft-ietf-httpbis-origin-frame.md
+++ b/draft-ietf-httpbis-origin-frame.md
@@ -186,6 +186,11 @@ Note that this does not prevent clients from performing other certificate checks
 allowed, either at connection time or when processing ORIGIN. See {{!RFC7540}} Section 9.1.1 for
 more information.
 
+Because ORIGIN can change the set of origins a connection is used for over time, it is possible
+that a client might have more than one viable connection to an origin open at any time. When this
+occurs, clients SHOULD not emit new requests on any connection whose Origin Set is a subset of
+another connection's Origin Set, and SHOULD close it once all outstanding requests are satisified.
+
 
 # IANA Considerations
 
--- a/draft-ietf-httpbis-origin-frame.md
+++ b/draft-ietf-httpbis-origin-frame.md
@@ -163,20 +163,20 @@ The Origin Set is also affected by the 421 (Misdirected Request) response status
 MUST create the ASCII serialisation of the corresponding request's origin (as per {{!RFC6454}},
 Section 6.2) and remove it from the connection's Origin Set, if present.
 
-Note that the Origin Set does not affect existing streams on a connection in any way.
-
 
 ## Authority, Push and Coalescing with ORIGIN {#authority}
 
 {{!RFC7540}}, Section 10.1 uses both DNS and the presented TLS certificate to establish the origin
 server(s) that a connection is authoritative for, just as HTTP/1.1 does in {{?RFC7230}}.
 Furthermore, {{!RFC7540}} Section 9.1.1 explicitly allows a connection to be used for more than one
-origin server, if it is authoritative. This affects what requests can be sent on the connection, both in HEADERS frame by the client and as PUSH_PROMISE frames from the server.
+origin server, if it is authoritative. This affects what requests can be sent on the connection,
+both in HEADERS frame by the client and as PUSH_PROMISE frames from the server.
 
-Once an Origin Set has been initialised for a connection, clients that implement this
-specification change these behaviors in the following ways:
+Once an Origin Set has been initialised for a connection, clients that implement this specification
+change these behaviors in the following ways:
 
-* Clients MUST NOT consult DNS to establish authority for origins in the Origin Set. The TLS certificate MUST be used to do so, as described in {{!RFC7540}} Section 9.1.1.
+* Clients MUST NOT consult DNS to establish the connection's authority for new requests. The TLS
+  certificate MUST be used to do so, as described in {{!RFC7540}} Section 9.1.1.
 
 * Clients sending a new request SHOULD use an existing connection if the request's origin is in that connection's Origin Set, unless there are operational reasons for creating a new connection.
 
--- a/draft-ietf-httpbis-origin-frame.md
+++ b/draft-ietf-httpbis-origin-frame.md
@@ -125,13 +125,13 @@ for the added origins (in the sense of {{!RFC7540}}, Section 10.1) on its own; s
 
 ## Establishing Authority and Coalescing with ORIGIN {#authority}
 
-{{RFC7540}}, Section 10.1 uses both DNS and the presented TLS certificate to establish the
-authority of an origin server, just as HTTP/1.1 does in {{RFC7230}}.
+{{!RFC7540}}, Section 10.1 uses both DNS and the presented TLS certificate to establish the
+authority of an origin server, just as HTTP/1.1 does in {{?RFC7230}}.
 
 Upon receiving an ORIGIN frame on a connection, clients that implement this specification are
 released from the requirement to establish authority for a given origin using DNS, for that
 connection. However, they MUST still establish authority using the certificate, as described in
-{{RFC7540}} Section 9.1.1.
+{{!RFC7540}} Section 9.1.1.
 
 Once such a frame is received, an implementing client MUST NOT use that connection for a given
 origin unless it appears in the connection's Origin Set. Implementing clients SHOULD use a
--- a/draft-ietf-httpbis-origin-frame.md
+++ b/draft-ietf-httpbis-origin-frame.md
@@ -186,7 +186,7 @@ order to coalesce connections to the target onto their existing connection.
 --- back
 
 
-# Non-Normative Processing Algoritm {#algo}
+# Non-Normative Processing Algorithm {#algo}
 
 The following algorithm illustrates how a client could handle received ORIGIN frames:
 
--- a/draft-ietf-httpbis-origin-frame.md
+++ b/draft-ietf-httpbis-origin-frame.md
@@ -242,3 +242,7 @@ used for with subsequent ORIGIN frames later (e.g., when the connection is idle)
 
 Senders should note that, as per {{!RFC6454}} Section 4, the values in an ORIGIN header need to be
 case-normalised before serialisation.
+
+Finally, servers that allow alternative services {{?RFC7838}} will need to explicitly advertise
+those origins when sending ORIGIN, because the default contents of the Origin Set (as per {{set}})
+do not contain any Alternative Services, even if they have been used previously on the connection.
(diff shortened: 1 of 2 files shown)


draft-ietf-httpbis-origin-frame.md
```diff
@@ -90,7 +95,7 @@ Origin:
 The ORIGIN frame defines the following flags:
 
 CLEAR (0x1):
-: Indicates that the Origin Set MUST be reset to an empty set before processing the contents of the frame it occurs upon.
+: Indicates that the Origin Set MUST be reset to an empty before processing the contents of the frame it occurs upon.
```
cabo: an empty what?

mnot:
Fixed in ac5a5e4, thanks

draft-ietf-httpbis-origin-frame.md
mnot: Fixed in ac5a5e4, thanks

draft-ietf-httpbis-origin-frame.md
```diff
@@ -101,49 +106,65 @@ REMOVE (0x2):
 The set of origins (as per {{!RFC6454}}) that a given connection might be used for is known in this
 specification as the Origin Set.
 
-When a connection is first established, its Origin Set is defined to be those origins that the client would normally consider the connection authoritative for; see {{!RFC7540}}, Section 10.1.
+When a connection is first established, its Origin Set is defined to be the origin composed from:
+  - Scheme: "https"
+  - Host: the value sent in Server Name Indication {{!RFC6066}}
+  - Port: the local port of the connection on the server
+
+If SNI is not sent, the Origin Set is empty when the connection is established.
```
martinthomson: This is a retroactive change to h2.  It says that you can't coalesce.  It also says that you can't *use* a connection if you didn't use SNI.  I think that you need to include some weasel-words about how this doesn't really apply until the first ORIGIN frame is sent or received.

draft-ietf-httpbis-origin-frame.md
```diff
@@ -101,49 +106,65 @@ REMOVE (0x2):
...
    2. If the REMOVE flag is set, remove any member of the Origin Set that is the same as `parsed_origin` (as per {{!RFC6454}}, Section 5), and continue to the next `parsed_origin`.
    3. Otherwise, add `parsed_origin` to the Origin Set.
 
 
 # Security Considerations
 
 Clients that blindly trust the ORIGIN frame's contents will be vulnerable to a large number of
-attacks; hence the reinforcement that this specification does not relax the requirement for server
-authority in {{?RFC7540}}, Section 10.1.
+attacks.
```
martinthomson: Rather than simple removal, I would add:

> Though this document removes the requirement on clients to ensure that the DNS refers to the same IP address, it does not remove the requirement for certificate-based authentication.  A client MUST validate a server IP address.

draft-ietf-httpbis-origin-frame.md
```diff
@@ -58,6 +58,11 @@ Using a status code in this manner allows clients to recover from misdirected re
 penalty of adding latency. To address that, this specification defines a new HTTP/2 frame type,
 "ORIGIN", to allow servers to indicate what origins a connection is usable for.
 
+Additionally, experience has shown that HTTP/2's requirement to establish server authority using
```
mcmanus: in discussions on this there are a couple of other motivators that may (or may not!) be worth including as introductory text.. one is simply that it can save a dns lookup by effectively inlining the result and the other is that dns is often a privacy leak that can be avoided by not making the lookup

draft-ietf-httpbis-origin-frame.md
```diff
@@ -70,80 +75,183 @@ The ORIGIN HTTP/2 frame ({{!RFC7540}}, Section 4) allows a server to indicate wh
...
+|         Origin-Len (16)       | ASCII-Origin? (*)           ...
 +-------------------------------+-------------------------------+
 ~~~~
 
 The ORIGIN frame's payload contains the following fields, sets of which may be repeated within the
 frame to indicate multiple origins:
 
 Origin-Len:
-: An unsigned, 16-bit integer indicating the length, in octets, of the Origin field.
+: An unsigned, 16-bit integer indicating the length, in octets, of the ASCII-Origin field.
```
deweerdt: Should this mention network ordering?

draft-ietf-httpbis-origin-frame.md
```diff
@@ -70,80 +75,183 @@ The ORIGIN HTTP/2 frame ({{!RFC7540}}, Section 4) allows a server to indicate wh
...
+6. For each Origin field `origin_raw` in the frame payload:
+   1. Parse `origin_raw` as an ASCII serialization of an origin ({{!RFC6454}}, Section 6.2) and let the result be `parsed_origin`. If parsing fails, skip to the next `origin_raw`.
+   2. If the `scheme` of `parsed_origin` is not "https", skip to the next `origin_raw`.
+   3. If the certificate presented by the server is not valid for the `host` of `parsed_origin` (see below), skip to the next `origin_raw`.
+   4. If the `port` of `parsed_origin` does not match the connection's remote port, skip to the next `origin_raw`.
+   5. Add `parsed_origin` to the Origin Set.
+
+The certificate presented by the server is valid for a host if it passes the checks that the client
+would perform when forming a new TLS connection to the origin. This includes verifying that the
+host matches a `dNSName` value from the certificate `subjectAltName` field (using the wildcard rules
```
deweerdt: typo: s/sNSName/DNSName/

draft-ietf-httpbis-origin-frame.md
```diff
@@ -70,80 +75,183 @@ The ORIGIN HTTP/2 frame ({{!RFC7540}}, Section 4) allows a server to indicate wh
...
+defined in {{!RFC2818}}).
+
+
+# Operational Considerations for Servers {#server-ops}
+
+The ORIGIN frame allows a server to indicate what for origins a given connection ought be used.
```
deweerdt: s/to indicate what for origins/to indicate for which origins/

draft-ietf-httpbis-origin-frame.md
```diff
@@ -70,80 +75,183 @@ The ORIGIN HTTP/2 frame ({{!RFC7540}}, Section 4) allows a server to indicate wh
...
-CLEAR (0x1):
-: Indicates that the Origin Set MUST be reset to an empty set before processing the contents of the frame it occurs upon.
+Likewise, the ORIGIN frame is only valid on connections with the "h2" protocol identifier, or when
+specifically nominated by the protocol's definition; it MUST be ignored when received on a
+connection with the "h2c" protocol identifier.
 
-REMOVE (0x2): 
-: Indicates that the origin(s) carried in the payload must be removed from the Origin Set, if present; if not present, it/they have no effect.
+This specification does not define any flags for the ORIGIN frame, but future updates might use
+them to change its semantics. The first four flags (0x1, 0x2, 0x4 and 0x8) are reserved for
```
mnot: Well, it has to be somewhere. I did it here because there seem to be some cases that require a breaking change (e.g., if we want to include a DELETE or CLEAR flag down the road), and some that don't (e.g., wildcard support).

Happy to take it out (there are always more ways to do it), but it seemed mostly harmless to try it out here.

draft-ietf-httpbis-origin-frame.md
mnot: Convention is not to in new frame types, since H2 defines it.

draft-ietf-httpbis-origin-frame.md
mnot: already fixed

draft-ietf-httpbis-origin-frame.md
```diff
@@ -58,6 +58,11 @@ Using a status code in this manner allows clients to recover from misdirected re
 penalty of adding latency. To address that, this specification defines a new HTTP/2 frame type,
 "ORIGIN", to allow servers to indicate what origins a connection is usable for.
 
+Additionally, experience has shown that HTTP/2's requirement to establish server authority using
+both DNS and the server's certificate is onerous. This specification relaxes the requirement to
+check DNS when the ORIGIN frame is in use.
+
+
```
mnot: already fixed

draft-ietf-httpbis-origin-frame.md
mnot: Is correct; see https://tools.ietf.org/html/rfc5280#section-4.2.1.6 (but will add a ref)

draft-ietf-httpbis-origin-frame.md
mnot: Done

draft-ietf-httpbis-origin-frame.md
```diff
@@ -70,80 +75,183 @@ The ORIGIN HTTP/2 frame ({{!RFC7540}}, Section 4) allows a server to indicate wh
...
+origin ({{!RFC6454}}, Section 6.2). If parsing fails, the field MUST be ignored.
+
+Senders should note that, as per {{!RFC6454}} Section 4, the values in an ORIGIN header need to be
+case-normalised before serialisation.
+
+Once parsed, the value MUST have:
+
+* a scheme of "https", 
+* a host that is reflected in a `subjectAltName` of the connection's TLS certificate (using the wildcard rules defined in {{!RFC2818}}, Section 3.1), and 
+* a port that reflects the connection's remote port on the client.
```
mnot: Yeah. need to think about how to best specify that.

mnot: @mcmanus see latest commit; I defer the cert checks until use of the origin, and I think that takes care of the concerns here.

draft-ietf-httpbis-origin-frame.md
```diff
@@ -70,80 +75,183 @@ The ORIGIN HTTP/2 frame ({{!RFC7540}}, Section 4) allows a server to indicate wh
...
 
-When a connection is first established, its Origin Set is defined to be those origins that the client would normally consider the connection authoritative for; see {{!RFC7540}}, Section 10.1.
+By default, a connections's Origin Set is uninitialised. When an ORIGIN frame is first received and
+successfully processed by a client, the connection's Origin Set is defined to contain a single
+origin, composed from:
 
-The ORIGIN frame allows the server to modify the Origin Set. In particular:
+  - Scheme: "https"
+  - Host: the value sent in Server Name Indication ({{!RFC6066}} Section 3), converted to lower case
+  - Port: the local port of the connection on the server
```
mnot: I'm inclined to think that if you want to host an alternative service and send ORIGIN, you need to explicitly list the alternative in ORIGIN. The client *could* make some assumptions, but that doesn't seem great. Thoughts?

mcmanus:
On Mon, Feb 6, 2017 at 9:15 PM, Mark Nottingham <notifications@github.com>
wrote:

> I'm inclined to think that if you want to host an alternative service and
> send ORIGIN, you need to explicitly list the alternative in ORIGIN. The
> client *could* make some assumptions, but that doesn't seem great.
> Thoughts?
>

wait - no. You wouldn't send the alternative in the ORIGIN frame, you would
send the origin that you are the alternative for, right? that would make
sense - if you want to use an alternative you can't rely on the SNI default.

Issue #306, opened 2017-02-27
Allow servers to specify max-age in the client hints header


tarunban:
The [Accept-CH](http://httpwg.org/http-extensions/client-hints.html) header is currently included by origin servers in HTTP responses if they are interested in receiving client hints from the browser (e.g., DPR). The spec currently does not require browsers to persist the opt-in across browser restarts. This means that if the origins are interested in receiving hints on all requests, then they need to include the Accept-CH header on all responses. This has two disadvantages: (i) Browsers currently fail to include the hints on the first request after the browser restart; and, (ii) It may not be possible for the origin servers to include the Accept-CH header on all responses due to responses being served out of different endpoints (CDNs etc.).

To fix this: The Accept-CH header should also support a [max-age](http://wicg.github.io/reporting/#max-age-member) header value which indicates for how long should the browser remember this opt-in (including across browser restarts). A value of max-age set to 0 should cause the browser to stop sending client-hints to the origin. As an example,

`Accept-CH: feature1, feature2, max-age=86400`

would imply that the browser should include hints `feature1` and `feature2` for next 86400 seconds for that origin.

michael-oneill:
It crossed my mind also that the response determines a future request, so how long in the future?

Why not create a well-known named cookie, similar to the cookie prefix idea in RFC6265bis, then use that to convey Accept-CH. Then you get the potential fingerprinting showing up in the cook,ie store where the user (or privacy extensions) can see it. You also get the expiry functionality  for free. 

And the SameSite attribute to help with privacy.

tarunban:
@michael-oneill I am not sure if cookies are the best place to store this information given that cookies are used by servers for state management, and this does not seem to fit that use case.

michael-oneill:
Accept-CH makes the UA retain the state on what Client Hints are to be sent in future requests. The problem with Client Hints is that they can (will) be used to identify users by fingerprinting their UAs. 

Users know to some extent that cookies are used to identify them, so why not hold this identifying styate their also.

It increases transparency,, enables user control and does not re-invent the wheel.

mnot:
Making this a new directive in `Accept-CH` is a bit problematic, since that header field is defined as a list of header field names. It's technically possible, you just would get into situations where you have to switch your parser based upon the format of the directive, and reserve keywords that now can't be used as headers (and thus, have to register them as headers -- see `Connection`).

You *could* do it as a new header, though -- e.g.,

`Accept-CH-Lifetime: 3600`

or similar.

[...]

PR #313, opened 2017-03-17
define Accept-CH-Lifetime

### LINKED ISSUE #307: Scoping of Accept-CH
The `Accept-CH` response header is defined as:

> When a client receives Accept-CH, or if it is capable of processing the HTML response and finds an equivalent HTML meta element, it can treat it as a signal that the server is interested in receiving the Client-Hint header fields that match the advertised field-values; subsequent requests initiated to the same server and, optionally any subresource requests initiated as a result of processing the response from the server that includes the Accept-CH opt-in, can include the Client-Hint header fields that match the advertised field-values.
>
> For example, based on Accept-CH example above, a user agent could append DPR, Width, Viewport-Width, and Downlink header fields to all subresource requests initiated by the page constructed from the response. Alternatively, a client can treat advertised support as a persistent origin preference and append same header fields on all future requests initiated to and by the resources associated with that origin.

"All future requests" is a long time indeed. While of course the client isn't going to literally do that (eventually it'll be garbage collected, or the user will reset state, or the machine will be recycled), from a server-side perspective, it'd be nice to have *some* level of control over how bloated your requests get.

Additionally, having `Accept-CH` possibly indicate that future references *to* **and** *from* this origin seems like a pretty broad brush to be painting with. While Chrome might make some reasonable decisions about tradeoffs here (and probably needs some wiggle room), if this ever does get implemented elsewhere, this broadness and freedom is a recipe for bad interop and worse server-side headaches.

So, can we nail this down at all? 

The most conservative approach is:

"Send CH headers on subrequests generated by this content, please."

However, the above also says "subsequent requests initiated to the same server" in the primary spot; the case above is relegated to "optionally."

Personally, my preference would be to drop "subsequent requests initiated to the same server and, optionally" and the corresponding text in the example below it. The primary use case for CH as I understand it is serving appropriate subresources, NOT dynamically modifying HTML, so that should do the trick.

Am I missing something, @igrigorik?

Regardless of that, "server" is used very loosely above; I think you mean "origin", right? And "can" should probably be "MAY".

CC @yoavweiss

igrigorik:
Closes #307 and #306.

### DIFF
--- a/draft-ietf-httpbis-client-hints.md
+++ b/draft-ietf-httpbis-client-hints.md
@@ -133,7 +133,7 @@ For example, based on Accept-CH example above, a user agent could append DPR, Wi
 
 ### The Accept-CH-Lifetime header field {#accept-ch-lifetime}
 
-Servers can ask the client to remember an origin-wide Accept-CH preference for a specified period of time to enable delivery of Client Hints on subsequent requests.
+Servers can ask the client to remember an origin-wide Accept-CH preference for a specified period of time to enable delivery of Client Hints on all subsequent requests to the origin, and on subresource requests initiated as a result of processing a response from the origin.
 
 ~~~ abnf7230
   Accept-CH-Lifetime = #delta-seconds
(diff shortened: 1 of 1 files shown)


mnot:
Looks good overall. 

`Accept-CH-Lifetime` changes the semantics of `Accept-CH`. Not only does it effectively put `Accept-CH` on subsequent responses from the same origin for the given lifetime, it **also** suggests that Client Hints should be added to requests _to_ that origin, no matter where the requesting page has come from.

Not a problem, but right now it's conveyed almost as an aside. This needs to be clearly specified so it isn't missed / misunderstood.

igrigorik:
@mnot updated, PTAL.

igrigorik:
Thanks Mark!

PR #323, opened 2017-04-11
change frame type of origin to 0xc to avoid conflict

mcmanus:
While testing firefox's implementation of this draft Vlad determined that some versions of Chrome throw PROTOCOL_ERROR and close the connection if it receives a type 0xB frame with a non zero length on stream 0.

Other frame type numbers, such as 0xC do not seem to have that problem.

testing indicates this is fixed in version 59 and is a problem previous to that. (57 is in release as I write this).

presumably this is related to historical code around BLOCKED which was proposed to use 0xB.

Prudence says to just use a different code point for the moment.

### DIFF
--- a/draft-ietf-httpbis-origin-frame.md
+++ b/draft-ietf-httpbis-origin-frame.md
@@ -78,7 +78,7 @@ for the connection it occurs within.
 
 ## Syntax {#syntax}
 
-The ORIGIN frame type is 0xb (decimal 11).
+The ORIGIN frame type is 0xc (decimal 12).
 
 ~~~~
 +-------------------------------+-------------------------------+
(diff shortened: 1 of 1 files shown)


martinthomson:
Question for @mcmanus and @mnot, should we somehow reserve 0xb to avoid this problem?

reschke:
If this is fixed in Chrome 59, why bother? Actually, I'd also revert this change.

mnot:
Neither. There are still old Chrome versions deployed (some people lag, for various reasons), so it's not safe to use now, but it should be in the not-to-distant future; we just need to wait a while to use it (presumably for another extension).

Reserving it for all time would be a bad precedent.

PR #326, opened 2017-04-18
Use stand_alone for markdown

martinthomson:
This includes a common set of processing instructions for the markdown files.  The results should be more consistent with other drafts in the repo.  

This is partly selfish.  I found that this is necessary on my machine because java is flaky and kramdown-rfc2629 creates entity references that it can't handle.


reschke:
I would recommend to remove all PIs (unless there is one that you think has an actual effect)

mnot:
OBE.

Issue #318, opened 2017-03-31
Definition of Expect-CT is a little unclear


martinthomson:
The convention for HTTP is to define header fields as follows:

```
<header-field-name> = #<header-field-name>-fields
```

Which leads to me suggesting:

```abnf
Expect-CT = #Expect-CT-Directives
Expect-CT-Directives = [ parameter *( OWS ";" OWS parameter ) ]
```

Note that this doesn't follow the common structure, which uses:

```abnf
common-structure = 1* ( identifier dictionary )
```

Ignore the obvious errors in the common structure syntax, the intent is to have a list of identifiers followed by dictionaries.

I've marked this as editorial, I think that the outcome will not affect the wire syntax at all.

royfielding:
If it is supposed to be a list of directives (some of which may be name=value), then it should be
```
   Expect-CT       = #directive
   directive       = directive-name [ "=" directive-value ]
   directive-name  = token
   directive-value = token / quoted-string
```
If it is supposed to be a list of directives with optional name=value parameters to each directive, then
```
   Expect-CT       = #directive
   directive       = directive-name *( OWS ";" OWS parameter ) 
   directive-name  = token
   parameter       = token / ( token "=" quoted-string )
```
I guess we could also use some examples in the specification.

reschke:
@martinthomson  what you suggested *does* change the wire syntax...

reschke:
I also disgree with the statement:

> The convention for HTTP is to define header fields as follows:

That depends on the type of the field. It's definitively not true universally right now.

royfielding:
BTW, that would be to match the standard HTTP/1.x syntax for most header fields. I don't think it is appropriate to use the experimental common structure, which is still just a proposal, unless folks want to make this header field a guinea pig.

reschke:
https://greenbytes.de/tech/webdav/rfc7838.html#rfc.section.3 might be a good template.

estark37:
Sorry for my delay here.

If I'm understanding correctly, then @royfielding's first suggestion is the intention: it's supposed to be a list of directives, some of which may be name=value:
```
   Expect-CT       = #directive
   directive       = directive-name [ "=" directive-value ]
   directive-name  = token
   directive-value = token / quoted-string
```

However this changes the format from semicolon-separated directives (`Expect-CT: enforce; max-age=23`) to comma-separated (`Expect-CT: enforce, max-age=23`); is that correct?

[...]

Issue #321, opened 2017-04-03
Expect CT: Add an example


craigfrancis:
Would it be possible to include an example in the spec?

Something like...

    Expect-CT: enforce; max-age=3600; report-uri="https://example.com/expect-ct-report"

As I believe most developers like to work with examples, rather than just using the spec definition as found under the heading "Response Header Field Syntax".

:-)

PR #327, opened 2017-04-18
Update Expect-CT syntax to match convention

estark37:
This changes Expect-CT to be a comma-separated list of directives (which may be
name=value pairs), and adds some examples.

Also allows the server to send multiple header fields so that they can be
combined, as per
https://github.com/httpwg/http-extensions/issues/318#issuecomment-294369176

Fixes issue #318, #321

### DIFF
--- a/draft-ietf-httpbis-expect-ct.md
+++ b/draft-ietf-httpbis-expect-ct.md
@@ -174,10 +174,10 @@ using the grammar defined in RFC 5234 {{!RFC5234}} and the rules defined in
 Section 3.2 of RFC 7230 {{!RFC7230}}.
 
 ~~~ abnf
-Expect-CT-Directives = directive *( OWS ";" OWS directive )
-directive            = directive-name [ "=" directive-value ]
-directive-name       = token
-directive-value      = token / quoted-string
+Expect-CT           = #expect-ct-directive
+expect-ct-directive = directive-name [ "=" directive-value ]
+directive-name      = token
+directive-value     = token / quoted-string
 ~~~
 {: #expect-ct-syntax title="Syntax of the Expect-CT header field"}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-expect-ct.md
```diff
@@ -274,6 +274,17 @@ delta-seconds = 1*DIGIT
 
 `delta-seconds` is used as defined in Section 1.2.1 of RFC 7234 {{!RFC7234}}.
 
+### Examples
+
+The following examples demonstrate valid Expect-CT response header fields:
+
+~~~ inline
+Expect-CT: max-age=86400,enforce
+Expect-CT: max-age=86400,enforce,report-uri="https://foo.test/report"
```
royfielding: For variation, I suggest adding a single space after the commas for this middle example just to remind readers that spaces here are ignored.

estark37: Done

draft-ietf-httpbis-expect-ct.md
```diff
@@ -274,6 +274,17 @@ delta-seconds = 1*DIGIT
...
 `delta-seconds` is used as defined in Section 1.2.1 of RFC 7234 {{!RFC7234}}.
 
+### Examples
+
+The following examples demonstrate valid Expect-CT response header fields:
+
+~~~ inline
+Expect-CT: max-age=86400,enforce
+Expect-CT: max-age=86400, enforce, report-uri="https://foo.test/report"
+Expect-CT: max-age=86400,report-uri="https://foo.test/report"
```
martinthomson: I would put a blank line between each example, lest this be confused for a single header block.

Also, use foo.example, not foo.test.

estark37: Done and done

PR #329, opened 2017-04-20
Two editorial nits.

bencebeky:
1. Fix one typo.
2. Technically, a set is always a subset of itself.  Therefore current wording implies that if two connections have identical Origin Sets, then neither should be used for new requests, and both should be closed.  Fix this by saying "proper subset" instead of "subset".


mnot:
Thanks!

Issue #333, opened 2017-04-26
internal section references should be dynamic


reschke:
Right now they use hardwired section numbers (likely from RFC6265); this is going to fail when numbers change.

PR #338, opened 2017-04-27
make internal references dynamic

reschke:
addresses #333


mikewest:
This is very cool. I didn't see this kind of linking mechanism in the kramdown-rfc2629 docs... is there a list of the supported syntatical sugar somewhere I should skim through?

Issue #343, opened 2017-05-17
text about why RFC 3490 is cited is missing


reschke:
See the annotation in https://greenbytes.de/tech/webdav/rfc6265.html#RFC3490

reschke:
Opened https://github.com/cabo/kramdown-rfc2629/issues/42

mikewest:
Will be addressed in the -02 I'll be uploading shortly, thanks to your patch in https://github.com/httpwg/http-extensions/commit/d5f7cb8c000f7a5c6d3545827883efe1a90eac99.

PR #344, opened 2017-05-17
augment downref to RFC 3490 with explanation (see #343)

reschke:
(taken from 6265)

Note this requires kramdown-rfc2629 version 1.2.2 or newer  - please make sure to use this when generating submissions


mikewest:
Thanks (and thanks for the note about Kramdown)

PR #348, opened 2017-05-23
Explain interactions with Alt-Svc

martinthomson:
The consequences for ORIGIN + Alt-Svc with a port change aren't awesome, but this should at least remove any confusion.

### DIFF
--- a/draft-ietf-httpbis-origin-frame.md
+++ b/draft-ietf-httpbis-origin-frame.md
@@ -149,7 +149,7 @@ and successfully processed by a client, the connection's Origin Set is defined t
 initial origin.  The initial origin is composed from:
 
   - Scheme: "https"
-  - Host: the value sent in Server Name Indication ({{!RFC6066}} Section 3), converted to lower case
+  - Host: the value sent in Server Name Indication (SNI, {{!RFC6066}} Section 3), converted to lower case
   - Port: the remote port of the connection (i.e., the server's port)
 
 The contents of that ORIGIN frame (and subsequent ones) allows the server to incrementally add new
@@ -162,11 +162,13 @@ Section 6.2) and remove it from the connection's Origin Set, if present.
 
 Note:
 
-: Using an alternative service {{?RFC7838}} can result in using a service that has a different
-  port to the original service (that is, the server that a client contacts when it has no active
-  alternative services).  In that case, the initial origin will have a different port number.  A
-  client could be unable to send requests for the origin that advertised the alternative
-  service.  Explicitly including origins in the ORIGIN frame avoids this problem.
+: When sending an ORIGIN frame to a connection that is initialised as an Alternative Service
+  {{?RFC7838}}, the initial origin set {{ref}} will contain an origin with the appropriate
+  scheme and hostname (since Alternative Services specifies that the origin's hostname be sent
+  in SNI). However, it is possible that the port will be different than that of the intended
+  origin, since the initial origin set is calculated using the actual port in use, which can be
+  different for the alternative service. In this case, the intended origin needs to be sent in
+  the ORIGIN frame explicitly.
 
 : For example, a client making requests for "https://example.com" is directed to an alternative
   service at ("h2", "x.example.net", "8443").  If this alternative service sends an ORIGIN
--- a/draft-ietf-httpbis-origin-frame.md
+++ b/draft-ietf-httpbis-origin-frame.md
@@ -162,11 +162,13 @@ Section 6.2) and remove it from the connection's Origin Set, if present.
 
 Note:
 
-: A server that acts as an alternative service {{?RFC7838}} might remove the origin that a
-  client creates the connection for by sending an ORIGIN frame.  The single value that is in the
-  newly-initialized Origin Set will contain the port of the alternative service rather than that
-  of the origin.  For instance, a client making requests for "https://example.com" that is
-  directed to an alternative at ("h2", "x.example.net", "8443") will seed the origin set with
+: A server that acts as an alternative service {{?RFC7838}} might invalidate the connection for
+  use with the origin that initiated use of the alternative by sending an ORIGIN frame.  If the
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-origin-frame.md
```diff
@@ -153,6 +153,16 @@ The Origin Set is also affected by the 421 (Misdirected Request) response status
 MUST create the ASCII serialisation of the corresponding request's origin (as per {{!RFC6454}},
 Section 6.2) and remove it from the connection's Origin Set, if present.
 
+Note:
+
+: A server that acts as an alternative service {{?RFC7838}} might remove the origin that a
+  client creates the connection for by sending an ORIGIN frame.  The single value that is in the
```
mnot: ORIGIN doesn't specify how to remove a member of the origin set.

draft-ietf-httpbis-origin-frame.md
```diff
@@ -153,6 +153,16 @@ The Origin Set is also affected by the 421 (Misdirected Request) response status
...
+  newly-initialized Origin Set will contain the port of the alternative service rather than that
+  of the origin.  For instance, a client making requests for "https://example.com" that is
+  directed to an alternative at ("h2", "x.example.net", "8443") will seed the origin set with
+  "https://example.com:8443".  This problem can be avoided by explicitly including an origin
+  with the correct port in the ORIGIN frame.
+
```
mnot: I'm not sure this is helping; what are you trying to accomplish?

draft-ietf-httpbis-origin-frame.md
martinthomson: It doesn't remove it from the origin set, that's unfortunate phrasing.  It does invalidate the connection for use with the origin.

draft-ietf-httpbis-origin-frame.md
martinthomson: I thought that the example was illustrative: if you Alt-Svc with a different port, then send ORIGIN you get in a bind unless you include the original origin in the ORIGIN frame because the implicitly-included origin uses the alternative service port.

I've been trying to find a word for a. the origin that is seed for the origin set, and b. the origin that the client is chasing with alternative services.  I don't think that "original origin" is very helpful.

mnot: Maybe define a new term "initial origin"?

martinthomson: That's not quite it, it's the default (alternative) service for the origin that matters here.  The origin doesn't change when talking to an alternative.

martinthomson: More noodling on this one... "initial origin" works, but that's not the hard one.  Does "original service" work for the service that is not an alternative?

draft-ietf-httpbis-origin-frame.md
```diff
@@ -153,6 +153,20 @@ The Origin Set is also affected by the 421 (Misdirected Request) response status
...
 Section 6.2) and remove it from the connection's Origin Set, if present.
 
+Note:
+
+: Using an alternative service {{?RFC7838}} can result in using a service that has a different
+  port to the original service (that is, the server that a client contacts when it has no active
+  alternative services).  In that case, the initial origin will have a different port number.  A
+  client could be unable to send requests for the origin that advertised the alternative
+  service.  Explicitly including origins in the ORIGIN frame avoids this problem.
+
```
mnot: I think you want to say something like this:

> Servers used as an alternative service {{?RFC7838}} will need to explicitly send the origin(s) that are available when sending ORIGIN frames, because the initial origin set (see {{ref}}) does not contain any alternative services. Because the origin includes the port number, this includes alternative services that are on the same host as the origin.

I.e., this is about more than just ports.

martinthomson: That doesn't ring true either.  I realize that there are cases where the alternative has its own independent identity (e.g., example.com might use alt.example.net as an alternative, but that server might also want to serve resources for alt.example.net), but I wanted to avoid dragging that into this.

The port thing is a trap because the initial origin is primed based on the port that was *used*.  At least with the SNI, the original origin (example.com) is sent, so there is no confusion.  If the port number is the same for origin and alternative, the initial origin covers the resources that caused the client to use this alternative.  Send an empty ORIGIN frame and you still have a usable alternative.  Change the port number and that is no longer true.

mnot: > When sending an ORIGIN frame to a connection that is initialised as an Alternative Service {{?RFC7838}}, the initial origin set {{ref}} will contain an origin with the appropriate scheme and hostname (since Alternative Services specifies that the origin's hostname be sent in SNI). However, it is possible that the port will be different than that of the intended origin, since the initial origin set is calculated using the actual port in use, which can be different for the alternative service. In this case, the intended origin ought to be sent in the ORIGIN frame explicitly.

mnot:
Are you going to take another go at this?

martinthomson:
Sorry, I let this one drop because I missed your suggest, which is good.

draft-ietf-httpbis-origin-frame.md
```diff
@@ -144,12 +144,12 @@ See {{algo}} for an illustrative algorithm for processing ORIGIN frames.
 The set of origins (as per {{!RFC6454}}) that a given connection might be used for is known in this
 specification as the Origin Set.
 
-By default, a connection's Origin Set is uninitialised. When an ORIGIN frame is first received and
-successfully processed by a client, the connection's Origin Set is defined to contain a single
-origin, composed from:
+By default, a connections's Origin Set is uninitialised. When an ORIGIN frame is first received
```
mnot: connections's?

martinthomson: Your text, not mine, but I'll fix it.

martinthomson: OK, it's mine apparently

mnot:
Thanks!

PR #362, opened 2017-06-27
Contributing

mnot:
Revise contributing.md to reflect our working practices more closely.

Specifically, editors can make proposals in the drafts that close issues, as long as they remember to label them `proposal`.

This is more aligned with the quicwg process; it's different (they use `has-consensus`) because we have a large number of existing issues with consensus.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,7 +43,7 @@ discussion and their resolution.
 
 Before filing a new issue, please consider a few things:
 
-* Issues should be just that; issues with our deliverables, **not questions or support requests**.
+* Issues should be just that; issues with our deliverables, **not proposals, questions or support requests**.
 
 * Please review the issues list to make sure that you aren't filing a duplicate.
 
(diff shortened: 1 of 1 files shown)


mnot:
@mcmanus I changed this to just be like the QUIC process, where we flag things with `has-consensus` after the issue is confirmed on-list, just like QUIC. I think that will cause less confusion (esp. since I see Martin starting to adopt `has-consensus` in his WGs), and we don't **have** to flag everything with consensus as such, especially for non-controversial drafts. WDYT?

mcmanus:
whoami to stand in the way of github progress? merge it!

wtarreau:
I'm seeing a few (but very few) occurrences of the word "we" which plays a bit against "you" and may make newcomers feel a bit of resistance (ie: we decide, you abide). I think it's important to keep in mind to use "the group" or something like this instead to avoid newcomers thinking that decisions are taken by a few people as is often the case in opensource projects for example, and make them realize that by participating they become part of this group.

mnot:
Hm. I read 'we' as being inclusive -- literally, "the reader along with others." 

Further up (in existing text) is:

> To be active in the Working Group, you can participate in any of these places. Most activity takes place on the mailing list, but if you just want to comment on and raise issues, that's fine too.

I'm happy to expand that to make it clear that "we" is "all participants" -- would that help?

wtarreau:
Semantically I agree with your view and share it as well, but keep in mind we're talking to people completely external to this group and thinking whether or not they're going to join by reading this file serving as a rule. Maybe indeed adding a line after the one you quoted saying 'Anyone participating to a discussion or an issue automatically becomes part of the Working Group, sometimes referred to as "we" in this document'. But in general I think that "we" should be avoided as it rarely includes "you", especially when there are probably 2 or 3 occurrences only.

mnot:
OK. I added an explicit statement about "membership" at the top -- that should clarify things. I don't think that with that context "we" is harmful or misleading.

wtarreau:
OK fine, thanks.

Issue #372, opened 2017-07-19
Accept-CH-Lifetime privacy concerns


arturjanc:
Based on https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/QHI3sio6--Q/v_zWX1O6AAAJ I have a couple of questions about the impact of `Accept-CH-Lifetime` as currently defined in https://tools.ietf.org/html/draft-ietf-httpbis-client-hints-04#section-2.2.2

The main risk, touched upon in the [Security Considerations](https://tools.ietf.org/html/draft-ietf-httpbis-client-hints-04#section-5) section, is the fact that providers of cross-origin subresources (e.g. images) loaded from pages with `Accept-CH` headers will start receiving fingerprinting-prone information about the configuration of a connecting client. Note that it will expose more information than what was available to these parties in the past because subresources such as images cannot use client-side logic to access information sent in Client Hints. A related issue is that based on the sets of headers sent in such subresource requests, the subresource owner might be able to identify the referring site even if it sets a Referrer Policy to prevent disclosing its URL or origin. For example, requests from a large site which sets `Referrer-Policy: no-referrer; Accept-CH: DPR` will be distinguishable from requests from sites with `Accept-CH: DPR, Viewport-Width` and from those without client hints. Depending on the chosen set of hints this can in practice uniquely identify the origin visited by the user.

The introduction of `Accept-CH-Lifetime` will extend this problem to all resources on a given origin -- if one page sets the header, then subresource requests from all pages in that origin will start carrying hint information. This can be undesirable for the user because it can disclose information about the visited origin and broadcasts fingerprintable information to all parties from which the given origin loads subresources.

It seems like this should be mitigated, possibly by one of the following:
- (Preferably) Not broadcasting client hints on cross-origin subresource loads.
- Obeying Referrer Policy and omitting hints if the referring page attempts to restrict data which is sent in the `Referer` header. Similarly, hints should probably also be stripped on HTTPS -> HTTP transitions.

tarunban:
> The introduction of Accept-CH-Lifetime will extend this problem to all resources on a given origin -- if one page sets the header, then subresource requests from all pages in that origin will start carrying hint information. 

IIUC, this is not correct. If the origin includes `Accept-CH-Lifetime` header, then the user agent is expected to send the client hints for requests to only that origin.

> Similarly, hints should probably also be stripped on HTTPS -> HTTP transitions.

[...]

PR #373, opened 2017-07-21
Scope Accept-CH opt-in to same origin

igrigorik:
WIP, based on discussion in #372.


draft-ietf-httpbis-client-hints.md
```diff
@@ -257,11 +257,13 @@ The Content-DPR response header field indicates to the client that the server ha
 
 # Security Considerations
 
-The request header fields defined in this specification expose information that is already available to Web applications in the browser runtime itself (e.g., using JavaScript and CSS). For example, the application can obtain viewport width, image display width, and device pixel ratio via JavaScript, or through the use of CSS media queries and unique resource URLs even if JavaScript is disabled. However, servers that gather this information through such mechanisms are typically observable (e.g., you can see that they're using JavaScript to gather it), whereas servers' use of the header fields introduced by this specification is not observable. Section 2.1 discusses potential mitigations.
+The request header fields defined in this specification, and those that extend it, expose information about the user's environment to enable proactive content negotiation. Such information may reveal new information about the user and implementers ought to provide policies and mechanisms to control how and when such hints are advertised.
```
arturjanc: I'm worried about this being a little vague. It seems that there would be a benefit to specifying the "policies and mechanisms" mentioned in this para because they might be crucial to ensuring the privacy properties of a given implementation.

draft-ietf-httpbis-client-hints.md
```diff
@@ -257,11 +257,13 @@ The Content-DPR response header field indicates to the client that the server ha
...
 
-For example, sending Client Hints on all requests can make information about the user's environment available to origins that otherwise did not have access to this data, which may or may not be the desired outcome - e.g. this may enable an image optimization service to deliver a tailored asset, and it may reveal same information about the user to other origins that may not have had access to it before. Similarly, sending highly granular data, such as image and viewport width may help identify users across multiple requests. Restricting such field values to an enumerated range, where the user agent advertises a threshold value that is close but is not an exact representation of the current value, can help mitigate the risk of such fingerprinting.
+The hint request headers ought not to provide new information that is otherwise not available to the application via HTML, CSS, or JavaScript. For example, this specification defines Viewport-Width, Width, and DPR header fields, all of which can be obtained via JavaScript, or through the use of CSS media queries and unique resource URLs even if JavaScript is disabled.
```
arturjanc: I would make this stronger:
1. Hints shouldn't provide information that isn't available via the usual channels.
2. Hints sent for subresource requests shouldn't carry information not available to the provider of the subresource unless the first-party opts into sending hints (note that this, combined with the same-origin restriction, would likely suggest some form of double-keying)

igrigorik:
@arturjanc thanks for the review. Took another run, with a few large updates:

- Accept-CH and Accept-CH-Lifetime should be processed for responses originating from potentially trustworthy origins (i.e. HTTPS-only)
- Accept-CH-Lifetime preference should be double-keyed, per discussion in #372.
- Consolidated security guidance in one section + tried to rewrite that in light of above updates.

I couldn't find any good existing spec examples on double-keying, so would appreciate any guidance on how to explain it here well. 

@mnot ptal as well, as this is a substantive update.

/cc @tarunban @yoavweiss

draft-ietf-httpbis-client-hints.md
```diff
@@ -101,9 +102,9 @@ A Client Hint request header field is a HTTP header field that is used by HTTP c
 
 ## Sending Client Hints
 
-Clients control which Client Hints are sent in requests, based on their default settings, user configuration and/or preferences. Implementers might provide user choice mechanisms so that users may balance privacy concerns with bandwidth limitations. Implementations specific to certain use cases or threat models might avoid transmitting these header fields altogether, or limit them to secure contexts or authenticated sessions. Implementers should be aware that explaining the privacy implications of passive fingerprinting or network information disclosure may be challenging.
+Clients control which Client Hints are sent in requests, based on their default settings, user configuration and/or preferences. The client and server, or an intermediate proxy, can use an opt-in mechanism outlined below to negotiate which fields should be sent to allow for efficient content adaption.
```
mnot: Is the intent of including 'intermediary proxy' to include CDNs, or to also include forward proxies? If it's just CDN/reverse proxies, I'd remove it; it's widely understood that they're acting as a server. 'intermediary proxy' means "forward proxy" and I don't think you want that.

draft-ietf-httpbis-client-hints.md
```diff
@@ -127,28 +128,28 @@ For example:
   Accept-CH: DPR, Width, Viewport-Width
 ~~~
 
-When a client receives Accept-CH, or if it is capable of processing the HTML response and finds an equivalent HTML meta element, it can treat it as a signal that the origin ({{RFC6454}}) is interested in receiving specified request header fields that match the advertised field-values; same-origin resource requests initiated as a result of processing the response from the server that includes the Accept-CH opt-in can include the request header fields that match the advertised field-values.
+When a client receives Accept-CH from a potentially trustworthy origin ({{SECURE-CONTEXTS}}), or if it is capable of processing the HTML response and finds an equivalent HTML meta element, it can treat it as a signal that the origin ({{RFC6454}}) is interested in receiving specified request header fields that match the advertised field-values; same-origin resource requests initiated as a result of processing the response from the server that includes the Accept-CH opt-in can include the request header fields that match the advertised field-values.
```
mnot: This reads as if you can either have a secure origin OR include the HTML meta tag to get Accept-CH processed; is that the intent?

draft-ietf-httpbis-client-hints.md
```diff
@@ -127,28 +128,28 @@ For example:
...
 
-Servers can ask the client to remember an origin-wide Accept-CH preference for a specified period of time to enable delivery of Client Hints on subsequent requests to the origin ({{RFC6454}}).
+Servers can ask the client to remember sent Accept-CH preference for a specified period of time, to enable delivery of Client Hints on subsequent requests to the server's origin ({{RFC6454}}).
 
 ~~~ abnf7230
   Accept-CH-Lifetime = #delta-seconds
 ~~~
 
-The field-value indicates that the Accept-CH preference SHOULD be considered stale after its age is greater than the specified number of seconds.
+When a client receives Accept-CH-Lifetime from a potentially trustworthy origin ("opt-in origin"), the field-value indicates that the Accept-CH preference SHOULD be considered stale after its age is greater than the specified number of seconds, and if applicable, persisted as a double-keyed preference that combines the values of the opt-in origin and the potentially trustworthy origin of the resource that initiated the request that received the opt-in preference.
```
mnot: Age in relation to what?

draft-ietf-httpbis-client-hints.md
```diff
@@ -127,28 +128,28 @@ For example:
   Accept-CH: DPR, Width, Viewport-Width
 ~~~
 
-When a client receives Accept-CH, or if it is capable of processing the HTML response and finds an equivalent HTML meta element, it can treat it as a signal that the application is interested in receiving specified request header fields that match the advertised field-values; subresource requests initiated as a result of processing the response from the server that includes the Accept-CH opt-in can include the request header fields that match the advertised field-values.
+When a client receives Accept-CH from a potentially trustworthy origin ({{SECURE-CONTEXTS}}), or if it is capable of processing the HTML response and finds an equivalent HTML meta element, it can treat it as a signal that the origin ({{RFC6454}}) is interested in receiving specified request header fields that match the advertised field-values; same-origin resource requests initiated as a result of processing the response from the server that includes the Accept-CH opt-in can include the request header fields that match the advertised field-values.
```
martinthomson: The addition of the reference to SECURE-CONTEXTS is new here, and unnecessary.  The requirement here is not that the context is secure (that's a notion that belongs to the web, not HTTP generally), but that the `Accept-CH` is both authentic and integrity protected.  You should limit the use of `Accept-CH` to HTTPS.

draft-ietf-httpbis-client-hints.md
```diff
@@ -127,28 +128,28 @@ For example:
...
 
-Servers can ask the client to remember an origin-wide Accept-CH preference for a specified period of time to enable delivery of Client Hints on all subsequent requests to the origin ({{RFC6454}}), and on any requests initiated as a result of processing a response from the origin.
+Servers can ask the client to remember sent Accept-CH preference for a specified period of time, to enable delivery of Client Hints on subsequent requests to the server's origin ({{RFC6454}}).
 
 ~~~ abnf7230
   Accept-CH-Lifetime = #delta-seconds
 ~~~
 
-The field-value indicates that the Accept-CH preference SHOULD be considered stale after its age is greater than the specified number of seconds.
+When a client receives Accept-CH-Lifetime from a potentially trustworthy origin ("opt-in origin"), the field-value indicates that the Accept-CH preference SHOULD be considered stale after its age is greater than the specified number of seconds, and if applicable, persisted as a double-keyed preference that combines the values of the opt-in origin and the potentially trustworthy origin of the resource that initiated the request that received the opt-in preference.
```
martinthomson: "double-keyed" is a new concept for the protocol here, and one that (again) really only makes sense on the web.  The property that you are looking for is that the persisted state is bound to the origin.  You should allow it to be more narrowly bound, but the point is to constrain its maximum scope.  Choosing a narrower scope is a discretionary thing that might be done by web browsers.

Also, I find the phrasing of this last bit cumbersome and hard to follow.  The point is that permission to persist might be scope (see above).

draft-ietf-httpbis-client-hints.md
```diff
@@ -257,11 +258,17 @@ The Content-DPR response header field indicates to the client that the server ha
...
+Transmitted Client Hints header fields should not provide new information that is otherwise not available to the application via HTML, CSS, or JavaScript.  Further, sending highly granular data, such as image and viewport width may help identify users across multiple requests. Restricting such field values to an enumerated range, where the advertised value is close but is not an exact representation of the current value, can help mitigate the risk of such fingerprinting as well as reduce possibility of unnecessary cache fragmentation.
```
martinthomson: The term you are looking for here is "linkability" or "linkable".  The more information that a request contains, the more likely an adversary is to be able to correctly identify two requests as belonging to the same user agent (i.e., link the two).

The claim here (reducing the accuracy of field values) isn't generically true.  I believe that it is true for the hints defined in this document, but the structure of this document is such that it might be taken as a universal truth.  For something like the proposed Geolocation header field, which is a continuous-valued field, this claim would be false (read [Section 13.5 of RFC 6772](https://tools.ietf.org/html/rfc6772#section-13.5) for background).

draft-ietf-httpbis-client-hints.md
```diff
@@ -257,11 +258,17 @@ The Content-DPR response header field indicates to the client that the server ha
...
+Implementers should consider both user and server controlled mechanisms and policies to control which Client Hints header fields are advertised:
+
+  - Implementers may provide user choice mechanisms so that users may balance privacy concerns with bandwidth limitations. However, implementers should also be aware that explaining the privacy implications of passive fingerprinting or network information disclosure to users may be challenging.
+  - Implementers should support double-keyed Client Hints opt-in requested by potentially trustworthy origins via Accept-CH and Accept-CH-Lifetime header fields, and clear remembered opt-in when site data, browsing history, browsing cache, or similar, are cleared.
```
martinthomson: should is not sufficient here: information that is persisted for an origin MUST be reset when any action that is designed to break linkability is executed (clearing cookies is one such action)

draft-ietf-httpbis-client-hints.md
```diff
@@ -257,11 +258,17 @@ The Content-DPR response header field indicates to the client that the server ha
...
+  - Implementations specific to certain use cases or threat models may avoid transmitting Client Hints header fields altogether or limit them to authenticated sessions only that already carry identifying information, such as cookies or referer data.
```
martinthomson: MAY is extremely weak and "may" is even weaker.  It seems like these are generally fixed values, akin to User-Agent, but others thus far proposed (see again Geolocation) aren't.  It seems like the advice here is intended to be generic, so what are the criteria an implementation might use to decide to send or suppress these fields if cookies are not being sent?

draft-ietf-httpbis-client-hints.md
```diff
@@ -257,11 +258,17 @@ The Content-DPR response header field indicates to the client that the server ha
...
+
+Following the above recommendations should significantly reduce the risks of linkability and passive fingerprinting.
```
martinthomson: You don't want to make this claim.  It's fairly easy to show it as false.  If your claim is that the incremental increase in linkability (over not sending hints) is reduced significantly, that might be true, but I doubt that also.  This is making a trade-off - if the information is basically already available via User-Agent or script, then this isn't a big increase, but in other cases it could be a big change.

draft-ietf-httpbis-client-hints.md
igrigorik: Ack, removed.

draft-ietf-httpbis-client-hints.md
igrigorik: Ack, updated.

draft-ietf-httpbis-client-hints.md
igrigorik: Ack, took a run at rewriting this.

draft-ietf-httpbis-client-hints.md
igrigorik: Ack, updated.

draft-ietf-httpbis-client-hints.md
igrigorik: Ack, updated.

draft-ietf-httpbis-client-hints.md
igrigorik: Ack, dropped.

draft-ietf-httpbis-client-hints.md
igrigorik: This is a tricky one. Took an attempt at rewriting it -- ptal.

igrigorik:
@martinthomson @mnot thanks for the feedback! Updated, ptal.

mnot:
Looks better.

draft-ietf-httpbis-client-hints.md
```diff
@@ -257,11 +257,15 @@ The Content-DPR response header field indicates to the client that the server ha
...
+Transmitted Client Hints header fields should not provide new information that is otherwise not available to the application via HTML, CSS, or JavaScript.  Further, sending highly granular data, such as image and viewport width may help identify users across multiple requests. Restricting such field values to an enumerated range, where the advertised value is close but is not an exact representation of the current value, can help reduce the risk of such linkability as well as reduce possibility of unnecessary cache fragmentation.
```
martinthomson: This is better, but it fails to recognize that some information can leak, even when attempts are made to obscure it.  I would add a note saying something like "Reducing the set of values that can be expressed in client hints can improve privacy by ensuring that the same value is sent by multiple users, but this can be insufficient for some types of data, especially data that can change over time."  That would address my concern regarding the possibility of a Geolocation header field that is a client hint.  (I would retain a host of other concerns about such a proposal, but you can't be expected to address all of those here.)

draft-ietf-httpbis-client-hints.md
```diff
@@ -257,11 +257,15 @@ The Content-DPR response header field indicates to the client that the server ha
 
 # Security Considerations
 
-The request header fields defined in this specification expose information that is already available to Web applications in the browser runtime itself (e.g., using JavaScript and CSS). For example, the application can obtain viewport width, image display width, and device pixel ratio via JavaScript, or through the use of CSS media queries and unique resource URLs even if JavaScript is disabled. However, servers that gather this information through such mechanisms are typically observable (e.g., you can see that they're using JavaScript to gather it), whereas servers' use of the header fields introduced by this specification is not observable. Section 2.1 discusses potential mitigations.
+The request header fields defined in this specification, and those that extend it, expose information about the user's environment to enable proactive content negotiation. Such information may reveal new information about the user and implementers ought to consider the following considerations, recommendations, and best practices.
```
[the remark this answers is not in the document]
igrigorik: I believe we use "proactive" and "server-driven" interchangeably, so current language is OK. That said, I'll defer to @mnot and @mcmanus.

draft-ietf-httpbis-client-hints.md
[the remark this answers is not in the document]
igrigorik: ack

igrigorik: Thanks good suggestions, tried to rewrite and work this into current text.

draft-ietf-httpbis-client-hints.md
```diff
@@ -257,11 +257,15 @@ The Content-DPR response header field indicates to the client that the server ha
...
+Implementers should consider both user and server controlled mechanisms and policies to control which Client Hints header fields are advertised:
+
+  - Implementers may provide user choice mechanisms so that users may balance privacy concerns with bandwidth limitations. However, implementers should also be aware that explaining the privacy implications of passive fingerprinting or network information disclosure to users may be challenging.
+  - Implementers should support Client Hints opt-in, delivered over secure transport, as advertised by Accept-CH and Accept-CH-Lifetime header fields, and must clear remembered opt-in when site data, browsing history, browsing cache, or similar, are cleared.
```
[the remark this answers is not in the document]
igrigorik: Good call, mixing may/must's here.. Moved it into a separate section.

draft-ietf-httpbis-client-hints.md
```diff
@@ -257,11 +257,15 @@ The Content-DPR response header field indicates to the client that the server ha
...
+  - Implementations specific to certain use cases or threat models may avoid transmitting some or all of Client Hints header fields. For example, by limiting them to authenticated sessions only that already carry identifying information, such as cookies or referer data, and/or avoid transmission of header fields that carry higher risks of linkability.
```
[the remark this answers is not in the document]
igrigorik: The HTTPS requirement we're adding in this update helps scope down our list of adversaries: before we were also considering passive observers, now we're scoped to the response server.  With that in mind, good point on regulating cookies: with server as adversary it's not a meaningful protection.

That said, I think the broader "may avoid transmitting some or all of CH header fields" can be a valid recommendation. I'll drop the cookies and referrer text.

igrigorik:
@martinthomson thanks for the thorough review (and patience :)), again! Updated, ptal.

draft-ietf-httpbis-client-hints.md
```diff
@@ -101,9 +101,9 @@ A Client Hint request header field is a HTTP header field that is used by HTTP c
...
+Clients control which Client Hints are sent in requests, based on their default settings, user configuration and/or preferences. The client and server can use an opt-in mechanism outlined below to negotiate which fields should be sent to allow for efficient content adaption.
 
-The client and server, or an intermediate proxy, can use an opt-in mechanism to negotiate which fields should be reported to allow for efficient content adaption.
+Implementers should be be aware of the passive fingerprinting and network information disclosure implications when implementing support for Client Hints, and follow the considerations outlined in "Security Considerations" section of this document.
```
yoavweiss: Nit: s/be be/be/

igrigorik: Ack. Good catch.

igrigorik:
> the opt-ins are limited to secure transport, but the actual hints are not. Is that correct? If so, should we change it?

I don't think that's necessary. This is implicit due to requirement for opt-in over secure transport — i.e. only secure origins can opt-in to receive hints and HTTPS-only is the new default policy for all hints.

igrigorik:
> @mnot I think it'd be good to see #373 in situ; from what I see in the patch, I think it's strictly better. No guarantees that there aren't other issues, or that this isn't completely covered, but personally I think it's better. - https://github.com/httpwg/http-extensions/issues/372#issuecomment-339513412

Squashed & merged. @martinthomson thanks for your help here and let me know if there is anything else I should address (in a separate PR :)).

PR #381, opened 2017-08-09
RFC 2616 -> 723x

reschke:
I left the term "request-uri" untouched (it was renamed in the base spec because it's not always a URI). We may want to change that separately, or even refer to the effective request URI (https://www.greenbytes.de/tech/webdav/rfc7230.html#effective.request.uri)


reschke:
@mikewest  - ping?

mikewest:
LGTM, thank you for the patch, and for the ping!

PR #390, opened 2017-09-06
Add HTTP replay draft

martinthomson:
Since our chairs have not seen the wisdom of separate repositories, I will acquiesce.


martinthomson:
Bah, it looks good.  No need to wait for Pat.

PR #398, opened 2017-09-13
reference format

reschke:


reschke:
Sorry, already committed due to mishap. Sorry.

reschke:
re-opened, as I reverted the unintended commit

PR #407, opened 2017-09-27
Manage out of order early data delivery

### LINKED ISSUE #391: Early data that arrives after the handshake completes
The conclusion in the draft is that early data is relatively "safe" (for some definition of the term, see the draft for details) if the handshake is complete. Early data in a transport that provides ordered delivery will always arrive before the handshake completes. But this is not the case in QUIC, or anything like DTLS. We should offer some advice for these situations.

From martinthomson/http-replay#43

martinthomson:
The best and easiest plan is to discard, but we know that throwing out
perfectly good packets can have bad effects on performance.  It sends a bad
signal to a congestion controller for instance.  So we need to say either
discard, or ensure that the data is properly attributed and consistently
processed.

Closes #391.

### DIFF
--- a/draft-ietf-httpbis-replay.md
+++ b/draft-ietf-httpbis-replay.md
@@ -327,6 +327,14 @@ data and selectively processing requests.  Generating a 503 (Service
 Unavailable) or 425 (Too Early) status code often leads to clients retrying
 requests, which could result in increased load.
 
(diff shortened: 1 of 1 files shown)


wtarreau:
I think you covered it all this way. +1.

PR #408, opened 2017-09-27
Intermediaries need to be consistent too

martinthomson:
This explains how consistent handling of requests can be applied by a gateway.

### DIFF
--- a/draft-ietf-httpbis-replay.md
+++ b/draft-ietf-httpbis-replay.md
@@ -251,9 +251,11 @@ Early-Data: 1
 
 ~~~
 
-An intermediary that forwards a request received in TLS early data MUST send it
-with the `Early-Data` header field set to "1" (i.e., it adds it if not present
-in the request).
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-replay.md
```diff
@@ -251,9 +251,11 @@ Early-Data: 1
 
 ~~~
 
-An intermediary that forwards a request received in TLS early data MUST send it
-with the `Early-Data` header field set to "1" (i.e., it adds it if not present
-in the request).
+An intermediary that forwards a request prior to the completion of the TLS
+handshake MUST send it with the `Early-Data` header field set to "1" (i.e., it
+adds it if not present in the request).  This is also true if the intermediary
```
mnot: s/true/required/? But if that's the case, it should really be a REQUIREMENT.

martinthomson: Sure. Reworded.

PR #409, opened 2017-09-27
Be extra consistent

martinthomson:
This text from Kyle Rose highlights the decision process.  If any node might
make a different decision about processing early data, then *this* node has to
wait.


wtarreau:
I like it, +1

PR #414, opened 2017-11-09
Incorporated comments from MartinT and some of my own edits to draft-ietf-httpbis-rand-access-live

craigpratt:
Also added a new Security Consideration for JavaScript-based clients and servers.

### DIFF
--- a/draft-ietf-httpbis-rand-access-live.xml
+++ b/draft-ietf-httpbis-rand-access-live.xml
@@ -166,16 +166,22 @@
     
     <section anchor="introduction" title="Introduction">
         <t>
-             Some Hypertext Transfer Protocol (HTTP) Clients use byte-range requests (Range requests using the "bytes" Range Unit) to transfer select portions of large representations. And in some cases large representations require content to be continuously or periodically appended - such as representations consisting of live audio or video sources, blockchain databases, and log files. Clients cannot access the appended/live content using a Range request with the bytes range unit using the currently defined byte-range semantics without accepting performance or behavior sacrifices which are not acceptable for many applications.
+             Some Hypertext Transfer Protocol (HTTP) clients use byte-range requests (Range requests using the "bytes" Range Unit) to transfer select portions of large representations. And in some cases large representations require content to be continuously or periodically appended - such as representations consisting of live audio or video sources, blockchain databases, and log files. clients cannot access the appended/live content using a Range request with the bytes range unit using the currently defined byte-range semantics without accepting performance or behavior sacrifices which are not acceptable for many applications.
         </t>
         <t>
-			For instance, HTTP Clients have the ability to access appended content by simply transferring the entire accessible portion of the representation from the beginning and continuing to read the appended content as it's made available. Obviously, this is highly inefficient for cases where the representation is large and only a portion of the randomly-accessible content is needed by the Client. And when bandwidth is limited, the Client may never be able to transfer all of the currently accessible potion and "catch up" with the appending content.
+			For instance, HTTP clients have the ability to access appended content on an indeterminate-length resource by transferring the entire representation from the beginning and continuing to read the appended content as it's made available. Obviously, this is highly inefficient for cases where the representation is large and only the most recently appended content is needed by the client. 
 		</t>
     	<t>
-			Alternatively, Clients can also access appended content by sending periodic open-ended bytes Range requests using the last-known end byte position as the range start. Performing low-frequency periodic bytes Range requests in this fashion (polling) introduces latency since the Client will necessarily be somewhat behind the aggregated content - mimicking the behavior (and latency) of segmented content representations such as HLS or MPEG-DASH. And performing these Range requests at higher frequency incurs more processing overhead and HTTP exchanges as the requests will often return no content - since content is usually aggregated in groups of bytes (e.g. a video frame, audio sample, block, or log entry).
+			Alternatively, clients can also access appended content by sending periodic open-ended bytes Range requests using the last-known end byte position as the range start. Performing low-frequency periodic bytes Range requests in this fashion (polling) introduces latency since the client will necessarily be somewhat behind the aggregated content - mimicking the behavior (and latency) of segmented content representations such as HLS or MPEG-DASH. And while performing these Range requests at higher frequency can reduce this latency, it also incurs more processing overhead and HTTP exchanges as many of the requests will return no content - since content is usually aggregated in groups of bytes (e.g. a video frame, audio sample, block, or log entry).
         </t>
         <t>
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-rand-access-live.xml
```diff
@@ -166,16 +166,22 @@
...
+             Some Hypertext Transfer Protocol (HTTP) clients use byte-range requests (Range requests using the "bytes" Range Unit) to transfer select portions of large representations. And in some cases large representations require content to be continuously or periodically appended - such as representations consisting of live audio or video sources, blockchain databases, and log files. clients cannot access the appended/live content using a Range request with the bytes range unit using the currently defined byte-range semantics without accepting performance or behavior sacrifices which are not acceptable for many applications.
```
martinthomson: "client" needs to be capitalized at the start of a sentence :)

draft-ietf-httpbis-rand-access-live.xml
```diff
@@ -166,16 +166,22 @@
...
         </t>
         <t>
-			To accommodate byte-range requests on large representations which have data appended over time efficiently and with low latency, this recommendation defines semantics whereby the HTTP Client performs byte-range requests using a combination of open-ended byte-range HEAD requests and GET requests using "Large Value" last-byte-pos values.
+            This document describes a usage model for range requests which enables
+            efficient retrieval of representations that are appended to over time
+            by utilizing large values and associated semantics for communicating
```
martinthomson: https://grammarpartyblog.com/2012/01/17/use-versus-utilize/

draft-ietf-httpbis-rand-access-live.xml
```diff
@@ -191,15 +197,31 @@
...
+            know from a range request that a client wishes to receive a response
+            that does not have a definite end. More critically, the header fields
+            do not allow the server to signal that a resource has indeterminate
+            length without also providing a fixed portion of the resource.
+
+            A client first learns that the resource has a representation of
+            indeterminate length by requesting a range of the resource. The server
+            responds with the range that is available, but indicates that the
+            length of the representation is unknown using the existing
+            Content-Range syntax. See Section 2.1 for details and examples.
```
martinthomson: <xref target="establishing-range"/> for the section reference.

draft-ietf-httpbis-rand-access-live.xml
```diff
@@ -191,15 +197,31 @@
...
+
+            Once the client knows the resource has indeterminate length, it can
+            request a range with a very large end position from the resource. The
+            client chooses an explicit end value larger than can be transferred in
+            the foreseeable term. A server which supports range requests of
+            indeterminate length signals its understanding of the client's
+            indeterminate range request by indicating that the range it is
+            providing has a range end that exactly matches the client's requested
+            range end rather than a range that is bounded by what is currently
+            available. See Sections 2.2 and 2.3 for details.
```
martinthomson: xrefs

draft-ietf-httpbis-rand-access-live.xml
```diff
@@ -246,139 +270,148 @@
...
         For example:
       </t>
 <figure><artwork type="example">
-   Range: bytes=1230000-999999999999
+    GET /resource HTTP/1.1
+    Range: bytes=1230000-999999999999
 </artwork></figure>
       <t>
-        could return
+        will return
```
martinthomson: maybe "will return a range covering only those bytes that are currently available:"

draft-ietf-httpbis-rand-access-live.xml
```diff
@@ -390,7 +423,9 @@
...
+      One potential issue with this recommendation is related to the use of very-large last-byte-pos values. Some client and server implementations may not be prepared to deal with byte position values of 2^^63 and beyond. So in applications where there's no expectation that the representation will ever exceed 2^^63, a value smaller than this value should be used as the Very Large last-byte-pos in a byte-seek request or content-range response.
+      
+      Also, some implementations (e.g. JavaScript-based clients and servers) are not able to represent all values beyond 2^^53. So similarly, if there's no expectation that a representation will ever exceed 2^^63 bytes, values smaller than this limit should be used for the last-byte-pos in byte-range requests.
```
martinthomson: it's less about representation for this, but about ease of parsing

I think that the second value here could be 2^53 instead of 2^63.

draft-ietf-httpbis-rand-access-live.xml
```diff
@@ -210,24 +232,26 @@
    byte-range-spec = first-byte-pos "-" [ last-byte-pos ]
 </artwork></figure>
       <t>
-      which allows a Client to send a HEAD request with a first-byte-pos and leave last-byte-pos absent. A Server that receives a satisfiable byte-range request (with first-byte-pos smaller than the current representation length) must respond with a 206 status code (Partial Content) with a Content-Range header indicating the currently satisfiable byte range. For example, a Client-issued HEAD request performed against a continuously aggregating representation hosted on a Server could contain a byte-range header of the form:
+      which allows a client to send a HEAD request with a first-byte-pos and leave last-byte-pos absent. A server that receives a satisfiable byte-range request (with first-byte-pos smaller than the current representation length) must respond with a 206 status code (Partial Content) with a Content-Range header indicating the currently satisfiable byte range. For example, a client-issued HEAD request performed against a continuously aggregating representation hosted on a server could contain a byte-range header of the form:
```
martinthomson: The "must" here is incorrect.  It can provide the entire representation (for instance, if it didn't support range requests).  Maybe "might".

martinthomson: Finally, the client doesn't necessarily know that the resource is continuously aggregating.  Maybe just say "For example, a client issues a request for an indefinite range of bytes as follows:"

draft-ietf-httpbis-rand-access-live.xml
craigpratt: Yeah - I think I recall hearing that rule in an English class once...

draft-ietf-httpbis-rand-access-live.xml
craigpratt: Guess I never really thought about it that much. No problem with changing it.

draft-ietf-httpbis-rand-access-live.xml
craigpratt: ok - simplified the setup...

draft-ietf-httpbis-rand-access-live.xml
craigpratt: fixed

draft-ietf-httpbis-rand-access-live.xml
craigpratt: fixed

PR #419, opened 2017-11-17
remove unused reference to IANA registry

reschke:


craigpratt:
Ahh yes - this was a leftover from the draft where I proposed adding a new range unit...

PR #421, opened 2017-11-17
Tune HTTP examples

reschke:
- add type
- consistently indent
- trailing lines for requests
- add mandatory Host header fields

### DIFF
--- a/draft-ietf-httpbis-rand-access-live.xml
+++ b/draft-ietf-httpbis-rand-access-live.xml
@@ -19,7 +19,8 @@
 <rfc
     category="exp"
     ipr="trust200902"
-    docName="draft-ietf-httpbis-rand-access-live-latest" >
+    docName="draft-ietf-httpbis-rand-access-live-latest"
+    xmlns:x='http://purl.org/net/xml2rfc/ext'>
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-rand-access-live.xml
```diff
@@ -238,16 +239,18 @@
...
       header indicating the currently satisfiable byte range. For example:
       </t>
-<figure><artwork type="example">
-      HEAD /resource HTTP/1.1
-      Range: bytes=0-
+<figure><artwork type="message/http; msgtype=&#34;request&#34;" x:indent-with="    ">
+HEAD /resource HTTP/1.1
+Host: example.com
+Range: bytes=0-
+
```
craigpratt: Do you think the blank lines at the end of the examples are really necessary? (Presume the intent is to show the terminating \n\n?)

reschke: We do it in the base specs as well...

craigpratt: Good enough reason for me. Consistency is good....

PR #427, opened 2017-11-26
Expand acronyms and cite HLS and DASH

reschke:


craigpratt:
Thanks much @reschke !

PR #428, opened 2017-11-26
Be more definite about early data

martinthomson:
This makes mitigation mandatory.  Either receive a signal that early data is safe for the given resource, or apply the mitigations.  That's consistent with the other text.


wtarreau:
LGTM

PR #429, opened 2017-11-26
Clarify intermediary TLS role better

martinthomson:
I think that these are the main places where we weren't clear enough about the relevant TLS role.


wtarreau:
LGTM

PR #430, opened 2017-11-26
New abstract

martinthomson:
John Mattsson pointed out that we didn't really discuss the risks of use of early data.  That's fair, but we don't need to.  This rewrites the abstract.


wtarreau:
Just one fix : s/mechanisms that allows/mechanisms that allow/
Nice overall.

PR #440, opened 2017-11-29
Nits from Ben Kaduk's review

martinthomson:
Hopefully none of this collides with other PRs in flight.


wtarreau:
I think it's fine with this. Adding a bit of pragmatism as you did in the last commit also reminds that the low cost of retrying is not worth a lot of trouble :-)

PR #442, opened 2017-12-01
Clarify what to do with out-of-order early data

martinthomson:
In response to review by @vasilvv.

### DIFF
--- a/draft-ietf-httpbis-replay.md
+++ b/draft-ietf-httpbis-replay.md
@@ -345,11 +345,12 @@ requests, which could result in increased load.
 
 ## Out of Order Delivery
 
-In protocols that deliver data out of order (such as QUIC {{HQ}}) early data
-can arrive after the handshake completes.  This leads to potential ambiguity
-about the status of requests and could lead to inconsistent treatment (see
+In protocols that deliver data out of order (such as QUIC {{HQ}}) early data can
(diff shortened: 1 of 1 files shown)


wtarreau:
It' indeed slightly clearer like this.

draft-ietf-httpbis-replay.md
```diff
@@ -345,11 +345,12 @@ requests, which could result in increased load.
 
 ## Out of Order Delivery
 
-In protocols that deliver data out of order (such as QUIC {{HQ}}) early data
-can arrive after the handshake completes.  This leads to potential ambiguity
-about the status of requests and could lead to inconsistent treatment (see
+In protocols that deliver data out of order (such as QUIC {{HQ}}) early data can
+arrive after the handshake completes.  This leads to potential ambiguity about
+the status of requests and could lead to inconsistent treatment (see
```
kaduk: Hmm, I wonder how we ended up with the gratuitous re-flowing.  (Different editors using different thresholds for which column to wrap at?)

martinthomson: I use both emacs and vim at different times, but both are configured with the same value.  ¯\\_(ツ)_/¯

martinthomson:
I've a new fix, based on discussion with davidben.

PR #446, opened 2017-12-08
Consistency fix

martinthomson:
Thanks @davidben and @vasilvv for pushing on this.


wtarreau:
I think this is good and can even emphasize how to consistently disable 0-RTT across a cluster for any reason then consistently re-enable it.

PR #448, opened 2017-12-09
Use the I-D Template

MikeBishop:
All the drafts appear to build cleanly on Circle in less than half the time currently required on Travis with the current setup.

**Benefits:**

- Doesn't require the installation of Java to build
- Auto-submit
- Under active development
- Doesn't require manual modification of the Makefile every time we add a new draft

~~**Drawbacks:**~~

- ~~New template doesn't use Julian's XSLT on Circle until Martin fixes https://github.com/martinthomson/i-d-template/issues/117~~


reschke:
I find xm2rfc's current HTML output disgusting, and as far as I can tell, it doesn't integrate properly into the page HTML (@mnot).

If the concern is running Java, then let's just switch to xsltproc.

MikeBishop:
You can find the current output [here](http://mikebishop.github.io/http-extensions/new-template/); I have trouble seeing this as "disgusting." They're almost exactly equivalent, from a cursory glance -- slightly different font, maybe?  And better styling on the TOC in the new template, but that's a matter of aesthetics.  Can you expand on "doesn't integrate properly into the page HTML"?

Java is certainly a pain every time I have to spin up a new machine, but much bigger is maintainability.  See the [equivalent page in the current repo](http://httpwg.github.io/http-extensions/), in which only two drafts are actually shown, and they're not even two that are currently active in the repo.  The Makefile has to be manually updated for each new draft, rather than simply detecting the drafts as they're added, and even then clearly only builds them rather than adding them to the index.

The current version never cleans up old drafts in gh-pages, doesn't build output for branches other than master, doesn't enable you to submit draft versions by tagging and pushing, and so on and so forth.

Sure, you/someone could build all those things into the current version, but... why?  Martin's already done it; let's just steal his code and be done.

reschke:
I was referring to the default output that we used to have in the repo recently for the live-ranges extension.

With "doesn't integrate properly into the page HTML" I was referring to "doesn't integrate properly into the page HTML" I meant http://httpwg.org/specs/ which I thought was including drafts as well.

I have no objects to changing the build process so that more of it is automated; but that should not affect which actual tools are used.

reschke:
FWIW, some measurements for converting XML to HTML (for secondary certs):

- rfc2629.xslt on Saxon: 3.4s
- xml2rfc: 1.7s
- rfc2629.xslt on xsltproc: 0.3s

So if the primary concern was performance, xsltproc seems to be a clear winner.

martinthomson:
@reschke, last time I ran an end-to-end test, the XSLT variants lost big time because of the dependencies.  That should go away with the docker image though.

BTW, i-d-template#117 should be fixed and xsltproc available.  Have you checked Mike?  It should be easy to see any speed difference, though you have to be careful with some of the caching rules (retriggers FTW).

MikeBishop:
I'm still failing with xsltproc on the Header Structure -- for some reason, it's not able to load the XMLs for the references, even though what appears to be the same documents are being referenced by other docs.  But that's not a problem with the Dockerfile, it's an issue running locally as well.

reschke:
How does it fail? Silently? Error messages?

MikeBishop:
Yep.  It looks like for the ones that are building properly, kramdown-rfc2629 is downloading the references and embedding them in the XML.  This draft gets these at the beginning...  `<!ENTITY RFC7230 SYSTEM "https://xml2rfc.tools.ietf.org/public/rfc/bibxml/reference.RFC.7230.xml">` followed by `&RFC7230;` in the references section.

reschke:
Just opened 

  https://bugzilla.gnome.org/show_bug.cgi?id=791491

(this was actually reported to me a few days ago, but I didn't make the connection...)

reschke:
Apparently, adding

    stand_alone: yes

does the in-lining trick.

martinthomson:
Yes, why this isn't the default is baffling.

reschke:
Might make sense if you convert from MD to XML with the goal to continue editing the XML.

MikeBishop:
Great catch -- added a fix to use stand_alone on the xslt branch.  So, with a sample size of one so far, new-template took 32 seconds total on the most recent build, while xslt took 28 seconds.  Resulting output for [xslt](http://mikebishop.github.io/http-extensions/xslt/) and [xml2rfc](http://mikebishop.github.io/http-extensions/new-template/) is very similar.

While I do like the second better aesthetically, the primary goal here is maintainability and shared code; I can live with that.  Should I change the PR to be off the xslt branch instead?

martinthomson:
Styling the XSLT output is easier, but it would a different stack to run.  I'd say that since it will make some of the things Julian relies on easier, we should pick xslt.

That reminds me, I can add a cache for the xslt to the template and get further speed gains.  Maybe some other time.

Makefile
```diff
@@ -1,143 +1,11 @@
-xml2rfc ?= "xml2rfc"
-saxpath ?= "lib/saxon9.jar"
-saxon ?= java -classpath $(saxpath) net.sf.saxon.Transform -l
-kramdown2629 ?= XML_RESOURCE_ORG_PREFIX=https://xml2rfc.tools.ietf.org/public/rfc kramdown-rfc2629
+LIBDIR := lib
```
martinthomson: Add USE_XSLT := true here.

MikeBishop: Yes, that's the extra commit on the other branch.  :-)

MikeBishop:
I've merged the XSLT branch into this one, so the PR now uses `xsltproc` within Martin's I-D template.  I agree with Martin that this needs chair approval:  @mnot, @mcmanus?

Note too that a lot of the speed gains come from running in Martin's Docker container on Circle -- the Travis build times for this branch are already improved (90-120 seconds vs. 120-150 seconds), but not as good as Circle (30-40 seconds).  I've been testing my fork of this repo against Circle.

I don't believe I have sufficient access to switch the repo from Travis to Circle, and you can't do that in a PR anyway, so that's definitely a chair action.

mnot:
Switching over; please stand by...

MikeBishop:
(Note that Circle won't see anything to do on master until this PR is merged.)

mnot:
All green. Only used one container.

mnot:
Travis hook deleted.

mnot:
I did a push, seems to have worked. However, the HTML files are in different places from where they used to be -- we *can* update the links, but changing links is kind of uncool.

Also, I think `SUBMITTING.md` needs an update...

mnot:
Maybe instead of generating `draft-ietf-httpbis-*.html` we could generate `*.html`? That would reduce the number of links we need to change.

mnot:
Also - I strongly suspect that building issues.json on every push is going to burn through Github's rate limiting. Can we limit it, or at least assure that we recover from 429 gracefully?

martinthomson:
I have rate limiting on issues.json, though we still hit the rate cap occasionally.

We can add rules for generating `*.html`.  That's a fairly simple exercise.  It's something like:

```make
latest:: a.html b.html c.html
a.html: draft-ietf-httpbis-a.html 
        cp $< $@
# etc...
```

We might do better, but it's a little sketchy.

mnot:
That should work; it's just for the existing drafts, going forward we can use the "full" names.

mnot:
ping. Links still broken.

mnot:
I've updated SUBMITTING.md; please review.

mnot:
Hmm. I just commited bcp56bis-00, and it builds locally, and on Circle, but doesn't upload the drafts to `gh-pages`. Any idea why @martinthomson?

martinthomson:
Have you set a value for GH_TOKEN in the build configuration?

MikeBishop:
Re: the *\*ahem\**, see https://github.com/martinthomson/i-d-template/pull/119.  It wound up being easier to do inside the template as something that the parent repo could opt into, than to have the parent repo try to force extra things into gh-pages.

mnot:
Ah, that's right - GH_TOKEN set, thanks.

It looks like gh_issues is built even for pull requests; is there a way to only do it for master?

martinthomson:
Yes, though I don't much see the point, other than as another form of rate limiting.

You can check the branch in the step.  I think that the variable is CIRCLE_BRANCH.

PR #465, opened 2017-12-30
fix missing word

reschke:


craigpratt:
Note this was manually incorporated into draft-ietf-httpbis-rand-access-live-03

PR #466, opened 2018-01-03
It's not 2017

MikeBishop:
Uses {DATE} to let the date be auto-added at build time.


martinthomson:
I see that you removed trailing whitespace here.  Do you want to add a .editorconfig so that this is done more consistently?

MikeBishop:
Perhaps, but I did that as an incidental so the template would let me check in the change.  Let's address that separately.

PR #467, opened 2018-01-03
Shortnames in gh-pages

MikeBishop:
Uses the added feature in https://github.com/martinthomson/i-d-template/pull/129 to drop shortname copies of the drafts in gh-pages to fix the broken links.  These are actual copies, not symlinks, but git keeps those relatively cheap.

Shouldn't break anything to have in the Makefile without that change, either.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 LIBDIR := lib
 USE_XSLT := true
 DISABLE_RIBBON := true
+GHPAGES_EXTRA = $(foreach ext,.html .txt,$(addsuffix $(ext),$(foreach draft,$(drafts),$(shell echo $(draft) | sed -e 's/draft-ietf-httpbis-//'))))
+
 include $(LIBDIR)/main.mk
 
 $(LIBDIR)/main.mk:
@@ -11,3 +13,6 @@ else
 	git clone -q --depth 10 $(CLONE_ARGS) \
 	    -b master https://github.com/martinthomson/i-d-template $(LIBDIR)
 endif
+
+$(GHPAGES_EXTRA):
+	ln -sf draft-ietf-httpbis-$@ $@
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -11,3 +13,12 @@ else
...
 endif
+
+$(GHPAGES_EXTRA):
+	ln -sf draft-ietf-httpbis-$@ $@
+
+.PHONY: clean-extra
+clean:: clean-extra
+
+clean-extra:
+	-rm -f $(GHPAGES_EXTRA)
```
martinthomson: Just do this:

```make
clean::
        -rm -f $(GHPAGES_EXTRA)
```

Makefile
```diff
@@ -1,6 +1,8 @@
 LIBDIR := lib
 USE_XSLT := true
 DISABLE_RIBBON := true
+GHPAGES_EXTRA = $(foreach ext,.html .txt,$(addsuffix $(ext),$(foreach draft,$(drafts),$(shell echo $(draft) | sed -e 's/draft-ietf-httpbis-//'))))
```
martinthomson: Use `:=` or the expansion of GHPAGES_EXTRA will run the loop every time that `$(GHPAGES_EXTRA)` appears.

Makefile
```diff
@@ -11,3 +13,12 @@ else
 	git clone -q --depth 10 $(CLONE_ARGS) \
 	    -b master https://github.com/martinthomson/i-d-template $(LIBDIR)
 endif
+
+$(GHPAGES_EXTRA):
+	ln -sf draft-ietf-httpbis-$@ $@
```
martinthomson: Cunning.  No dependencies, but then it doesn't need any.

Makefile
MikeBishop: That's intentional, actually.  $(drafts) is undefined until after the import, at which point it's too late.  By delaying expansion, it *is* defined by the time GHPAGES_EXTRA is used.

martinthomson: Ugh, oh well.  There should only be two expansions, so that isn't so bad.

PR #469, opened 2018-01-12
Add SETTINGS_ prefix to parameter name.

bencebeky:
* Add SETTINGS_ prefix to parameter name ENABLE_CONNECT_PROTOCOL.
* Fix one typo ("s/a entry/an entry/").
* Remove "(type = 0x8)" which is ambigous (what is type?) and redundant
  with code listed in IANA Considerations section.


mcmanus:
thanks

PR #472, opened 2018-01-17
Fix the number of members in a list.

jyasskin:
The counts forgot to include the initial required member.

The parsing algorithm also doesn't enforce any bounds on the number of items, but I haven't fixed that here.


mnot:
Thanks, had to manually incorporate.

PR #481, opened 2018-02-05
Incorporate exported authenticator requests

MikeBishop:
Tracking changes from tlswg/tls-exported-authenticator#9 to create an exported authenticator request message at the TLS layer as well.

This is a straight port of https://github.com/MikeBishop/http2-certs/pull/18/; feedback there has already been addressed, modulo one issue I'll be adding to this repo.

### DIFF
--- a/draft-ietf-httpbis-http2-secondary-certs.md
+++ b/draft-ietf-httpbis-http2-secondary-certs.md
@@ -522,7 +522,7 @@ error of type `PROTOCOL_ERROR`.
 The referenced certificate chain needs to conform to the requirements expressed
 in the `CERTIFICATE_REQUEST` to the best of the sender's ability, or the
 recipient is likely to reject it as unsuitable despite properly validating the
-authenticator.  If the recipicent considers the certificate unsuitable, it MAY
+authenticator.  If the recipient considers the certificate unsuitable, it MAY
 at its discretion either return an error at the HTTP semantic layer, or respond
 with a stream error {{RFC7540}} on any stream where the certificate is used.
 {{errors}} defines certificate-related error codes which might be applicable.
--- a/draft-ietf-httpbis-http2-secondary-certs.md
+++ b/draft-ietf-httpbis-http2-secondary-certs.md
@@ -520,10 +520,7 @@ necessary frames have been received on stream zero MUST also result in a stream
 error of type `PROTOCOL_ERROR`.
 
 The referenced certificate chain MUST conform to the requirements expressed in
-the `CERTIFICATE_REQUEST` to the best of the sender's ability. Specifically, if
-the `CERTIFICATE_REQUEST` contained a non-empty `Cert-Extensions` element, the
-end-entity certificate MUST match with regard to the extensions recognized by
-the sender.
+the `CERTIFICATE_REQUEST` to the best of the sender's ability.
 
 If these requirements are not satisfied, the recipient MAY at its discretion
 either return an error at the HTTP semantic layer, or respond with a stream
@@ -532,11 +529,10 @@ defines certificate-related error codes which might be applicable.
 
 ## The CERTIFICATE_REQUEST Frame {#http-cert-request}
 
-TLS 1.3 defines the `CertificateRequest` message, which prompts the client to
-provide a certificate which conforms to certain properties specified by the
-server.  This draft defines the `CERTIFICATE_REQUEST` frame (0xFRAME-TBD2),
-which uses the same set of extensions to specify a desired certificate, but
-can be sent over any TLS version and can be sent by either peer.
+The `CERTIFICATE_REQUEST` frame (id=0xFRAME-TBD2) provides a exported
+authenticator request message from the TLS layer that specifies a desired
+certificate.  This describes the certificate the sender wishes to have
+presented.
 
 The `CERTIFICATE_REQUEST` frame SHOULD NOT be sent to a peer which has not
 advertised support for HTTP-layer certificate authentication.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -563,20 +557,22 @@ Request-ID:
...
-Some extensions used for certificate selection allow multiple values (e.g.
-oid_filters on Extended Key Usage). If the sender has included a non-empty
-Extensions list, the certificate MUST match all criteria specified by extensions
-the recipient recognizes. However, the recipient MUST ignore and skip any
-unrecognized certificate selection extensions.
+The Exported Authenticator `request` API defined in
+[I-D.ietf-tls-exported-authenticator] takes as input a set of desired
+certificate characteristics and a `certificate_request_context`. When generating
+exported authenticators for use with this extension, the
+`certificate_request_context` MUST be the two-octet Request-ID.
```
martinthomson: Exported authenticators require that the certificate_request_context be unpredictable.  I think that we can address this by prefixing a random sequence with the two-octet Request-ID.  Then we can ensure that we have a mapping key, but also enough entropy to avoid guessing.

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -640,16 +636,18 @@ received on any other stream MUST be rejected with a stream error of type
...
+takes as input a request, a set of certificates, and supporting information
+about the certificate (OCSP, SCT, etc.).
+
+Upon receipt of a completed authenticator, an endpoint MUST perform the
+following steps:
+ - Using the `get context` API, retrieve the `certificate_request_context` used
+   to generate the authenticator, if any.
+ - Verify that the `certificate_request_context` is the Request-ID of a
+   previously-sent `CERTIFICATE_REQUEST` frame.  Alternatively, on clients the
+   `certificate_request_context` MAY also be empty.
```
martinthomson: Verify that the authenticator has either an empty `certificate_request_context` (for clients only), or contains a Request-ID from a previously-sent `CERTIFICATE_REQUEST` frame.

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -640,16 +636,18 @@ received on any other stream MUST be rejected with a stream error of type
...
-generating exported authenticators for use with this extension, the
-`certificate_request_context` MUST be the two-octet Cert-ID.
-
-Upon receipt of a completed authenticator, an endpoint MUST check that:
-
- - the `validate` API confirms the validity of the authenticator itself
- - the `certificate_request_context` matches the Cert-ID of the frame(s) in
-   which it was received
+takes as input a request, a set of certificates, and supporting information
+about the certificate (OCSP, SCT, etc.).
```
martinthomson: This statement applies to the *creation* of an authenticator.  But the following paragraph talks about validation of one.  I think that you need to make that distinction a lot clearer.

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -520,10 +520,7 @@ necessary frames have been received on stream zero MUST also result in a stream
 error of type `PROTOCOL_ERROR`.
 
 The referenced certificate chain MUST conform to the requirements expressed in
-the `CERTIFICATE_REQUEST` to the best of the sender's ability. Specifically, if
-the `CERTIFICATE_REQUEST` contained a non-empty `Cert-Extensions` element, the
-end-entity certificate MUST match with regard to the extensions recognized by
-the sender.
+the `CERTIFICATE_REQUEST` to the best of the sender's ability.
```
martinthomson: This is really a requirement on exported authenticators.  Can we turn this into a statement of truth instead?  Or remove it entirely.

MikeBishop: I'm not sure that's true -- the exported authenticator functionality is responsible for communicating the requirements and the certificate, but it's not responsible for selecting the certificate to respond with.  I think it's a requirement on whomever does certificate selection, which I believe is the HTTP layer.

draft-ietf-httpbis-http2-secondary-certs.md
MikeBishop: Should we specify the length?

draft-ietf-httpbis-http2-secondary-certs.md
martinthomson: It's not really actionable though.  If your intent is to state an interoperability requirement (why else do you use MUST), what do we lose if someone violates that?  I mean, I know that NSS could try harder to comply, but we do have better things to do than add full support for oid_filter (just for example, which is really hard to implement in practice).

draft-ietf-httpbis-http2-secondary-certs.md
martinthomson: We probably need a minimum size to stop someone from inducing lots of requests until the right number comes up (for the stolen credentials they have).

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -637,15 +638,15 @@ received on any other stream MUST be rejected with a stream error of type
...
-following steps:
+Upon receipt of a `CERTIFICATE` frame, an endpoint MUST perform the following
+steps to validate the token it contains:
  - Using the `get context` API, retrieve the `certificate_request_context` used
    to generate the authenticator, if any.
- - Verify that the `certificate_request_context` is the Request-ID of a
-   previously-sent `CERTIFICATE_REQUEST` frame.  Alternatively, on clients the
-   `certificate_request_context` MAY also be empty.
+ - Verify that the `certificate_request_context` is either empty (clients only)
+   or the Request-ID of a previously-sent `CERTIFICATE_REQUEST` frame.
```
martinthomson: Or *contains* the Request-ID.

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -566,9 +566,10 @@ Request:
 
 The Exported Authenticator `request` API defined in
 [I-D.ietf-tls-exported-authenticator] takes as input a set of desired
-certificate characteristics and a `certificate_request_context`. When generating
-exported authenticators for use with this extension, the
-`certificate_request_context` MUST be the two-octet Request-ID.
+certificate characteristics and a `certificate_request_context`, which needs to
+be unpredictable. When generating exported authenticators for use with this
+extension, the `certificate_request_context` MUST be the two-octet Request-ID,
+followed by at least six random octets.
```
martinthomson: I don't think that we need to stipulate order.  Can we say that it contains both the two-octet Request-ID and at least 96 bits of additional entropy.   (64-bits total isn't enough to avoid the problem I was talking about, especially when you consider that there is a good reason to choose a sequential Request-ID, which reduces that to 48 bits).

BTW, those extra bits can be constructed using HMAC(secret, Request-ID), rather than randomness.

PR #506, opened 2018-03-12
Added description of caching limitations for Range-less shift-buffer requests

craigpratt:


draft-ietf-httpbis-rand-access-live.xml
```diff
@@ -463,6 +464,16 @@ Content-Range: bytes 1020000-1254567/*
       <t>
         from the server to indicate that the response will start at byte 1020000, end at byte 1254567, and will not include any aggregated content. This is the response expected from a typical HTTP server - one that doesn't support byte-range requests on aggregating content.
       </t>
+      <t>
+          Note that responses to GET requests against shift-buffer representations using Range 
+          can be cached by intermediaries, since the Content-Range response header indicates 
+          which portion of the representation is being returned in the response body. However 
+          GET requests without a Range header cannot be cached since the first byte of the response
```
martinthomson: either that or the resource provided might be identified differently

that is, if the start of `/live` moves by 10M, then you might redirect to `/live+10M`, all assuming that the redirect works (content-location isn't especially nice for this because it isn't any more cacheable

craigpratt: Interesting idea @martinthomson. I'll see if I can reword to not preclude this option.

So guess I need to sprinkle in "successful responses" on both of these sentences and maybe include redirect as an example.

I'll revise tonight. Thx!

PR #509, opened 2018-03-16
avoid markdown syntax in spec text examples

reschke:
(this change will be in conflict with the JSON reference update though)

### DIFF
--- a/draft-ietf-httpbis-bcp56bis.md
+++ b/draft-ietf-httpbis-bcp56bis.md
@@ -239,7 +239,7 @@ their task cannot be completed).
 
 For example, an application can specify that it uses HTTP like this:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-bcp56bis.md
```diff
@@ -239,7 +239,7 @@ their task cannot be completed).
 
 For example, an application can specify that it uses HTTP like this:
 
-    Foo Application uses HTTP {{!RFC7230}}. Implementations MUST support
+    Foo Application uses HTTP [RFC7230]. Implementations MUST support
```
martinthomson: If the intent is to avoid creating a reference, then you want `\[RFC7230]`.  Functionally, this isn't a change at all.

BTW, if the intent is to blockquote this section, it should use `> Foo Application [...]`.

Issue #520, opened 2018-03-18
"_" is invalid in ABNF rule names, also: rule names are case-insensitive


reschke:
...so we can't have both CHAR and char.

reschke:
FWIW, the ABNF below would work:

~~~

dictionary        = dictionary-member 0*1023( OWS "," OWS dictionary-member )
dictionary-member = identifier "=" item

list = list-member 0*1023( OWS "," OWS list-member )
list-member = item

param-list = param-id 0*255( OWS "," OWS param-id )
param-id   = identifier 0*256( OWS ";" OWS identifier [ "=" item ] )

item = integer / float / string / identifier / binary

integer   = ["-"] 1*19DIGIT

float    = ["-"] (
             DIGIT "." 1*14DIGIT /
            2DIGIT "." 1*13DIGIT /
            3DIGIT "." 1*12DIGIT /
            4DIGIT "." 1*11DIGIT /
            5DIGIT "." 1*10DIGIT /
            6DIGIT "." 1*9DIGIT /
            7DIGIT "." 1*8DIGIT /
            8DIGIT "." 1*7DIGIT /
            9DIGIT "." 1*6DIGIT /
           10DIGIT "." 1*5DIGIT /
           11DIGIT "." 1*4DIGIT /
           12DIGIT "." 1*3DIGIT /
           13DIGIT "." 1*2DIGIT /
           14DIGIT "." 1DIGIT )

string    = DQUOTE 0*1024(character) DQUOTE
character = unescaped / escape ( DQUOTE / "\" )
unescaped = %x20-21 / %x23-5B / %x5D-7E
escape    = "\"

identifier = lcalpha 0*255( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
lcalpha    = %x61-7A ; a-z

binary = "*" 0*21846(base64) "*"
base64 = ALPHA / DIGIT / "+" / "/" / "="
~~~

PR #521, opened 2018-03-19
fix ABNF (see #520)

reschke:
- underscores
- case insensitivity (char vs CHAR)
- also attempts to clarify ch(a)r / unescaped production a bit


mnot:
Thanks!

PR #555, opened 2018-04-03
Get and side effects fix

martinthomson:
This just fixes a typo, but this text has more problems.

I'm having trouble connecting this final sentence to the text that precedes it.  I think that the point is that including a body on GET is unwise because it might cause explosions.  A point about side effects needs far more exposition, especially since GET is defined to be free from overt side effects (safe, idempotent, and all that business).  If this were just another statement and the "As a result" is dropped, that's different.


mnot:
Barg, I think that's a copy/paste error, will fix. thx.

PR #559, opened 2018-04-03
replay: fix TLS section ref

reschke:
I believe Section 8 has moved to 9 (FWIW: it's dangerous to use section number references to drafts without specifying the draft number)


reschke:
oh - sorry for the confusion then

PR #573, opened 2018-04-09
Fixed a couple references to "Very Large values" lower case and "^^" to "^"

craigpratt:


craigpratt:
@martinthomson 

Thanks Martin,

Do you think some of the language from Section 5 needs to be mentioned/reiterated earlier?

Guess I could say something like "a value much larger than the expected maximum size of the representation should be provided by the client (see section 5 for security considerations)"?

Here's most of Section 5

`...in applications where there's no expectation that the representation will ever exceed 2^63, a value smaller than this value should be used as the very large last-byte-pos in a byte-seek request or content-range response. Also, some implementations (e.g. JavaScript-based clients and servers) are not able to represent all values beyond 2^53. So similarly, if there's no expectation that a representation will ever exceed 2^53 bytes, values smaller than this limit should be used for the last-byte-pos in byte-range requests.`

martinthomson:
That would be perfect.  Note that the security considerations here are more "compatibility considerations" than anything else.

PR #574, opened 2018-04-09
Clarify handling of invalid SameSite values (fixes #389)

### LINKED ISSUE #389: SameSite Cookie Attribute - Is attribute-value required?
https://bugs.chromium.org/p/chromium/issues/detail?id=635882

Section 3.1 of the spec states that "SameSite" alone (without an attribute value) is a valid token: https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00#section-3.1

Section 4.1 https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00#section-4.1 states 

   1.  If "cookie-av"'s "attribute-value" is not a case-insensitive
       match for "Strict" or "Lax", ignore the "cookie-av".

fmarier:
There is currently a discrepancy in the spec regarding invalid values
in the SameSite cookie attribute.

For example, this cookie:

  Set-Cookie: foo=bar; SameSite=bogus

is expected to be dropped entirely according to the "Server Requirements"
under Section 4.1.2.7:

  If the "SameSite" attribute's value is neither of these [ "Lax",
  "Strict" ], the cookie will be ignored.

whereas under Section 5.3.7 of "User Agent Requirements", the cookie
is to be kept but the attribute is ignored:

  If cookie-av's attribute-value is not a case-insensitive match for
  "Strict" or "Lax", ignore the "cookie-av".

Additionally, the end of Section 4.1.2 also matches the behavior
described in Section 5.3.7:

  User agents ignore unrecognized cookie attributes (but not the entire
  cookie).

From a forward-compatibility point of view, the behavior described in
section 5.3.7 is the ideal one since it allows for future expansion
of this feature such as:

  Set-Cookie: foo=bar; SameSite=medium

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -667,7 +667,7 @@ If the "SameSite" attribute's value is "Strict", the cookie will only be sent
 along with "same-site" requests. If the value is "Lax", the cookie will be sent
 with same-site requests, and with "cross-site" top-level navigations, as
 described in {{strict-lax}}. If the "SameSite" attribute's value is neither of
-these, the cookie will be ignored.
+these, the attribute will be ignored.
 
 ### Cookie Name Prefixes
 
@@ -2049,6 +2049,11 @@ Specification document:
 
    *   Fixed the "site for cookies" algorithm to return something that makes
        sense: <https://github.com/httpwg/http-extensions/issues/302>
+
+## draft-ietf-httpbis-rfc6265bis-03
+
+*  Clarified handling of invalid SameSite values:
+   <https://github.com/httpwg/http-extensions/issues/389>
 
 # Acknowledgements
 {:numbered="false"}


fmarier:
@mikewest

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -667,7 +667,7 @@ If the "SameSite" attribute's value is "Strict", the cookie will only be sent
 along with "same-site" requests. If the value is "Lax", the cookie will be sent
 with same-site requests, and with "cross-site" top-level navigations, as
 described in {{strict-lax}}. If the "SameSite" attribute's value is neither of
-these, the cookie will be ignored.
+these, the `cookie-av` will be ignored.
```
mikewest: ```s/`cookie-av`/attribute/```: `cookie-av` doesn't have any meaning until the algorithm in 5.3.

fmarier:
Done. Let me know if I missed anything else.

PR #575, opened 2018-04-09
Abstracts are small

### LINKED ISSUE #562: The abstract is really an introduction
5 paragraphs is probably too much.  We should check to see if anything needs to be salvaged and then shorten to:

A use of TLS exported authenticators is described that enables HTTP/2 clients and servers to offer additional certificate-based credentials after the connection is established.  The means by which these credentials are use with requests is defined.

MikeBishop:
Fixes #562.

### DIFF
--- a/draft-ietf-httpbis-http2-secondary-certs.md
+++ b/draft-ietf-httpbis-http2-secondary-certs.md
@@ -45,24 +45,10 @@ informative:
 
 --- abstract
 
-TLS provides fundamental mutual authentication services for HTTP, supporting up
-to one server certificate and up to one client certificate associated to the
-session to prove client and server identities as necessary. This draft provides
-mechanisms for providing additional such certificates at the HTTP layer when
-these constraints are not sufficient.
-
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -80,7 +66,7 @@ code and issues list for this draft can be found at
 
 HTTP clients need to know that the content they receive on a connection comes
 from the origin that they intended to retrieve in from. The traditional form of
-server authentication in HTTP has been in the form of X.509 certificates
+server authentication in HTTP has been in the form of the single X.509 certificate
```
martinthomson: s/the/a/ ?

MikeBishop: Perhaps; "the" when there's only one, and the point of the statement is that there's only one.  But maybe I'm over-emphasizing that?

martinthomson: Yeah, "the single" is redundant in that sense.

PR #576, opened 2018-04-10
Added a small blurb referencing client range-end considerations from section 5

craigpratt:
Also considering if "Security Considerations" can be "Compatibility Considerations".

### DIFF
--- a/draft-ietf-httpbis-rand-access-live.xml
+++ b/draft-ietf-httpbis-rand-access-live.xml
@@ -291,7 +291,7 @@ Range: bytes=1230000-999999999999
 
 </artwork></figure>
       <t>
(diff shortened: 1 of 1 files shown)


craigpratt:
I haven't renamed "Security Considerations" in this version. And I have no issue with calling this "Compatibility Considerations". But I have some recollection that this is a required section...

martinthomson:
Yeah don't do that or various people will start to scream at you :)

craigpratt:
Call me a perfectionist, but on reread, I realized there was a cleaner/shorter way to clarify. 

Let me know if you have any objection...

craigpratt:
Thanks Martin!

PR #585, opened 2018-04-13
replay: refer to actual header field registry

reschke:
....instead of RFC defining it.


wtarreau:
looks good to me. Thanks Julian.

PR #586, opened 2018-04-13
Stream-linking frames can reference Stream Zero

### LINKED ISSUE #564: Client probes for server certificates are awkward
In the current draft, the client ties up a stream for a pending request on an unproven domain so that it can send a `NEEDED` frame, permitting the server to respond with `USE`.  This is rather awkward.

However, the alternative is also somewhat unappealing.  In the process of updating #480, I'm adding a suggestion that servers SHOULD respond to `REQUEST` with an appropriate certificate if they have one, which makes this flow much cleaner.  However, there's current no way to *reject* a certificate request without a request stream.

MikeBishop:
Fixes #564.

If clients need a server certificate before making requests for an origin, they can send `CERTIFICATE_NEEDED` for stream zero, indicating the connection.  Servers respond with `USE_CERTIFICATE`.

Known drawbacks:
  - Since `USE_CERTIFICATE` has to be sent in order, if the client requests a batch of domains, the server can't provide the requested certificates out-of-order (though it can outright reject some of them).
  - This is fine for H2; in HTTP/QUIC, we'll need to decide if this is still zero, or if it's 3/4, or something else.


MikeBishop:
@kazuho, @martinthomson, I would appreciate it if you could give this a look.  Do you think this is a good approach?

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -367,19 +368,23 @@ the connection.
...
+then sends a `CERTIFICATE_NEEDED` frame for stream zero referencing the request,
+indicating that the connection cannot be used for that origin until the
+certificate is provided.
 
-If the server does not have the desired certificate, it MUST \[see issue #564].
-In this case, or if the server has not advertised support for HTTP-layer
-certificates, the client MUST NOT send any requests for resources in that origin
-on the current connection.
+If the server does not have the desired certificate, it MUST send an empty
+`USE_CERTIFICATE` frame for stream zero.  In this case, or if the server has not
```
martinthomson: When you say empty, I think that you mean that the server sends an empty CERTIFICATE frame and then references that certificate when it wants to reject a CERTIFICATE_NEEDED request.

That is:

```
Client Sends:
CERTIFICATE_REQUEST (id=N) -->
CERTIFICATE_NEEDED (S=0, R=M) -->
...
Server can't answer, so:
<-- CERTIFICATE (Cert-ID=Q, <empty>)
<-- USE_CERTIFICATE (S=0, R=M)
```

The problem here is that the empty CERTIFICATE frame isn't bound in any way to the CERTIFICATE_REQUEST, unless your definition of "empty" isn't that the authenticator is completely absent, but that it includes a valid authenticator, but no certificate in the TLS Certificate message.  Then the `certificate_request_context` can be used to correlate CERTIFICATE with CERTIFICATE_REQUEST in the usual fashion.  That's how TLS does it - it sends Certificate without the certificate and omits the CertificateVerify message entirely if it doesn't want to authenticate.  (I just realized that I used CERTIFICATE, Certificate and certificate in the same paragraph to mean three different things, which is not at all confusing.)

I just read exported authenticators again and I don't see that use case being addressed at all.  @grittygrease should be aware of that given that TLSEA is in WGLC.  I think that TLSEA needs to address that case one way or other.  If the intent is to have the using application manage rejection, then it needs to be clear where the responsibility lies.  However, I think that having authenticated rejection is superior and that should be expressly included in the draft with a description of how that is done.

MikeBishop: Actually, "empty" probably needs to be clarified here.  In the previous draft (i.e. where things were on-stream with the request), you could omit the CertID and send a USE_CERTIFICATE with a length of zero, which meant "I don't have / choose not to provide the requested certificate.  Deal."

If ExpAuth gives us an authenticated denial, I'd be fine using it.  But I'm also fine if it doesn't -- the point of ExpAuth is proving possession, and I think declining to provide an authenticator is sufficient for our purposes.  This isn't DNSsec where we want to authoritatively prove non-existence, we're just declining to supply a certificate (or, therefore, proof of possession).

MikeBishop:
Note that, while I have changed the text to indicate that an Empty Authenticator should be used when asked for a certificate you don't have, I haven't removed the text that says a USE_CERTIFICATE without a Cert-ID points to the cert provided in TLS.  Do you think that needs to go as well?

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -373,8 +373,11 @@ then sends a `CERTIFICATE_NEEDED` frame for stream zero referencing the request,
 indicating that the connection cannot be used for that origin until the
 certificate is provided.
 
-If the server does not have the desired certificate, it MUST send an empty
-`USE_CERTIFICATE` frame for stream zero.  In this case, or if the server has not
+If the server does not have the desired certificate, it MUST send an Empty
```
martinthomson: Why only one side?

MikeBishop: This section is talking specifically about clients probing servers.  But you're right -- there should be a corresponding statement about clients.

PR #589, opened 2018-04-17
bcp56bis: split message exchange into two figures

reschke:


mnot:
The point here is to illustrate an entire exchange, so I think it's one example.

reschke:
Yes, but it displays as a single piece of artwork which I find confusing. I don't believe we have a single spec in HTTPbis space where we made request and response part of a single piece of artwork.

mnot:
That make sense, because HTTPbis generally illustrates how to implement HTTP, not applications on top of it. I suspect that most applications are going to want to demonstrate complete request/response interactions, not messages in isolation. It's also better practice to show both IMO.

reschke:
Yes, both should be shown. This is about *how*.

PR #592, opened 2018-04-24
Document some of the properties of Early-Data

martinthomson:
According to RFC 7231, Section 8.3.1.

I didn't include anything here about Vary because I wanted to discuss that first and these seemed uncontroversial.  It seems OK to say that Vary is permitted here because we expressly talk about varied reactions based on the presence of Early-Data.  If that's right, I will amend this with a second commit.  (I'm jetlagged, so being a little cautious.)

### DIFF
--- a/draft-ietf-httpbis-replay.md
+++ b/draft-ietf-httpbis-replay.md
@@ -269,6 +269,7 @@ the `Early-Data` header field if it might have forwarded the request prior to
 handshake completion (see {{be-consistent}} for details).
 
 An intermediary MUST NOT remove this header field if it is present in a request.
+`Early-Data` MUST NOT appear in a `Connection` header field.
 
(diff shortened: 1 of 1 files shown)


wtarreau:
Your proposal looks very clear to me, despite your jetlag :-)  I agree with you that Vary should be OK. I don't even think we need to say anything about it then.

draft-ietf-httpbis-replay.md
```diff
@@ -282,6 +283,13 @@ marked with Early-Data was sent in early data on a previous hop. Requests that
 contain the Early-Data field and cannot be safely processed MUST be rejected
 using the 425 (Too Early) status code.
 
+The `Early-Data` header field carries a single bit of information and clients
+MUST include at most one instance.  However multiple instances MUST be treated
```
MikeBishop: Comma after "However", or just drop the word.

PR #602, opened 2018-05-03
Mention rejection of 0-RTT

martinthomson:
Our basic list of mitigations missed a fairly obvious one.

Mentioning it should help with the class of confusion Magnus N. had with the draft.

I decided not to include a note about the server being unable to examine early data before making this decision.  That's just something people will need to discover for themselves.  Generally, you have to decide whether you want 0-RTT without seeing any of it.  Partly this is because it avoids a potential deadlock, but mostly it's because the TLS stack will not even decrypt 0-RTT if it is rejected.


wtarreau:
Looks good to me, and indeed clearer. Thanks.

mcmanus:
+1

PR #603, opened 2018-05-03
Attempt to clarify the replay attack

martinthomson:
This was a little obtuse in the previous iteration, likely as a result of trying to be parsimonious.  Expand it a little and clarify the attack.


wtarreau:
It's indeed better like this in my opinion.

PR #616, opened 2018-05-08
Reword exclusion regarding retries

martinthomson:
@MikeBishop noted that automated retries can be initiated by other enties than the client, which might be without client knowledge.  That's a fair point and an opportunity to improve text more.

### DIFF
--- a/draft-ietf-httpbis-replay.md
+++ b/draft-ietf-httpbis-replay.md
@@ -68,8 +68,9 @@ be used to reduce the likelihood that an attacker can successfully replay a
 request, but these techniques can be difficult to deploy, and still leave some
 possibility of a successful attack.
 
-Note that this is different from automated or user-initiated retries; replays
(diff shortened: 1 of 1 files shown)


MikeBishop:
You're still glossing over the fact that retries can be initiated by something other than the client, but this focuses specifically on client-initiated actions, so that's probably okay.

martinthomson:
The glossing-over is intentional. :)

Getting into what an intermediary might do here is what the rest of the document covers.  The primary point here is to separate replay from intentional retries.

wtarreau:
I'm not seeing the benefit of the rewording here. How about this instead ?

The difference between replay attacks and retries is that retries may be performed by a user, a client or an intermediary for the purpose of getting a response that will be delivered to the client, while a replay attack is generally performed by a third party not part of the chain from the client to the server, which is not interested in receiving a response, will often not receive it, and not even know whether or not such a response was provided or not.

PR #622, opened 2018-05-12
Treat invalid header fields conservatively

martinthomson:
As discussed on the list.


wtarreau:
looks good, OK for me. Thanks Martin!

PR #625, opened 2018-05-22
expect-ct: tune examples

reschke:
...avoid too wide artwork by changing the second example to use multiple field instances (which, in itself, is good to have in the examples anyway)


estark37:
Thanks!

PR #626, opened 2018-05-23
Expect-CT: update text about http-equiv

estark37:


estark37:
@reschke @mnot thanks for the explanation/suggestion on the list. Does this look about right to you?

mnot:
LGTM.

reschke:
- avoid lowercase "should" unless you refer to [RFC8174] in "Terminology"

FWIW, I still think that citing both HTML specs does not help the reader, and that the whole section isn't needed. If it was needed, why don't we say the same thing in draft-ietf-httpbis-replay-03?

estark37:
I don't have any strong feelings on this, I'm happy to just remove it. I cribbed it from HPKP originally. (https://tools.ietf.org/html/rfc7469#section-2.3.4)

estark37:
This is removed now.

reschke:
LGTM, and thanks for pointing to RFC 7469. @mnot - should we raise an erratum for that spec?

reschke:
FWIW: https://www.rfc-editor.org/errata/eid5377

PR #645, opened 2018-06-05
Ben's comments

martinthomson:
@kaduk had a few comments in the review of this that are easy to solve.


wtarreau:
All changes are OK to me as well. Thanks!

PR #648, opened 2018-06-07
Spencer's q

martinthomson:
This will conflict with the changes ekr suggested, but I integrated them, so merging shouldn't lose the important changes.


wtarreau:
For "sent in early data on another connection", I think Spencer's proposal "sent in early data on a previous hop" would make it even clearer regarding the hop chain so as not to confuse this with another connection from the same user-agent. OK for the rest.

PR #651, opened 2018-06-12
Man-in-the-middle protection via SETTINGS

### LINKED ISSUE #617: MitM detection
Secondary certs depends on there not being a MitM.  And sadly, they still exist in some places.

Now, as an advisory extra, the failure mode isn't fatal to the connection, but it could mean a lot of wasted effort.  So we could pack the setting with 32 bits extracted from the exporter as confirmation that the mechanism works, or design a new mechanism to reduce the chance that we waste effort on creating authenticators that we can't use.

Either way, it's probably worth mentioning this possibility.

MikeBishop:
Fixes #617.

Specifies an exporter-derived value in SETTINGS to prove that the TLS connection is between the same endpoints as the HTTP connection.  If there is a TLS-terminating proxy intermediating the connection, the expected value will not match the received value, and the work of producing exported authenticators can be skipped.

One open question is around the use of early exporters.  I think this is probably something we want, so that servers can send additional certificates in their 0.5-RTT flight to 0-RTT clients.  However, should clients be required to supply a value derived from a regular exporter after the handshake finishes?  This is a minor pain in H2, but in HTTP/QUIC would require defining a separate frame type to carry the updated token.  The obvious value is that the server could still validate the token after forgetting the 0-RTT secret, but it seems like the server could as easily validate the setting early and remember the result even after discarding the secret.  Is there any additional security benefit to the client re-proving the connection is end-to-end after the handshake completes?

### DIFF
--- a/draft-ietf-httpbis-http2-secondary-certs.md
+++ b/draft-ietf-httpbis-http2-secondary-certs.md
@@ -317,8 +317,8 @@ In order to ensure that the TLS connection is direct to the server, rather than
 via a TLS-terminating proxy, each side will separately compute and confirm the
 value of this setting.  The setting is derived from a TLS exporter (see Section
 7.5 of [I-D.ietf-tls-tls13] and {{?RFC5705}} for more details on exporters).
-Clients MAY use an early exporter during their 0-RTT flight, but MUST send an
-updated SETTINGS frame using a regular exporter after the TLS handshake
+Clients MUST NOT use an early exporter during their 0-RTT flight, but MUST send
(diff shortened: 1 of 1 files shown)


MikeBishop:
Without early data, the client can't include this in the initial SETTINGS frame, which means servers don't know whether the client supports things when it decides whether to supply certificates on the 0.5-RTT flight.  I'm not sure how important that is, though.

martinthomson:
In h2 at least, we can defer enabling the feature. And if neither side can provide certificates until after the handshake, that is ok.

That isn't an answer for hq. I don't know how I feel about that.

PR #653, opened 2018-06-14
Clients use Empty Authenticator as well

MikeBishop:
Follow-on to #586; in the process of fixing this, I noticed that I'd added a "server" header but missed adding a corresponding "client" header, so the client discussion appeared in the server section.


MikeBishop:
For some reason, that seems to be endemic in this doc.  Let's do that cleanup separately.

PR #656, opened 2018-06-27
Accept-CH+Lifetime opt-in is for same-origin resources

igrigorik:
Per discussions in fetch#726 [1], update the opt-in example to apply to
same-origin resources only, with Accept-CH-Lifetime being processed on
navigation requests.

[1] https://github.com/whatwg/fetch/issues/726

/cc @yoavweiss @tarunban

### DIFF
--- a/draft-ietf-httpbis-client-hints.md
+++ b/draft-ietf-httpbis-client-hints.md
@@ -100,7 +100,7 @@ A Client Hint request header field is a HTTP header field that is used by HTTP c
 
 ## Sending Client Hints
 
-Clients control which Client Hints are sent in requests, based on their default settings, user configuration, and server preferences. The client and server can use an opt-in mechanism outlined below to negotiate which fields should be sent to allow for efficient content adaption, and optinally use additional mechanisms to negotiate delegation policies that control access of third parties to same fields.
+Clients control which Client Hints are sent in requests, based on their default settings, user configuration, and server preferences. The client and server can use an opt-in mechanism outlined below to negotiate which fields should be sent to allow for efficient content adaption, and optionally use additional mechanisms to negotiate delegation policies that control access of third parties to same fields.
 
 Implementers should be aware of the passive fingerprinting implications when implementing support for Client Hints, and follow the considerations outlined in "Security Considerations" section of this document.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-client-hints.md
```diff
@@ -148,7 +148,7 @@ The preference MUST be delivered over a secure transport, and MUST NOT be persis
   Accept-CH-Lifetime: 86400
 ~~~
 
-For example, based on the Accept-CH and Accept-CH-Lifetime example above, which is received from "https://bar.example.com" in response to a resource request initiated by "https://foo.example.com", both delivered over a secure transport: a user agent SHOULD persist an Accept-CH preference bound to "https://foo.example.com", for requests initiated to "https://bar.example.com" from "https://foo.example.com", for up to 86400 seconds (1 day). This preference SHOULD NOT extend to requests initiated to "https://bar.example.com" from other origins.
+For example, based on the Accept-CH and Accept-CH-Lifetime example above, which is received in response to a user agent navigating to "https://example.com", and delivered over a secure transport: a user agent SHOULD persist an Accept-CH preference bound to "https://example.com" for up to 86400 seconds (1 day), and use it for user agent navigations to "https://example.com" and any same-origin resource requests initiated by the page constructed from the navigation's response. This preference SHOULD NOT extend to resource requests initiated to "https://example.com" from other origins.
```
yoavweiss: I think this example works, but would be good to also include the same rules in the processing definition above. e.g. s/When a client receives an HTTP response/When a client receives an HTTP navigation response, which is the same origin as the request which initiated the navigation/

Would such a language work for non-browser scenarios?

igrigorik: > Would such a language work for non-browser scenarios?

My hunch is "no" and we shouldn't impose this restriction here / scope it to browsers only. UA's have a concept of user-initiated "navigation", but this doesn't extend to other potential clients — e.g. a native app using CH negotiation with a CDN should be able to define own logic for how to process the opt-in; FP doesn't apply for other clients.

My suggestion is to keep the text as is, otherwise we'll quickly find ourselves redefining HTML concepts, which we should instead defer to the HTML spec.

WDYT?

yoavweiss: OK, so the normative text is for all clients and the examples is UAs only? I guess that makes sense.

igrigorik: Yep, that's how we've structured it so far.

That aside, good to merge?

yoavweiss: yup :)

PR #661, opened 2018-06-29
Reword as discussed on list

martinthomson:
This moves to SHOULD from MAY for automatic retry and shift the subject of the tricky sentence about potential replays at an intermediary.


wtarreau:
OK to me, thanks!

PR #665, opened 2018-07-13
Fix VCHAR references in draft-ietf-httpbis-header-structure.md

### LINKED ISSUE #664: draft-ietf-httpbis-header-structure.md#parse-string incorrectly references VCHAR
#parse-string Step 4.4 equates "the range %x00-1f or %x7f" to "not in VCHAR", but VCHAR doesn't include %x20.

#ser-string Step 0 says to fail serialization if the input "contains characters outside the range allowed by VCHAR or SP".

twifkak:
Fix #664. Step 4.4 equates "the range %x00-1f or %x7f" to "not in VCHAR", but VCHAR doesn't include %x20; this change adds "or SP" and imports SP from RFC5234.


mnot:
Thanks!

PR #667, opened 2018-07-14
Disallow trailing whitespace in #parse-param-list

twifkak:
This corresponds with the ABNF in the #param section.


twifkak:
@mnot

mnot:
Sorry for the wait.

I think we should be forgiving about whitespace around commas and at the end of headers, since there will be HTTP implementations (outside the control of this spec) that add them / combine them, etc.

twifkak:
Thanks for the feedback! (No worries re: delay.) Should the ABNF in #param be updated to match, or are you okay with the difference between #param and #parse-param-list? Perhaps the difference should be documented as intentional?

mnot:
I've noted that the parsing algorithms can be "greedy" regarding trailing whitespace. As the notational conventions section mentions, the ABNF is the "ideal" form, but the algorithms define conformance.

Issue #669, opened 2018-07-16
Ambiguity in 4.1.1 Structured Headers Draft v07


sharma95:
Section 4.1.1 (Serialising a Dictionary) seems inaccurate. It makes no mention of where the delimiter between successive key-value pairs (a comma) should be placed when serializing a dictionary, and how much whitespace to add before/after the delimiter

PR #672, opened 2018-07-26
[SH] add commas to dict,plist algorithms

phluid61:
Serialisation algorithms for dictionary and parameterised-list didn't include commas (?)

This copies the equivalent steps from the `list` algorithm


phluid61:
Just noticed that this is #669

mnot:
Thx.

PR #675, opened 2018-07-31
[SH] make 'parse number from text' less greedy

### LINKED ISSUE #668: Ambiguity in 4.2.6, Structured Headers Draft v07
Section 4.2.6 (Parsing a Number from Text) is ambiguous. It makes no mention of the exit condition from the while loop in step 7 on parsing a number from an input buffer.

phluid61:
Makes "Parse a Number from Text" not be greedy, and return (not fail) when it encounters a non-digit character.

Resolves #668


phluid61:
I'm pretty sure it works.

* Steps 5 & 6 ensure there's always at least one digit.
* All the calling contexts of "Parsing an Item from Text" (including the top-level algorithm in §4.2) inspect the next character, and handle it appropriately.

mnot:
SGTM

PR #676, opened 2018-08-03
[SH] make Serialising an Item return a value

phluid61:
Serialised `value` wasn't getting appended to `output`


mnot:
Would it be better to just have each of the steps before return their result?

phluid61:
I did wonder that when writing it.  Or make them all `s/let value/let output`.  Or however you think looks best.

I probably should have just filed it as an Issue. :thinking:

mnot:
👍

PR #677, opened 2018-08-03
[SH] multiply int by sign before testing range

phluid61:
When parsing a number, multiply `sign` back in to `output_number`
*before* testing against the range defined in Section 3.5.


mnot:
Thanks!

PR #678, opened 2018-08-03
[SH] remove rogue 'Otherwise' from string parser

phluid61:


draft-ietf-httpbis-header-structure.md
```diff
@@ -663,7 +663,7 @@ Given an ASCII string input_string, return an unquoted string. input_string is m
    3. Else, if char is DQUOTE, return output_string.
    4. Else, if char is in the range %x00-1f or %x7f (i.e., is not in VCHAR or SP), fail parsing.
    4. Else, append char to output_string.
-6. Otherwise, fail parsing.
+6. Unterminated string; fail parsing.
```
mnot: Maybe make this a tiny bit less terse -- e.g., "Fail parsing (due to the string being unterminated)."?

phluid61:
How's that?  It looks kind of like the equivalent case in Parse Dict/Parse List, but more descriptive than it was.

PR #690, opened 2018-08-22
Adding a note about 303 See Other

evert:
I noticed in my testing that 303 See Other works pretty much everywhere. A very common case for application developers is to send browsers a redirect after a POST. Almost every framework will default to using 302 for this, but really they should be using 303.

Let me know if you don't like the wording, happy to make edits.

### DIFF
--- a/draft-ietf-httpbis-bcp56bis.md
+++ b/draft-ietf-httpbis-bcp56bis.md
@@ -558,6 +558,11 @@ This table summarises their relationships:
 | Allows changing the request method from POST to GET | 301       | 302       |
 | Does not allow changing the request method          | 308       | 307       |
 
+A common use-case for redirects that change the request method to GET, is communicate to a browser
+that the result of an operation can be found at a different location. The appropriate status code
+for this is 303 See Other. This status code is widely supported and the recommended status code
+for these purposes.
+
 As noted in {{?I-D.ietf-httpbis-semantics}}, a user agent is allowed to automatically follow a 3xx
 redirect that has a Location response header field, even if they don't understand the semantics of
 the specific status code. However, they aren't required to do so; therefore, if an application
--- a/draft-ietf-httpbis-bcp56bis.md
+++ b/draft-ietf-httpbis-bcp56bis.md
@@ -558,10 +558,10 @@ This table summarises their relationships:
 | Allows changing the request method from POST to GET | 301       | 302       |
 | Does not allow changing the request method          | 308       | 307       |
 
-A common use-case for redirects that change the request method to GET, is communicate to a browser
+A common use-case for redirects that change the request method to GET is to communicate to the browser
 that the result of an operation can be found at a different location. The appropriate status code
-for this is 303 See Other. This status code is widely supported and the recommended status code
-for these purposes.
+for this is 303 See Other. This status code is widely supported and is the recommended status code
+for such purposes.
 
 As noted in {{?I-D.ietf-httpbis-semantics}}, a user agent is allowed to automatically follow a 3xx
 redirect that has a Location response header field, even if they don't understand the semantics of


draft-ietf-httpbis-bcp56bis.md
```diff
@@ -545,6 +545,11 @@ This table summarises their relationships:
 | Allows changing the request method from POST to GET | 301       | 302       |
 | Does not allow changing the request method          | 308       | 307       |
 
+A common use-case for redirects that change the request method to GET, is communicate to a browser
```
ericlaw1979:
```suggestion
A common use-case for redirects that change the request method to GET is to communicate to the browser
```

draft-ietf-httpbis-bcp56bis.md
```diff
@@ -545,6 +545,11 @@ This table summarises their relationships:
...
+that the result of an operation can be found at a different location. The appropriate status code
+for this is 303 See Other. This status code is widely supported and the recommended status code
```
ericlaw1979:
```suggestion
for this is 303 See Other. This status code is widely supported and is the recommended status code
```

draft-ietf-httpbis-bcp56bis.md
```diff
@@ -545,6 +545,11 @@ This table summarises their relationships:
...
+for these purposes.
```
ericlaw1979:
```suggestion
for such purposes.
```

ioggstream:
@evert reading https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html I have doubts about that interpretation.

Do you still propose this change? cc: @mnot

reschke:
I believe the proposed text is correct.

ioggstream:
Does it apply only to POST -> GET or to a generic redirect from an once-gone resource to a new one?

reschke:
"This status code is applicable to any HTTP method. It is primarily used to allow the output of a POST action to redirect the user agent to a selected resource, since doing so provides the information corresponding to the POST response in a form that can be separately identified, bookmarked, and cached, independent of the original request."

Not sure where you found "once-gone"?

evert:
Time flies! If there's still interest in this I can fix the conflict

ioggstream:
@reschke let me clarify. In this old  "Rest ws cookbook" https://github.com/tain335/tain335/blob/master/books/RESTful%20Web%20Services%20Cookbook.en%E5%AE%8C%E5%85%A8%E7%89%88.pdf 

There are various examples of 303, eg for implementing asynchronous flows.
I was guessing whether those recipes are correct or not and - in case they are not formally incorrect - they can be considered best practices. iiuc the 303 payload should only contain information about the "Other" resource and not eg. processing data.

Which is your opinion on this and on the other examples in the book? Thx++, R.

cc: @mfortini

```
After the server successfully completes image processing, it can redirect the client to
the result. In this example, the result is a new image resource:

# Request
GET /images/task/1 HTTP/1.1
Host: www.example.org

# Response
HTTP/1.1 303 See Other
Location: http://www.example.org/images/1
Content-Location: http://www.example.org/images/task/1

<status xmlns:atom="http://www.w3.org/2005/Atom">
 <state>done</state>
 <atom:link href="http://www.example.org/images/task/1" rel="self"/>
 <message xml:lang="en">Your request has been processed.</message>
</status> 
```

reschke:
I don't see how the suggested text and the example are in conflict. Please elaborate...

ioggstream:
@reschke the text suggested by @evert is aligned with the example. 

it is unclear to me whether the provided example is consistent with [semantics](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.15.4.4.p.4) because the example payload contains not only the link, but further information too (Eg. `<state>done</state>` ) 

In case it is consistent, is it ok to return a generic content-type with a 303 See Other ? eg.

```
# Request
GET /images/task/1 HTTP/1.1
Host: www.example.org

# Response
HTTP/1.1 303 See Other
Location: https://www.example.org/images/1
Content-Location: https://www.example.org/images/task/1

{
  "status": "ok",
  "href":   
}


```

reschke:
So we're talking about:

> Except for responses to a HEAD request, the representation of a 303 response ought to contain a short hypertext note with a hyperlink to the same URI reference provided in the Location header field.

right?

First of all, this is only "ought to". Not sure what exactly you mean by "generic" content type, but I agree that JSON is not hypertext. I don't see how the text can be read that it's not allowed to return additional information besides the link.

ioggstream:
>> the representation of a 303 response ought to contain a short hypertext
> right?

yes, that was the point.

> this is only "ought to"

Yes, but I was guessing whether the suggested implementations could be considered best practices or not.

>I don't see how the text can be read that it's not allowed

Since the spec says that the payload [ought to (RFC 6919)](https://tools.ietf.org/html/rfc6919#section-4) convey just an hypertext with a link, implementeres might tend to just ignore the payload and just get the `Location`.

reschke:
First of all, it's "ought to", not what you linked to.

Second, the spec doesn't say "just an hyperlink".

ioggstream:
> First of all, it's "ought to", not what you linked to.
yes, it was a joke :) 


> Second, the spec doesn't say "just an hyperlink".
true, but I think that `ought` +  `a short hypertext note with a hyperlink ` is not a good background  for an interoperable behavior,

If it's ok to provide more information in the 303 payload, it could have sense to explicit that.

mnot:
Hey @evert - sorry for the delay. Would you mind rebasing?

mnot:
See 5059b7cc.

Issue #199, opened 2016-06-14
Change definition of unique cookie to include host-only-flag to match browser behavior


macoxms:
RFC 6265 section 5.3 defines a cookie by the name, domain, and path.  However, given these two headers in a response from a request to http://contoso.com/:

Set-Cookie: mycookie=nothostonly; domain=contoso.com
Set-Cookie: mycookie=hostonly

Most browsers will have two cookies since they include host-only-flag in their definition of a unique cookie.  I'd like to update section 4.1.2 and 5.3 to reflect the behavior of modern browsers with regards to host-only-flag.

mnot:
Discussed in Berlin; seems reasonable.

PR #693, opened 2018-08-27
RFC6265bis: Include 'host-only-flag' when calculating uniqueness.

mikewest:
Closes https://github.com/httpwg/http-extensions/issues/199

@johnwilander, mind taking a look?

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1456,12 +1456,13 @@ user agent MUST process the cookie as follows:
     3.  The cookie-attribute-list contains an attribute with an attribute-name
         of "Path", and the cookie's path is `/`.
 
-17. If the cookie store contains a cookie with the same name, domain, and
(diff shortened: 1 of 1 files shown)


mikewest:
(@mnot, would you mind adding @johnwilander to the repo as a collaborator so I can ask him to review via the GitHub tool? :) )

mikewest:
Ping.

mnot:
Invite was sent a while back. Still listed as "pending."

mikewest:
> I probably need to read more of the context but would it be useful to point out that this is an internal flag for the user agent? It is, right? Inferred from whether the setting of the cookie included the domain or not.

It is an internal flag, but I don't think the PR needs additional documentation. The change we're making is in the middle of an algorithm, and `host-only-flag` is basically a variable that was set earlier in the process. I think folks reading the document will understand that. :)

mikewest:
Thanks for the review! I'll merge this, and look forward to your PR adding yourself as an editor.

PR #696, opened 2018-09-24
Define policy-controlled features

eeeps:
This PR is part of an attempt to address https://github.com/WICG/feature-policy/issues/129 – using Feature Policy to let authors opt-into sending specific Client Hints to specific origins.

The overall, basic idea is:

1. A new set of policy-controlled features – one for each Client Hint header field – is defined in the Client Hints spec. Those are defined in this PR.
2. A new algorithm which determines whether a request is able to use a policy-controlled feature is [defined in the Feature Policy spec](https://github.com/WICG/feature-policy/pull/220).
3. The step in Fetch that appends Client Hint headers [calls the new algorithm](https://github.com/whatwg/fetch/pull/811), checking if a request is allowed to use a Client Hints’ policy-controlled feature before appending the Client Hint header.

### DIFF
--- a/draft-ietf-httpbis-client-hints.md
+++ b/draft-ietf-httpbis-client-hints.md
@@ -255,6 +255,27 @@ Implementers ought to consider both user and server controlled mechanisms and po
 
 Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persisted opt-in preferences when site data, browsing history, browsing cache, or similar, are cleared.
 
+## Opt-in via Feature Policy
+
+Implementers that implement [Fetch](http://fetch.spec.whatwg.org) SHOULD provide authors the ability to opt-into sending specific Client Hints to specific origins using [Feature Policy](https://wicg.github.io/feature-policy/).
+
+This document defines [policy-controlled features](https://wicg.github.io/feature-policy/#features) for the following Client Hints header fields:
+
+| Client Hints header field | policy-controlled feature | [default allowlist](https://wicg.github.io/feature-policy/#default-allowlists) |
+|--|--|--|
+| `DPR` | `ch-dpr` | `'self'` |
+| `Width` | `ch-width` | `'self'` |
+| `Viewport-Width` | `ch-viewport-width` | `'self'` |
+
+These feature policies are set on the [client](https://fetch.spec.whatwg.org/#concept-request-client), and enforced on requests made by that client.
+
+For example, after receiving a document with the following [policy directives](https://wicg.github.io/feature-policy/#policy-directives)
+
+~~~ example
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-client-hints.md
```diff
@@ -255,6 +255,27 @@ Implementers ought to consider both user and server controlled mechanisms and po
 
 Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persisted opt-in preferences when site data, browsing history, browsing cache, or similar, are cleared.
 
+## Opt-in via Feature Policy
```
igrigorik: @eeeps this entire section should live in Feature Policy spec, not here. Could you move it over to your FP PR?

eeeps: I'm not sure it belongs in the Feature Policy spec. The Feature Policy spec only defines a generic set of mechanisms, and is not in the business of defining (or even, to date, [enumerating](https://github.com/WICG/feature-policy/issues/244)) specific policy-controlled features like this.

Can you explain why you don't think these feature policy definitions belong in the Client Hints spec?

Perhaps there's somewhere in the HTML or Fetch specs that'd be a better home? 

cc: @yoavweiss

yoavweiss: I think you're both right! :)

Cross-origin fetches is a web concept, so it has no place here. At the same time, FP does indeed just define the mechanisms, not the features.

HTML and Fetch are probably the right places for this, where the features are defined in HTML and initialize an object keeping track of the Document's CH delegation to certain third-parties, which is later used in Fetch to actually make a decision on sending the hints themselves using it.

eeeps: Do you think this policy-controlled feature definition belongs in (or is blocked by?) https://github.com/whatwg/html/pull/3774, or should I create a new PR on the HTML spec?

mnot: If we're going to make the CH spec generic (which still needs to be properly explained/discussed, btw :), I'd think that the CH spec should have a mention of FP and some general guidance, whereas the specific feature definitions / guidance should go wherever those particular hints are defined. YMMV, of course.

igrigorik: Fair enough. We could mention FP as an example of how delegation can happen for a browser use case. For other environments.. they may or may not use a similar pattern. 

@mnot does that sound reasonable to you?

martinthomson: If feature policy were a done deal, that might be OK, but it's still very unstable.

eeeps:
@yoavweiss is adding these definitions to the HTML spec here: https://github.com/whatwg/html/pull/3774/commits/335f566e0162a8fdb71e0acd513cff148e32eb53

@mnot @igrigorik Maybe spin up a separate issue to discuss the level which delegation mechanisms should be explained/exampled in the CH spec?

PR #709, opened 2018-10-23
Required Primary Domain

### LINKED ISSUE #522: Issues with server detection of attacks
ORIGIN removed the need for DNS lookups for servers, and IP connectivity.  Servers can look for bad DNS records or BGP hijacking to detect attacks without ORIGIN, but now those techniques don't work.  In particular, key compromise is a concern; misissuance is perhaps less of a concern.

Secondary certificates makes the entire situation worse.

In particular, it means that you can get a standalone certificate for the victim.  Previously, with coalescing, you would need to have a certificate that has both your own origin and the victim's origin, so that would be visible in a CT log.

Subcerts can help with this (they have much shorter lifetimes).

MikeBishop:
Fixes #522.

Defines ~~two~~ certificate extension:
- ~~**Affiliated Domain** effectively operates as a tag for all certificates managed by an entity (pick a domain you control), but doesn't make the certificate authoritative for that domain.~~
- **Required Primary Domain** says that you can only use *this* cert as a Secondary Certificate if you've already presented a cert that says you control the indicated domain.

In the simple case, you'd set this to a domain you include on all your certificates, "coalesce.example.com", and any of your certificates can be coalesced with any of your other certificates.  A CDN which requests certs on behalf of its customers can do the same thing, adding an extra domain owned by the CDN to each certificate.

In the more complex case, customers create certificates with RPD="coalesce.cdn.com".  Now any customer's certificate can be coalesced onto any connection to the CDN *provided that* the CDN first presents a certificate for "coalesce.cdn.com" where RPD is either the primary certificate or "*".

(One comment which has been raised in offline discussion:  Some origins might not want the CDN name human-readable in the certificate.  CDNs could address this by using a less-branded domain, but we could also address this in the extension design by using hashes of domain names if desired.)


draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -753,6 +753,80 @@ any stream which references the failing certificate in question or process the
 requests as unauthenticated and provide error information at the HTTP semantic
 layer.
 
+# Certificate Extensions {#extension}
+
```
martinthomson: Don't leave this bit empty.  A brief summary of the options might help.

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -753,6 +753,80 @@ any stream which references the failing certificate in question or process the
...
+verified by the CA.  Conforming CAs SHOULD mark the allowedPrimaryDomain
+extension as non-critical.
+
+Each allowed primary domain is represented as an IA5String. The name MUST be in
+the "preferred name syntax", as specified by Section 3.5 of {{!RFC1034}} and as
+modified by Section 2.1 of {{!RFC1123}}.  Note that while uppercase and
+lowercase letters are allowed in domain names, no significance is attached to
+the case. In addition, while the string " " is a legal domain name,
+affiliatedDomain extensions MUST NOT contain " ".  Rules for encoding
+internationalized domain names are specified in Section 7.2 of {{!RFC5280}}.
```
martinthomson: Does this crib from text elsewhere?  Can that text be included by reference instead?  The reason I ask is that this entire section seems to be copied from something, and I don't know if it is the same as that other thing, or subtly different.

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -753,6 +753,80 @@ any stream which references the failing certificate in question or process the
...
+
+If the affiliatedDomain extension is present, the sequence MUST contain at least
+one entry.  Unlike the subject field, conforming CAs MUST NOT issue certificates
+with an affiliatedDomain extension containing empty GeneralName fields.  Clients
+that encounter such a certificate when processing a certification path MUST
+ignore that extension.
```
martinthomson: I know that the effect of tolerance here is disabling the feature, but do we really need to mandate this particular form of tolerance?  Wouldn't it be better to consider the certificate invalid?  That is, if the extension is present and understood, it is an error for it to violate the rules.  That way we avoid weirdnesses accumulating.

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -753,6 +753,80 @@ any stream which references the failing certificate in question or process the
 requests as unauthenticated and provide error information at the HTTP semantic
 layer.
 
+# Certificate Extensions {#extension}
+
+## Allowed Primary Domain Certificate Extension {#extension-primary}
+
+The Allowed Primary Domain extension allows certificates to limit their use with
+Secondary Certificate Authentication.  The identities listed in this extension
```
martinthomson: This doesn't say *how* the use is limited.

I assume that this means that the certificate won't be accepted as a secondary certificate unless the certificate offered in the TLS handshake contains one of the identified names  in subjectAltName?

The name of the extension implies that there is only one, but this appears to allow many-to-many matching, which might be a bit gross to implement.

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -753,6 +753,80 @@ any stream which references the failing certificate in question or process the
...
+If the affiliatedDomain extension is present, the sequence MUST contain at least
+one entry.  Unlike the subject field, conforming CAs MUST NOT issue certificates
+with an affiliatedDomain extension containing empty GeneralName fields.  Clients
+that encounter such a certificate when processing a certification path MUST
+ignore that extension.
+
+The wildcard character "*" MAY be used to represent that any set of subdomains
+is acceptable.  This character MUST be the entirety of the first label if used,
+and MUST NOT occur elsewhere.  (That is, "*.example.com" and "*" are acceptable,
+but "www.*.com" and "w*.example.com" are not).
```
martinthomson: If the primary connection is to foo.example.com and the primary certificate contains *.example.com, does this extension match if it includes bar.example.com?  I can't tell.

I think that this only addresses the scenario where the primary connection is to foo.example.com and the primary certificate contains foo.example.com, allowing this extension to match if it includes *.example.com.

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -753,6 +753,80 @@ any stream which references the failing certificate in question or process the
...
+
+    id-ce-allowedPrimaryDomain OBJECT IDENTIFIER ::=  { id-ce TBD1 }
+
+    AllowedPrimaryDomain ::= SEQUENCE SIZE (1..MAX) OF IA5String
```
martinthomson: Why not structure this like subjectAltName and allow alternative identifier types as well?  Too much feature?  The advantage would be that you could cite the subjectAltName definition whole cloth and concentrate on matching rules.

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -753,6 +753,80 @@ any stream which references the failing certificate in question or process the
...
+
+## Affiliated Domain Certificate Extension {#extension-affiliated}
+
+The Affiliated Domain extension allows groups of certificates under the same
+control to be identified without giving each certificate full identity with the
+domain.  The identities stated as affiliated domains are not bound to the
+subject of the certificate, but this extension indicates that there is a common
+management relationship between the subject and the affiliated domains.
```
martinthomson: Here I think that the list of names is a genuine problem.  For this, I would recommend instead that this include a single name.  Keep the requirement that it be a domain name under the control of the entity that is serving these connections, but don't constrain it.  Basically, say that it must be validated as though it were subjectAltName (see below for more on this point.)

The problem with multiple values is that you can have a primary with A and B, a secondary with B and C, and a third secondary with C and D and that might be allowed, even though the third secondary shares no common points with the primary.

Similarly, you might have a secondary with A and D, which means that you can have secondaries with no common affiliated domains.

The reason you validate this is so that an attacker who compromises a key for foo.example.com and can cause a connection to be established to example.attacker can't get the two together because example.attacker because both certificates no longer match up.

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -753,6 +753,80 @@ any stream which references the failing certificate in question or process the
...
+"preferred name syntax", as specified by Section 3.5 of {{!RFC1034}} and as
+modified by Section 2.1 of {{!RFC1123}}.  Note that while uppercase and lowercase
+letters are allowed in domain names, no significance is attached to the case. In
+addition, while the string " " is a legal domain name, affiliatedDomain
+extensions MUST NOT contain " ".  Rules for encoding internationalized domain
+names are specified in Section 7.2 of {{!RFC5280}}.
+
+Affiliated domains MAY be constrained in the same manner as subject
+distinguished names using the name constraints extension as described in Section
+4.2.1.10 of {{!RFC5280}}.
```
martinthomson: If they are validated, then I believe that name constraints are necessary, not optional.

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -770,6 +844,13 @@ its control in order to present the compromised certificate. As recommended in
 {{?RFC8336}}, clients opting not to consult DNS ought to employ some alternative
 means to increase confidence that the certificate is legitimate.
 
+One such means is the Allowed Primary Domain certificate extension defined in
+{extension}. Clients SHOULD require that server certificates presented via this
+mechanism contain the Allowed Primary Domain extension and require that a
+certificate previously accepted on the connection (including the certificate
+presented in TLS) lists one of the Allowed Primary Domains in the Subject field,
+the Subject Alternative Name extension, or the Affiliated Domains extension.
```
martinthomson: This isn't very clear to me.  The point here is that you can allow either mechanism.

draft-ietf-httpbis-http2-secondary-certs.md
MikeBishop: Cribs a lot from the definition of SAN, but cuts out the allowance for different name types since this is specifically a domain name.

draft-ietf-httpbis-http2-secondary-certs.md
MikeBishop: Yes, that restriction is down in the Security Considerations.  I can add a ref, unless you think it belongs here (or elsewhere).

draft-ietf-httpbis-http2-secondary-certs.md
MikeBishop: Yeah, mostly too much feature.  The security properties are probably similar -- if you can hijack a domain to fool DV, you can hijack any e-mail address on the domain to fool an e-mail validation, etc.  We expect domains to be the common case and they're sufficient, so let's not multiply entities without need.

draft-ietf-httpbis-http2-secondary-certs.md
MikeBishop: Yes, but I think the chaining is necessary (and that's why the affiliated and the allowed are separate), or at least strongly desirable.  Consider:

- Primary:  (Affiliated:  M)
- First Secondary:  (Affiliated:  A, Allowed:  *)
- Second Secondary:  (Allowed:  A, Affiliated:  Z)
- Third Secondary:  (Allowed:  A, Affiliated: Y)

It's fine to have secondaries without overlapping affiliations, provided that there's a clear chain in the allowed path.

draft-ietf-httpbis-http2-secondary-certs.md
MikeBishop: Imported text from SAN, but I read this as saying that name constraints are permissible behavior (and the converse is that they MUST be supported).  I'll reword to clarify, unless you're arguing for name constraints being required to be present.

draft-ietf-httpbis-http2-secondary-certs.md
martinthomson: OK, so I *really* don't want chain building here.  But I can see how you get there and it sucks.  Need to think more about this.

draft-ietf-httpbis-http2-secondary-certs.md
martinthomson: If the CA is responsible for validating a name, then it can't validate a name outside of its name constraints.  For instance, if an enterprise were delegated *.enterprise.example, then it wouldn't be permitted to say that foo.example.com was OK.

draft-ietf-httpbis-http2-secondary-certs.md
martinthomson: I think that you could just copy the SAN grammar and say that the rules are inherited.  I know that means that you get other name types, but I don't think that's fatal (we still accept IP SANs, and probably wouldn't have a problem if they were included in this).

grittygrease:
It's not clear to me that Affiliated Domain adds anything. In what situation would someone use an affiliated domain that they are not authoritative for?

grittygrease:
I don't think that's an issue. We currently have <certid>.cloudflaressl.com as the CN on all Cloudflare-issued certs. It wouldn't be a stretch to put coalesce.cloudflare.com or something similar on all certificates.

draft-ietf-httpbis-http2-secondary-certs.md
MikeBishop: I think it should, yes.

draft-ietf-httpbis-http2-secondary-certs.md
MikeBishop: I'm okay with tightening here, but this seems at odds with importing the full set of SAN rules by reference.  I'd rather just state the rules for this extension than be "like SAN, except...."

draft-ietf-httpbis-http2-secondary-certs.md
MikeBishop: I'm okay either way, but I think I'd like broader WG opinion on whether we want the extra feature of non-domain requirements.

MikeBishop:
**UPDATES:**
- Required Domain is no longer a list, per @grittygrease's suggestion
- It borrows the SAN syntax wholesale (`GeneralName`), per @martinthomson's suggestion
- Per @martinthomson's observation that matching two sets is annoying, it can no longer be a partial wildcard.  It's either a total wildcard ("Use with anything") or a specific name.  That specific name could be matched by a wildcard in the primary certificate, of course.
- Requires CAs to verify the presence of a CAA record before issuing, so the issue of a misissued certificate is somewhat better constrained.

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -758,6 +758,34 @@ any stream which references the failing certificate in question or process the
...
+
+The Required Domain extension allows certificates to limit their use with
+Secondary Certificate Authentication.  A client MUST verify that the server has
+proven ownership of the indicated identity before accepting the limited
+certificate over Secondary Certificate Authentication.
+
+The identity in this extension is a restriction asserted by the requester of the
+certificate and is not verified by the CA.  Conforming CAs SHOULD mark the
+requiredDomain extension as non-critical.  Conforming CAs MUST require the
+presence of a CAA record {{!RFC6844}} prior to issuing a certificate with this
```
martinthomson: The design in signed exchanges also requires a CAA parameter for the issuance to be allowed.  Because we are interested in limiting the scope of misissuance, perhaps it would be good to define a new CAA parameter and require that the name match the value of the parameter.  That doesn't have to happen here, but we should discuss whether that helps any.

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -758,6 +758,34 @@ any stream which references the failing certificate in question or process the
...
+The required domain is represented as a GeneralName, as specified in Section
+4.2.1.6 of {{!RFC5280}}. Unlike the subject field, conforming CAs MUST NOT issue
+certificates with a requiredDomain extension containing empty GeneralName
+fields.  Clients that encounter such a certificate when processing a
+certification path MUST consider the certificate invalid.
+
+The wildcard character "*" MAY be used to represent that any previously
+authenticated identity is acceptable.  This character MUST be the entirety of
+the name if used and MUST have a type of "dNSName".  (That is, "*" is
+acceptable, but "*.com" and "w*.example.com" are not).
```
martinthomson: We discussed having stricter validation for certificates that use "\*".  It's not appropriate to mandate that, but this could mention that there are far greater risks to issuing certificates with a required domain of "\*".

draft-ietf-httpbis-http2-secondary-certs.md
```diff
@@ -775,6 +803,13 @@ its control in order to present the compromised certificate. As recommended in
 {{?RFC8336}}, clients opting not to consult DNS ought to employ some alternative
 means to increase confidence that the certificate is legitimate.
 
+One such means is the Required Domain certificate extension defined in
+{extension}. Clients MUST require that server certificates presented via this
+mechanism contain the Required Domain extension and require that a certificate
+previously accepted on the connection (including the certificate presented in
+TLS) lists the Required Domain in the Subject field or the Subject Alternative
+Name extension.
```
martinthomson: I think that you want to explicitly address the question ekr raised at the meeting (which might not have been minuted).

He asks whether you can produce an EA that has a certificate without a required domain and then use that as the target of required domain in other certificates.  I don't think that this is a good idea because it allows an attacker to muddy the audit trail if they are able to acquire a certificate they control. But we should be very clear either way and this isn't quite clear enough.

"Any certificate presented in an exported authenticator MUST be rejected if it does not include a RequiredDomain extension." might be enough.

MikeBishop: Except that's not quite correct, because we're not requiring this in the client->server direction.  I thought ekr's concern was already covered by this requirement:

> Clients MUST require that server certificates presented via this
mechanism contain the Required Domain extension and require that a certificate
previously accepted on the connection (including the certificate presented in
TLS) lists the Required Domain in the Subject field or the Subject Alternative
Name extension.

That is, a certificate with no Required Domain will never be an acceptable secondary to clients and therefore can't have been previously accepted (except in the TLS handshake), but can still be acceptable to servers for client verification.

MikeBishop: I added a reference to this requirement in the section about validating the contents of an EA.  Hopefully that addresses the concern.

PR #722, opened 2018-11-19
[SH] fix key=item serialisation

phluid61:
Serialisation of dictionary and param-list currently say `identifier=key` instead of `key=item`

This fixes that.


mnot:
Blah, good catch; thanks.

PR #726, opened 2018-11-22
[SH] make boolean ABNF case-sensitive

phluid61:
the `boolean` ABNF rule is case-*in*sensitive, but the algorithm for parsing (and the test case in the separate repo) enforce capitals only.


mnot:
Thanks!

phluid61:
Oh! It should have been `%54 / %46`

I'm going to stop typing anything at all today.

mnot:
No worries - fixed :)

Issue #734, opened 2018-12-07
Variants' incoming-request assumes all content-negotiation systems' Accept headers are lists


jyasskin:
https://httpwg.org/http-extensions/draft-ietf-httpbis-variants.html#cache says:

> Given incoming-request (a mapping of field-names to lists of field values)

It doesn't specify how to calculate this from the HTTP request (https://fetch.spec.whatwg.org/#requests? I don't see a version-independent definition of a request in the HTTP specs.) that the cache actually receives, and it doesn't deal with request headers like https://httpwg.org/http-extensions/client-hints.html#dpr that hold a single value instead of a list.

mnot:
Are you looking for text that effectively says:

> If the header is defined to be a single value, take the first instance of the header field and put its value into a list. If the header is defined to be a list (e.g., using the #rule), parse each instance of the header field into a list and combine all of those lists.

?

note that work on http-core might help here; e.g., see https://github.com/httpwg/http-core/issues/116

jyasskin:
#735 changed the *use* of `incoming-request` to say

> Let request-value be the field-value associated with field-name in incoming-request (after being combined as allowed by Section 3.2.2 of [RFC7230])

which I think is good (but I would since I wrote it). The declaration of the variable could say the same thing.

I'd rather have this part of this spec not worry about the definition of the header, and instead isolate that to the call into the "content negotiation mechanism that the implementation supports". Then headers like DPR can define for themselves how they react to responses that are invalid by including multiple header instances.

PR #735, opened 2018-12-07
Handle absent request headers.

### LINKED ISSUE #731: Cache behavior assumes browser request includes all Variant'ed headers
https://httpwg.org/http-extensions/draft-ietf-httpbis-variants.html#cache says "... For each variant in variants-header ... Let request-value be the field-value(s) associated with field-name in incoming-request." I don't see any handling of the case where there is no `field-name` header in `incoming-request`. The algorithms in https://httpwg.org/http-extensions/draft-ietf-httpbis-variants.html#backports then also don't handle the case where request-value is empty.

Sorry if I've just missed where you handle the empty case.

jyasskin:
This change passes 'null' to the content-negotiation algorithm if the
request header is absent. For list headers, the algorithms can always
turn that into the empty list, but for non-list headers they need to
come up with something appropriate to do.

Fixes #731. This also tries to do better about #734, but may not completely fix it.


mnot:
Thanks.

PR #744, opened 2018-12-18
Convert Variants to use Structured Headers.

### LINKED ISSUE #570: Adopt Structured Headers

### LINKED ISSUE #743: Cache behavior examples are out of date wrt new Variant-Key format
https://httpwg.org/http-extensions/draft-ietf-httpbis-variants.html#example-of-cache-behaviour shows a Variant-Key of

```http
Variant-Key: fr, gzip
```

But that's now [spelled](https://httpwg.org/http-extensions/draft-ietf-httpbis-variants.html#variant-key)

```http
Variant-Key: fr;gzip
```

This also has implications for the https://httpwg.org/http-extensions/draft-ietf-httpbis-variants.html#gen-variant-key algorithm, which currently ignores semicolons.

jyasskin:
Fixes #570. I think this also fixes #743, but I might have missed something.

Misc changes that aren't actually necessary to use Structured Headers:

* "variant-item" became "variant-axis".
* I defined what clients ("caches"?) MUST do when a server violates one of its MUSTs.
* I removed the (incorrect) `token / "/" / "?" / "\" / "[" / "]" / ":" / "@" / "(" / ")"` definition of available-value in favor of allowing arbitrary [Structured Header strings](https://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html#string). It wasn't clear what excluding characters like `{` was gaining us, and the restriction would have required extra words when parsing the headers.
* I removed {{gen-variant-key}} entirely and had Compute Possible Keys return a list of lists that could be directly compared against the Structured Header parse of `Variant-Key`. It would be difficult to keep the old behavior of returning a list of strings without re-adding some restriction on an available-value, for example by requiring that they don't contain spaces.
* I simplified Compute Possible Keys a bit.

### DIFF
--- a/draft-ietf-httpbis-variants.md
+++ b/draft-ietf-httpbis-variants.md
@@ -219,6 +219,8 @@ Variant-Key: gzip;fr, "identity";fr
 
 indicates that this response can be used for requests whose Accept-Encoding algorithm selects "gzip" or "identity", as long as the Accept-Language algorithm selects "fr" -- perhaps because there is no gzip-compressed French representation.
 
+When more than one Variant-Key value is in a response, the first one present MUST correspond to the request that caused that response to be generated.
+
 Parsing is strict. For example:
 
 ~~~ example
--- a/draft-ietf-httpbis-variants.md
+++ b/draft-ietf-httpbis-variants.md
@@ -115,24 +115,26 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 described in BCP 14 {{!RFC2119}} {{!RFC8174}} when, and only when, they appear in all capitals, as
 shown here.
 
-This specification uses the Augmented Backus-Naur Form (ABNF) notation of {{!RFC5234}} with a list extension, defined in Section 7 of {{!RFC7230}}, that allows for compact definition of comma-separated lists using a '#' operator (similar to how the '*' operator indicates repetition).
+This specification uses the Augmented Backus-Naur Form (ABNF) notation of {{!RFC5234}} but relies on Structured Headers from {{!I-D.ietf-httpbis-header-structure}} for parsing.
 
-Additionally, it uses the "field-name", "OWS" and "token" rules from {{!RFC7230}}, and "type", "subtype", "content-coding" and "language-range" from {{!RFC7231}}.
+Additionally, it uses the "field-name" rule from {{!RFC7230}}, and "type", "subtype", "content-coding" and "language-range" from {{!RFC7231}}.
 
 
 # The "Variants" HTTP Header Field {#variants}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-variants.md
```diff
@@ -115,24 +115,26 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
-Variants        = 1#variant-item
-variant-item    = field-name *( OWS ";" OWS available-value )
-available-value = token
-                  / "/" / "?" / "\" / "[" / "]"
-                  / ":" / "@" / "(" / ")"
+Variants        = sh-list-of-lists
 ~~~
 
-Each "variant-item" indicates a request header field that carries a value that clients might proactively negotiate for; each parameter on it indicates a value for which there is an available representation on the origin server.
+If Structured Header parsing fails or a list-member has the wrong type, the client MUST treat the representation as having no Variants header field.
```
jyasskin: Should a malformed `Variants` header be ignored or should it disable caching for that resource?

mnot: Treated as absent, I think; Vary should still ensure correct (if suboptimal) behaviour.

draft-ietf-httpbis-variants.md
```diff
@@ -209,33 +214,32 @@ If the response can be used to satisfy more than one request), they can be liste
 
 ~~~ example
 Variants: Accept-Encoding;gzip;br, Accept-Language;en ;fr
-Variant-Key: gzip;fr, identity;fr
+Variant-Key: gzip;fr, "identity";fr
 ~~~
 
 indicates that this response can be used for requests whose Accept-Encoding algorithm selects "gzip" or "identity", as long as the Accept-Language algorithm selects "fr" -- perhaps because there is no gzip-compressed French representation.
 
-When more than one Variant-Key value is in a response, the first one present MUST indicate the variant-key for the response it occurs within.
```
mnot: This requirement addresses an implementation constraint; if they can't easily link a response to multiple potential requests, this allows them to still behave correctly (if not optimally). Why did you remove it?

mnot: The wording above doesn't reflect the intent well. It should be:

> When more than one Variant-Key value is in a response, the first one present MUST correspond to the request that caused that response to be generated.

jyasskin: Ah, that makes sense. I've added your wording back.

I think it could be wordsmithed a little better now that Variant-Key is a structured header, but I don't want to delay merging the overall change. Specifically, "more than one Variant-Key value" implies to me that the Variant-Key header appears more than once, but we actually mean that the list parsed for Variant-Key from a response contains more than one inner-list, even if they're all parsed from a single instance of the header.

PR #749, opened 2019-01-10
Define support requirements for integers

### LINKED ISSUE #737: Integer limits
Working on evert/structured-headers, @evert [noticed](https://github.com/httpwg/structured-header-tests/issues/9) that representing the full range of SH integers is [hard](http://2ality.com/2013/10/safe-integers.html), and going beyond it can cause [problems](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER).

Given JS's place in the Web ecosystem and the relative rarity of headers that need integers that big (as well as the availability of floats), should we consider limiting the range of integer to match JS?

mnot:
Fixes #737

### DIFF
--- a/draft-ietf-httpbis-header-structure.md
+++ b/draft-ietf-httpbis-header-structure.md
@@ -781,8 +781,8 @@ NOTE: This algorithm parses both Integers {{integer}} and Floats {{float}}, and 
    6. If type is "float" and input_number contains more than 16 characters, fail parsing.
 8. If type is "integer":
    1. Parse input_number as an integer and let output_number be the product of the result and sign.
-   2. If output_number is outside the range for integers defined in {{integer}}, fail parsing.
-   3. If output_number is not supported accurately by the implementation (as noted in {{integer}}), fail parsing.
+   2. If output_number is not supported accurately by the implementation (as noted in {{integer}}), fail parsing.
+   3. If output_number is outside the range for integers defined in {{integer}}, fail parsing.
 9. Otherwise:
    1. If the final character of input_number is ".", fail parsing.
    2. Parse input_number as a float and let output_number be the product of the result and sign.
--- a/draft-ietf-httpbis-header-structure.md
+++ b/draft-ietf-httpbis-header-structure.md
@@ -932,6 +932,7 @@ _RFC Editor: Please remove this section before publication._
 * Added Lists of Lists (#721).
 * Rename Identifier to Token (#725).
 * Add implementation guidance (#727).
+* Require minimum range for integers, and predictable failure (#737).
 
 
 ## Since draft-ietf-httpbis-header-structure-07
--- a/draft-ietf-httpbis-header-structure.md
+++ b/draft-ietf-httpbis-header-structure.md
@@ -305,6 +305,10 @@ For example:
 Example-IntegerHeader: 42
 ~~~
 
+Parsers MUST accurately support integers in the range -9,007,199,254,740,991 to
+9,007,199,254,740,991 inclusive (i.e., -2^53+1 to 2^53-1) at a minimum, and MUST fail parsing on
+values outside this range if they are not supported by the parser.
+
 
 ## Floats {#float}
 
@@ -777,7 +781,8 @@ NOTE: This algorithm parses both Integers {{integer}} and Floats {{float}}, and 
    6. If type is "float" and input_number contains more than 16 characters, fail parsing.
 8. If type is "integer":
    1. Parse input_number as an integer and let output_number be the product of the result and sign.
-   2. If output_number is outside the range defined in {{integer}}, fail parsing.
+   2. If output_number is outside the range for integers defined in {{integer}}, fail parsing.
+   3. If output_number is not supported accurately by the implementation (as noted in {{integer}}), fail parsing.
 9. Otherwise:
    1. If the final character of input_number is ".", fail parsing.
    2. Parse input_number as a float and let output_number be the product of the result and sign.
--- a/draft-ietf-httpbis-header-structure.md
+++ b/draft-ietf-httpbis-header-structure.md
@@ -306,8 +306,8 @@ Example-IntegerHeader: 42
 ~~~
 
 Parsers MUST accurately support integers in the range -9,007,199,254,740,991 to
-9,007,199,254,740,991 inclusive (i.e., -2^53+1 to 2^53-1) at a minimum, and MUST fail parsing on
-values outside this range if they are not supported by the parser.
+9,007,199,254,740,991 inclusive (i.e., -2^53+1 to 2^53-1) at a minimum. Parsers MUST fail parsing
+on values not within the range supported accurately.
 
 
 ## Floats {#float}


draft-ietf-httpbis-header-structure.md
```diff
@@ -305,6 +305,10 @@ For example:
 Example-IntegerHeader: 42
 ~~~
 
+Parsers MUST accurately support integers in the range -9,007,199,254,740,991 to
+9,007,199,254,740,991 inclusive (i.e., -2^53+1 to 2^53-1) at a minimum, and MUST fail parsing on
+values outside this range if they are not supported by the parser.
```
martinthomson: I think that this is clearer:

```suggestion
Parsers MUST fail parsing if values are not within the range supported by the parser.
```

I think that's simpler.

That is, say that parsers need to support up to 2^53-1.  Allow more.  Require parsing failure for numbers outside that range (whatever it is).

draft-ietf-httpbis-header-structure.md
```diff
@@ -777,7 +781,8 @@ NOTE: This algorithm parses both Integers {{integer}} and Floats {{float}}, and
    6. If type is "float" and input_number contains more than 16 characters, fail parsing.
 8. If type is "integer":
    1. Parse input_number as an integer and let output_number be the product of the result and sign.
-   2. If output_number is outside the range defined in {{integer}}, fail parsing.
+   2. If output_number is outside the range for integers defined in {{integer}}, fail parsing.
```
martinthomson: The two lines could be one, as follows:
```suggestion
   2. If output_number is outside the range for integers supported by the parser (see {{integer}}), fail parsing.
```

mnot: I'd rather be explicit.

draft-ietf-httpbis-header-structure.md
mnot: Took a slightly different approach.

draft-ietf-httpbis-header-structure.md
martinthomson: In which case, your two points need to be switched.  As written, this would prohibit supporting a larger range.

martinthomson: Belay that.  The comment stands.  There is no way to have a fail parsing on this condition.

mnot: I don't think it prohibits supporting a larger range directly, but it does beg the question of how you can check that the number falls within the greater range if you can't accurately represent it. Will fix.

draft-ietf-httpbis-header-structure.md
```diff
@@ -777,7 +781,8 @@ NOTE: This algorithm parses both Integers {{integer}} and Floats {{float}}, and
    6. If type is "float" and input_number contains more than 16 characters, fail parsing.
 8. If type is "integer":
    1. Parse input_number as an integer and let output_number be the product of the result and sign.
-   2. If output_number is outside the range defined in {{integer}}, fail parsing.
+   2. If output_number is not supported accurately by the implementation (as noted in {{integer}}), fail parsing.
+   3. If output_number is outside the range for integers defined in {{integer}}, fail parsing.
```
martinthomson: As noted, this needs another condition ...and the implementation does not support integers outside this range.

mnot: Where?

draft-ietf-httpbis-header-structure.md
```diff
@@ -777,7 +781,8 @@ NOTE: This algorithm parses both Integers {{integer}} and Floats {{float}}, and
    6. If type is "float" and input_number contains more than 16 characters, fail parsing.
 8. If type is "integer":
    1. Parse input_number as an integer and let output_number be the product of the result and sign.
```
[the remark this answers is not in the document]
wtarreau: I totally agree with Kazuho here. It's mandatory to check for overflows during the parsing, like we do here for example : http://git.haproxy.org/?p=haproxy.git;a=blob;f=src/h1.c;hb=HEAD#l52
Many integer parsers will reliably provide the information about the fact that an overflow was detected, but when it comes to floats it might not necessarily be the case (since very large numbers can still fit in a float), however the check against bounds is expected to work.

This means that we should mandate the use of a reliable parser known for detecting overflows, AND then check for the output number to fall within our bounds.

We should also warn against the risk of using generic parsers which as a byproduct of checking for errors, also happen to support base prefixes like "0x", or float parsers supporting exponents like "1e32" or special values like "inf", "-inf", "nan", all of which defeat length checks.

wtarreau:
All this suddenly makes me think that we have another integer representation used in H1, which is the chunk size expressed as an hexadecimal value, which can also trigger an overflow during parsing and which can be represented inaccurately on some implementations for the same reasons. Thus probably the block above should mention that it is valid for decimal numbers only ? Or we must take care of the base in use ?

draft-ietf-httpbis-header-structure.md
```diff
@@ -305,6 +305,10 @@ For example:
 Example-IntegerHeader: 42
 ~~~
 
+Parsers MUST accurately support integers in the range -9,007,199,254,740,991 to
+9,007,199,254,740,991 inclusive (i.e., -2^53+1 to 2^53-1) at a minimum. Parsers MUST fail parsing
+on values not within the range supported accurately.
```
annevk: This reads a little awkward to me. Perhaps something like "Implementations must have a range for supported integers. At a minimum this range must be X to Y. Parsers must fail on parsing values outside the range for supported integers."

draft-ietf-httpbis-header-structure.md
martinthomson: Let's say that I support -2^63 to 2^63-1, something that isn't unreasonable if you aren't javascript.  This code turns into:

```js
if (n < -2^63 || n > 2^63-1) { // Step 2.
  return fail_parsing;
}
if (n < -2^53+1 || n > 2^53-1) { // Step 3.
  return fail_parsing;
}
```

Pass n = 2^60 into that code and you get an error.  Hence my insistence that the first test is adequate.

Now, you could read "the range for integers" as meaning  -2^63 to 2^63-1 in this case, but then you just repeated yourself.

Does that help clarify the concern?

draft-ietf-httpbis-header-structure.md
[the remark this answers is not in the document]
phluid61: > We should also warn against the risk of using generic parsers which as a byproduct of checking for errors, also happen to support base prefixes like "0x", or float parsers supporting exponents like "1e32" or special values like "inf", "-inf", "nan", all of which defeat length checks.

At this point in the algorithm the string has been preprocessed and is guaranteed to have no more than 19 characters, all of which are decimal digits. Or do you mean explicit warning against replacing the entire integer-parsing algorithm with a generic `to_int()` or equivalent?

wtarreau:
> Do you mean adding a note to the "Parsers MUST accurately support integers in the range..." chunk?
> Because the ABNF sh-integer = ["-"] 1*19DIGIT already implies as much.

Yes, so that implementers are warned about this problem they don't necessarily think about while implementing. Quite frankly, who really verifies each character in a number before passing it to strtol() ? Not that many...

However, my point about chunk sizes is irrelevant to structured headers since it's not in headers that we find it.

bsdphk:
Having turned this one around in my head for days, I am on board for a clear <±2^53-1 limit.

I cannot come up with any plausible need for more than 8 peta-anything in HTTP headers which are not better served by other means.

However, this removes the need to have the Integer type in the first place, we added that because of the 15 digit limitation of what is now called "sh-float."

A much simpler edit is:

1. Remove sh-integer

2. Rename sh-float to sh-number

3. add "15DIGIT" to sh-numbers ABNF

This way we do not need to spend any text on the limit or its enforcement, because it is implicit in the 15 digit upper limit.

Issue #757, opened 2019-01-15
bcp56bis: terminology URL vs URI


reschke:
See https://lists.w3.org/Archives/Public/ietf-http-wg/2019JanMar/0002.html

PR #758, opened 2019-01-15
bcp56bis: URI vs URL (see #757)

reschke:
This change switches from "URL" to "URI" specifically for two contexts:

- URI schemes
- URI syntax

(It might be good to also add an explanation somewhere in the Introduction; I can make a proposal if needed)


reschke:
It is a PR, no?

mnot:
Weird, it didn't show the UX for PRs before, but does now. *shrug*

PR #759, opened 2019-01-24
Added a section with more detailed description of Very Large Value to rand-access-live

craigpratt:
Also removed some of the VLV description from the Security section
 and put some more on-point security considerations in the section.

These edits are in response to Adam Roach's feedback in IESG.


draft-ietf-httpbis-rand-access-live.xml
```diff
@@ -439,6 +439,9 @@ Content-Range: bytes 1020000-999999999999/*
       <t>
         with the response body returning bytes 1020000-1254567 immediately and aggregated ("live") data being returned as the content is aggregated.
       </t>
+      <t>
+        While it would be ideal to define a single numerical Very Large Value, there's no single value that would work for all applications and platforms. e.g. JavaScript numbers cannot represent integer values above 2^^53, so a JavaScript application may want to use 2^^53-1 for a Very Large Value. This value, however, would not be sufficient for all applications. So this draft recommends 2^^53-1 (9007199254740991) as a Very Large Value unless an application has a good justification to use a smaller or larger value. e.g. If it's always known that the resource won't exceed a value smaller than the recommended Very Large Value for an application, a smaller value can be used for efficiency. And if it's likely that an application will utilize resources larger than the recommended Very Large Value, a larger value should be used.
```
martinthomson: "e.g." is followed by a comma: "e.g.,"

And I think that you can drop "for efficiency" on the second to last sentence.

draft-ietf-httpbis-rand-access-live.xml
```diff
@@ -478,16 +481,23 @@ Content-Range: bytes 1020000-1254567/*
 
 <!-- Possibly a 'Contributors' section ... -->
 
+    <section anchor="RecommendedVLV" title="Recommendations for Very Large Values">
+        <t>
+            While it would be ideal to define a single numerical Very Large Value, there's no single value that would work for all applications and platforms. e.g. JavaScript numbers cannot represent integer values above 2^^53, so a JavaScript application may want to use 2^^53-1 for a Very Large Value. This value, however, would not be sufficient for all applications. So this draft recommends 2^^53-1 (9007199254740991) as a Very Large Value unless an application has a good justification to use a smaller or larger value. e.g. If it's always known that the resource won't exceed a value smaller than the recommended Very Large Value for an application, a smaller value can be used for efficiency. And if it's likely that an application will utilize resources larger than the recommended Very Large Value, a larger value should be used.
```
martinthomson: Is this duplicated text?  It's almost the same as the above.  You should probably just forward-reference this section from above.

Given that you have a dedicated section, I would split the paragraph at "So this draft recommends" and drop the "So".

draft-ietf-httpbis-rand-access-live.xml
```diff
@@ -478,16 +481,23 @@ Content-Range: bytes 1020000-1254567/*
...
-
-      Also, some implementations (e.g. JavaScript-based clients and servers) are not able to represent all values beyond 2^53. So similarly, if there's no expectation that a representation will ever exceed 2^53 bytes, values smaller than this limit should be used for the last-byte-pos in byte-range requests.
-    </t>
+            As described above, servers need to be prepared to receive values of last-byte-pos in Range requests that may be larger than the numerical representations of the language or platform the server is implemented upon - and return these values in Content-Range response header fields. Servers should check the last-byte-pos value before converting and storing them into numeric form. The simplest way to satisfy the live-range semantics defined in this document without potential overflow issues is to store the last-byte-pos as a string value and return it in the byte-range Content-Range response header's last-byte-pos field.
```
martinthomson: s/may be/are/

draft-ietf-httpbis-rand-access-live.xml
craigpratt: Oops - thought I moved the section. Guess I copied during my late-night editing. Fixing...

draft-ietf-httpbis-rand-access-live.xml
craigpratt: Agreed. Thanks as always Martin!

craigpratt:
OK - think I'm done tweaking @martinthomson. Let me know if I've addressed your comments.

thx again!

PR #760, opened 2019-01-24
Removed the "Requirements Language" section

craigpratt:
This draft isn't using MUST/SHOULD/MAY/etc as the normative
 language is supplied by the referenced HTTP RFCs. So removed
 the "Requirements Language" section defining them.

These edits are in response to Adam Roach's feedback in IESG.


craigpratt:
this was incorporated into PR #759

PR #763, opened 2019-01-28
cache-header: fix title

reschke:


mcmanus:
https://github.com/httpwg/http-core/issues/111  .. that issue is still open, but sense of room in bkk was to greenlight the common parlance of header

reschke:
I would prefer the WG to stick to the current terminology until we decide to change/extend it.

tfpauly:
@mnot can this be closed?

PR #772, opened 2019-02-20
Clarify TLS 1.3 post-handshake auth with HTTP/2

davidben:
HTTP/2 blocked renegotiation for a number of reasons. Among other
problems, as this document notes in section 1.2.3, there is no defined
way to correlate application-visible renegotiation events
(CertificateRequest) with HTTP/2 requests. In essence, the prohibition
is a reflection of TLS+renegotiation being the wrong API shape for
HTTP/2 (and most protocols).

TLS 1.3 removed renegotiation, but added post-handshake client
authentication to replace some of renegotiation's use cases. This
retains the problems above. However, with the feature renamed,
RFC7540's prohibition no longer applies and we're stuck with an
incompatibility between RFC8446 and RFC7540.

Post-handshake authentication is optional in TLS 1.3 and advertised with
an extension, but this is not sufficient. Most HTTP/2 clients are also
HTTP/1.1 clients and TLS parameters are negotiated in parallel. An
HTTP/1.1+HTTP/2 client that advertises the extension for HTTP/1.1 (see
section 1.2.2 of this document) also advertises it for HTTP/2.

By a strict composition of the existing specification text,
HTTP/2-capable clients cannot offer post-handshake authentication to
HTTP/1.1 servers.

In reality, existing HTTP/2 servers won't use this mechanism because no
one has defined a request correlation. Thus we can fix all this by
clarifying that TLS 1.3 post-handshake auth is forbidden whether or not
the client offered it.

If some future use case comes up, it can define an overriding extension
with a correlation mechanism. It would already need to do so, because
existing clients wouldn't know about the correlation. This is unlikely
to happen anyway because this document *is* that use case, and it
already defines a separate mechanism.

### DIFF
--- a/draft-ietf-httpbis-http2-secondary-certs.md
+++ b/draft-ietf-httpbis-http2-secondary-certs.md
@@ -39,6 +39,7 @@ normative:
   I-D.ietf-tls-exported-authenticator:
 
 informative:
+  RFC7301:
   RFC7838:
 
 
@@ -231,6 +232,14 @@ identifier could be added to TLS 1.2 by means of an extension to TLS.
 Unfortunately, many TLS 1.2 implementations do not permit application data to
 continue during a renegotiation. This is problematic for a multiplexed protocol
 like HTTP/2.
+
+TLS 1.3's post-handshake client authentication mechanism shares similar problems
+for use with HTTP/2. HTTP/2 clients that offer multiple ALPN {{RFC7301}}
+protocols in a TLS ClientHello MAY include the `post_handshake_auth` extension
+to support those other protocols, but this does not indicate support for
+post-handshake client authentication in HTTP/2. HTTP/2 servers MUST NOT send
+post-handshake TLS CertificateRequest messages, even if the client offered the
+`post_handshake_auth` extension.
 
 ## HTTP-Layer Certificate Authentication
 


MikeBishop:
Two points here:
- Is there an issue for this?  I'd prefer to do non-editorial discussion on the issue or the list.
- I'd be fine saying that clients which support this extension MUST NOT use post-handshake auth in HTTP/2 over TLS 1.3 even if negotiated; saying that for all HTTP/2 clients oversteps some bounds.  This document defines an extension -- pretty much by definition, it can only impose requirements scoped to implementations which support the extension.  This isn't the right place to make a prohibition on the behavior of HTTP/2 clients in general.

martinthomson:
I would say that the reasons we prohibited renegotiation apply to post-handshake authentication.  So unless you have negotiated an extension (the likes of which @MikeBishop defined), then this is a good idea.

Yes, that oversteps bounds, so I would say we should discuss whether it needs a document of its own.  That seems like it is overkill though.

davidben:
Using this document seems natural given it already talks about the gamut of reactive client certificates, and indeed by doing so has a dependency on this bugfix. Section 1.2.2 of draft-ietf-httpbis-http2-secondary-certs does not work without this fix in it.

But if folks want a separate document, that works too I suppose. Though, note that due to the section 1.2.2 problem, we'd then need to queue that new document up in front of this one.

MikeBishop:
I don't see that 1.2.2 is impacted, since it merely describes the use of post-handshake auth with HTTP/1.1 (which is perfectly valid).  It's worth mentioning that post-handshake auth doesn't fix HTTP/2's problems with renegotiation in 1.2.3; that's currently a gap.  I don't think filling that gap with a change to HTTP/2 all-up is appropriate.

I agree, a separate RFC for this text seems overkill.  It's certainly not an erratum, as the mechanism in question didn't exist at the time of 7540. But I don't think this document is the right home and I don't know of another in-progress document that would be either.  So my inclination would still be a super-short RFC.

I suspect we could sidestep making a blocking reference pretty easily.  Your proposed text focuses on client behavior when negotiating which protocol to use, which the document currently doesn't address.  If it publishes first, it could crib some introductory content from here and we'll add an informative reference to it here (close the gap above, "...and it is therefore prohibited by XXXX."); if later, it can point to this doc for a description of the problem it's solving and be even shorter.

davidben:
Remember that HTTP/1.1 and HTTP/2 share a ClientHello. A client which implements both cannot offer the post_handshake_auth extension to HTTP/1.1 without also offering it to HTTP/2. But it fundamentally doesn't work because there is no request correlation.

This means that modern clients effectively cannot implement the scheme described in 1.2.2. Chrome doesn't currently implement it, but if we ever do, the bugfix in this PR is a hard requirement. I'd rather it be codified in a spec, as this is a general problem, than for us to just make up a rule ad-hoc.

(This is why I wanted post-handshake auth to be conditioned on an application profile. Alas, it wasn't, so now it is unusable for even HTTP/1.1 without this fix.)

davidben:
So, what are the next steps here?

MikeBishop:
Since this is a WG doc, the fact that _I_ don't want this text in this doc doesn't trump the WG's decision.  My recommendation is a separate draft; if we had more time before the submission deadline, I'd say write a simple draft and ask for time in Prague.

If that doesn't happen, I'll put at least one slide on this in the update on Secondary Certs; we can have the discussion there about whether this needs to be in a separate doc.

davidben:
I'm rather hosed this week and am unlikely to put together a whole new document in time.

To clarify, do you not want this text to exist at all, or do you just want it in a separate document for editorial reasons? I agree with MT the latter is overkill, but I wouldn't object if it materialized in its own document. The former is just a fact of life given how ALPN and TLS works. If post-handshake auth in HTTP/1.1 is to exist, it *must* be forbidden in HTTP/2.

MikeBishop:
I don't think an extension to HTTP/2 ought to make a requirement on all of HTTP/2 in the absence of the extension.  So, process reasons rather than editorial reasons -- this isn't the right doc, and in the absence of one that is, make one.  I don't object to the text itself.

davidben:
Closing in favor of https://tools.ietf.org/html/draft-davidben-http2-tls13-00

PR #775, opened 2019-03-08
Remove image-related Client Hints

yoavweiss:
As previously discussed, the current state where some CH hints are defined in the IETF draft and some in accompanying specifications is confusing. Furthermore, defining the image related hints in web specifications would be a better fit, as they have direct implications to image processing there.

This PR removes those hints from the IETF draft. This hints are [being added](https://github.com/yoavweiss/html/commit/7bd134ec67472303d06700686be5d37d45ada048) to the HTML specification as part of a larger Client Hints related PR.

### DIFF
--- a/draft-ietf-httpbis-client-hints.md
+++ b/draft-ietf-httpbis-client-hints.md
@@ -93,7 +93,7 @@ This document defines a new response header, Accept-CH, that allows an origin se
 
 Client Hints mitigate the performance concerns by assuring that clients will only send the request headers when they're actually going to be used, and the privacy concerns of passive fingerprinting by requiring explicit opt-in and disclosure of required headers by the server through the use of the Accept-CH response header.
 
-This document defined the Client Hints infrastructure, but does not define any specific features that will use it. Those features will be defined in their respective specifications.
+This document defines the Client Hints infrastructure, a framework that enables servers to opt-in to specific proactive content negotiation features, which will enable them to adapt their content accordingly. However, it does not define any specific features that will use that infrastructure. Those features will be defined in their respective specifications.
 
 This document does not supersede or replace the User-Agent header field. Existing device detection mechanisms can continue to use both mechanisms if necessary. By advertising user agent capabilities within a request header field, Client Hints allow for cache friendly and proactive content negotiation.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-client-hints.md
```diff
@@ -93,9 +93,9 @@ This document defines a new response header, Accept-CH, that allows an origin se
 
 Client Hints mitigate the performance concerns by assuring that clients will only send the request headers when they're actually going to be used, and the privacy concerns of passive fingerprinting by requiring explicit opt-in and disclosure of required headers by the server through the use of the Accept-CH response header.
 
-This document also defines an initial set of Client Hints.
+This document defined the Client Hints infrastructure, but does not define any specific features that will use it. Those features will be defined in their respective specifications.
```
igrigorik: s/defined/defines? Past tense doesn't make sense here, I think?

Is "infrastructure" well understood term, or should we elaborate a bit on what this means? Something like, a framework for advertising hints that the client wants to receive, persisting them, and corresponding server logic? Ish?

draft-ietf-httpbis-client-hints.md
```diff
@@ -133,12 +132,12 @@ Servers can advertise support for Client Hints using the Accept-CH header field
 For example:
 
 ~~~ example
-  Accept-CH: DPR, Width, Viewport-Width
+  Accept-CH: Example-CH-Hint, Example-CH-Hint-2
```
igrigorik: Add `Sec-` prefix? `Sec-CH-Example1`, ...

draft-ietf-httpbis-client-hints.md
yoavweiss: > s/defined/defines? Past tense doesn't make sense here, I think?

Oops. definitely "defines".

I'll also elaborate on "infrastructure"

draft-ietf-httpbis-client-hints.md
yoavweiss: yup

igrigorik:
Merging. 

@mnot if you have any nits or suggestions, please chime in here and we can address in a separate PR. In the meantime, I'll push a new draft so we have the updated version to discuss at upcoming IETF.

Issue #767, opened 2019-02-12
Client-Hints exposes fingerprint values to additional parties and logging sensitive locations


snyderp:
From PING:

We're concerned about the privacy implications of moving these attributes to header values, specifically since header values are more likely to wind up in passive / middle man / etc logs. Existing approaches require active techniques, and so (partially) reduce the fingerprinting risk.

The most on point issue I can find addressing this issue is https://github.com/httpwg/http-extensions/issues/215, but this isn't quite on point (does not address increased risk from moving to passive collection).

I see the text added / modified in https://github.com/httpwg/http-extensions/commit/2ba199856ae926284fe868de7c257009f01fe3b8 that mentions that "implementors can do otherwise for privacy", but PING is uncomfortable with such text ( such text dissolve the point of the standard; a standard that says "its w/in this standard to vary arbitrarily", then all that is introduced is web compatibility problems for privacy oriented parties).

* What discussion has been had regarding increased information leak into logs? 
* What measurements / data exists to suggest this is not a problem?

mnot:
HI @snyderp. Be aware that people participate in the IETF as individuals, not organisations; if people in PING (aside for other readers: this is the W3C Privacy Interest Group) want to engage here, they'll need to do so individually. 

My understanding is that the current thinking around CH considers the requirement for a site to request a specific CH (in the form of the Accept-CH) as effectively "converting" passive collection to active collection; a researcher, for example, can measure how many sites are collecting such metrics, and a browser can alert the user when such collection is taking place.

There was a substantial amount of discussion around this, including in #215, #372 and elsewhere.

Are you questioning that, or are you only focusing on information leakage into logs? 

Regarding the latter, can you illustrate how that would happen, and what the impact would be?

snyderp:
@mnot thank you for your reply.

> My understanding is that the current thinking around CH considers the requirement for a site to request a specific CH (in the form of the Accept-CH) as effectively "converting" passive collection to active collection; a researcher, for example, can measure how many sites are collecting such metrics, and a browser can alert the user when such collection is taking place.

Researchers and clients already do exactly the things you mentioned.  CH doesn't help make preventing / measuring / notifying of FP & tracking any easier.  I'm similarly not familiar (and don't see any related comment in CH tagged issues) to deprecate / remove the existing methods for retrieving the same information, so the responsibility / burden on privacy focused parities is strictly increased.

> There was a substantial amount of discussion around this, including in #215, #372 and elsewhere.

[...]

Issue #716, opened 2018-11-06
Consider a `Sec-CH-` prefix for client hint headers


mikewest:
In https://github.com/w3ctag/design-reviews/issues/320, @annevk (re-)raised the question of Client Hints' integration with CORS. One suggestion in that thread is to prefix headers with `Sec-CH-`, which on the one hand makes them trivial to add _en masse_ to the CORS-safelisted list, and on the other prevents JavaScript from setting their values (which in turn limits the risk associated with adding them to the CORS-safelisted list in the first place).

I have two concrete proposals for hints ([`UA-*`](https://github.com/mikewest/ua-client-hints) and [`Lang`](https://github.com/mikewest/lang-client-hint)) that adopt this pattern. Perhaps it's one that could be baked more deeply into the Client Hints infrastructure?

/cc @igrigorik @yoavweiss @arturjanc

reschke:
Please no additional overloading of names.

yoavweiss:
> Please no additional overloading of names.

OK. Can you expand on that?

reschke:
Names are names. I realize that there's a precedent with "Sec-". But it really doesn't scale.

If we're serious about this, it needs to be done with a plan that explains how this would work with the next keyword being added. In particular, if the "next" extension is completely orthogonal to this one.

mikewest:
`Sec-` exists, and has meaning (at least in browsers). Are you objecting to that in itself? Or are you objecting to adding an additional `CH-`?

reschke:
I object to this unless we have a documented and agreed-upon plan how this scheme would work with future keywords.

mikewest:
`Sec-CH-Next-`. :)

reschke:
What if "Next" doesn't have the properties of "Sec"?

yoavweiss:
The reason we realize now that it is in fact needed is that we want client-hints to get a free-pass when it's comes to the CORS safelist. That is, we don't want to add every new header to the list.
Having a prefix will enable us to give all these headers the same treatment, but will also enable us to have a clearer mind when doing so, because the prefix reduces the probability that some existing server is already using those request header values in their logic in an unpredicted way.

mikewest:
> we want client-hints to get a free-pass when it's comes to the CORS safelist

FWIW, I don't actually care about this. It's fine with me if Fetch just lists all the headers, and y'all have to send PRs to Fetch to add them. :)

[...]

PR #776, opened 2019-03-11
Add Sec- prefix to security considerations

yoavweiss:
As discussed in https://github.com/httpwg/http-extensions/issues/716 and https://github.com/httpwg/http-extensions/pull/775, this PR adds a recommended `Sec-` prefix to Client Hints request header values.

Open questions:
* Should it be `Sec-CH`?
* Should it be a `MUST`?

/cc @igrigorik @reschke @mnot @mikewest @arturjanc


draft-ietf-httpbis-client-hints.md
```diff
@@ -266,6 +267,11 @@ Implementers ought to consider both user and server controlled mechanisms and po
 Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persisted opt-in preferences when any one of site data, browsing history, browsing cache, or similar, are cleared.
 
 
+## Backend Security
+Adding new request headers increases the risk that legacy server systems already use those values for a different purpose. Changing the request header values such legacy systems receive may result in server bugs.
```
igrigorik: Hmm, I believe a big part of the motivation for `Sec` prefix is that it's a reserved prefix that only the browser can emit, which seems like something worth mentioning here? This allows the servers to treat such headers with different policies; have some confidence in the data being delivered.


> Adding new request headers introduces potential deployment risks. For example, some servers may already use the same header for a different purpose, which may result in conflicts or other undesired behaviors.
>
>While this risk is significantly mitigated by the opt-in mechanisms of Client Hints, specifications and features relying on Client Hints should use the `Sec-` prefix for request header names. User agents reserve this namespace for headers that can only be emitted by the browser and thus can guarantee and enforce the type and format of the data communicated in the header value fields.

wdyt, warmer?

arturjanc:
My opinion doesn't carry a lot of weight here, but I like both `MUST` and `Sec-CH` (or potentially some other namespace-like thing, e.g. `Sec-Hint`) -- this mirrors what @mikewest  did for `Sec-Fetch-*`.

Ilya's suggestion also LGTM.

draft-ietf-httpbis-client-hints.md
yoavweiss: Yup, SGTM

yoavweiss:
Apologies for my slowness on this. Adopted Ilya's suggestion.
I know that @reschke was wary of adding more namespace like things though. 
@reschke - can you repeat your concerns around this?

reschke:
Well, the discussion is in #716 - maybe this is a good topic for the WG meetings in Montreal?

yoavweiss:
OK, I'll revive the discussion there

yoavweiss:
Fixes #767 and #716

draft-ietf-httpbis-client-hints.md
```diff
@@ -184,6 +215,22 @@ Implementers ought to consider both user and server controlled mechanisms and po
...
+
+The former concern is a deployment risk, slightly mitigated by the opt-in mechanisms of Client Hints.
+
+The latter is a security concern. CORS {{FETCH}} would have helped to eleviate the latter concern, but for performance purposes, we want to make sure Client Hints request headers are considered CORS-safe.
+
+
+In order to address both of these concerns, specifications and features relying on Client Hints MUST use the "Sec-" prefix for request header names. User agents reserve this namespace for headers that can only be emitted by the browser and thus can guarantee and enforce the type and format of the data communicated in the header value fields. That restriction also means that servers are not likely to be using such prefixed-headers for other purposes.
+
+Features relying on Client Hints are also encouraged to add "CH-" to the header name, in order to make it easier to distinguish Client Hint request headers from others. Servers and intermediaries handling those headers should take special care when storing them, as they are likely to contain entropy that can contribute to user fingerprinting.
```
yoavweiss: @sleevi - can you take a look to see that this language reflects what you had in mind when we discussed this at the IETF105 HTTPWG meeting?

igrigorik: Deployment of new request headers requires several considerations:

- Potential conflicts due to existing use of field name
- Properties and validation of communicated data in field value

Specifications and features relying on Client Hints MUST use the "Sec-" prefix for request header names. User agents reserve this namespace that can only be emitted by the browser and thus can guarantee and enforce the type and format of the data communicated in the header value fields. Further, specifications and features relying on Client Hints are encouraged to include "CH-" in the header name, in order to make it easier to distinguish Client Hint request headers from others. 

The combination of the above recommendations and requirements, plus the opt-in mechanism of Client Hints, helps minimize deployment risks. Further, these requirements enable server and intermediaries handling requests with Client Hints to identify such request headers and take precautions in how they are used and processed — e.g. opt out from storing them as they may expose information about the user.

draft-ietf-httpbis-client-hints.md
```diff
@@ -216,20 +216,14 @@ Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persis
...
+Deployment of new request headers requires several considerations:
 
-  - Will user-agent emitted values impact current servers which could be using these headers for a different purpose?
-  - Can malicious content impact servers on the user's network by crafting such request headers with malicious header values?
+  - Potential conflicts due to existing use of field name
+  - Properties of the data communicated in field value
 
-The former concern is a deployment risk, slightly mitigated by the opt-in mechanisms of Client Hints.
+In order to avoid conflicts and provide some guaranties regarding field values sent by user agents, specifications and features relying on Client Hints MUST use the "Sec-" prefix for request header names. User agents reserve this namespace to headers that can only be emitted by the browser. Therefore, they can guarantee and enforce the type and format of the data communicated in the header value fields. Further, specifications and features relying on Client Hints are encouraged to include "CH-" in the header name, in order to make it easier to distinguish Client Hint request headers from others.
```
sleevi: So, the first half of your recommendation, the "Sec-" prefix, is immediately explained as to why such a requirement exists.

However, the second half, "CH-", does not explain _why_ it's useful to distinguish Client Hint request headers from others.

**BEGIN**
Specifications and features relying on Client Hints MUST prefix request header names with a "Sec-" prefix. User agents, such as web browsers, ensure such prefixed headers may only be emitted by the browser, and application content, such as scripts, are forbidden from emitting such headers {{FETCH}}. This provides assurance that, when transmitted by user agents, that both the type and format of data communicated in such request headers are enforced by the user agent.

Additionally, specifications and features relying on Client Hints are RECOMMENDED to include the "CH-" in the header name, for example, in the form `Sec-CH-Foo` for a hypothetical Client Hint `Foo`. Client Hints provide additional information about a user agent [1] that previously may only have been accessible to scripts or content [2], and not necessarily transmitted in requests. Servers that wish to minimize the data collected in intermediate steps may wish to filter out all unrecognized or unprocessed "CH-" values from its records or intermediate steps, and a consistent use of "CH-" helps facilitate filtering new and unrecognized Client Hint Values.
**END**

[1] I'm not sure if this is strictly about UAs?
[2] I'm not entirely happy with that wording, but *shrug*

However, from our IETF 105 conversation, @mnot aptly pointed out that such concerns arguably apply to any/all new headers, and so perhaps the 'CH-' infix may be entirely unnecessary, because it's a generalized problem. The language I offer above attempts to provide explanation/justification for the RECOMMENDation, but I could also see with doing away with 'CH-' entirely.

yoavweiss:
@igrigorik - integrated your suggestions and then suggestions from @sleevi. PTAL?

yoavweiss:
mnot@ - Friendly ping :) Do you think the "CH-" bit is required?

draft-ietf-httpbis-client-hints.md
```diff
@@ -169,6 +200,17 @@ Implementers ought to consider both user and server controlled mechanisms and po
 Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persisted opt-in preferences when any one of site data, browsing history, browsing cache, or similar, are cleared.
 
 
+## Deployment and Security Risks
+Deployment of new request headers requires several considerations:
+
+  - Potential conflicts due to existing use of field name
```
mnot: That's managed by registering the header field, and is well-documented in HTTP. Why is this important to call out here?

draft-ietf-httpbis-client-hints.md
```diff
@@ -169,6 +200,17 @@ Implementers ought to consider both user and server controlled mechanisms and po
...
+  - Properties of the data communicated in field value
+
+Specifications and features relying on Client Hints MUST prefix request header names with a "Sec-" prefix. User agents, such as web browsers, ensure such prefixed headers can only be emitted by the browser, and application content, such as scripts, are forbidden from emitting such headers {{FETCH}}. This provides assurance that, when transmitted by user agents, the type, format and value of data communicated in such request headers are enforced by the user agent.
+
+Additionally, specifications and features relying on Client Hints are RECOMMENDED to include the "CH-" in the header name, for example, in the form Sec-CH-Foo for a hypothetical Client Hint Foo. Client Hints can provide additional information about the user that previously only have been accessible to scripts or content, and not necessarily transmitted in requests. Servers that wish to minimize the data collected in intermediate steps MAY filter out all unrecognized or unprocessed "CH-" values from its records or intermediate steps, and a consistent use of "CH-" helps facilitate filtering new and unrecognized Client Hint Values.
```
mnot: Instead of this graph, I'd say something like:

By convention, request headers that are client hints use a `CH-` prefix, to make them easier to identify as using this framework; for example, `Sec-CH-Foo`. Doing so makes them easier to identify programmatically (e.g., for stripping unrecognised hints from requests by privacy filters).

draft-ietf-httpbis-client-hints.md
```diff
@@ -169,6 +200,17 @@ Implementers ought to consider both user and server controlled mechanisms and po
 Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persisted opt-in preferences when any one of site data, browsing history, browsing cache, or similar, are cleared.
 
 
+## Deployment and Security Risks
+Deployment of new request headers requires several considerations:
+
+  - Potential conflicts due to existing use of field name
+  - Properties of the data communicated in field value
+
+Specifications and features relying on Client Hints MUST prefix request header names with a "Sec-" prefix. User agents, such as web browsers, ensure such prefixed headers can only be emitted by the browser, and application content, such as scripts, are forbidden from emitting such headers {{FETCH}}. This provides assurance that, when transmitted by user agents, the type, format and value of data communicated in such request headers are enforced by the user agent.
```
mnot: You're going to get pushback that MUST is only for targeting implementations, not specifications. Make this advisory.

draft-ietf-httpbis-client-hints.md
yoavweiss: The example I had in mind were the conflicts @mikewest ran into when deploying the initial version of `upgrade-insecure-requests` as `https`. I believe that resulted in some server side conflicts with some servers that already used that name internally, and read all incoming request headers to be global variables. (@mikewest - please correct me if I remember that wrong)

draft-ietf-httpbis-client-hints.md
yoavweiss: OK

draft-ietf-httpbis-client-hints.md
yoavweiss: :+1:

yoavweiss:
Friendly ping! :)

draft-ietf-httpbis-client-hints.md
mikewest: Some versions of PHP do strange things with certain headers. `https` was one of them. I'm not sure that has much weight in either direction, except as an anecdote that gave me a very bad week a few years ago. :)

mnot: Sure. I just don't know why we're doing this specifically here, rather than relying on the generic HTTP text (DRY and all that). Not a huge deal, just struck me as odd.

draft-ietf-httpbis-client-hints.md
```diff
@@ -169,6 +200,17 @@ Implementers ought to consider both user and server controlled mechanisms and po
 Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persisted opt-in preferences when any one of site data, browsing history, browsing cache, or similar, are cleared.
 
 
+## Deployment and Security Risks
+Deployment of new request headers requires several considerations:
+
+  - Potential conflicts due to existing use of field name
+  - Properties of the data communicated in field value
+
+Specifications and features relying on Client Hints are advised to prefix request header names with a "Sec-" prefix. User agents, such as web browsers, ensure such prefixed headers can only be emitted by the browser, and application content, such as scripts, are forbidden from emitting such headers {{FETCH}}. This provides assurance that, when transmitted by user agents, the type, format and value of data communicated in such request headers are enforced by the user agent.
```
mnot: Once again -- this text assumes that EVERY Client Hint is designed such that making it settable from code is an attack, and that there isn't any value in allowing it happen. I'd be much more comfortable saying that specs defining (not relying upon, btw) a Client Hint _carefully consider_ whether it should be exposed to code, and if it's a risk, to prefix with `Sec-`.

draft-ietf-httpbis-client-hints.md
```diff
@@ -169,6 +200,17 @@ Implementers ought to consider both user and server controlled mechanisms and po
...
+
+By convention, request headers that are client hints use a CH- prefix, to make them easier to identify as using this framework; for example, Sec-CH-Foo. Doing so makes them easier to identify programmatically (e.g., for stripping unrecognised hints from requests by privacy filters).
```
mnot: ... are encouraged to use...?

yoavweiss: :+1:

draft-ietf-httpbis-client-hints.md
yoavweiss: changed to "Specifications and features ... are advised to carefully consider prefixing..."

yoavweiss:
mnot@ - apologies for the delay. Addressed your comments. PTAL?

draft-ietf-httpbis-client-hints.md
```diff
@@ -169,6 +200,17 @@ Implementers ought to consider both user and server controlled mechanisms and po
 Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persisted opt-in preferences when any one of site data, browsing history, browsing cache, or similar, are cleared.
 
 
+## Deployment and Security Risks
+Deployment of new request headers requires several considerations:
+
+  - Potential conflicts due to existing use of field name
+  - Properties of the data communicated in field value
+
+Specifications and features relying on Client Hints are advised to carfully consider prefixing request header names with a "Sec-" prefix. User agents, such as web browsers, ensure such prefixed headers can only be emitted by the browser, and application content, such as scripts, are forbidden from emitting such headers {{FETCH}}. This provides assurance that, when transmitted by user agents, the type, format and value of data communicated in such request headers are enforced by the user agent.
```
martinthomson: carfully?

Also, the number, of, commas, here, is, starting to, get tiresome.  Maybe there are two sentences in that second one.

draft-ietf-httpbis-client-hints.md
martinthomson: So Yoav explained that the main reason here is that the reason for the Sec- prefix was to avoid problems like the one where header fields are turned into environment variables in a non-injective fashion by some frameworks.  So you get the value of the "Foo" header field mixed in with the "foo" query parameter.  That motivates **a** prefix, but it doesn't motivate the "Sec-" one.

I hadn't heard that reason properly before.  It took me a while to internalize it.

And then it didn't make sense.  The more I think about this, the more that I think that the "CH-" prefix is useful, but the "Sec-" one isn't.  On the contrary, I see value in giving script the ability to query for different content using these header fields.  For one, it might allow an app that is preparing for going offline in the browser the ability to request content based on anticipated rather than current needs.

For example, assuming for a moment that I thought `Save-Data` was useful (it's not), then the content could request the lightweight content that this targets, to avoid expending resources on the fetch or storage.  At that point, I don't have to have an opinion on `Save-Data` from the browser perspective, it might as well be yet another proprietary header field.

draft-ietf-httpbis-client-hints.md
mnot: I think it's much more clear to say something like:

Authors of new Client Hints are advised to carefully consider whether they should be able to be added by client-side content (e.g., scripts), or whether they should be exclusively set by the user agent. In the latter case, the `Sec-` prefix on the header field name has the effect of preventing scripts and other application content from setting them in many user agents. See {{FETCH}} for more information.

martinthomson: That is definitely better phrasing.  Assuming that you still want to do this based on [my earlier comments](https://github.com/httpwg/http-extensions/pull/776#discussion_r342875878).

yoavweiss: Oops regarding "carfully"...

Otherwise, I like @mnot's proposed text, assuming we are fine with user-added Client-Hints. I need to give them some more thought.

arturjanc:
The `Sec-` prefix guarantees that application developers are not currently sending requests with that header to their server (i.e. they're not adding it in XHR/Fetch because it's a forbidden header name.)  If a browser starts sending a new header without the prefix, it can result in unexpected behavior if the header name clashes with a custom header already in use by an application. In some cases this can create a security issue, e.g. if an application uses the presence of such a header to identify trusted requests -- this approach is a relatively common CSRF protection because the server can be sure that requests with a custom header were sent same-origin, or the server had opted into them in response to a CORS preflight.

I don't feel very strongly about this particular case, but in general I'm worried about browsers sending new unprefixed headers by default for this reason.

arturjanc:
I believe @annevk had some comments about similar situations in the past.

mnot:
@arturjanc that basically means we can't define any new request HTTP headers, which is pretty demonstrably untrue. What we shouldn't do is choose especially generic names, like `HTTPS`. Don't generalise from one bad experience.

martinthomson:
Also, the establishment of the convention to use a "CH-" prefix would seem to avoid that problem amply.  But @mnot's point is the stronger one.  I realize that having a standard bulldoze your established use of "My-Header" is an experience people are reluctant to be exposed to the possibility of, but they could always [register header field names](https://www.iana.org/assignments/message-headers/message-headers.xhtml).

annevk:
Each new request header that a browsers is to transmit cross-origin and is not prefixed with `Sec-` (or subject to a CORS preflight) would be another enshrined exception to the same-origin policy and a possible security issue until servers are updated to expect it.

Furthermore, some new request headers are not supposed to be spoofable same-origin and for those the prefix is important too. We have not had to update https://fetch.spec.whatwg.org/#forbidden-header-name for about a decade now and it would be nice to keep it that way.

martinthomson:
(OK, disclaimer: I'm not 100% clear on the model for cross-origin use of client hints then, but this is my operating model.  If this model is incorrect somehow, I think that the fetch changes here are what needs adjusting...)

Given that these aren't sent unless the server specifically requests them, I think that the cross-origin concern isn't that significant.  Rather than insisting on a one-by-one exception, a systemic exception for client hints would be preferable.

The way that this draft models CH header fields is that the site collects the information and injects the header field in every request as it leaves for the server (as if by service worker fetch interception).  The browser is just helping out.  If we are able to adhere to that model, then allowing insertion is different than header fields like `User-Agent` or `Origin` or other security-critical fields.

If this model is correct, then it doesn't make sense for these to be forbidden or for them to require special processing for exemption.  Note that this changes how the information needs to be generated (it can't access privileged information; that's already in this doc), and it means that these are treated like any other header field.

As for concerns about interaction with ambient authority that a server might associate with the request, I think that I will point back at the bit where the server explicitly asked for the hint to be provided.  I concede that cross-resource concerns might be a problem (resource A asks, resource B is unprepared), but the set of fields will be small, carefully groomed by browser makers, mostly prefixed with `CH-`, and carrying information that is not obviously useful in security-critical decisions.

annevk:
Agreed that if the server asks for the browser to set these headers on subsequent requests the cross-origin concern does not apply. However, that does assume these headers are not attacker-controlled and depending on when they are set (and the ask has been that they are set early to expose them to service workers as normal headers) they would also have to bypass CORS and we'd have to ensure they cannot be manipulated, etc.

And writing that down I realize there's a security problem there as now A can figure out that B requested client hints (and see what hints B will get).

Also, does the spoofing concern not apply?

It's also not entirely clear if server opt-in is tenable as we've seen requests for client hint like data to be available on the first request.

yoavweiss:
> Given that these aren't sent unless the server specifically requests

That is true for the same origin case, but not for the cross-origin one. For privacy purposes, we've limited the Client Hints opt-in to top-level documents, so a cross-origin resource cannot opt-in to client hints. And with Feature Policy, the top-level document can delegate hints to certain origins.

That means that those origins will get the hints without opting into them, which increases the risk.

At the same time, we want to avoid Client Hints request headers from triggering preflights, while being sure that it's indeed safe to do so.

martinthomson:
> A can figure out that B requested client hints (and see what hints B will get).

A seeing the values isn't an issue as it is A that put the hints there (and A sees the same values).

I don't see A having influence over values (viewport width maybe...).

B being unprepared for the hints does seem to be a problem.  I know that avoiding preflight is appealing from a performance perspective, but that's how we've decided to operate for these things.  Because A is essentially forcing the inclusion of header fields into cross-origin requests, I think that we have a few options:

1. perform some sort of analysis for each client hint to verify that no server conditions behaviour on the newly introduced header field in a way that would expose security-critical information

2. preflight always 

3. find a better alternative to preflights (origin policy, the TLS extension, etc...)

4. pretend that `Sec-`-prefixed header fields aren't subject to these same concerns

5. require the use of `CH-` prefixes and perform generic analysis

FWIW, for the first reason, we can't remember that B was OK with hints when it was top-level or framed in C because that lets A observe that B has been visited elsewhere.

arturjanc:
> * pretend that `Sec-`-prefixed header fields aren't subject to these same concerns

I'm not sure we need to do a lot of pretending here ;-) We know that applications can't get `Sec-`prefixed headers with ambient authority, so developers don't use them in their applications (unlike other headers which may be set same-origin or allowed via a response to a preflight). I believe delivering hints with a `Sec-` prefix would indeed avoid the class of problems discussed above; is there a specific issue that you're still worried about in this case?

yoavweiss:
@martinthomson - does @arturjanc's response answer concerns you had around option (4)?
As I'm not sure how to go about either (1) or (5), (2) would be a significant performance regression and (3) seems somewhat orthogonal, (4) seems like the best path forward.

yoavweiss:
At IETF 106, we agreed that it might be better to modify this so that the infrastructure draft indicates that features need to make sure that adding new request headers would be safe for servers. Individual features would then be able to decide if they enforce `Sec-`, live with a preflight and its performance regressions, or otherwise add their request headers to the CORS safe list. An alternative PR outlining that will follow.

yoavweiss:
> At IETF 106, we agreed that it might be better to modify this so that the infrastructure draft indicates that features need to make sure that adding new request headers would be safe for servers. Individual features would then be able to decide if they enforce `Sec-`, live with a preflight and its performance regressions, or otherwise add their request headers to the CORS safe list. An alternative PR outlining that will follow.

@martinthomson - looking at the current language, it seems to well-align with that decision without much modification. Can you take a look and let me know if you agree?

yoavweiss:
@martinthomson - friendly ping! :)

draft-ietf-httpbis-client-hints.md
martinthomson: Yeah, Mark's text seems fine to me.

yoavweiss:
Post holidays ping! :)

draft-ietf-httpbis-client-hints.md
martinthomson: > Using the "Sec-" prefix signals to servers that the user agent - and not application content - generated the values.

draft-ietf-httpbis-client-hints.md
```diff
@@ -196,12 +219,26 @@ Servers SHOULD take that into account when opting in to receive Client Hints, an
...
 Due to request byte size increase, features relying on this document to define Client Hints MAY consider restricting sending those hints to certain request destinations {{FETCH}}, where they are more likely to be useful. 
 
+
+## Deployment and Security Risks
+Deployment of new request headers requires several considerations:
+
+  - Potential conflicts due to existing use of field name
+  - Properties of the data communicated in field value
+
+Authors of new Client Hints are advised to carefully consider whether they should be able to be added by client-side content (e.g., scripts), or whether they should be exclusively set by the user agent. In the latter case, the Sec- prefix on the header field name has the effect of preventing scripts and other application content from setting them in user agents.
```
yoavweiss: @martinthomson I incorporated mnot's text (I thought I done so earlier :/) as well as your suggestion. PTAL?

martinthomson: WFM, one tweak below.

draft-ietf-httpbis-client-hints.md
```diff
@@ -212,7 +212,8 @@ Deployment of new request headers requires several considerations:
...
+Authors of new Client Hints are advised to carefully consider whether they should be able to be added by client-side content (e.g., scripts), or whether they should be exclusively set by the user agent. In the latter case, the Sec- prefix on the header field name has the effect of preventing scripts and other application content from setting them in user agents. 
+Using the "Sec-" prefix signals to servers that the user agent - and not application content - generated the values. See {{FETCH}} for more information.
 
 By convention, request headers that are client hints are encouraged to use a CH- prefix, to make them easier to identify as using this framework; for example, Sec-CH-Foo. Doing so makes them easier to identify programmatically (e.g., for stripping unrecognised hints from requests by privacy filters).
```
martinthomson: By convention, request headers that are client hints are encouraged to use a CH- prefix, to make them easier to identify as using this framework; for example, CH-Foo or, with a "Sec-" prefix, Sec-CH-Foo. Doing so makes them easier to identify programmatically (e.g., for stripping unrecognised hints from requests by privacy filters).

draft-ietf-httpbis-client-hints.md
yoavweiss: Thanks, added!

PR #777, opened 2019-03-11
Rename Cache header to Cache-Status

ghedo:
Cache-Status seems to better describe the information this header carries
and is potentially less confusing. It also seems more consistent with other
Cache-related headers (e.g. Cache-Control).

FWIW, some variants of *-Cache-Control are also somewhat popular in the
wild, e.g. the NGINX caching guide [0] uses X-Cache-Status and Cloudflare
uses CF-Cache-Control (though I'm not sure popularity matters much, since
either way servers will need to be updated).

[0] https://www.nginx.com/blog/nginx-caching-guide/

---

@mnot As proposed a while ago on the mailing list https://lists.w3.org/Archives/Public/ietf-http-wg/2018JulSep/0318.html Happy to discuss this more in Prague.

### DIFF
--- a/draft-ietf-httpbis-cache-header.md
+++ b/draft-ietf-httpbis-cache-header.md
@@ -1,6 +1,6 @@
 ---
-title: The Cache HTTP Response Header
-abbrev: Cache Header
+title: The Cache-Status HTTP Response Header
+abbrev: Cache-Status Header
 docname: draft-ietf-httpbis-cache-header-latest
 date: 2019
 category: std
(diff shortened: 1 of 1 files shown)


OlliV:
I support the one that is mentioned in the draft because I don't see how it could be misleading. `X-Something-Cache` but I don't think the popularity matters because everybody will need to do some adjustment if they want to go for this.

wimleers:
The current name is also clear to me. Can you explain what you find confusing about `Cache`? 

I think it can be argued that `Cache-Status` would be more confusing since this header effectively is a log of _actions_ by intermediaries, it doesn't convey any status. At most, it helps explain the values of the HTTP response.

A strong potential reason I can see to rename this to `Cache-Status` is to convey this exists to provide cache debugging information just like `Proxy-Status` exists to provide proxy debugging information (`Proxy-Status` is in progress at https://tools.ietf.org/html/draft-nottingham-proxy-status-00). Similarity in naming should then probably also imply similarity in structure though.

PiotrSikora:
@wimleers `Cache[-Status]` and now-adopted `Proxy-Status` both use parameterised list from [draft-ietf-httpbis-header-structure](https://tools.ietf.org/html/draft-ietf-httpbis-header-structure).

Now that `Proxy-Status` has been adopted by the WG, I'm also in favor of renaming `Cache` to `Cache-Status`.

wimleers:
Thanks, then the last paragraph of my comment applies, so I now agree with renaming :)

tfpauly:
Singapore: Already renamed, closing

PR #796, opened 2019-05-02
proxy-status: fix section reference formatting

reschke:
See <https://www.rfc-editor.org/styleguide/part2/>, "Referring to specific sections within another document"


mnot:
Thanks!

Issue #594, opened 2018-04-25
Setting samesite cookies when not same site context


aliams:
What happens when a page tries to set a samesite cookie when we are not in a samesite context? For instance, a top-level a.com has an iframe to b.com and b.com tries to set a cookie with the samesite attribute. It sounds like we would want the cookie to not be writeable since it's not readable in that case.

cc @mikewest

fmarier:
In Firefox, we [prevent same-site cookies from being set in a third-party context](https://bugzilla.mozilla.org/show_bug.cgi?id=1452496).

johnwilander:
I vaguely remember discussions on this, i.e. whether the whole cookie should be dropped or just the SameSite attribute. Anyone on this thread who remembers?

ericlaw1979:
Blocking SameSite cookie creation in cross-origin POST contexts means that an OAuth cookie that is set as a result from a POST from the auth provider site will fail. This seems undesirable.

mikewest:
I think it's pretty clear that we ought to be blocking the creation of `SameSite={Lax,Strict}` cookies in cross-site subresource requests. Firefox does that today, and Chrome [just landed a patch](https://chromium-review.googlesource.com/c/chromium/src/+/1534237) to do the same.

Navigations are more interesting. We send `SameSite=Lax` cookies along with "safe" HTTP methods (e.g. `GET`, but not `POST`). Chrome's implementation applies the same restriction to `SameSite=Lax` creation. I assumed that Firefox's implementation did the same, but @ericlaw1979 suggests in https://bugs.chromium.org/p/chromium/issues/detail?id=837412#c18 that it doesn't. That seems strange to me, though perhaps justifiable given the use case Eric's pointing to.

I think I could live with a world in which top-level navigations (regardless of method) are empowered to _create_ `SameSite` cookies (but keep the rules around _sending_ them as they are). Would that address your concerns, Eric? Does it open us up to attacks I'm not thinking about?

/cc @mozmark @morlovich

ericlaw1979:
I've built a reduced repro for this scenario at https://webdbg.com/test/cookie/samesite/post.aspx .

Firefox Nightly ignores the attempt to set the SameSite cookie for the cross-origin image, but permits the cookie to be set during the form post.

mikewest:
> Firefox Nightly ignores the attempt to set the SameSite cookie for the cross-origin image, but permits the cookie to be set during the form post.

[...]

PR #800, opened 2019-05-02
RFC6265bis: Allow `SameSite` cookies to be set for all top-level navigation

mikewest:
Based on the discussion in httpwg/http-extensions#594, this patch aligns
the specification with Firefox's status quo behavior: allowing
`SameSite=*` cookies to be set from all top-level navigations, not just
navigations for which `SameSite=*` cookies would also be sent.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1463,7 +1463,7 @@ user agent MUST process the cookie as follows:
 
         Note: Top-level navigations can create a cookie with any `SameSite`
         value, even if the new cookie wouldn't have been delivered along with
-        the request had it existed.
+        the request had it already existed prior to the navigation.
 
     4.  Abort these steps and ignore the newly created cookie entirely.
 
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -2089,6 +2089,11 @@ Specification document:
 *  Introduced an explicit "None" value for the SameSite attribute:
    <https://github.com/httpwg/http-extensions/issues/788>
 
+## draft-ietf-httpbis-rfc6265bis-04
+
+*  Allow `SameSite` cookies to be set for all top-level navigations.
+   <https://github.com/httpwg/http-extensions/issues/594>
+
 # Acknowledgements
 {:numbered="false"}
 This document is a minor update of RFC 6265, adding small features, and
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1462,7 +1462,7 @@ user agent MUST process the cookie as follows:
         and continue processing the cookie.
 
         Note: Top-level navigations can create a cookie with any `SameSite`
-        value, even if the new cookie wouldn't have been delivered along with
+        value, even if the new cookie wouldn't have been sent along with
         the request had it already existed prior to the navigation.
 
     4.  Abort these steps and ignore the newly created cookie entirely.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1444,10 +1444,29 @@ user agent MUST process the cookie as follows:
...
-14. If the cookie's `same-site-flag` is not "None", and the cookie is being set
-    from a context whose "site for cookies" is not an exact match for
-    request-uri's host's registered domain, then abort these steps and ignore
-    the newly created cookie entirely.
+14. If the cookie's `same-site-flag` is not "None":
+
+    1.  If the cookie was received from a "non-HTTP" API, and the API was called
+        from a context whose "site for cookies" is not an exact match for
+        request-uri's host's regisgered domain, then abort these steps and
+        ignore the newly created cookie entirely.
```
morlovich: Does this mean that in a context that's not same-site-strict but same-site-lax, a same-site=lax cookie will not be settable by JavaScript? I am not sure such difference between JS and HTTP is desirable.

Hmm. I may be confused here between context's "site for cookies" and "request"'s client's "site for cookies" as used in section 5.2.

(Also a typo in line 1451, regisgered)

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1444,10 +1444,29 @@ user agent MUST process the cookie as follows:
...
+    2.  If the cookie was received from a "same-site" request (as defined in
+        {{same-site-requests}}), skip the remaining substeps and continue
+        processing the cookie.
+
+    3.  If the cookie's `same-site-flag` is "Lax", and the cookie was received
+        from a request which is navigating a top-level browsing context {{HTML}}
+        (e.g. if the request's "reserved client" is either `null` or an
+        environment whose target browsing context is a top-level browing context
+        {{FETCH}}), skip the remaining substeps and continue processing the
+        cookie.
```
morlovich: So I think this takes a different decision than the checking CL --- both in your and my takes on it --- on whether setting a strict cookie in a lax context is allowed.

draft-ietf-httpbis-rfc6265bis.md
mikewest: It's hard to be precise here, because this document doesn't have much relation to HTML. But, my goal is something like this:

When the `document.cookie` setter is executed, execute the "site for cookies" algorithm from https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-03#section-5.2.1. That'll give you either the top-level document's origin's host's registrable domain, or an empty string.

In the case we're debating, it'll give you the registrable domain of the document that was just navigated-to. That is, both `samesite=strict` and `samesite=lax` will be settable.

draft-ietf-httpbis-rfc6265bis.md
mikewest: Yes, I'm not sure why I wrote this. The proposal I made in the bug was that `SameSite=*` would be settable from a top-level navigation. Let me take another stab at it. :)

mikewest:
I've taken another pass at it, @morlovich. Mind taking another look?

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1444,10 +1444,28 @@ user agent MUST process the cookie as follows:
...
+        processing the cookie.
+
+    3.  If the cookie was received from a request which is navigating a
+        top-level browsing context {{HTML}} (e.g. if the request's "reserved
+        client" is either `null` or an environment whose "target browsing
+        context" is a top-level browing context), skip the remaining substeps
+        and continue processing the cookie.
+
+        Note: Top-level navigations can create a cookie with any `SameSite`
+        value, even if the new cookie wouldn't have been delivered along with
```
ericlaw1979:
```suggestion
        value, even if the new cookie wouldn't have been sent along with
```

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1444,10 +1444,28 @@ user agent MUST process the cookie as follows:
...
+        the request had it existed.
```
ericlaw1979:
```suggestion
        the request had it already existed prior to the navigation.
```

PR #804, opened 2019-05-03
Proxy-Status: remove error restriction in http_response_status.

PiotrSikora:


PiotrSikora:
This is effectively editorial, since the error restriction was removed in https://github.com/mnot/I-D/commit/1be85db7a3ae0c8c93b970d1d2946831d292308a, where this type was renamed from `origin_http_error_status` to `http_response_status`, but the description didn't reflect that change.

cc @mnot

PR #820, opened 2019-07-01
Structured Headers: Allow lists as dictionary values.

### LINKED ISSUE #816: Structured Headers: Lists as dictionary values.
_I kinda expect that this is a dupe, as I know there's been conversation on this topic (https://github.com/httpwg/http-extensions/issues/476 is close, for example). But I didn't find anything specifically on point, so I'm erring on the side of another issue._

The discussion of substructures in https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-10#appendix-B.2 is somewhat unsatisfying, as it seems there are a number of cases in which a little support for additional structure would be helpful. In particular, allowing the values of dictionary members to be lists would open up a syntax for a few of the policy headers I'm somewhat responsible for that seems difficult to replicate otherwise.

Even the example in that section:

> ```
> Example-Thing: name="Widget", cost=89.2, descriptions="foo bar"
> Example-Description: foo; url="https://example.net"; context=123,
>                      bar; url="https://example.org"; context=456
> ```

implicitly treats the `descriptions` member as a space-separated list which requires parsing logic above and beyond what structured headers themselves offer. It seems like a very common pattern, and one which would be hard to layer in after the fact in a v2 if we decide that syntax really would be nice.

I am 100% sure that y'all considered a syntax like:

```
Example-Dictionary-Header: ..., memberName=[thing1, thing2, thing3]
```

(which, of course, could also be used for [lists of lists](https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-10#section-3.3)). I'm not sure why you rejected it. The discussion in #476 seemed to be about the narrow value of allowing parsing without context. I happen to believe that that's actually valuable, but I also think there's both semantic and aesthetic value in explicitly demarcating lists as a primitive that can exist basically anywhere `sh-item` exists in the current spec.

I'm happy to send a PR and add tests to https://github.com/httpwg/structured-header-tests/ if you'd accept them, conceptually. :)

mikewest:
This patch allows 'inner-list' as a valid dictionary member value. Fixes #816.

### DIFF
--- a/draft-ietf-httpbis-header-structure.md
+++ b/draft-ietf-httpbis-header-structure.md
@@ -181,7 +181,7 @@ This section defines the abstract value types that can be composed into Structur
 
 ## Dictionaries {#dictionary}
 
-Dictionaries are ordered maps of key-value pairs, where the keys are short, textual strings and the values are items ({{item}}) or lists thereof ({{list}}). There can be one or more members, and keys are required to be unique.
+Dictionaries are ordered maps of key-value pairs, where the keys are short, textual strings and the values are items ({{item}}) or arrays thereof. There can be one or more members, and keys are required to be unique.
 
 Implementations MUST provide access to dictionaries both by index and by key. Specifications MAY use either means of accessing the members.
 
@@ -191,7 +191,7 @@ The ABNF for dictionaries in HTTP/1 headers is:
 sh-dictionary  = dict-member *( OWS "," OWS dict-member )
 dict-member    = member-name "=" member-value
 member-name    = key
-member-value   = sh-item | inner-list
+member-value   = sh-item / inner-list
 key            = lcalpha *( lcalpha / DIGIT / "_" / "-" )
 lcalpha        = %x61-7A ; a-z
 ~~~
--- a/draft-ietf-httpbis-header-structure.md
+++ b/draft-ietf-httpbis-header-structure.md
@@ -896,7 +896,7 @@ Structured headers intentionally limits the complexity of data structures, to as
 Sometimes, this can be achieved by creating limited substructures in values, and/or using more than one header. For example, consider:
 
 ~~~
-Example-Thing: name="Widget", cost=89.2, descriptions="foo bar"
+Example-Thing: name="Widget", cost=89.2, descriptions=foo;bar
 Example-Description: foo; url="https://example.net"; context=123,
                      bar; url="https://example.org"; context=456
 ~~~
--- a/draft-ietf-httpbis-header-structure.md
+++ b/draft-ietf-httpbis-header-structure.md
@@ -427,13 +427,13 @@ This section defines how to serialize and parse Structured Headers in HTTP/1 tex
 
 ## Serializing Structured Headers into HTTP/1 {#text-serialize}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-header-structure.md
```diff
@@ -191,15 +191,15 @@ The ABNF for dictionaries in HTTP/1 headers is:
 sh-dictionary  = dict-member *( OWS "," OWS dict-member )
 dict-member    = member-name "=" member-value
 member-name    = key
-member-value   = sh-item
+member-value   = sh-item | inner-list
```
mnot: ABNF alteration is `/` (sorry, I probably misled you into that)

draft-ietf-httpbis-header-structure.md
```diff
@@ -181,7 +181,7 @@ This section defines the abstract value types that can be composed into Structur
 
 ## Dictionaries {#dictionary}
 
-Dictionaries are ordered maps of key-value pairs, where the keys are short, textual strings and the values are items ({{item}}). There can be one or more members, and keys are required to be unique.
+Dictionaries are ordered maps of key-value pairs, where the keys are short, textual strings and the values are items ({{item}}) or lists thereof ({{list}}). There can be one or more members, and keys are required to be unique.
```
mnot: Hmm, I don't think the reference to `list` is appropriate here; probably just say "arrays thereof."

draft-ietf-httpbis-header-structure.md
mnot: Big question here is whether we should refactor List and ListOfLists to be just List with this kind of production as well -- but that's a separable issue.

phluid61:
What about the algorithms? Particularly the parsing one...

mnot:
^^ what @phluid61 said.

draft-ietf-httpbis-header-structure.md
mikewest: :+1:

draft-ietf-httpbis-header-structure.md
mikewest: :+1:

mikewest:
Ah. Yes. That was dumb. I took a first pass at the algorithms in https://github.com/httpwg/http-extensions/pull/820/commits/2ab04f852459cb135f65288ad37da592b06ac757. The core of it is to reuse "Parsing/Serializing a List"'s logic by passing in "Inner" or "Outer". I'm not sure if this makes things more or less complex to follow. :)

Two stylistic questions:

1.  Some places in the text use the "result of [ALGORITHM]". Others the "result of applying [ALGORITHM]", others "running [ALGORITHM]". Some pass in arguments, others leave them inferred. What would you like this patch to do? :)

2.  COMMA is spelled out in algorithms, while ";" is not (nor are "-", "_", "-", etc). There's probably a reason for this, but I don't understand it. Is it an ABNF grammar thing? Is COMMA actually a set of characters?

And, as noted in the issue, these algorithm rely on quite a bit of magical knowledge of type expectations. That makes me uncomfortable as someone who might someday need to write a parser in Chromium...

mnot:
OBE

Issue #827, opened 2019-07-03
Digest headers: typo in reference


ioggstream:
## I expect

references to FIPS180-3

## Instead

I have a reference to FIPS-180-3

ioggstream:
fixed in #829

PR #829, opened 2019-07-03
Fix: #827. typo in reference: FIPS-180 -> FIPS180

ioggstream:
## This PR

fixes typo in reference #827


martinthomson:
FYI, there is (at least) a 180-4 now.

PR #860, opened 2019-07-18
Clarify defined headers are response headers

### LINKED ISSUE #761: Client Hints doesn't distinguish clearly between request and response header fields
I've been reading Client Hints (both [draft-ietf-httpbis-client-hints-06](https://tools.ietf.org/html/draft-ietf-httpbis-client-hints-06) and the [version in this repository](https://github.com/httpwg/http-extensions/blob/master/draft-ietf-httpbis-client-hints.md)), and I'm having a little trouble understanding how things fit together in terms of which header fields are sent when and lead to what.

My confusion started when I looked at the [IANA Considerations](https://tools.ietf.org/html/draft-ietf-httpbis-client-hints-06#section-6) section and noticed (to my surprise) that `Accept-CH` was being registered as *both* a request header field *and* a response header field (when, having read the spec quickly, I was expecting it to be only a response header field, and wasn't sure what it would mean as a request header field).

Then I looked back over the spec, and noticed that section 2 of the spec is called [Client Hint Request Header Fields](https://tools.ietf.org/html/draft-ietf-httpbis-client-hints-06#section-2) but that the entirety of the section appears (I think) to be about *response* header fields.

I think it would help the spec be easier to understand if it:
* more carefully distinguished between request header fields and response header fields, and
* more clearly explained what the `Accept-CH` *request* header field is.

yoavweiss:
Closes #761 

@dbaron - can you take a look and see if that clarifies things?

### DIFF
--- a/draft-ietf-httpbis-client-hints.md
+++ b/draft-ietf-httpbis-client-hints.md
@@ -103,9 +103,13 @@ When presented with a request that contains one or more client hint header field
 
 Further, depending on the hint used, the server can generate additional response header fields to convey related values to aid client processing.
 
-### Advertising Support via Accept-CH Header Field {#accept-ch}
+# Advertising Server Support
 
-Servers can advertise support for Client Hints using the Accept-CH header field or an equivalent HTML meta element with http-equiv attribute ({{HTML5}}).
+Servers can advertise support for Client Hints using the mechnisms described below.
+
+## The Accept-CH Response Header Field {#accept-ch}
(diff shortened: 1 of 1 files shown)


dbaron:
This seems helpful given what's currently in the spec.

However, I think some of the material I was talking about when I filed #761 is no longer there.  But I suspect that material may now be in other specs, and could perhaps still gain from being clarified at its current location.

PR #862, opened 2019-07-18
[Client Hints] Replace ABNF with Structured Headers

### LINKED ISSUE #700: Clarify behavior for multiple values for CH headers
Migrating [feedback from Kari Hurtta](https://lists.w3.org/Archives/Public/ietf-http-wg/2018JulSep/0137.html).. 

> On Wed, Jul 25, 2018 at 9:14 AM HTTP Working Group <ietf-http-wg@w3.org>
> wrote:
> 
> > |   If DPR occurs in a message more than once, the last value overrides
> > |   all previous occurrences.
> 
> So is that saying that if there is several DPR values on
> > DPR Header Field, last value is used ?
> >
> 
> Yes, that's the intent.
> 
> If that in interpreted only that if last DPR header field is used,
> > and header field is ignored if it does not fit
> >
> >    DPR = 1*DIGIT [ "." 1*DIGIT ]
> >
> > syntax, result is different is recipient combines multiple
> > DPR headers.
> 
> 
> The intent was to specify that last valid value (regardless if it is within
> a combined or separate header) is the one that should be used by the
> client. My read of the current wording is that it matches this intent, but
> I'll defer to the experts here on how to best word this if that's not the
> case.
> 
> Mark, Julian: any guidance on this one?
> 
> 
> > |   If <X> occurs in a message more than once, the last value
> > |   overrides all previous occurrences.
> >
> > So it should mention that on these header fields also last value
> > is used when values are ',' -separated.
> >
> 
> My understanding is that this is implicit, but if that's not the case,
> happy to rework the wording.

--- 

Julian:
> No, it's not implicit, so yes, a clarification would be good. I don't think we have good text yet.

yoavweiss:
Closes #700

@mnot - can you take a look?

### DIFF
--- a/draft-ietf-httpbis-client-hints.md
+++ b/draft-ietf-httpbis-client-hints.md
@@ -106,10 +106,10 @@ Further, depending on the hint used, the server can generate additional response
 ### Advertising Support via Accept-CH Header Field {#accept-ch}
 
 Servers can advertise support for Client Hints using the Accept-CH header field or an equivalent HTML meta element with http-equiv attribute ({{HTML5}}).
-Accept-CH is a Structured Header {{!I-D.ietf-httpbis-header-structure}}. Its value MUST be a list-of-tokens (Section 3.3 of {{!I-D.ietf-httpbis-header-structure}}) whose members are tokens (Section 3.9 of {{!I-D.ietf-httpbis-header-structure}}). Its ABNF is:
+Accept-CH is a Structured Header {{!I-D.ietf-httpbis-header-structure}}. Its value MUST be an sh-list (Section 3.1 of {{!I-D.ietf-httpbis-header-structure}}) whose members are tokens (Section 3.7 of {{!I-D.ietf-httpbis-header-structure}}). Its ABNF is:
 
 ~~~ abnf7230
-  Accept-CH = sh-list-of-tokens
+  Accept-CH = sh-list
 ~~~
 
 For example:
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-client-hints.md
```diff
@@ -106,9 +106,10 @@ Further, depending on the hint used, the server can generate additional response
 ### Advertising Support via Accept-CH Header Field {#accept-ch}
 
 Servers can advertise support for Client Hints using the Accept-CH header field or an equivalent HTML meta element with http-equiv attribute ({{HTML5}}).
+Accept-CH is a Structured Header {{!I-D.ietf-httpbis-header-structure}}. Its value MUST be a list-of-tokens (Section 3.3 of {{!I-D.ietf-httpbis-header-structure}}) whose members are tokens (Section 3.9 of {{!I-D.ietf-httpbis-header-structure}}). Its ABNF is:
 
 ~~~ abnf7230
-  Accept-CH = #field-name
+  Accept-CH = #sh-list-of-tokens
```
phluid61: Does this want the `#`?

yoavweiss: Nope. Thanks for catching that!

draft-ietf-httpbis-client-hints.md
```diff
@@ -109,7 +109,7 @@ Servers can advertise support for Client Hints using the Accept-CH header field
 Accept-CH is a Structured Header {{!I-D.ietf-httpbis-header-structure}}. Its value MUST be a list-of-tokens (Section 3.3 of {{!I-D.ietf-httpbis-header-structure}}) whose members are tokens (Section 3.9 of {{!I-D.ietf-httpbis-header-structure}}). Its ABNF is:
 
 ~~~ abnf7230
-  Accept-CH = #sh-list-of-tokens
+  Accept-CH = sh-list-of-tokens
```
mnot: `sh-list-of-tokens` hasn't been defined as ABNF; you probably need to use just `sh-list` and say its contents are tokens in prose.

yoavweiss: Thanks, fixed it!

PR #866, opened 2019-07-20
Add an Alt-Svc extension

yoavweiss:
In order to enable first view Client Hints with [HTTPSSVC](https://tools.ietf.org/html/draft-nygren-httpbis-httpssvc-01), this PR adds an Alt-Svc extension to Client Hints

### DIFF
--- a/draft-ietf-httpbis-client-hints.md
+++ b/draft-ietf-httpbis-client-hints.md
@@ -144,6 +144,13 @@ For example, based on the Accept-CH and Accept-CH-Lifetime example above, which 
 If Accept-CH-Lifetime occurs in a message more than once, the last value overrides all previous occurrences.
 
 
+### The "accept-ch" Alt-Svc extension
+
+Origins can advertise support for Client Hints using `accept-ch` parameter in their alternative service entry.
+
+Syntax:
+   accept-ch = sh-list
+
 ### Interaction with Caches
 
 When selecting an optimized response based on one or more Client Hints, and if the resource is cacheable, the server needs to generate a Vary response header field ({{RFC7234}}) to indicate which hints can affect the selected response and whether the selected response is appropriate for a later request.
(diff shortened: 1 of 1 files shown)


davidben:
We probably need to think a bit about the security model for sticking random things into HTTPSSVC by way of Alt-Svc. (I'm a little torn on whether HTTPSSVC should actually align with the Alt-Svc HTTP header. They do very similar things, but the authentication is very different.)

When the client fetches HTTPSSVC from DNS, there is no authentication from the actual origin server. Even if using DoH, DoH only authenticates the resolver. The resolver could still inject things. (I'm ignoring DNSSEC here as that's not really a thing on the web today.)

Of the features in HTTPSSVC right now, alias records and remote Alt-Svc are just funny CNAMEs. The resolver can deny you access to the server, up to the record's TTL, but that is already true. Injecting a bad QUIC advertisement is similar. Injecting an invalid ESNI key can likewise just deny access to the server or cause the client to encrypt the SNI badly. The DNS resolver, already learns the SNI, so that's consistent. (It would not be okay to, e.g., just use the ESNI key as-is to replace the TLS certificate because of this.)

Client hints affects the HTTP requests we'd actually send over TLS, so that has different properties to think through.

martinthomson:
I think that this can be deferred.  I appreciate the thinking behind this, but it seems like it might jeopardize the possible near-term success of CH.

yoavweiss:
related issue is #871

yoavweiss:
> I think that this can be deferred. I appreciate the thinking behind this, but it seems like it might jeopardize the possible near-term success of CH.

@martinthomson - Thanks. That's useful feedback!

yoavweiss:
Closing as it seems we can defer this to a later time and I've lost confidence that Alt-SVC is necessarily the right mechanism for this.

Issue #872, opened 2019-07-23
Use cases for Digest


ioggstream:
## I expect

More descriptive Digest use-cases.

## Note

Digest is used:

- [x] Digest is usually combined with some mechanism to protect HTTP metadata. 
This allows for L7 protections that transcend transport. See [draft-ietf-httpbis-digest-headers.md#L129](https://github.com/httpwg/http-extensions/blame/master/draft-ietf-httpbis-digest-headers.md#L129)

- [x] you could compose a resource from partial content 
   retrieved from different sources and validate 
  the resource's integrity before passing it to
  application for processing. See [draft-ietf-httpbis-digest-headers.md#L686](https://github.com/httpwg/http-extensions/blame/master/draft-ietf-httpbis-digest-headers.md#L686)

- [ ] in web payments APIs, including EU PSD2 directive via [http-signatures](https://tools.ietf.org/html/draft-cavage-http-signatures-11)
- [x] in MICE content-coding;
- [ ] in [webpackage](https://github.com/WICG/webpackage) (via MICE)

## Questions

- We can highlight more the concept of "Digest as a Signature building block" in the use-cases
- Should we reference other I-D like http-signatures (now under http-extension), MICE and webpackage in the spec
  to clarify use cases?

@jyasskin @royfielding issue reopened here from https://github.com/ioggstream/draft-polli-resource-digests-http/issues/15

LPardue:
We've made progress on this. My fear is that mention of other WIP specs like packaging and could tie us up. Therefore, I suggest spinning out the two remaining use cases into separate issues and asking the WG for feedback if we should work on them.

ioggstream:
imho the current setup is fine and we can close this issue without further actions.
The current reference to MICE algorithm is informative and is related to Trailers.

PR #873, opened 2019-07-23
Add use cases section.

ioggstream:
## This PR

Adds an `Use Case` section

#872


ioggstream:
I added the discussion we had on #872. Feel free to add whatever you think is appropriate.

ioggstream:
Addressed in #971

Issue #832, opened 2019-07-03
Digest: update references for SHA algorithms.


ioggstream:
## Question

Should we update the FIPS180 reference to the latest one?

## Context

NIST updated the document defining SHA-* algoritms.

The latest reference is

```
  FIPS180-4:
    title: NIST FIPS 180-4, Secure Hash Standard
    author:
      name: NIST
      ins: National Institute of Standards and Technology, U.S. Department of Commerce
    date: 2015-08
    target: https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf
```

digest-headers references preserve the ones cited in RFC3230 and subsequent.

LPardue:
cross-posting from elsewhere, I said

> Note that references to FIPS180-1 are due to historical reference, e.g. an algorithm that was removed from later drafts.

I think we probably want to use FIPS180-4 for SHA-256 and SHA-512, and FIPS180-1 for SHA-1.

martinthomson:
There is also [RFC 6234](https://tools.ietf.org/html/rfc6234).

PR #876, opened 2019-07-25
Fix: #832. Reference RFC6234 instead of FIPS180.

ioggstream:
## This PR

Simplifies references to sha algorithms pointing to RFC 6234
instead of FIPS180 documents.

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -26,6 +26,7 @@ author:
 
 normative:
   RFC1321:
+  RFC3174:
   RFC3230:
   RFC2119:
   RFC5789:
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -32,25 +32,12 @@ normative:
   RFC5843:
   RFC4648:
   RFC5234:
+  RFC6234:
```
martinthomson: You can drop this and use `{{!SHA2=RFC6234}}` in the first use of the reference.  You can use the same afterwards, or even just `{{!RFC6234}}`.

ioggstream: Cool, thanks! I'll use that syntax in an editorial review of the whole doc.

ioggstream:
@LPardue ping :)

draft-ietf-httpbis-digest-headers.md
```diff
@@ -334,11 +321,11 @@ Response:
 
   SHA:
   :  
-    * Description:  The SHA-1 algorithm [FIPS180-1].  The output of this
```
LPardue: Strictly, SHA-1 is specified in https://tools.ietf.org/html/rfc3174 

I got there by following the link from RFC 6234, both are still active RFCs.

ioggstream: 3174 is updated by 6234. Do you think we should change the PR and reference both?

LPardue: Personally I'd like to reference both.

"Updates" or "Updated by" is additive (in contrast, "Obsoletes" is replacement) so both documents are still active and IMO if we reference 3174 we avoid one level of indirection for people that want to see the active spec for SHA-1.

ioggstream: should be fixed now, WDYT?

Issue #828, opened 2019-07-03
Digest Headers: what about the other algorithms?


phluid61:
ADLER32 and CRC32c are in the registry, but not mentioned in the text (although they are covered by this blanket statement)

> ## Changes compared to RFC5843
>
> [...]
>
> The status for all other algorithms have been updated to "standard".

Is it ok to gloss them this way?

ioggstream:
@phluid61 thanks for your question! I really appreciate some guidance here!

What should we put there instead of `standard`?

Here's the rationale for the current choice:

- I wasn't able to find the RFCs registering CRC32c and ADLER32 in the IANA table;
- RFC5843 doesn't mention them;
- eg. for Adler32, [RFC3309 has no iana considerations](https://tools.ietf.org/html/rfc3309#section-4)

This draft:

- does not remove those adler32 and crc32c from https://www.iana.org/assignments/http-dig-alg/http-dig-alg.xhtml
- just updates their status to `standard`

Thanks for your time!

phluid61:
I'm pretty sure those RFCs predate standardised registration, so it's definitely going to be a mess.

I think you could mark ADLER32 as 'obsoleted' (see: https://tools.ietf.org/html/rfc4460#section-2.38 ) I think the references in the registry are still the most relevant.

CRC32c is 'standard', as far as I can see.

LPardue:
WDYT of adding a subsection to section 9 - "Obsolete ADLER32 Digest Algorithm" that cites Stream Control Transmission Protocol (SCTP) Checksum Change https://tools.ietf.org/html/rfc3309

ioggstream:
Ok for me. Do you agree to leave this to `-01` ?

mnot:
Please keep -00 as close to the draft we adopted as possible.

ioggstream:
Seems that ADLER32 is still used by various projects, like dcache.org, for a weak form of validation.

@paulmillar is mentioned in the IANA table too :) Can you share your view on the subject?

manger:
ADLER32 and CRC32c are well-defined, standardised, algorithms. But there is no spec specifically registering them for use in the HTTP Digest header. My judgement at the time (as the designated expert for that IANA table of HTTP Digest Algs) was that it was sufficient for the _algorithms_ to have specs, without demanding an extra spec just to added them to the table.

[...]

PR #877, opened 2019-07-25
Obsolete ADLER-32 but don't forbid it. It's now NOT RECOMMENDED #828 

ioggstream:
## This PR

Obsolete ADLER-32 but don't forbid it. #828

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -28,7 +28,9 @@ author:
 normative:
   RFC1321:
   RFC3174:
+  RFC1950:
   RFC3230:
+  RFC3309:
   RFC2119:
   RFC5789:
   RFC5843:
@@ -764,6 +766,12 @@ However, these rely on collision-resistance for their security proofs [CMU-83606
 The MD5 and SHA-1 algorithms are vulnerable to collisions attacks,
 so MD5 MUST NOT be used and SHA-1 is NOT RECOMMENDED.
 
+## Other deprecated algorithms
+
+The ADLER32 algorithm defined in [RFC1950] has been deprecated
+by [RFC3309] because under certain conditions it provides
+weak detection of errors and is now NOT RECOMMENDED.
+
 ## Digest for end-to-end integrity
 
 `Digest` alone does not provide end-to-end integrity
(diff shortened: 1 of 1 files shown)


ioggstream:
@manger will this work for you?

draft-ietf-httpbis-digest-headers.md
```diff
@@ -760,6 +768,20 @@ registry:
...
+Algorithm
+Values](https://www.iana.org/assignments/http-dig-alg/http-dig-alg.xhtml)
+registry:
+
+* Digest Algorithm: ADLER32
+* Description: The ADLER32 algorithm is a checksum specified in [RFC1950] "ZLIB Compressed Data Format".
+  The 32-bit output is encoded in hexadecimal (using between 1 and 8 ASCII characters from 0-9, A-F, and a-f; leading 0's are allowed).
+  For example, ADLER32=03da0195 and ADLER32=3DA0195 are both valid checksums for the 4-byte message "Wiki".
+  This algorithm is obsoleted and SHOULD NOT be used.
+* Status: obsoleted
```
ioggstream: Add `Reference` to this document.

paulmillar:
I think the pull request is OK as-is, but might be improved slightly.

In particularly, I'm wondering if SHOULD NOT is too strong for the server.  A server could support ADLER32 without that support impacting on cryptographic data integrity provided that server also supports a cryptographically secure hash; e.g., SHA-512.  Clients that wish to protect against malicious actor changes would use the SHA-512 hash and not request (or ignore) the ADLER32 value.  Therefore, I would suggest that, for the server, the document says a server MAY implement ADLER32.

I think it is perfectly reasonable for the document to say the client SHOULD NOT use ADLER32, for the reasons stated.

That said, I can't decide if this distinction is really helpful, so I defer to your judgement, here.

LPardue:
I'm not so sure we want to make a distinction between client and server as you write it. Both client and server can attach Digest header to request or response. It is the receiver of Digest that is exposed to the risk of using a weak digest.

The current language avoids mentioning either endpoint so we leave it to implementers. I'm not opposed to spelling things out more clearly but we'd really need to nail down what we mean by "using Digest  algorithms" because that relates to requesting them, generating them, regenerating them (proxy), validating them (endpoint or proxy) etc.

paulmillar:
That's a good point about client/server.  Sorry, I had that wrong.  As you say, it's the recipient of a Digest that must decide whether the algorithm is sufficient.  This recipient may be either the client or the server.

I think the document is right to point out the dangers of accepting a weak digest.

Perhaps there is a useful distinction between generating a digest and accepting a digest; e.g., an agent MAY support generating a weak digest but SHOULD NOT accept a weak digest.

LPardue:
> That's a good point about client/server. Sorry, I had that wrong.

Not wrong, just partially true. 

> Perhaps there is a useful distinction between generating a digest and accepting a digest; e.g., an agent MAY support generating a weak digest but SHOULD NOT accept a weak digest.

That could work for me. I'll leave it to @ioggstream to decide what he things makes sense to do.

ioggstream:
@paulmillar @LPardue as this issue is for ADLER, I moved the `weak digest` discussion  to #894

ioggstream:
Merging as for the positive feedback from @phluid61 and @LPardue comment https://github.com/httpwg/http-extensions/pull/877#issuecomment-520372044

Issue #786, opened 2019-04-04
Active vs. passive fingerprinting


yoavweiss:
At IETF 104's HTTPWG meeting, @ekr raised concerns about the categorization of Client Hints as an active fingerprinting vector, claiming that they are in-fact passive, because (IIUC) they can be persisted beyond the point where the information was requested, and persisted for future requests on current and future sessions to the same host.

My response to that was that this is something origins already do with cookies today, and that Client Hints access can be monitored by browsers in similar ways to other active fingerprinting vectors. On top of that, Client Hints opt-in persistence is purged when the user deletes their cookies, and support is disabled when JS is disabled.

Talking to @martinthomson, it seems like there's another concern around the fact that the hints change when the values does. That seems similar to me to various change events around the exposed values (which do exist for most if not all Client Hints features). Browsers can track CH value changes in similar ways to active fingerprinting vector change events, and treat registration as if the origin accessed the equivalent API and subscribed to receive future events. Similarly, browsers can also mitigate that change reporting in similar ways (e.g. omit the hints or lie about them in some cases).

@martinthomson & @ekr - Does the above properly captures the concerns you raised? If so, do the mitigations make sense and properly address them?

If they do, I'm happy to modify the security & privacy consideration sections to reflect the above

snyderp:
I believe there is a good deal of overlap between this issue and the concerns raised in https://github.com/httpwg/http-extensions/issues/786

yoavweiss:
> I believe there is a good deal of overlap between this issue and the concerns raised in #786

#786 is this issue. I'm guessing you meant something different :)

snyderp:
Derp you are exactly right, apologies! 

Here is the one I meant: https://github.com/httpwg/http-extensions/issues/767

yoavweiss:
I believe those issues are rather different. The threat model here is persistence of the hints in the browser, where for #767 is seems like the threat model is persistence on the server side. As they are rather different, mitigations for either can be fairly different as well.

snyderp:
I see, point taken. I meant mostly in terms of the discussion of whether or not CH is an increase in FP surface or not.  But I'm happy to deal with that discussion in both issues if thats preferred.

However, I'd like to reiterate my concern / position that its not sufficient to deal with concerns like these in non-normative security & privacy consideration sections; that to be an appropriate addition to the web, the normative portions of the document should be negative or zero regarding fingerprinting surface / privacy risk.

yoavweiss:
[...]

PR #879, opened 2019-07-26
Detail dimensions of information exposure

yoavweiss:
A first pass at resolving #786 

@igrigorik & @martinthomson - Can you take a look and see if this properly sums up our Monday conversations?

### DIFF
--- a/draft-ietf-httpbis-client-hints.md
+++ b/draft-ietf-httpbis-client-hints.md
@@ -189,7 +189,7 @@ Features relying on this document to define Client Hints headers SHOULD take int
   - Change over time
     - The feature SHOULD NOT expose user information that changes over time, unless the state change itself is also exposed (e.g. through JavaScript callbacks).
 
-Different features will be positioned in different spots on the spectrum between low-entropy, non-sensitive and static information (e.g. user agent information), and high-entropy, sensitive and dynamic information (e.g. geolocation). User agents SHOULD carefully consider the feature's position on that spectrum before exposing it as a Client Hint.
+Different features will be positioned in different spots on the spectrum between low-entropy, non-sensitive and static information (e.g. user agent information), and high-entropy, sensitive and dynamic information (e.g. geolocation). User agents SHOULD consider the value provided by a particular feature vs these considerations, and MAY have different policies regarding that tradeoff on a per-feature basis.
 
 Implementers ought to consider both user and server controlled mechanisms and policies to control which Client Hints header fields are advertised:
 
--- a/draft-ietf-httpbis-client-hints.md
+++ b/draft-ietf-httpbis-client-hints.md
@@ -203,7 +203,7 @@ Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persis
 ## Abuse Detection
 A user agent that tracks access to active fingerprinting information SHOULD consider emission of Client Hints headers similarly to the way it would consider access to the equivalent API.
 
-User agents which detect abuse through forensic means can observe HTTP responses to requests containing Client Hints, and their variance based on those hints. As Client Hints are provided in order to allow variation and optimization of the content to the provided Client-Hints, resource responses that do not vary on the Client Hints provided can indicate that such Hints are being used for nefarious purposes. Detection of variance can be done using the explicit Vary header, or by observing content changes as a result of different Client Hints sent.
+Research into abuse of Client Hints might look at how HTTP responses that contain Client Hints differ from those with different values, and from those without. This might be used to reveal which Client Hints are in use, allowing researchers to further analyze that use.
 
 
 # IANA Considerations
--- a/draft-ietf-httpbis-client-hints.md
+++ b/draft-ietf-httpbis-client-hints.md
@@ -179,7 +179,17 @@ Above example indicates that the cache key needs to include the Sec-CH-Example a
 
 The request header fields defined in this document, and those that extend it, expose information about the user's environment to enable proactive content negotiation. Such information may reveal new information about the user and implementers ought to consider the following considerations, recommendations, and best practices.
 
-Transmitted Client Hints header fields SHOULD NOT provide new information that is otherwise not available to the application via other means, such as using HTML, CSS, or JavaScript. Further, sending highly granular data, such as image and viewport width may help identify users across multiple requests. Reducing the set of field values that can be expressed, or restricting them to an enumerated range where the advertised value is close but is not an exact representation of the current value, can improve privacy and reduce risk of linkability by ensuring that the same value is sent by multiple users. However, such precautions can still be insufficient for some types of data, especially data that can change over time.
+Features relying on this document to define Client Hints headers SHOULD take into account the following aspects of the information exposed by the feature: 
+
+  - Entropy
+    - The feature SHOULD NOT provide new information that is otherwise not available to the application via other means, such as existing request headers, HTML, CSS, or JavaScript.
+    - Exposing highly granular data may help identify users across multiple requests. Reducing the set of field values that can be expressed, or restricting them to an enumerated range where the advertised value is close but is not an exact representation of the current value, can improve privacy and reduce risk of linkability by ensuring that the same value is sent by multiple users.
+  - Sensitivity
+    - The feature SHOULD NOT expose user sensitive information. To that end, information available to the application, but gated behind specific user actions (e.g. a permission prompt or user activation) SHOULD NOT be exposed as a Client Hint.
+  - Change over time
+    - The feature SHOULD NOT expose user information that changes over time, unless the state change itself is also exposed (e.g. through JavaScript callbacks)
+
+Different features will be positioned in different places on the spectrum between low-entropy, non-sensitive and static information (e.g. user agent information), and high-entropy, sensitive and dynamic information (e.g. geolocation). User agent SHOULD carefully consider the feature's position on that spectrum before exposing it as a Client Hint.
 
 Implementers ought to consider both user and server controlled mechanisms and policies to control which Client Hints header fields are advertised:
 
@@ -188,6 +198,12 @@ Implementers ought to consider both user and server controlled mechanisms and po
   - Implementations specific to certain use cases or threat models MAY avoid transmitting some or all of Client Hints header fields. For example, avoid transmission of header fields that can carry higher risks of linkability.
 
 Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persisted opt-in preferences when any one of site data, browsing history, browsing cache, or similar, are cleared.
+
+
+## Abuse Detection
+A user agent that tracks access to active fingerprinting information SHOULD consider emission of Client Hints headers similarly to the way it would consider access to the equivalent API.
+
+User agents which track abuse through forensic means can observe HTTP responses to requests containing Client Hints, and their variance based on those hints. Origins which opt-in but don't vary their responses are likely using that information for nefarious purposes.
 
 
 # IANA Considerations
--- a/draft-ietf-httpbis-client-hints.md
+++ b/draft-ietf-httpbis-client-hints.md
@@ -179,10 +179,13 @@ Above example indicates that the cache key needs to include the Sec-CH-Example a
 
 The request header fields defined in this document, and those that extend it, expose information about the user's environment to enable proactive content negotiation. Such information may reveal new information about the user and implementers ought to consider the following considerations, recommendations, and best practices.
 
-Features relying on this document to define Client Hints headers SHOULD take into account the following aspects of the information exposed: 
+The underlying assumption is that exposing information about the user as a request header is equivalent to the capability of that request's origin to access that information by other means and transmit it to itself.
+
+Therefore, features relying on this document to define Client Hint headers MUST NOT provide new information that is otherwise not available to the application via other means, such as existing request headers, HTML, CSS, or JavaScript.
+
+Such features SHOULD take into account the following aspects of the information exposed: 
 
   - Entropy
-    - The feature SHOULD NOT provide new information that is otherwise not available to the application via other means, such as existing request headers, HTML, CSS, or JavaScript.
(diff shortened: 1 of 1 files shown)


yoavweiss:
@sleevi - care to take a look as well?

draft-ietf-httpbis-client-hints.md
```diff
@@ -190,6 +200,12 @@ Implementers ought to consider both user and server controlled mechanisms and po
 Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persisted opt-in preferences when any one of site data, browsing history, browsing cache, or similar, are cleared.
 
 
+## Abuse Detection
+A user agent that tracks access to active fingerprinting information SHOULD consider emission of Client Hints headers similarly to the way it would consider access to the equivalent API.
+
+User agents which detect abuse through forensic means can observe HTTP responses to requests containing Client Hints, and their variance based on those hints. Origins which opt-in to Client Hints but don't vary any of their responses accordingly are likely using that information for nefarious purposes.
```
sleevi: This last sentence is a little tricky. I don't think we have any data to support the 'likely' claim - we're assuming that, but it may be that it turns out folks just have trouble with CH and implementing it (in)correctly.


> As Client Hints are provided in order to allow variation and optimization of the content to the provided Client-Hints, resource responses that do not vary on the Client Hints provided may indicate that such Hints are being used for nefarious purposes.

I mean, it makes it weaker in terms of *is* being used nefariously, so perhaps that's not worth it. I'm just a little uncomfortable stating such things without data, and without explanation about why the conclusion is there.

yoavweiss: yeah, that's fair

draft-ietf-httpbis-client-hints.md
```diff
@@ -190,6 +200,12 @@ Implementers ought to consider both user and server controlled mechanisms and po
 Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persisted opt-in preferences when any one of site data, browsing history, browsing cache, or similar, are cleared.
 
 
+## Abuse Detection
+A user agent that tracks access to active fingerprinting information SHOULD consider emission of Client Hints headers similarly to the way it would consider access to the equivalent API.
+
+User agents which detect abuse through forensic means can observe HTTP responses to requests containing Client Hints, and their variance based on those hints. As Client Hints are provided in order to allow variation and optimization of the content to the provided Client-Hints, resource responses that do not vary on the Client Hints provided can indicate that such Hints are being used for nefarious purposes.
```
igrigorik: > resource responses that do not vary on the Client Hints provided can indicate that such Hints are being used for nefarious purposes

Hmm.. I can see `no-cache` + `no-store` responses not emitting Vary by default today, and that doesn't directly imply "nefarious" use. If we're making that claim, perhaps we should also be recommending that servers always emit such Vary responses, regardless of cacheability of the response? Or, am I overthinking it?

yoavweiss: I'm not talking about using the `Vary` header, but about actually varying the resource served when different hints are included in the request. Maybe I should clarify that?

igrigorik: Ah, I see. Yeah, I was picturing an implementation looking for `Vary` in the response to asses use: seems like a straightforward + explicit signal? Probably worth being (slightly) more explicit.

yoavweiss: Added another sentence to clarify that.

draft-ietf-httpbis-client-hints.md
```diff
@@ -190,6 +200,12 @@ Implementers ought to consider both user and server controlled mechanisms and po
 Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persisted opt-in preferences when any one of site data, browsing history, browsing cache, or similar, are cleared.
 
 
+## Abuse Detection
+A user agent that tracks access to active fingerprinting information SHOULD consider emission of Client Hints headers similarly to the way it would consider access to the equivalent API.
+
+User agents which detect abuse through forensic means can observe HTTP responses to requests containing Client Hints, and their variance based on those hints. As Client Hints are provided in order to allow variation and optimization of the content to the provided Client-Hints, resource responses that do not vary on the Client Hints provided can indicate that such Hints are being used for nefarious purposes. Detection of variance can be done using the explicit Vary header, or by observing content changes as a result of different Client Hints sent.
```
martinthomson: "their" variance?

The problem here is that it is difficult to identify variance from a single instance.  You need to establish patterns of behaviour, which means comparing the response to a baseline.  Maybe "Research into abuse of Client Hints might look at how HTTP responses that contain Client Hints differ from those without.  This might be used to reveal which Client Hints are in use, allowing researchers to further analyze that use."

I don't think that you need the mention of Vary here.  Bad actors won't use that, or might use `Cache-Control: no-store` and friends.

draft-ietf-httpbis-client-hints.md
```diff
@@ -179,7 +179,17 @@ Above example indicates that the cache key needs to include the Sec-CH-Example a
...
+  - Sensitivity
+    - The feature SHOULD NOT expose user sensitive information. To that end, information available to the application, but gated behind specific user actions (e.g. a permission prompt or user activation) SHOULD NOT be exposed as a Client Hint.
+  - Change over time
+    - The feature SHOULD NOT expose user information that changes over time, unless the state change itself is also exposed (e.g. through JavaScript callbacks).
+
+Different features will be positioned in different spots on the spectrum between low-entropy, non-sensitive and static information (e.g. user agent information), and high-entropy, sensitive and dynamic information (e.g. geolocation). User agents SHOULD consider the value provided by a particular feature vs these considerations, and MAY have different policies regarding that tradeoff on a per-feature basis.
```
martinthomson: "spectrum" implies linear.  Consider rewording.

draft-ietf-httpbis-client-hints.md
yoavweiss: Cool, incorporated your language (while mentioning different values, on top of a with/without comparison)

draft-ietf-httpbis-client-hints.md
yoavweiss: Would "range" be better?

igrigorik: Hmm, I don't think "spectrum" connotes linear; spectrum can be linear and nonlinear.

image

That said, if a different term or turn of phrase is a better fit.. 

Riffing.. 

> Each feature will have different characteristics with respect to key considerations such as: static vs. dynamic information, low vs. high entropy, and user sensitive vs non-sensitive information. For example, user agent version number that is shared by many users is static, low entropy, and does not reveal user sensitive information, whereas geolocation is dynamic, high entropy, and user sensitive. User agents SHOULD consider the value provided by each particular feature on a per-feature basis and MAY have different policies on whether and how the feature is made available.

Warmer?

yoavweiss: @martinthomson friendly ping! :)

yoavweiss:
@ekr - could you please take a look? :)

draft-ietf-httpbis-client-hints.md
```diff
@@ -179,7 +179,17 @@ Above example indicates that the cache key needs to include the Sec-CH-Example a
...
+Features relying on this document to define Client Hints headers SHOULD take into account the following aspects of the information exposed: 
+
+  - Entropy
+    - The feature SHOULD NOT provide new information that is otherwise not available to the application via other means, such as existing request headers, HTML, CSS, or JavaScript.
+    - Exposing highly granular data may help identify users across multiple requests to different origins. Reducing the set of field values that can be expressed, or restricting them to an enumerated range where the advertised value is close but is not an exact representation of the current value, can improve privacy and reduce risk of linkability by ensuring that the same value is sent by multiple users.
+  - Sensitivity
+    - The feature SHOULD NOT expose user sensitive information. To that end, information available to the application, but gated behind specific user actions (e.g. a permission prompt or user activation) SHOULD NOT be exposed as a Client Hint.
+  - Change over time
+    - The feature SHOULD NOT expose user information that changes over time, unless the state change itself is also exposed (e.g. through JavaScript callbacks).
+
```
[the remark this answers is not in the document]
yoavweiss: This does not mandate exposing every JS state change in Client Hints as well, but the inverse. If something is not exposed in JS, it SHOULD NOT be exposed through CH.

Your example of geo "moved" callback would be covered by the bullet above this one: "The feature SHOULD NOT expose user sensitive information. To that end, information available to the application, but gated behind specific user actions (e.g. a permission prompt or user activation) SHOULD NOT be exposed as a Client Hint"

I don't think that we can be overly descriptive here (e.g. call out geo or other features specifically). At the same time, any feature that will use this framework will go through a separate privacy review process, so we'd be able to catch more specific issues at that point as well.

Does that make sense? Or do you think further clarifications are required?

draft-ietf-httpbis-client-hints.md
yoavweiss: SGTM

draft-ietf-httpbis-client-hints.md
```diff
@@ -190,6 +200,12 @@ Implementers ought to consider both user and server controlled mechanisms and po
 Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persisted opt-in preferences when any one of site data, browsing history, browsing cache, or similar, are cleared.
 
 
+## Abuse Detection
+A user agent that tracks access to active fingerprinting information SHOULD consider emission of Client Hints headers similarly to the way it would consider access to the equivalent API.
+
```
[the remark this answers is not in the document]
yoavweiss: I was imagining both the "privacy budget" work as well as Mozilla's forensic work we discussed at IETF105 as potential ways to "track access to active fingerprinting information". I'm happy to drop the normative SHOULD here, if you think that'd be better. Alternatively, I can add examples pointing to those two methods of fingerprinting tracking.

yoavweiss:
@ekr - friendly ping :) Would be great to drive this to conclusion

/cc @mnot as we discussed this last week at TPAC

draft-ietf-httpbis-client-hints.md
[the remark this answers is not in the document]
ekr: I think my point is broader than this: While the current client hints may be innocuous, this document does not provide any rubric for rationally evaluating abuse or for assessing new client hints, and this guidance doesn't help. So this just creates an open door for new dangerous features.

As I said in Montreal: I'm generally OK with taking features that are currently sent in headers and client-hintizing them, because that seems better. I'm not fine with a document that suggests that it is a good thing to take things that currently are acquired by active JS APIs and turn them into client hints.

draft-ietf-httpbis-client-hints.md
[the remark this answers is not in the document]
ekr: I don't think further clarifications are required. I think it should remove everything after the comma. I.e., "The feature SHOULD NOT expose user information that changes over time".

ekr:
See above. I think this text is still far too positive about this feature.

draft-ietf-httpbis-client-hints.md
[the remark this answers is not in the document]
martinthomson: After Yoav walked me through this, I think that this specific problem isn't what we want to fixate on.

Basically, though we might agree about the relative utility of the different methods for accounting for information use, I can understand how you might conclude that there is an equivalence here.  This all rides on having strong coupling between information access controls and the network query.  That is, it is assumed that sites that can access information are also able to send themselves that information and that having those two events conjoined in the browser is equivalent to that in every way.  As long as that is true - it's an assumption here that probably needs to be more clearly articulated - then this is OK.

The problem here is that this list isn't all that needs to be considered when deciding whether it is appropriate to expose information as a client hint.  I opened #962 to track the most obvious unaddressed aspect of the decision; I can't think of any others right now.

draft-ietf-httpbis-client-hints.md
```diff
@@ -179,7 +179,17 @@ Above example indicates that the cache key needs to include the Sec-CH-Example a
...
 
-Transmitted Client Hints header fields SHOULD NOT provide new information that is otherwise not available to the application via other means, such as using HTML, CSS, or JavaScript. Further, sending highly granular data, such as image and viewport width may help identify users across multiple requests. Reducing the set of field values that can be expressed, or restricting them to an enumerated range where the advertised value is close but is not an exact representation of the current value, can improve privacy and reduce risk of linkability by ensuring that the same value is sent by multiple users. However, such precautions can still be insufficient for some types of data, especially data that can change over time.
+Features relying on this document to define Client Hints headers SHOULD take into account the following aspects of the information exposed: 
+
+  - Entropy
+    - The feature SHOULD NOT provide new information that is otherwise not available to the application via other means, such as existing request headers, HTML, CSS, or JavaScript.
```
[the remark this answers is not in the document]
yoavweiss: Makes sense!

draft-ietf-httpbis-client-hints.md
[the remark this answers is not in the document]
yoavweiss: @martinthomson so if I understand correctly, I need to better articulate the fact the underlying assumption( that sites that can access the information can also send it to themselves) and resolve #962. I'll add the former here, and the latter as a followup, if that works for you.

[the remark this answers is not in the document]
martinthomson: Yep.  A follow-up is fine.  I'm happy to defer to you on how you want to proceed.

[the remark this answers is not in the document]
yoavweiss: If this PR looks good to you, I'd love to land it :)
Added the MUST and clarified the underlying assumptions. Let me know if there are other things to address here.

yoavweiss:
@martinthomson - friendly ping! :) I believe we agreed at the HTTPWG meeting at IETF106 that this is good to go

martinthomson:
Yep.  I just assumed you were going ahead.

yoavweiss:
@martinthomson In that case, can you land it? (I don't have the permissions)

mnot:
Well that's an oversight; you should have an invite in your inbox.

yoavweiss:
> Well that's an oversight; you should have an invite in your inbox.

Thanks! :)

Issue #880, opened 2019-07-30
Examples for id-sha-256 and id-sha-512


manger:
draft-ietf-httpbis-digest-headers includes examples of sha-256 in the Digest header value, but no examples of the two new algorithms it introduces: id-sha-256 and id-sha-512.

Examples would be extremely useful to illustrate the difference between sha-256 and id-sha-256.

LPardue:
That sounds reasonable to me.

ioggstream:
@manger WDYT? #883

PR #883, opened 2019-07-31
Fix: #880. Add id-sha-* algorithm examples.

ioggstream:
## This PR

Adds id-sha-* algorithm examples throughout the document.

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -492,6 +492,13 @@ A sender MAY send a representation-data-digest using a digest-algorithm without
 knowing whether the recipient supports the digest-algorithm, or even
 knowing that the recipient will ignore it.
 
+Examples:
+
+~~~
+  Digest: sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
+  Digest: id-sha-512=WZDPaVn/7XgHaAy8pmojAkGWoRx2UFChF41A2svX+TaPm+AbwAgBWnrIiYllu7BNNyealdVLvRwE\nmTHWXvJwew==
+  Digest: sha-256=4REjxQ4yrqUVicfSKYNO/cF9zNj5ANbzgDZt3/h3Qxo=, id-sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
+~~~
 ...
 
 # Deprecate Negotiation of Content-MD5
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -531,6 +538,9 @@ Response:
 
 ###  Representation data is not contained in the payload
 
+As there is no content coding applied, the `id-sha` and the `id-sha-256`
```
LPardue:
```suggestion
As there is no content coding applied, the `sha-256` and the `id-sha-256`
```

ioggstream: 1- thanks for the fix :) I mis-typed the algorithm name

2- Here and below some `Content-Encoding: identity` + `sha-256`:

- https://github.com/httpwg/http-extensions/blame/76f77c9ce2e9ffa51eb1fe01c8656b9b41b52584/draft-ietf-httpbis-digest-headers.md#L571-L573
- https://github.com/httpwg/http-extensions/blame/76f77c9ce2e9ffa51eb1fe01c8656b9b41b52584/draft-ietf-httpbis-digest-headers.md#L533-L534

Let me know if that works for you!

Thanks++,
R:

LPardue: Ah yeah I missed those sorry. Changes look good to me, lets see if the issue creator has any further suggestions

draft-ietf-httpbis-digest-headers.md
```diff
@@ -492,6 +492,13 @@ A sender MAY send a representation-data-digest using a digest-algorithm without
...
 knowing that the recipient will ignore it.
 
+Examples:
+
+~~~
+  Digest: sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
+  Digest: id-sha-512=WZDPaVn/7XgHaAy8pmojAkGWoRx2UFChF41A2svX+TaPm+AbwAgBWnrIiYllu7BNNyealdVLvRwE\nmTHWXvJwew==
+  Digest: sha-256=4REjxQ4yrqUVicfSKYNO/cF9zNj5ANbzgDZt3/h3Qxo=, id-sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
+~~~
 ...
```
manger: I'd drop the first line (sha-256=X48...) and keep the other two.
It is probably legal to have multiple Digest headers so this looks like it might be 1 HTTP message (lines from 3 separate messages). But having sha-256 twice with different values in 1 message wouldn't make sense.

ioggstream: I can clarify those are three different HTTP messages.
Ok for leaving just two.

PR #893, opened 2019-08-13
Clamp floats

### LINKED ISSUE #848: Serializing sh-float is hard
The current definition of sh-float is surprisingly hard to serialize correctly, the best C-code I could come up with is:

        int
        serialize_sh_float(char buf[18], double d)
        {
                int p = 0, l, dig;

                buf[0] = '*';
                buf[1] = '\0';
                if (isnan(d))
                        return (-1);
                if (fabs(d) >= 99999999999999.95)
                        return (-1);

                if (d < 0) {
                        buf[p++] = '-';
                        d = -d;
                        l = 18;
                } else {
                        l = 17;
                }
                dig = trunc(log10(d));
                p += snprintf(buf + p, l - p, "%.*f", 15-dig, d);
                while (buf[p-2] == '0') {
                        buf[p-1] = '\0';
                        p--;
                }
                return (p);
        }

And that is horrible in severale levels.

If we insist instead that sh-float can have maximum six digits after the decimal point, people can rely on the default "6 decimals" behaviour of "%f" formatters, and use the simpler:

        int
        serialize_sh_float(char buf[18], double d)
        {
                int p = 0, l, dig;

                if (isnan(d))
                        return (-1);
                if (fabs(d) >= 99999999999999.95)
                        return (-1);

                if (d < 0) {
                        buf[p++] = '-';
                        d = -d;
                }
                p += snprintf(buf + p, 18 - p, "%-16f", d);
                while (buf[p-1] == ' ')
                        p -= 1;
                buf[p] = '\0';
                return (p);
        }

mnot:
Fixes #848

### DIFF
--- a/draft-ietf-httpbis-header-structure.md
+++ b/draft-ietf-httpbis-header-structure.md
@@ -37,6 +37,18 @@ author:
 normative:
 
 informative:
+  IEEE754:
+     target: http://ieeexplore.ieee.org/document/4610935/
+     title: IEEE Standard for Floating-Point Arithmetic
+     author:
+     -
+       organization: IEEE
+     date: 2008-08
+     seriesinfo:
+       IEEE: 754-2008
+       DOI:  10.1109/IEEESTD.2008.4610935
+       ISBN: 978-0-7381-5752-8
+     annotation: See also <http://grouper.ieee.org/groups/754/>.
 
 
 --- abstract
--- a/draft-ietf-httpbis-header-structure.md
+++ b/draft-ietf-httpbis-header-structure.md
@@ -217,15 +217,13 @@ Header specifications can constrain the types of individual list values (includi
 
 Dictionaries are ordered maps of name-value pairs, where the names are short, textual strings and the values are items ({{item}}) or arrays of items. There can be zero or more members, and their names are required to be unique within the scope of the dictionary they occur within.
 
-Each member of the dictionary can also have associated parameters -- an ordered map of key-value pairs where the keys are short, textual strings and the values are items ({{item}}). There can be zero or more parameters on a member, and their keys are required to be unique within that scope.
-
 Implementations MUST provide access to dictionaries both by index and by name. Specifications MAY use either means of accessing the members.
 
 The ABNF for dictionaries in textual HTTP headers is:
 
 ~~~ abnf
 sh-dictionary  = dict-member *( OWS "," OWS dict-member )
-dict-member    = member-name "=" member-value *parameter
+dict-member    = member-name "=" member-value
 member-name    = key
 member-value   = sh-item / inner-list
 ~~~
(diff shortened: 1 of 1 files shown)


phluid61:
Is this PR meant to include #892?

mnot:
Yeah, oops. Need more coffee.

mnot:
Try now.

draft-ietf-httpbis-header-structure.md
```diff
@@ -856,6 +851,8 @@ _RFC Editor: Please remove this section before publication._
 ## Since draft-ietf-httpbis-header-structure-11
 
 * Allow \* in key (#844).
+* Allow dictionary members to have parameters (#842).
```
phluid61: Close enough

phluid61:
Otherwise, looks alright to me.

draft-ietf-httpbis-header-structure.md
```diff
@@ -498,13 +491,14 @@ Given an integer as input_integer:
...
 
-0. If input_float is not a IEEE 754 double precision number, fail serialisation.
-1. Let output be an empty string.
-2. If input_float is less than (but not equal to) 0, append "-" to output.
-3. Append input_float's integer component represented in base 10 using only decimal digits to output; if it is zero, append "0".
-4. Append "." to output.
-5. Append input_float's decimal component represented in base 10 using only decimal digits to output; if it is zero, append "0".
-6. Return output.
+0. If input_float's fractional part has more than six digits of precision, fail serialisation.
+1. If the number of digits of precision in input_float's fractional part plus those in its integer part add to more than fifteen digits, fail serialisation.
```
bsdphk: Not sure I agree with this one, what if the value is 1/3 ?

mnot: Perhaps something like:

> Append up to six digits of input_float's decimal component...

phluid61: 1/3 doesn't match "integers with a fractional part, which can be at most six digits of precision." This isn't the algorithm for serialising an IEEE-754 float, nor an arbitrary rational number.

Although maybe guidance for converting those to an SH-float data object could be useful.

draft-ietf-httpbis-header-structure.md
```diff
@@ -278,26 +278,19 @@ Note that commas in integers are used in this section's prose only for readabili
 
 ## Floats {#float}
 
-Floats are integers with a fractional part, that can be stored as IEEE 754 double precision numbers (binary64) ({{IEEE754}}).
+Floats are integers with a fractional part, which can be at most six digits of precision. Additionally, like integers, it can have no more than fifteen digits in total, which in some cases further constrains its precision.
```
phluid61: Might be an idea to say "decimal digits" here.

ioggstream: > Floats are integers with a fractional part

I would prefere "floats are numbers with a fractional part"

phluid61: Or "numbers with integer and fractional parts"

I still think saying "decimal digits" would help, too.

draft-ietf-httpbis-header-structure.md
```diff
@@ -278,7 +278,7 @@ Note that commas in integers are used in this section's prose only for readabili
 
 ## Floats {#float}
 
-Floats are integers with a fractional part, which can be at most six digits of precision. Additionally, like integers, it can have no more than fifteen digits in total, which in some cases further constrains its precision.
+Floats are decimal numbers with an integer and a fractional part. The fractional part has at most six digits of precision. Additionally, like integers, it can have no more than fifteen digits in total, which in some cases further constrains its precision.
```
ioggstream: LGTM

draft-ietf-httpbis-header-structure.md
```diff
@@ -520,13 +513,14 @@ Given an integer as input_integer:
...
 Given a float as input_float:
 
-0. If input_float is not a IEEE 754 double precision number, fail serialisation.
-1. Let output be an empty string.
-2. If input_float is less than (but not equal to) 0, append "-" to output.
-3. Append input_float's integer component represented in base 10 using only decimal digits to output; if it is zero, append "0".
-4. Append "." to output.
-5. Append input_float's decimal component represented in base 10 using only decimal digits to output; if it is zero, append "0".
-6. Return output.
+0. If input_float's fractional component has more than six digits of precision, fail serialisation.
```
ioggstream: Q: here we use fractional component, above we use fractional part. Is that ok?

mnot: Good catch.

PR #910, opened 2019-08-30
Use payload body.

ioggstream:
## This PR

Uses payload body as iiuc:

- we want to be transparent to transfer-coding;
- rfc 7540 doesn't seem to use `message body`

## Note

If you agree I can replace elsewhere too.


reschke (approved):

ioggstream:
It's stale now ;)

PR #920, opened 2019-09-02
[SH] say 1=true, 0=false

### LINKED ISSUE #916: SH: boolean definition doesn't say what is what
Boolean values can be conveyed in Structured Headers.

    The ABNF for a Boolean in textual HTTP headers is:

    sh-boolean = "?" boolean
    boolean    = "0" / "1"

Either say "0" is false and "1" is true, or maybe rephrase the ABNF to:

    sh-boolean = sh-false / sh-true
    sh-false = "?0"
    sh-true = "?1"

phluid61:
fixes #916


mnot:
thx!

Issue #918, opened 2019-08-30
SH: broken link in IEEE reference


reschke:
<http://grouper.ieee.org/groups/754/> 404s.

PR #921, opened 2019-09-03
[SH] reference IEEE 754-2019

phluid61:
Incidentally fixes in #918 by removing the annotation.

According to IEEE's site and references the target should be <https://ieeexplore.ieee.org/servlet/opac?punumber=8766227> but I've kept the equivalent of the current address.


mnot:
ta!

PR #923, opened 2019-09-06
Fix #853. Use when acting on resources

### LINKED ISSUE #853: Investigate the use of Digest with PATCH
## About PATCH

1- `PATCH` is not in http 723x
2- `PATCH` media type should be something like https://tools.ietf.org/html/rfc7396 because 
servers SHOULD NOT assume PATCH semantics for generic
   media types that don't define them, such as "application/xml" or
   "application/json".  Doing so will cause interoperability issues,
   because the semantics of PATCH become specific to that resource,
   rather than general.
3- it's difficult to use patch + checksums, like @LPardue said

## As of now we

1- warn people about PATCH+Digest
2- don't forbid PATCH+Digest

## Notes

-  it could be useful to validate the PATCH request body before applying it to the resource
- guidance of the usage of Digest+PATCH is welcome to confirm  if the current expression of Digest support that use.

ioggstream:
## This PR

Clarifies the relations b/w Digest,  PATCH and POST:

- PATCH and POST does not convey partial representations;
- they instead specify actions on resources

In requests:
- Digest should refer to the representation-data of the action

In responses:
- if the enclosed representation refers to the status of the action, then it's complete and Digest is computed on it;
- if the enclosed representation referers to another resource (eg. via content-location), Digest is computed on the selected representation of that resource;
- the above is true even in case of an empty payload body.

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -527,57 +527,34 @@ requires complete representation metadata (see {{representation-digest}}).
 In responses,
 
 - if the representation describes the status of the request,
-  `Digest` MUST be computed on the enclosed representation;
+  `Digest` MUST be computed on the enclosed representation
+   (see {{post-referencing-action}} );
 
-- if the referenced resource target is different from the effective request URI
-  or the response does not reference the target resource
-  `Digest` MUST be computed on the referenced resource.
+- if there is a referenced resource
+  `Digest` MUST be computed on the selected representation of the referenced resource
+   even if that is different from the target resource.
+   That may or may not result in computing `Digest` on the enclosed representation.
 
 The latter case might be done accordingly to the HTTP semantics of the given method,
 for example using the `Content-Location` header field.
 
+Differently from `Content-Location`, which is representation metadata,
+the `Location` header field does not affect `Digest`.
+
 ## Digest and PATCH
 
-In a PATCH request, the representation metadata refers
+In PATCH requests
+the representation digest MUST be computed on the patch document.
+
+This is because the representation metadata refers
 to the patch document and not to the target resource (see Section 2 of {{?RFC5789}}).
 
-This makes `Digest` usage with PATCH very similar to the POST one,
-where the resource's own semantic is partly implied by the method and by the patch document.
+In PATCH responses
+the representation digest MUST be computed on
+the selected representation of the patched resource.
 
-The following example shows a PATCH request:
-
-- using an appropriate content-type defined in {{?RFC7396}};
-- with the representation digest of the enclosed payload.
-
-The response contains the representation digest
-of the patched resource together with the complete
-resource representation.
-
-Request:
-
-~~~
-PATCH /books/123 HTTP/1.1
-Content-Type: application/merge-patch+json
-Accept: application/json
-Accept-Encoding: identity
-Digest: sha-256=bWopGGNiZtbVgHsG+I4knzfEJpmmmQHf7RHDXA3o1hQ=
-
-{"title": "New Title"}
-~~~
-
-Response:
-
-~~~
-HTTP/1.1 200 OK
-Content-Type: application/json
-Digest: id-sha-256=BZlF2v0IzjuxN01RQ97EUXriaNNLhtI8Chx8Eq+XYSc=
-
-{"id": "123", "title": "New Title"}
-~~~
-
-Note that a `202 No Content` response without a payload body but with the same `Digest` header
-field value would have been legitimate too.
-
+`Digest` usage with PATCH is thus very similar to the POST one,
+but with the resource's own semantic partly implied by the method and by the patch document.
 
 # Deprecate Negotiation of Content-MD5
 
@@ -814,17 +791,15 @@ Digest: sha-256=4REjxQ4yrqUVicfSKYNO/cF9zNj5ANbzgDZt3/h3Qxo=, id-sha-256=X48E9qO
 iwiAeyJoZWxsbyI6ICJ3b3JsZCJ9Aw==
 ~~~
 
-## Digest with POST Referencing a Resource Different from the Request URI
+## POST Response does not Reference the Request URI {#post-not-request-uri}
 
-As described in {{acting-on-resources}}, the request representation digest is computed
-on the enclosed representation.
+Request `Digest` value is computed on the enclosed representation (see {{acting-on-resources}}). 
 
-The response does not reference the target resource
-so the `Digest` is computed on the referenced resource.
+The representation enclosed in the response
+refers to the resource identified by `Content-Location`
+(see [RFC7231] Section 3.1.4.2 and Section 3.1.4.1 point 4).
 
-In the example, the payload is a representation of the resource identified by `Content-Location`
-(see [RFC7231] Section 3.1.4.2 and Section 3.1.4.1 point 4),
-but other cases might apply.
+`Digest` is thus computed on the enclosed representation.
 
 Request:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -501,6 +500,47 @@ to collision attacks [CMU-836068].
 The SHA algorithm is NOT RECOMMENDED as it's now vulnerable
 to collision attacks [IACR-2019-459].
 
+# Use of Digest when acting on resources
+
+As the representation-data-digest is calculated on the selected representation,
+you need all the representation-metadata to compute it.
```
LPardue: Alternative suggestion:

A representation-data-digest is calculated on the selected representation, which requires complete representation metadata.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -501,6 +500,47 @@ to collision attacks [CMU-836068].
...
+PATCH representation metadata apply to the patch document
+and not to the target resource, as stated in Section 2 of {{?RFC5789}},
+and should instead use a proper content-type, eg {{?RFC7396}}.
+
+In these cases, the representation-data-digest
+MUST be computed on the action of the payload.
+
+The response instead can convey the representation-data-digest
+of the representation as long as it contains all the required
+representation metadata.
```
LPardue: Alternative suggestion for above 4 pargraphs:

Methods such as PATCH or POST may appear to convey partial representations but are semantically acting on resources. In these cases the representation-data-digest MUST be computed on the action of the payload. The response to such requests can convey the representation-data-digest.

In the case of PATCH a request conveys representation metadata that applies to the patch document and not the target resource (see ,Section 2 of {{?RFC5789}}). 

The following example shows a PATCH request (using a valid content-type defined in {{?RFC7396}} and a response that contains the representation-data-digest of the patched representation.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -874,10 +914,6 @@ RFC EDITOR PLEASE DELETE THIS SECTION.
 
    The PATCH verb brings some complexities (eg. about representation metadata header fields, patch document format, ...),
```
LPardue: I think we can remove all of Q3 because it has been covered in the new section.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -732,6 +774,14 @@ integrity at transport layer that protects HTTP header fields.
 
 A `Digest` header field using NOT RECOMMENDED digest-algorithms SHOULD NOT be used in signatures.
 
+## Usage with PATCH
+
+A `200 OK` response to a PATCH request would contain
+the `Digest` of the patched item, and the etag of the new object.
+
+This behavior is tighly coupled to the application logic
```
LPardue: I think this sentence comes from me on some earlier thread and I'm not really sure what you're aiming to achieve by including it in this context.

Lets figure out the core intent and write a clear message (I'll help), or just remove it because the new section "Use of Digest when acting on resources" does the job now.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -732,6 +774,14 @@ integrity at transport layer that protects HTTP header fields.
 
 A `Digest` header field using NOT RECOMMENDED digest-algorithms SHOULD NOT be used in signatures.
 
+## Usage with PATCH
+
+A `200 OK` response to a PATCH request would contain
+the `Digest` of the patched item, and the etag of the new object.
```
LPardue: What about other status codes? Perhaps we can give an example that failed and that returns the Digest and etag of the unpatch item (if that is what happens).

ioggstream:
@LPardue main refactoring with examples. Still needs a lot of work but seems more understandable.

Added python snippets for creating examples in FAQ.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -511,6 +510,109 @@ to collision attacks [CMU-836068].
 The SHA algorithm is NOT RECOMMENDED as it's now vulnerable
 to collision attacks [IACR-2019-459].
 
+# Use of Digest when acting on resources {#acting-on-resources}
+
+POST and PATCH requests may appear to convey partial representations
```
ioggstream: This statement seems to me the best way to introduce the vision.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -511,6 +510,109 @@ to collision attacks [CMU-836068].
...
+but are semantically acting on resources.
+The enclosed representation, including its metadata refers to that action.
+
+In these cases the representation digest MUST be computed on
+the representation-data of that action.
+
+This is the only possible choice because representation digest
```
ioggstream: We may remove this line if it's redundant.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -962,6 +1055,29 @@ Specification document(s):  {{digest-header}} of this document
    and adds two more algorithms: ID-SHA-256 and ID-SHA-512 which allows to
    send a checksum of a resource representation with no content codings applied.
 
+8. How can I test digest output? Use the following python3 function.
```
LPardue: Editorial suggestion:

It would be nice to split the line after the question, and provide a little bit more decription in the. For example

```
8. How can I generate and validate the Digest values shown in the examples throughout this document?

The following python3 script can be used to generate digests using SHA algorithms for a range of encodings. Note that these are formatted as base64. This function could be adapted to other algorithms and should take into account their specific formatting rules.

File: draft-ietf-httpbis-digest-headers.md [hunk 9]
```diff
@@ -962,6 +1055,29 @@ Specification document(s):  {{digest-header}} of this document
...
+item = {"hello": "world"}
+print("Identity encoding, sha256", digest(item))
+
+# Out: Identity encoding, sha256 4REjxQ4yrqUVicfSKYNO/cF9zNj5ANbzgDZt3/h3Qxo=
+
+print("Brotli encoding, sha256", digest(item, encoding=brotli.compress))
+
+# Out: Brotli encoding, sha256 4REjxQ4yrqUVicfSKYNO/cF9zNj5ANbzgDZt3/h3Qxo=
+
+print("Identity encoding, sha512", digest(dict(hello='world'), algorithm=hashlib.sha512)
```
LPardue: why does this example use a temporary dict rather than `item`?

draft-ietf-httpbis-digest-headers.md
ioggstream: moved to PR

draft-ietf-httpbis-digest-headers.md
ioggstream: addressing in #938

draft-ietf-httpbis-digest-headers.md
```diff
@@ -507,7 +506,51 @@ Two examples of its use are
...
+
+This is because the representation metadata refers
+to the patch document and not to the target resource (see Section 2 of {{?RFC5789}}).
+
+In PATCH responses
+the representation digest MUST be computed on
+the selected representation of the patched resource.
+
+`Digest` usage with PATCH is thus very similar to the POST one,
+but with the resource's own semantic partly implied by the method and by the patch document.
```
```suggestion
but with the resource's own semantic is partly implied by the method and by the patch document.
```

ioggstream:
@reschke could you PTAL :)

ioggstream:
>I haven't looked at the details; in general, this looks right.

That's generally great 

> My main concern is that this is written as if POST and PATCH are special. They are not.
> Shouldn't the rules be the same for any HTTP request that has a request payload?

As I just investigated on PATCH and POST didn't want to extend that too far.
For example, `an hypothetical method where the request payload body contains a
partial representation` would behave differently.

draft-ietf-httpbis-digest-headers.md
ioggstream: Further reasoning moved to a new issue #958

reschke:
That's ok.

I just don't think that it would be a good idea to vary the semantics based on the request method. But go ahead with the changes you have...

Issue #936, opened 2019-09-24
Explain reference to `cache-validators`


ioggstream:
## I expect

to clarify the following passage

> The resource is specified by the effective
> Request-URI, and any cache-validator contained in the message.

for defining the behavior of digest with caching (eg. when Digest is returned together with
Last-Modified and ETag).

and to better extend the definition of Digest to POST and PATCH requests.
This is because [POST is cacheable for rfc 7231 §4.2.3](https://tools.ietf.org/html/rfc7231#section-4.2.3) though only when they include explicit freshness information.

If we go on the new http-core definition (which is cleaner) we have that

> Responses to POST requests are only cacheable when they include
>   explicit freshness information (see Section 4.2.1 of [Caching]) 
>   and a
>   Content-Location header field that has the same value as the POST's
>   effective request URI (Section 6.2.5).
 
eg.
```
>POST /books

< Content-Location: /books
< Cache-Control: max-age=3600

```

## note
According to https://tools.ietf.org/html/draft-ietf-httpbis-semantics-04#section-7.3.3

POST responses identify returned resources in different ways according to the following:

### referenced resource target != effective request uri

* the primary resource is identified by Content-Location
* the enclosed representation matches the primary resource
eg. 1

 ```
POST /books

{"title": "foo"}
---
HTTP/1.1 201
Content-Location /books/123

{"title": "foo", "id": 123}
```

eg. 2

 ```
POST /books

{"title": "foo"}
---
HTTP/1.1 204 No Content
Content-Location /books/123

```
### representation describe the status of the request

* primary resource identified by `Location`
* enclosed representation matches the status of the request
  according to the resource's own semantic

eg.

 ```
POST /books

{"title": "foo"}
---
HTTP/1.1 200
Location /books/123

{"status": "updated", "id": 123}
```

ioggstream:
re-reading the old RFC my understanding is that validator fields should just be aligned with Digest (that is: digest-value should be aligned with the values of last-modified and etag). While it is trivial, the original editor could have considered that a potential issue.

Issue #850, opened 2019-07-09
Clarify digest-algorithm syntax with parameters (or drop parameters).


ioggstream:
## I expect

This spec either:

* to provide an RFC3230 compatible example of digest-algorithm with parameters;
* explicitly drop parameters because they are not widely used and augment the attack surface;

RFC3230 stated that:

```
   For some algorithms, one or more parameters may be
   supplied.

      digest-algorithm = token

   The BNF for "parameter" is as is used in RFC 2616 [4].  All digest-
   algorithm values are case-insensitive.
```

but:

- provided no example of parameterized algorithms

In this draft we *preserved that sentence* but:

- still did not provide an example of parameterized algorithm;
- still did not update the reference to BNF to the latest http spec.

In #1213 there's a possible fix to support parameters, like the following

```
Digest: myhash=dcRDgR2GM35DluAV13PzgnG6+pvQwPywfFvAu1UeFrs=; param1=40

```

## Notes

Use cases are parametrizable digest-algorithms like mi-sha256.

Another solution could be to *deprecate* field parameters in digest: parameters could still be adopted serializing them in the digest-value, and leaving the specification/serialization to the definition of the digest-algorithm. eg:
```
Digest: myhash=40/dcRDgR2GM35DluAV13PzgnG6+pvQwPywfFvAu1UeFrs
```

ioggstream:
@LPardue @reschke I don't know of any parameter in the wild, but removing that will deviate from RFC 3230.

If we want to remove parameters, we should explicit:

- that we will deprecate that feature (eg. because parameters can make things less secure);
- that parameters are forbidden;
- that if you want to use a parameterized algorithm you should encode their values in the digest-value and take care of the security considerations inside the definition of your digest-algorithm;

`bikeshed=bs=8|offset=2|8E9q0okq`

reschke:
FWIW, deviating would fine in this case. One point in revising specs is to remove unused stuff.

ioggstream:
@reschke I agree. My question is whether it's better to:

- leave digest-algorithm implementors to define a way to pass parameters via the serialization
- standardize the way parameters are passed.

Guidance on that is welcome.

LPardue:
@mnot removal of Digest parameters would be a case where our update work deviates a bit further from what we might have initially had in mind. As Julian points out its fine to do so, however I want to make this more visibile before we take a decision.

ioggstream:
> I want to make this more visibile before we take a decision

[...]

PR #937, opened 2019-09-24
Fix: #936. Cache semantics impacts on Digest and on the representation.

ioggstream:
## This PR

Clarifies the following points:

- resource identification is inherited from SEMANTICS
- validator fields (ETags & Last Modified) which apply to the identified resource, apply to Digest too.

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -237,7 +237,7 @@ transformations (eg. transfer codings for HTTP/1.1 see 6.1 of
 
 A representation digest consists of
 the value of a checksum computed on the entire selected `representation data`
-(see Section 7 of {{SEMANTICS}}) of a resource identified as per (see Section 7.3.2 of {{SEMANTICS}})
+(see Section 7 of {{SEMANTICS}}) of a resource identified according to Section 7.3.2 of {{SEMANTICS}}
 together with an indication of the algorithm used (and any parameters)
 
 ~~~ abnf
@@ -281,9 +281,6 @@ validation instead of verifying every received representation-data-digest.
 A sender MAY send a representation-data-digest using a digest-algorithm without
 knowing whether the recipient supports the digest-algorithm, or even knowing
 that the recipient will ignore it.
-
-Digest MAY be cached and the freshness information of the resource, eventually conveyed by validator
- header fields or other cache headers, apply to its value too.
 
 Digest can be sent in a trailer section. When using incremental digest-algorithms
 this allows the sender and the receiver to dynamically compute the digest value
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -282,8 +282,8 @@ A sender MAY send a representation-data-digest using a digest-algorithm without
 knowing whether the recipient supports the digest-algorithm, or even knowing
 that the recipient will ignore it.
 
-Digest MAY be cached and the freshness information of the resource eventually conveyed by validator
- header fields or other cache headers apply to its value too.
+Digest MAY be cached and the freshness information of the resource, eventually conveyed by validator
+ header fields or other cache headers, apply to its value too.
 
 Digest can be sent in a trailer section. When using incremental digest-algorithms
 this allows the sender and the receiver to dynamically compute the digest value
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -236,7 +236,8 @@ transformations (eg. transfer codings for HTTP/1.1 see 6.1 of
 {{resource-representation}} contains several examples to help illustrate those effects.
 
 A representation digest consists of
-the value of a checksum computed on the entire selected `representation data` of a resource
+the value of a checksum computed on the entire selected `representation data`
+(see Section 7 of {{SEMANTICS}}) of a resource identified as per (see Section 7.3.2 of {{SEMANTICS}})
 together with an indication of the algorithm used (and any parameters)
 
 ~~~ abnf
@@ -263,9 +264,6 @@ response.
    Digest = "Digest" ":" OWS 1#representation-data-digest
 ~~~
 
-The resource is specified by the effective request URI and any `validator field`
-contained in the message.
-
 The relationship between Content-Location (see Section 7.2.5 of
 {{SEMANTICS}}) and Digest is demonstrated in
 {{post-not-request-uri}}. A comprehensive set of examples showing the impacts of
@@ -283,6 +281,9 @@ validation instead of verifying every received representation-data-digest.
 A sender MAY send a representation-data-digest using a digest-algorithm without
 knowing whether the recipient supports the digest-algorithm, or even knowing
 that the recipient will ignore it.
+
+Digest MAY be cached and the freshness information of the resource eventually conveyed by validator
+ header fields or other cache headers apply to its value too.
 
 Digest can be sent in a trailer section. When using incremental digest-algorithms
 this allows the sender and the receiver to dynamically compute the digest value


draft-ietf-httpbis-digest-headers.md
```diff
@@ -284,6 +284,29 @@ Content-Encoding: gzip
...
+Accept: application/json
+Content-Type: application/json
+
+{"author": "Camilleri"}
+~~~
+
+Response:
+
+~~~
+HTTP/1.1 201 OK
```
LPardue: `201 OK` ??

draft-ietf-httpbis-digest-headers.md
```diff
@@ -469,7 +492,9 @@ The Digest header field provides a digest of the representation data.
 - or not at all contained in the message body.
 
 The resource is specified by the effective
-Request-URI and any cache-validator contained in the message.
+Request-URI,
+the HTTP semantics of the message
+and any cache-validator contained in the message.
```
ioggstream: @LPardue I need some guidance here:

- iiuc cache validators are ETags & Last Modified [RFC2616 s. 13.3.3 ](https://tools.ietf.org/html/rfc2616#section-13.3.3)

it's not clear to me:

* how a cache validator specify a resource;
* how can we translate the sentence according to RFC723x

LPardue: sorry, not a caching expert so my opinion might not help much here.

ioggstream: As @reschke had similar considerations in https://github.com/httpwg/http-extensions/pull/940#pullrequestreview-299297863

I asked him to help.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -479,7 +506,8 @@ The Digest header field provides a digest of the representation data.
 - or not at all contained in the message body.
 
 The resource is specified by
-the effective request URI
+the effective request URI,
```
ioggstream: Can that work?
```
selected representation-data is identified accordingly to Section 3.1.4.1 of [RFC7231].
https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#identifying.payload
```

ioggstream:
I'll merge the content-location stuff and leave the resource identification to this issue (which is the hard part). This will push forward the ongoing work on POST/PATCH.

draft-ietf-httpbis-digest-headers.md
ioggstream: I thought twice. Probably rfc 3230 refers to receiving a cached `Digest`.

`validators` do not identify the resource on the origin server 
 but it's cached representation. Let me know, R.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -812,6 +812,13 @@ integrity at transport layer that protects HTTP header fields.
 
 A `Digest` header field using NOT RECOMMENDED digest-algorithms SHOULD NOT be used in signatures.
 
+## Relation with validators
+
+`Digest` field-value may be cached by intermediaries.
```
```suggestion
`Digest` field-value can be cached by intermediaries.
```

ioggstream:
@reschke @LPardue PTAL :)

reschke:
Consider me confused.

Currently, the draft says:

"The resource is specified by the effective request URI and any validator field contained in the message."

...and then, in that section, doesn't mention "resource" anymore.

What was the intent of that statement? Can't it be just removed?

ioggstream:
> Currently, the draft says:
> 
> "The resource is specified by the effective request URI and any validator field contained in the message."
> 
> ...and then, in that section, doesn't mention "resource" anymore.
> 
> What was the intent of that statement?

The statement originates from https://tools.ietf.org/html/rfc3230#section-4.3.2 . 
My understanding is that the original author stated explicitly that Digest must be aligned with  Last-Modified & co.
This could be trivial though...

> Can't it be just removed?

It is reasonable to me to remove the whole statement 

```
￼ The resource is specified by the effective request URI and any validator field contained in the message
```

and delegate the resource specification to https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#target.resource

@LPardue  @mnot WDYT?

ioggstream:
My bad, Mark replied https://github.com/httpwg/http-extensions/issues/949#issuecomment-673931768 with the above proposal. Let's do it then!

draft-ietf-httpbis-digest-headers.md
```diff
@@ -284,6 +282,9 @@ A sender MAY send a representation-data-digest using a digest-algorithm without
 knowing whether the recipient supports the digest-algorithm, or even knowing
 that the recipient will ignore it.
 
+Digest MAY be cached and the freshness information of the resource eventually conveyed by validator
+ header fields or other cache headers apply to its value too.
```
ioggstream, line 286:
```suggestion
Digest MAY be cached and the freshness information of the resource, eventually conveyed by validator
 header fields or other cache headers, apply to its value too.
```

ioggstream:
Current wording integrates @reschke and @mnot suggestion (iiuc).

draft-ietf-httpbis-digest-headers.md
```diff
@@ -284,6 +282,9 @@ A sender MAY send a representation-data-digest using a digest-algorithm without
 knowing whether the recipient supports the digest-algorithm, or even knowing
 that the recipient will ignore it.
 
+Digest MAY be cached and the freshness information of the resource, eventually conveyed by validator
```
mnot, line 285: Generally, we try to avoid re-stating requirements in other specs, or constraining how they're interpreted, because that makes the composition of the specs more brittle -- there's more chance of misinterpretation or unintentional misalignment (either as written or after some future change in the dependency).

So if you say anything at all, it shouldn't be a RFC2119 requirement; just some prose about the relationship for reader convenience.

ioggstream: Thanks @mnot. Let's drop those lines then!

draft-ietf-httpbis-digest-headers.md
```diff
@@ -284,6 +282,9 @@ A sender MAY send a representation-data-digest using a digest-algorithm without
...
+ header fields or other cache headers, apply to its value too.
+
```
ioggstream, line 287:
```suggestion
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -236,7 +236,8 @@ transformations (eg. transfer codings for HTTP/1.1 see 6.1 of
 {{resource-representation}} contains several examples to help illustrate those effects.
 
 A representation digest consists of
-the value of a checksum computed on the entire selected `representation data` of a resource
+the value of a checksum computed on the entire selected `representation data`
+(see Section 7 of {{SEMANTICS}}) of a resource identified according to Section 7.3.2 of {{SEMANTICS}}
 together with an indication of the algorithm used (and any parameters)
```
LPardue, line 241: This new text look technically fine but the sentence is a bit long winded now, especially with the
parameter comment in brackets. How about some editorializing like this, which intentially avoids saying anything about parameters.

```
A representation digest consists of a `digest-algorithm` followed by an `encoded digest output`.

~~~ abnf
   representation-data-digest = digest-algorithm "="
                                <encoded digest output>
~~~

The `digest-algorithm` identifies both an algorithm and output encoding format which are applied
to a representation. A checksum value is first computed on the entire selected `representation data`
(see Section 7 of {{SEMANTICS}}) of a resource identified according to Section 7.3.2 of {{SEMANTICS}}.
This is then encoded using the required format.

The list of defined `digest-algorithms` is provided in {{algorithms}}.

The example below shows the `sha-256` digest-algorithm which uses base64 encoding.

~~~ example
   sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
~~~
```

ioggstream: If this is technically ok, I suggest to:

- merge and close the issue;
- resolve #850 (parameters);
- rewording in another PR.

In the rewording, we could even avoid mentioning the "encoding" part as I think it should only be mentioned in the `digest-algorithm` section.

WDYT?

LPardue:
Yeah I forgot about the clash with 850. Let's land this and I see you've made a signpost back to my editorial suggestion so we don't lose it.

ioggstream:
Ok, to me this is complete. I'll leave to @LPardue  the decision to merge or to wait for further feedback.

draft-ietf-httpbis-digest-headers.md
LPardue: sgtm

LPardue (approved):

PR #940, opened 2019-09-25
Replace cache-validator with validator (rfc 7231 section 7.2)

ioggstream:
## I expect

to use the new "validator" name. See: 

- https://tools.ietf.org/html/rfc7231#section-7.2
- https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#response.validator


ioggstream:
@reschke I'm quite weak on that side: is this PR correct? If not feel free to fix at will :)

ioggstream:
@reschke I agree... this comes from the original RFC 3230. I'm trying to address that in #937 but your guidance will be great. See #936 too.

Issue #945, opened 2019-10-05
EU Cookie Law: Mechanisms enabling user agents to aid in compliance absent


ttacker:
A European Union (EU) [directive](https://en.wikipedia.org/wiki/Directive_(European_Union)) requires member states to achieve a result, without dictating specific means. In contrast, EU [regulations](https://en.wikipedia.org/wiki/EPrivacy_Regulation_(European_Union)) are binding law in all member states. Recital 25 of the Preamble and Article 5(3) of the [ePrivacy Directive](https://en.wikipedia.org/wiki/Privacy_and_Electronic_Communications_Directive_2002) (ePD)--often referred to as the EU Cookie Law--regulate the use of cookies. The [ePrivacy Regulation](https://en.wikipedia.org/wiki/EPrivacy_Regulation_(European_Union)) (ePR) is expected to repeal and replace the ePD in 2019, but it's expected the ePR will regulate the use of cookies in substantially the same manner as the ePD.

To paraphrase [one summary](https://www.iubenda.com/en/help/5525-cookies-gdpr-requirements), the EU Cookie Law requires the following:

- Unless an exempt type of cookie, informed consent, based on explicit affirmative action, before storing a cookie on the user's device.
- Conspicuously providing the option for obtaining informed consent.
- A means for withdrawal of consent.
- Ability to prove consent occurred, even if subsequently withdrawn.
- A guarantee, via prior blocking, no tracking performed before consent.
- Informing users of the use of third-party cookies.
- Stating the category and purpose of third-party cookies, even if not listed individually.
- Linking to relevant third-party privacy/cookie policies.

[RFC 6265](https://www.rfc-editor.org/info/rfc6265) doesn't include mechanisms enabling user agents to aid in compliance with the above requirements, and this has resulted in presentation of inconsistently placed, and inconsistently formatted, cookie banners worldwide. Such inconsistency degrades user experience, works against conspicuousness, and is annoying to the point that some have attempted to create [workarounds](https://cookiesok.com/) to address the lack of desired user agent functionality. Internet Draft (I-D) [draft-ietf-httpbis-rfc6265bis](https://datatracker.ietf.org/doc/draft-ietf-httpbis-rfc6265bis/) proposes to obsolete RFC 6265; however, it also fails to address the issues raised above.

Attributes, such as the following, could be added to the Set-Cookie Header:

- Category (Allows grouping for purpose of category consent.)
- Purpose (Provides information to ensure consent is informed.)
[...]

PR #946, opened 2019-10-06
[WIP] EU Cookie Law Mechanisms

ttacker:
* Add attributes to Set-Cookie syntax
* Update Set-Cookie semantics
* Add subcomponent algorithms and todo throughout

I had started work on this PR to address issue #945 before it was closed. I thought other people might like to see my initial thoughts, so I created this PR as a WIP. I'm not requesting it be merged, and it can be closed; but if anyone wants to provide feedback, that would be welcome. Thanks!

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -207,6 +207,15 @@ code and issues list for this draft can be found at <https://github.com/httpwg/h
 
 # Introduction
 
+TODO: Update to include explanation of global benefit of adding mechanisms
+      to enable user agents to aid in compliance with EU Cookie Law, such
+      as enabling standard placement and formatting of cookie banners per
+      user agent and ability to consent in advance, and by category and
+      domain, via settings or other means.
+TODO: Add references to ePrivacy Directive, GDPR, and secondary sources.
+      
+TODO: Update to explain new consent requirement prior to storage by user agent.
+
 This document defines the HTTP Cookie and Set-Cookie header fields. Using
 the Set-Cookie header field, an HTTP server can pass name/value pairs and
 associated metadata (called cookies) to a user agent. When the user agent makes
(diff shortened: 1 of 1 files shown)


petkivim:
It might me a good idea to include the action in the Consent attribute's value or alternatively, define separate attributes for the action and time, e.g.:
```
consent-av        = "Consent=" consent-action; consent-date
consent-action    = "Unconfirmed" / "Granted" / "Revoked"
consent-date      = <IMF-fixdate>
                      ; IMF-fixdate defined in [RFC7231], Section 7.1.1.1
```

OR

```
consent-action-av  = "ConsentAction=" consent-action
consent-action     = "Unconfirmed" / "Granted" / "Revoked"
consent-date-av    = "ConsentDate=" consent-date
consent-date       = <IMF-fixdate>
                      ; IMF-fixdate defined in [RFC7231], Section 7.1.1.1
```

In this way it is possible to prove that consent occurred, even if subsequently withdrawn. In case the Consent attribute is just unset when consent is withdrawn, there's no trace that consent occurred earlier.

ioggstream:
As suggested by @mnot I propose to create a new draft for addressing that, which could be easily integrated or extend current one.

I'm volunteering as co-editor and I can setup the repo with an automatic draft build in minutes. @ttacker

awwright:
Doesn't consent occur when the user-agent saves the cookie to storage? Saving cookies is something the user-agent must opt into on behalf of the user; not by the server.

The one thing that seems to be missing is a way to link to the cookie policy; but didn't P3P already standardize a way to do this?

pbowyer:
> Doesn't consent occur when the user-agent saves the cookie to storage? Saving cookies is something the user-agent must opt into on behalf of the user; not by the server.

Yes that is correct.

ioggstream:
> didn't P3P already standardize a way to do this?

afaik P3P is somewhat deprecated, see https://bugzilla.mozilla.org/show_bug.cgi?id=225287#c12

chlily1:
> I'm not requesting it be merged, and it can be closed

Thanks, I will go ahead and close this.

Issue #907, opened 2019-08-28
[SH] Allow parameterized inner-list values.


mikewest:
As discussed around https://github.com/httpwg/http-extensions/issues/839#issuecomment-508367966, it would be lovely if we could address the use case raised in https://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html#structured-headers-dont-fit-my-data by allowing list items themselves to contain parameters. This would allow the example in the specification:

```
Example-Thing: name="Widget", cost=89.2, descriptions=(foo bar)
Example-Description: foo; url="https://example.net"; context=123,
                     bar; url="https://example.org"; context=456
```

to be written as:

```
Example-Thing:
    name="Widget",
    cost=89.2,
    descriptions=(foo;url="https://example.net";context=123
                  bar;url="https://example.org";context=456)
```

mnot:
Hm. We put parameters on (item or inner-list), where maybe we should have just put them on item; I suspect your use case is more common than the one that's currently enabled.

Would you be against making that change (i.e., moving parameters "inside" the list, as opposed to enabling them both on the list as a whole and individual items)?

mikewest:
> Would you be against making that change (i.e., moving parameters "inside" the list, as opposed to enabling them both on the list as a whole and individual items)?

That sounds fine to me. I, personally, don't yet have a use for parameterizing the `inner-list` itself, but I would quite appreciate being able to add parameters to the items an `inner-list` contains.

mnot:
@bsdphk @phluid61 any concern about trying the above change?

bsdphk:
If we keep this up, I can dust of my original proposal which had recursion :-)

As long as we can pick it apart consistently, I'm fine.

mikewest:
> If we keep this up, I can dust of my original proposal which had recursion :-)

I'm fine with recursion, FWIW. More complexity === better. ;)

mnot:
So to summarise, I think the proposal is:

* Allow parameters on Item
* Allow parameters on inner-list

This removes the need to make parameters specific to Dictionaries and lists, and effectively allows parameters "everywhere". This is probably a very good story for extensibility, if we document it right.

The only downside that I see is that implementations will no longer map "simple" types (e.g., integers, strings) to their analogies in-language; it'll all need to be objects, to accommodate parameters.

I think that's acceptable -- especially given the token/string schism -- but wanted to make sure everyone understood this. 

For the test suite, it means that the JSON representation of an item will probably be something like:

[...]

PR #954, opened 2019-10-09
Allow parameters on Items and Inner Lists

mnot:
For #907

### DIFF
--- a/draft-ietf-httpbis-header-structure.md
+++ b/draft-ietf-httpbis-header-structure.md
@@ -305,7 +305,7 @@ bare-item = sh-integer / sh-float / sh-string / sh-token / sh-binary
 ~~~
 
 
-## Integers {#integer}
+### Integers {#integer}
 
 Integers have a range of −999,999,999,999,999 to 999,999,999,999,999 inclusive (i.e., up to fifteen digits, signed), for IEEE 754 compatibility ({{IEEE754}}).
 
@@ -324,7 +324,7 @@ Example-IntegerHeader: 42
 Note that commas in integers are used in this section's prose only for readability; they are not valid in the wire format.
 
 
-## Floats {#float}
+### Floats {#float}
 
 Floats are decimal numbers with an integer and a fractional component. The fractional component has at most six digits of precision. Additionally, like integers, it can have no more than fifteen digits in total, which in some cases further constrains its precision.
 
@@ -348,7 +348,7 @@ Example-FloatHeader: 4.5
 ~~~
 
 
-## Strings {#string}
+### Strings {#string}
 
 Strings are zero or more printable ASCII {{!RFC0020}} characters (i.e., the range %x20 to %x7E). Note that this excludes tabs, newlines, carriage returns, etc.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-header-structure.md
```diff
@@ -211,32 +203,64 @@ In HTTP headers, each member is separated by a comma and optional whitespace. Fo
...
+
+
+### Inner Lists {#inner-list}
+
+An inner list is an array of zero or more items ({{item}}). Both the individual items and the inner-list itself can be parameterised ({{param}}).
+
+The ABNF for inner-lists in HTTP headers is:
+
+~~~ abnf
+inner-list    = "(" OWS [ sh-item *( SP sh-item ) OWS ] ")" *parameter
```
phluid61: Not related to this PR per se, but I just realised the ABNF for inner list says exactly `SP` between items, but the parsing algorithm accepts `SP` followed by any other `OWS`

mnot: Hm. IIRC my thinking when I wrote this was that the ABNF was illustrating the "correct" on-wire form, rather than being a way to write a parser. Still, we could import `RWS` and use that...

phluid61: I've absolutely no doubt someone will assume the ABNF and the algorithms match, and write a regexp parser based on the former.

(_sotto voce_) `RWS` still doesn't work, because it must start with `SP`

mnot: So maybe we should make the parsing algorithm more strict, either `SP` or `1*SP`

phluid61: Just `SP` works for me. That way it couldn't mask something like a null value that gets written as a zero-length string, for example.

phluid61:
I'm having some other issues with my roughed-in implementation, but I think the algorithm to parse parameters aggressively consumes leading OWS, so when you have an inner-list it consumes the spaces between items.

mnot:
Ah, that makes sense, will have a look.

phluid61:
Maybe removing the 'SP or )' check from the bottom of the inner-list parsing algorithm also lets you change the ABNF to `RWS`.

phluid61:
I can't think of a way around the it without adding a delimiter (comma?) in inner-lists. Or being super strict about whitespace and cutting down on OWS.

phluid61:
I was playing around with a regular expression parser to check the ABNF, and noticed this:

```
parameter     = OWS ";" OWS param-name [ "=" param-value ]
param-value   = sh-item
sh-item   = bare-item *parameter
```

`param-value` should be a `bare-item`

mnot:
Weird; I'd seen that and thought I'd fixed it; huh.

Fixed; thanks.

mnot:
It's the `OWS` before `;`, isn't it?

I'm kinda fine ditching that; WDYT?

phluid61:
Oh yeah, that's right. I'm happy removing it.

It invalidates the ABNF slightly, but I think that's easy enough to resolve.

mnot:
OK, that's in. I'm going to take one more look over this, merge it, and clean up a few other things, then get another draft out.

Issue #851, opened 2019-07-09
Using Digest in signatures


ioggstream:
## I expect

To better detail the usage of `Digest` in signatures, including:

- [x] you need a transport-layer integrity mechanism for the whole message;
- [x] what should be included in signature for integrity (representation metadata affecting digest-algorithms)
- [ ] issues with signing partial representations (eg. Range-Request, PATCH, ..)
- [x] using strong crypto algorithms

This should be done:

- [x] respecting the dependency model is that a signature doesn't need to include a digest but it can;
- [x] don't explain too much of signature risks inverting the dependency;
- [x] add useful guidance that is relevant to Digest and expected usages but without also defining the expected usages completely.

## Questions

- consider signing `Content-Length` too ([it's a representation-metadata in http-core](https://tools.ietf.org/html/draft-ietf-httpbis-semantics-04#section-6.2)), though it would eliminate the ability to do things like chunked transfer-encoding, which is quite common. @dlongley

## Previous discussions

What we have today is probably sufficient but we're open to compelling reasons on what would be really helpful to define in this document.

The followings may be out-of-scope:

  - [length-extension attack on the signature](https://en.wikipedia.org/wiki/Length_extension_attack): Digest should not be subject to length extension attacks because **it doesn't provide a proof of origin**;
  - which other signature data unrelated to `Digest` (eg. timestamp, method & co)

LPardue:
What does it take to close this issue?

Since we opened it, the HTTP WG has adopted a signautes draft. Does that move the needle on any questions we have?  Can we spin out anything or makr them as won't do?

ioggstream:
"Usage in signatures" covers the goals enough.

https://github.com/httpwg/http-extensions/blame/master/draft-ietf-httpbis-digest-headers.md#L1012

- we provide considerations but won't specify what other headers should be signed (eg. content-length);
- as https://github.com/httpwg/http-extensions/blob/master/draft-ietf-httpbis-message-signatures.md should be independent from Digest, the best we can do is to write a BCP on Digest and Signatures.

This is the repository placeholder https://github.com/ioggstream/draft-polli-digest-signatures (still not edited).

@richanna @jyasskin @jricher  WDYT? 
@LPardue feel free to close if you agree.

ioggstream:
Last considerations integrated via #1334 

The point on partial representations is related on HTTP semantics and cannot be addressed by this spec.

ioggstream:
Last considerations integrated via #1334 

[...]

PR #971, opened 2019-10-31
digest: editorial sweep and address use cases and SRI

### LINKED ISSUE #868: Relationship to SRI
It would be awesome if this could relate directly to [SRI](https://w3c.github.io/webappsec-subresource-integrity/).  Ideally, if SRI uses the same hash function, the value in the `Digest` header field is the same.

LPardue:
My intention of this PR was to fix #868 (relationship to SRI) and #872 (use case).

Having reviewed the I-D, IMO it already explained a few general use cases, so I did some word smithing to make them jump out some more. We can expand on these further if needs be.

I wanted to break out the SRI stuff, but ran out of time today to fully complete that so the PR is WIP.

Finally, the usage in signatures section seemed to benefit from some tidy up too. I don't think we've fully completed the ask on #851 and I'm inclined to leave it that way for the time being.

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -111,18 +111,22 @@ The source code and issues list for this draft can be found at
 
 # Introduction
 
-Integrity protection for HTTP content is multi layered and
-is usually achieved across the protocol stack:
-TCP checksums and TLS [RFC2818] record to name but some.
+The core specification of HTTP does not define a means to protect the integrity
+of resources. When HTTP messages are transferred between endpoints, the protocol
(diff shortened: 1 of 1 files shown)


ioggstream:
Thanks, I think the wording is significantly improved! I would split the editorial changes from the rest to be able to merge it soon.

LPardue:
Thanks. To be clear I plan to close this out tomorrow, subject to review.

LPardue:
This is now ready to be reviewed. The changeset relies on changes throughout the document but IMO these are quality of life improvements to the document and I'd be a lot happier with them in.

The most controversial changes are some slight shuffling of the examples, i I really have to I can break them out but I'd prefer not to.

PR #979, opened 2019-11-04
Adding whitespace to correct formatting

thpts:
This will now correctly render the unordered lists.
Before:
image

After:
image


mnot:
Thanks!

PR #988, opened 2019-11-25
Variants dictionary keys must be lowercase

essen:
The `Variants` header is a dictionary and in the current version of
structured headers member names are `key` which only accepts
lowercase letters:

* https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-14#section-3.2
* https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-14#section-3.1.2


mnot:
Variants is no longer active; closing.

PR #989, opened 2019-11-25
[SH] Remove repetition

essen:
Not an English native so might not be the best wording but it sounds better than repeating `within` twice.


phluid61:
Might be better to swap them the other way. (in the scope...occur within)

mnot:
Fixed in source, thanks for the report.

PR #996, opened 2019-12-04
Change byte sequence delimiters from "*" to ":"

### LINKED ISSUE #991: The best use for `*`
I'm running my SH implementation against the HTTP Archive dumps (400,000,000+ HTTP requests) to see how the [Directly Represented Fields](https://mnot.github.io/I-D/binary-structured-headers/#direct) from BSH parse in something like the real world.

It looks like some headers are failing because `*` is commonly used in things that are parsed as tokens; e.g., the `Accept*` headers, and `Access-Control-Allow-Origin`. 

If we were to use a different character to delimit Byte Sequences, we could include `*` in Token (necessarily as a first character, as well as in the rest, I assume).

Worthwhile? If so, what should we use for Byte Sequences?

mnot:
Fixes #991.


ioggstream:
fyi @jyasskin impacts on https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#section-3.1

PR #1001, opened 2019-12-04
Clear CH persistence when session cookies are cleared

### LINKED ISSUE #957: Not maximum lifetime for server opt-in preferences specified
The client hint spec does not seem to specify if there's a maximum lifetime for the origin opt-in preferences. There seem to be spec that the data should be cleared when site data, cookies, cache, etc. are cleared, but barring that should there be some sort of reasonable limit?

yoavweiss:
Closes #957 

@igrigorik - can you review?

### DIFF
--- a/draft-ietf-httpbis-client-hints.md
+++ b/draft-ietf-httpbis-client-hints.md
@@ -174,7 +174,7 @@ Implementers ought to consider both user and server controlled mechanisms and po
   - Implementers MAY provide user choice mechanisms so that users may balance privacy concerns with bandwidth limitations. However, implementers should also be aware that explaining the privacy implications of passive fingerprinting to users may be challenging.
   - Implementations specific to certain use cases or threat models MAY avoid transmitting some or all of Client Hints header fields. For example, avoid transmission of header fields that can carry higher risks of linkability.
 
-Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persisted opt-in preferences when any one of site data, browsing history, browsing cache, or similar, are cleared.
+Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persisted opt-in preferences when any one of site data, browsing history, browsing cache, session cookies, or similar, are cleared.
 
 # Cost of Sending Hints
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-client-hints.md
```diff
@@ -174,7 +174,7 @@ Implementers ought to consider both user and server controlled mechanisms and po
   - Implementers MAY provide user choice mechanisms so that users may balance privacy concerns with bandwidth limitations. However, implementers should also be aware that explaining the privacy implications of passive fingerprinting to users may be challenging.
   - Implementations specific to certain use cases or threat models MAY avoid transmitting some or all of Client Hints header fields. For example, avoid transmission of header fields that can carry higher risks of linkability.
 
-Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persisted opt-in preferences when any one of site data, browsing history, browsing cache, or similar, are cleared.
+Implementers SHOULD support Client Hints opt-in mechanisms and MUST clear persisted opt-in preferences when any one of site data, browsing history, browsing cache, session cookies, or similar, are cleared.
```
igrigorik: Sanity check: session cookie, or cookie? Why session cookie?

yoavweiss: There are two distinct cookie related scenarios in which I think we want to clear CH persistence:
1) User clears their cookies by an explicit action. I think this is already covered by "site data".
2) Session cookies are cleared because the user restarted their browser. IIRC, our reasoning to remove ACHL was that we could tie the hints to a session cookie lifetime, and this is where I was going with this change. 

Does that clarify things? Or would it be better to spell that out further in the text?

igrigorik: (1) makes sense. What I'm after is to clarify (2): do we have a clear definition of what a session cookie is? E.g. [Wikipedia](https://en.wikipedia.org/wiki/HTTP_cookie#Session_cookie)..

> A session cookie, also known as an in-memory cookie, transient cookie or non-persistent cookie, exists only in temporary memory while the user navigates the website.[18] Web browsers normally delete session cookies when the user closes the browser.[19] Unlike other cookies, session cookies do not have an expiration date assigned to them, which is how the browser knows to treat them as session cookies.

One way to read the above is that opt-in should be cleared anytime the user navigates away from the site... Is that what we intend to imply here?

yoavweiss: I was thinking more of "session cookies" as  the term is used by (but not strictly defined in) [RFC 6265](https://tools.ietf.org/html/rfc6265#section-4.1.2.2). Cookies that have user agent defined lifetime, and in practical term only get cleared on browser restarts (when the session is not restored).

igrigorik: Would it be clearer if we omit "session" and just talk about cookies, since that's what RFC is referring to? That would address concerns from my end.

yoavweiss: Done

Issue #973, opened 2019-11-02
Add example of unpatched response with its digest.


ioggstream:
## I expect

an example with a  failed PATCH 

## Note

Depending on the PATCH semantic it may return:
- the Digest and etag of the unpatch item (if that is what happens);
- the Digest of the error.

PR #1004, opened 2019-12-06
Fix: #973. Error response is unrelated to the effective request URI

ioggstream:
## This PR

Adds an example of error response. Digest is computed
on the enclosed representation of the error.

## Note

imho this example is inferred from the current normative part, clarifies the expected behavior and does not require a change in the normative part.

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -927,8 +927,12 @@ In error responses, the representation-data does not necessarily refer to
 the target resource.
 Instead it refers to the representation of the error.
 
-In the following example, the `Digest` of the response
-is computed on the enclosed representation.
+In the following example a client attempts to patch
+the resource located at /books/123.
+However, the resource does not exist and the server generates a 404 response
+with a body that describes the error in accordance with {{?RFC7807}}.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -921,6 +921,37 @@ Digest: id-sha-256=BZlF2v0IzjuxN01RQ97EUXriaNNLhtI8Chx8Eq+XYSc=
...
 but with the same `Digest` field-value would have been legitimate too.
 
+## Error responses
+
+In error responses, the representation-data does not necessarily refer to
+the target resource.
+Instead it refers to the representation of the error.
+
+In the following example, the `Digest` of the response
+is computed on the enclosed representation.
```
LPardue: Lets describe the example a little more, something like the following would help me understand better e.g.

In the following example a client attempts to patch the resource located at /books/123. However, the resource does not exist and the server generates a 404 response with a body that describes the error in accordance with {{?RFC7807}}. The digest of the response is computed on this enclosed representation.

ioggstream: thx++. fixed

ioggstream:
PTAL @LPardue

PR #1008, opened 2019-12-11
Float -> Decimal

### LINKED ISSUE #982: Round-tripping float is still hard
(Related to now-closed #896)

Should implementations be able to store parsed floats in IEEE754 binary64? And if so, should we expect that they round trip between parsing and serialization? It seems like the current limits were designed to support that notion, but there are cases where they do not.

(Note, this isn't about being able to serialize arbitrary C++ doubles; the spec sensibly only supports a tiny subset of all possible 64-bit floats. This is about whether `x == Serialize(Parse(x))` should always be true)

IEEE 754 guarantees that if a string with no more than 15 digits is converted to a double and then back to a string, it should result in the original string, but that appears to require rounding in some cases; truncation, which the serialization algorithm calls for, can change the string representation.

Parsing the string "12345678901234.1" as a float, for instance, results in a binary64 representation with value 0x42a674e79c5fe433, whose exact decimal expansion is 12345678901234.099609375. Following the text of the serialization algorithm, `fractional_digits_avail` will be 1, and so the number will be serialized as "12345678901234.0"

It appears, as an implementer, that the only way to ensure that a float parsed from a valid string can be serialized back to the original string is to store it as a fixed point decimal in two parts, or else use a float with >64 bits of precision for storage.

mnot:
Fixes #982

### DIFF
--- a/draft-ietf-httpbis-header-structure.md
+++ b/draft-ietf-httpbis-header-structure.md
@@ -145,7 +145,7 @@ However, both Items and Inner Lists allow parameters as an extensibility mechani
 
 Conversely, inner lists are only valid when a header definition explicitly allows them.
 
-Note that a header field definition cannot relax the requirements of this specification because doing so would preclude handling by generic software; they can only add additional constraints (for example, on the numeric range of integers and floats, the format of strings and tokens, the types allowed in a dictionary's values, or the number of items in a list). Likewise, header field definitions can only use Structured Headers for the entire header field value, not a portion thereof.
+Note that a header field definition cannot relax the requirements of this specification because doing so would preclude handling by generic software; they can only add additional constraints (for example, on the numeric range of integers and decimals, the format of strings and tokens, the types allowed in a dictionary's values, or the number of items in a list). Likewise, header field definitions can only use Structured Headers for the entire header field value, not a portion thereof.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-header-structure.md
```diff
@@ -853,17 +845,17 @@ NOTE: This algorithm parses both Integers ({{integer}}) and Floats ({{float}}),
...
-   6. If type is "float" and input_number contains more than 16 characters, fail parsing.
+   6. If type is "decimal" and input_number contains more than 16 characters, fail parsing.
 8. If type is "integer":
    1. Parse input_number as an integer and let output_number be the product of the result and sign.
    2. If output_number is outside the range −999,999,999,999,999 to 999,999,999,999,999 inclusive, fail parsing.
 9. Otherwise:
    1. If the final character of input_number is ".", fail parsing.
-   2. If the number of characters after "." in input_number is greater than six, fail parsing.
-   2. Parse input_number as a float and let output_number be the product of the result and sign.
+   2. If the number of characters after "." in input_number is greater than three, fail parsing.
```
martinthomson: Your algorithm doesn't explicitly mark the placement of the "." or count characters.  I think that's OK, it's just imprecise.

phluid61: It's not clear whether `1234567890123.0` or `0034567890123.0` should fail.

mnot: Right now they don't. Separate issue?

phluid61: Yeah, they both fail now with 087d6b1, and the second one is a bit silly anyway.

PR #1011, opened 2019-12-16
make #parse-number algorithm fail on bad decimals

phluid61:
Explicitly disallow thirteen digits before the dot.


phluid61:
@mnot I think this is the right way to validate the number of digits in a decimal.

Issue #159, opened 2016-03-15
Allow cookies without key or value


inikulin:
As a part of discussion in whatwg/html#804 I've made a [research](http://inikulin.github.io/cookie-compat/) of the modern browsers compatibility with the RFC 6265. It appears that all browsers nowadays allows cookies without key or (in case of Safari) without value, thus making it de facto standard. However, it's debatable how cookies like `foo;` should be treated: as the cookie without value or without key. Thinking of cookie jar as some kind of key/value store makes it more logical to treat such cookie as cookies without value, but on the other hand, currently most implementers treats them as the cookies with the special empty key.

mnot:
Discussed at Berlin IETF; Julian points out that we should check previous discussion in cookie WG.
Martin points out that we should talk to the implementation that has divergent behaviour.

reschke:
Context: around <https://mailarchive.ietf.org/arch/msg/http-state/7UX_L6HCNbMIyuI2Q9LQ3V7wDgk>, ticket in <https://trac.tools.ietf.org/wg/httpstate/trac/ticket/1>

johnwilander:
For clarity, are we saying Chrome, Firefox, and Edge treat such a cookie as key: "" and value "foo", and Safari treats it like key: "foo" and value: ""?

annevk:
Per the table linked in OP Safari serializes input `foo` as `foo=` whereas others serialize as `foo`.

Firefox attempted to be more strict on cookie setting lacking `=` and found it broke a router: https://bugzilla.mozilla.org/show_bug.cgi?id=1551729. 😟

mikewest:
https://wpt.fyi/results/cookies/http-state/general-tests.html?label=experimental&label=master&aligned is the relevant test, in particular the following subtests:

* [0004](https://github.com/web-platform-tests/wpt/blob/master/cookies/http-state/resources/test-files/0004-test) sends `Set-Cookie: foo`. Chromium stores a cookie with an empty name and value of `foo`, serialized as `Cookie: foo`. Firefox does the same. Safari drops the cookie entirely. Given Firefox's experience with <https://bugzilla.mozilla.org/show_bug.cgi?id=1551729> and related bugs above, this seems like it might be baked into some user-visible sites that are unlikely to update. I'd suggest that we change this test expectation to `Cookie: foo`, and update the spec accordingly.

[...]

PR #1018, opened 2020-01-03
[RFC6265bis] Accept nameless cookies.

mikewest:
This patch alters the cookie parsing algorithm to treat
`Set-Cookie: token` as creating a cookie with an empty name and a value
of "token". It also rejects cookies with neither names nor values (e.g.
`Set-Cookie:` and `Set-Cookie: =`.

Tests in web-platform-tests/wpt#21093.

Closes httpwg/http-extensions#159.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1102,19 +1102,20 @@ parse a set-cookie-string:
         the set-cookie-string, and the unparsed-attributes is the empty
         string.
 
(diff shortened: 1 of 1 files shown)


mikewest:
@johnwilander: Ping?

mikewest:
Landing this based on the discussion in #159. @johnwilander: If the wording needs more cleanup, I'd appreciate feedback!

triple-underscore:
The cookie-pair ABNF rule still requires presence of `=`.
Is it intentional?

PR #1030, opened 2020-01-13
Add Wilander as editor and move Barth from Authors to Acknowledgements.

johnwilander:


mikewest:
I'll land this and tweak some nits in a followup.

PR #1017, opened 2019-12-27
[RFC6265bis, EDITORIAL]: Use `.example` TLD for examples.

mikewest:
This patch replaces `example.com` with `site.example`, and corrects
inadvertant usage of non-example domains in the hopes of avoiding
confusion.

Closes httpwg/http-extensions#706.

PR #1035, opened 2020-01-21
[RFC6265bis, Editorial]: Actually use `example` TLD.

mikewest:
https://github.com/httpwg/http-extensions/pull/1017 was incomplete,
leaving a number of inadvertant references to `*.com` sites, or the
`com` TLD. This patch, hopefully, is less incomplete.

h/t @triple-underscore and @bagder for pointing this out.


mikewest:
WDYT, @bagder?

PR #1038, opened 2020-01-22
[RFC6265bis, Editorial]: Define "public suffix" in the terminology section.

### LINKED ISSUE #1033: 6265bis: public suffix def
The spec currently says:

> The term “public suffix” is defined in a note in Section 5.3 of [RFC6265]

This should actually point to Section 5.4 (right now) of 6265bis.

That said, that note is hidden there very well; instead the contents should me moved up to the terminology section.

mikewest:
Closes #1033.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -326,16 +326,12 @@ same" matching algorithm for origins are defined in {{RFC6454}}.
 "Safe" HTTP methods include `GET`, `HEAD`, `OPTIONS`, and `TRACE`, as defined
 in Section 4.2.1 of {{RFC7231}}.
 
-A "public suffix" is a domain that is controlled by a public registry, such as
-"com", "co.uk", and "pvt.k12.wy.us". For example, `site.example`'s public
-suffix is `example`. User agents SHOULD use an up-to-date public suffix list,
-such as the one maintained at {{PSL}}.
-
-An origin's "registered domain" is the origin's host's public suffix plus the
-label to its left. That is, for `https://www.site.example`, the public suffix is
-`example`, and the registered domain is `site.example`. This concept is defined more
-rigorously in {{PSL}}, and is also known as "effective top-level domain plus one"
-(eTLD+1).
+A domain's "public suffix" is the portion of a domain that is controlled by a
+public registry, such as "com", "co.uk", and "pvt.k12.wy.us" {{PSL}}. A domain's
+"registerable domain" is the domain's public suffix plus the label to its left.
+That is, for `https://www.site.example`, the public suffix is `example`, and the
+registrable domain is `site.example`. This concept is defined more rigorously in
+{{PSL}}, which specifies a formal algorithm to obtain both.
 
 The term "request", as well as a request's "client", "current url", "method",
 and "target browsing context", are defined in {{FETCH}}.
@@ -948,7 +944,7 @@ following conditions holds:
 
 ## "Same-site" and "cross-site" Requests  {#same-site-requests}
 
-A request is "same-site" if its target's URI's origin's registered domain
+A request is "same-site" if its target's URI's origin's registrable domain
 is an exact match for the request's client's "site for cookies", or if the
 request has no client. The request is otherwise "cross-site".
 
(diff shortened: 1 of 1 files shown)


mikewest:
Would you mind taking a look, @reschke?

mikewest:
I think this document has the conceptual conceit that it might be possible to have distinct understandings of public suffixes. Browsers certainly have coalesced around https://publicsuffix.org/, but it's not clear to me that it's "normative" in any sense other than practice. A _de facto_ standard, not _de jure_.

I'm happy to just say "Use the PSL", as that's most likely to be most compatible?

reschke:
Well, if we have normative requirements attached to whether something is a public suffix, then we need to define how to decide that.

If there's a risk that the list ever moves, we could introduce an indirection through a new IANA registry (just holding the location of the public list, not the list itself).

mikewest:
The current normative requirements boil down to "Hey, user agent. Go figure out which suffixes you think are public. The PSL might be helpful. Have fun!" I'm happy to make that suggestion normative by bringing in the PSL in the same way URL does (https://url.spec.whatwg.org/#host-public-suffix). Would you be happy with that outcome?

reschke:
Yes, this seems to be needed (maybe not in the same words, but...).

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -326,11 +326,10 @@ same" matching algorithm for origins are defined in {{RFC6454}}.
 "Safe" HTTP methods include `GET`, `HEAD`, `OPTIONS`, and `TRACE`, as defined
 in Section 4.2.1 of {{RFC7231}}.
 
-The term "public suffix" is defined in a note in Section 5.3 of {{RFC6265}} as
-"a domain that is controlled by a public registry", and are also known as
-"effective top-level domains" (eTLDs). For example, `site.example`'s public
+A "public suffix" is a domain that is controlled by a public registry, such as
+"com", "co.uk", and "pvt.k12.wy.us". For example, `site.example`'s public
```
gibson042: Would it be worth making a comparison to "delegation-centric zone" from [RFC 8499](https://tools.ietf.org/html/rfc8499#page-26)? It's certainly possible for a public suffix to not be delegation-centric (e.g., `github.io`) and there are some delegation-centric zones that are not public suffixes (most notably the DNS root zone itself), but there is nevertheless substantial overlap in both concept (different authority between parent and child domain) and contents.

mikewest: I'm not sure that bringing another concept in would clarify, especially given the discrepancies you note between "delegation-centric zone" and existing user agent understanding of "public suffix". I think I'd prefer to continue using the term that user agents have generally applied to cookies.

mikewest:
@reschke I've update the patch to more explicitly point to the PSL's algorithm, and switched from "registered domain" to "registrable domain" to match URL. WDYT?

reschke:
Works for me...

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -326,17 +326,12 @@ same" matching algorithm for origins are defined in {{RFC6454}}.
...
-An origin's "registered domain" is the origin's host's public suffix plus the
-label to its left. That is, for `https://www.site.example`, the public suffix is
-`example`, and the registered domain is `site.example`. This concept is defined more
-rigorously in {{PSL}}, and is also known as "effective top-level domain plus one"
-(eTLD+1).
+A domain's "public suffix" is the portion of a domain that is controlled by a
+public registry, such as "com", "co.uk", and "pvt.k12.wy.us" {{PSL}}. A domain's
+"registerable domain" is the domain's public suffix plus the label to its left.
+That is, for `https://www.site.example`, the public suffix is `example`, and the
+registrable domain is `site.example`. This concept is defined more rigorously in
```
annevk: registrable*

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -326,17 +326,12 @@ same" matching algorithm for origins are defined in {{RFC6454}}.
...
-such as the one maintained by Mozilla at {{PSL}}.
-
-An origin's "registered domain" is the origin's host's public suffix plus the
-label to its left. That is, for `https://www.site.example`, the public suffix is
-`example`, and the registered domain is `site.example`. This concept is defined more
-rigorously in {{PSL}}, and is also known as "effective top-level domain plus one"
-(eTLD+1).
+A domain's "public suffix" is the portion of a domain that is controlled by a
+public registry, such as "com", "co.uk", and "pvt.k12.wy.us" {{PSL}}. A domain's
+"registerable domain" is the domain's public suffix plus the label to its left.
```
annevk: registrable*

mikewest: Gah. Thank you. :) Fixed in https://github.com/httpwg/http-extensions/commit/6f299dd3386c2126682c744d53db3e9c6a5f5340.

PR #1040, opened 2020-01-24
[RFC6265bis, Editorial] Consistently use 'last-access-time'.

### LINKED ISSUE #1032: Remove "MUST store last-access-time" for each cookie?
The original RFC 6265 as well as the current draft (-04) says that a user-agent **MUST** store last-access-time and then goes on to explain how to store a cookie... but never says how that access time would even be used.

I find no argument for saying it MUST be used. We can implement fully compliant cookie user-agents without storing that information.

mikewest:
Closes #1032.

@johnwilander @bagder, mind taking a look? John, in particular, you noted "But we should also have an explicit reference in my opinion.", which I didn't really understand. Do you want more than this patch?

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1537,7 +1537,7 @@ agent MUST evict cookies in the following priority order:
 4.  All cookies.
 
 If two cookies have the same removal priority, the user agent MUST evict the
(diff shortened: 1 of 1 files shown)


mikewest:
@johnwilander: Ping.

mikewest:
I'm just going to land this and close #1032, @johnwilander. If you have concerns, I'd love to review a PR!

PR #1044, opened 2020-01-30
make serialise decimal algo round early, fail fast

### LINKED ISSUE #1043: Decimal rounding
... still isn't right, as Kari noticed. E.g., a value of `1.9998` won't get rounded to `2.0`.

phluid61:
Round to three decimal places first, then check the domain, then do all the actual serialising.

It's potentially not very l13n-friendly; feel free to change "decimal point" or "to the left" etc., if something is more appropriate.

Fixes #1043

### DIFF
--- a/draft-ietf-httpbis-header-structure.md
+++ b/draft-ietf-httpbis-header-structure.md
@@ -621,17 +621,18 @@ Given an integer as input_integer, return an ASCII string suitable for use in a 
 
 ### Serializing a Decimal {#ser-decimal}
 
-Given a decimal_number as input_decimal, return an ASCII string suitable for use in a HTTP header value.
+Given a decimal number as input_decimal, return an ASCII string suitable for use in a HTTP header value.
 
-1. Let output be an empty string.
-2. If input_decimal is less than (but not equal to) 0, append "-" to output.
-3. Append input_decimal's integer component represented in base 10 (using only decimal digits) to output; if it is zero, append "0".
-4. If the number of characters appended in the previous step is greater than 12, fail serialisation.
-5. Append "." to output.
-6. If input_decimal's fractional component is zero, append "0" to output.
-7. Else if input_decimal's fractional component has up to three digits, append them represented in base 10 (using only decimal digits) to output.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-header-structure.md
```diff
@@ -621,17 +621,18 @@ Given an integer as input_integer, return an ASCII string suitable for use in a
...
-1. Let output be an empty string.
-2. If input_decimal is less than (but not equal to) 0, append "-" to output.
-3. Append input_decimal's integer component represented in base 10 (using only decimal digits) to output; if it is zero, append "0".
-4. If the number of characters appended in the previous step is greater than 12, fail serialisation.
-5. Append "." to output.
-6. If input_decimal's fractional component is zero, append "0" to output.
-7. Else if input_decimal's fractional component has up to three digits, append them represented in base 10 (using only decimal digits) to output.
-8. Otherwise, append the first three digits of input_decimal's fractional component (represented in base 10, using only decimal digits) to output, rounding the final digit to the nearest value, or to the even value if it is equidistant.
-9. Return output.
+Given a decimal number as input_decimal, return an ASCII string suitable for use in a HTTP header value.
```
clelland: This should probably just be "number", or "real number" to be extra pedantic.

(If we wanted to restrict to `Decimals`, as defined in [Section 3.3.2](https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-15#section-3.3.2), then the previous algorithm is probably correct, since rounding isn't even an issue.)

phluid61: That's a bigger issue than just removing the underscore. It potentially has far-reaching implications, too. (What is a 'key' if not an SH-key? Why is Boolean written with a capital? etc.)

clelland: Does this algorithm even accept input that is not a valid "Decimal", as defined in the spec? If not, then there's no reason to consider rounding at all. Decimals can have at most 3 fractional digits. If it needs rounding, it's not a Decimal.

But you're right, those questions do have bigger impact. I'll open a different issue to ask that question.

PR #1045, opened 2020-01-31
Fix "is can be" (header structure)

essen:
Small editorial fix.

I'm also not comfortable with "a integer", maybe remove "a"? Anyway that's a nitpick.


mnot:
Thanks @essen.

PR #1047, opened 2020-02-05
[RFC6265bis, Editorial] Typo'd a section number.

### LINKED ISSUE #1046: Bad link in 6265bis samesite enforcement section
In https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-04#section-5.3.7.1 under sub heading # 1 there is a paragraph, "Attackers can still pop up new windows or trigger...", which references section 2.1 for more information on that topic. However section 2.1 is "Conformance Criteria" so that doesn't seem the intended target.

mikewest:
Closes #1046.


mikewest:
@johnwilander I'm just going to land this (as it's trivial), and roll an -05 today.

reschke:
This should use a named section reference, no? Otherwise it can break again when section numbers change.

PR #1052, opened 2020-02-06
Minor structured header nits

clelland:
* Change "removing the first character" to "consuming the first character" in `#parse-key`
* Add "as" for consistency in `#ser-item`
* Correct `parameter-name` to `param_name` in `#ser-params`

### DIFF
--- a/draft-ietf-httpbis-header-structure.md
+++ b/draft-ietf-httpbis-header-structure.md
@@ -590,7 +590,7 @@ Given an ordered Dictionary as input_dictionary (each member having a member_nam
 
 ### Serializing an Item {#ser-item}
 
-Given an Item as bare_item and Parameters as item_parameters, return an ASCII string suitable for use in a HTTP header value.
+Given an Item as bare_item and Parameters as item_parameters, return an ASCII string suitable for use in a HTTP field value.
 
 1. Let output be an empty string.
 2. Append the result of running Serializing a Bare Item {{ser-bare-item}} with bare_item to output.
--- a/draft-ietf-httpbis-header-structure.md
+++ b/draft-ietf-httpbis-header-structure.md
@@ -565,7 +565,7 @@ Given a key as input_key, return an ASCII string suitable for use in a HTTP fiel
 
 0. Convert input_key into a sequence of ASCII characters; if conversion fails, fail serialization.
 1. If input_key contains characters not in lcalpha, DIGIT, "\_", "-", ".", or "\*" fail serialisation.
-2. If the first character of input_key is not lcalpha, fail parsing.
+2. If the first character of input_key is not lcalpha, fail serialisation.
 3. Let output be an empty string.
 4. Append input_key to output.
 5. Return output.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-header-structure.md
```diff
@@ -587,7 +587,7 @@ Given an ordered dictionary as input_dictionary (each member having a member_nam
 
 ### Serializing an Item {#ser-item}
 
-Given an item bare_item and parameters item_parameters as input, return an ASCII string suitable for use in a HTTP header value.
+Given an item as bare_item and parameters item_parameters as input, return an ASCII string suitable for use in a HTTP header value.
```
phluid61: I think this was right. Otherwise it would be "Given an item as base_item and parameters as item_parameters as input ..."

mnot: I tend to agree; that's the style we're using elsewhere.

clelland: Ahh, I didn't read this sentence correctly.

I read it as "GIven (an item named bare_item) and (parameters named item_parameters) as input to this function, ..."

but it's actually saying "Given an item (which is a tuple made up of bare_item and (parameters named item_parameters)) named input, ..."

The name "input" isn't used in the algorithm, which I think threw my parser off. Maybe the style used in #ser-dict would be appropriate here:

"Given an item (having a bare_item and item_parameters), return an ASCII string..."

phluid61: I've just realised the other algorithms use "as" to mean "named", and don't say "as input". For example, bare item:

> Given an item as input_item, return...

The right fix would probably be:

> Given an item as bare_item and parameters as item_parameters, return...

clelland: Yeah, "as input" isn't used at all elsewhere.

> Given an item as bare_item and parameters as item_parameters, return...

That would work, except that an item already has parameters, according to `{#item}`, I think.
But then again, `{#ser-bare-item}` also claims to take an item, so maybe it just works.

clelland: Fixed now.

draft-ietf-httpbis-header-structure.md
```diff
@@ -590,7 +590,7 @@ Given an ordered Dictionary as input_dictionary (each member having a member_nam
 
 ### Serializing an Item {#ser-item}
 
-Given an Item bare_item and Parameters item_parameters as input, return an ASCII string suitable for use in a HTTP field value.
+Given an Item as bare_item and Parameters as item_parameters, return an ASCII string suitable for use in a HTTP header value.
```
mnot: Needs to be "field value", not "header value".

clelland:
Found one more - step 2 of {{ser-key}} reads

    2. If the first character of input_key is not lcalpha, fail parsing.

It should say "fail serialisation"

mnot:
Many thanks!

Issue #1053, opened 2020-02-06
Structured headers uses COMMA without defining it


clelland:
The structured headers draft imports and uses a number of productions from RFC 5234 for typographically awkward characters and character classes (`SP`, `DQUOTE`, etc), but there is no definition of `COMMA` provided anywhere.

Similar to the way that `":"`, `";"`, `"\"`, `"("`, and `")"` are used without special treatment, `COMMA` can probably be replaced with `","` wherever it occurs.

PR #1054, opened 2020-02-06
Replace COMMA with ","

clelland:
The term COMMA is not defined anywhere in this document, or in any of the documents it imports. But we can just use the literal `","` instead.

Fixes: #1053


mnot:
Thanks.

PR #1063, opened 2020-02-18
Change Client Hints revision to 09

yoavweiss:


yoavweiss:
> 👍 ... but didn't you just push -10?

Indeed. 09 had some merge issues. Abandoning this one

martinthomson:
You should be able to use "-latest" instead of a specific version number in the document. That might save some of this type of messing around.

yoavweiss:
> You should be able to use "-latest" instead of a specific version number in the document. That might save some of this type of messing around.

TIL - thanks! :)

PR #1072, opened 2020-02-24
[client hints] LC feedback from @reschke

yoavweiss:
This PR addresses Client Hints [LC feedback](https://lists.w3.org/Archives/Public/ietf-http-wg/2020JanMar/0141.html) from @reschke.

### DIFF
--- a/draft-ietf-httpbis-client-hints.md
+++ b/draft-ietf-httpbis-client-hints.md
@@ -319,7 +319,7 @@ Features that define Client Hints will need to specify the related variants algo
 * PR 1064: Fix merge issues with "cost of sending hints".
 
 ## Since -10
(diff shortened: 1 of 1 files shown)


yoavweiss:
@reschke - I'm assuming you were referring to the xref. It creates a link in the HTML version that looks like "considerations outlined in the Security Considerations (Section 4) section of this document" where "Section 4" is linking to the right section.

I'm also happy to change it if you have an example of a better way to link.

reschke:
In that case, fine.

(I personally prefer consistency in these references; we ususally say "...outlined in the Security Considerations (Section x)..." which would translate to simpler MD.

But this is your choice :-)

yoavweiss:
Thanks for the feedback and for reviewing! :)

PR #1082, opened 2020-03-01
[client-hints] Move Variants reference to the Variants draft

yoavweiss:
This is moving the Variants-Client Hints interaction section from the Client Hints draft to the Variants one, as [requested](https://lists.w3.org/Archives/Public/ietf-http-wg/2020JanMar/0158.html)

### DIFF
--- a/draft-ietf-httpbis-variants.md
+++ b/draft-ietf-httpbis-variants.md
@@ -676,7 +676,10 @@ Variant-Key: (gold europe)
 Vary: Cookie
 ~~~
 
+## Client Hints {#client-hints}
 
+Client Hints {{CLIENT-HINTS}} might be combined with Variants response header field to enable fine-grained control of the cache key for improved cache efficiency.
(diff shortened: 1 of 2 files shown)


draft-ietf-httpbis-variants.md
```diff
@@ -676,7 +676,10 @@ Variant-Key: (gold europe)
 Vary: Cookie
 ~~~
 
+## Client Hints {#client-hints}
 
+Client Hints {{CLIENT-HINTS}} might be combined with Variants response header field to enable fine-grained control of the cache key for improved cache efficiency.
```
yoavweiss: The {{CLIENT-HINTS}} reference fails to build for me locally, but I'm not sure why...

yoavweiss:
@mnot  - friendly ping :)

yoavweiss:
> I don't think CH needs to be mentioned explicitly in Variants.

Removed the reference from Variants, so this is just deleting the Variants ref from the CH draft.

Issue #1068, opened 2020-02-21
* for dictionary member name


mnot:
In https://github.com/w3c/reporting/issues/177, @clelland asks for `*` as a dictionary member name.

I think this just means adding it to valid start characters in key.

PR #1084, opened 2020-03-03
Allow "*" for sh-dict member and parameter names

clelland:
This change to the key syntax allows dictionaries and parameters to use "*" as a starting character for names, allowing keys like

    *
    *value*
    *my*god*its*full*of*stars*

Closes: #1068


mnot:
LGTM. Do you want to add the change note, or shall I do that after merging?

mnot:
BTW - would take a PR with an example key that looks like the last one above :)

phluid61:
`*8J+Rhg==*=:8J+RjQ==:;*4pi6*`

PR #1091, opened 2020-03-03
digest: editorial reordering

LPardue:
@ioggstream I've taken an editorial stab at juggling the document a bit in order to improve its presentation and readability. It's still a bit WIP.

This is mainly just borrowing text from different places, and eliding some duplication. However, I may have inadvertently made changes or ommissions so PTAL and tell me.

Note it builds on top of #1086, so you might find it easier to compare the single commit https://github.com/httpwg/http-extensions/commit/43529457c31d4d660027d782ec2fff2b0466e767

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -216,6 +216,158 @@ interpreted as described in [RFC7230] and [RFC7231].
 The definition "validator" in this document is to be interpreted as described in
 Section 7.2 of [RFC7231].
 
+# The Digest Header Field {#digest-header}
+
+The Digest header field can appear in requests or responses. It contains a
+digest value of HTTP `representation data`, together with an indication of the
+algorithm used to produce the value.
+
+~~~
+   Digest = "Digest" ":" OWS 1#representation-data-digest
+   representation-data-digest = digest-algorithm "="
+                                <encoded digest output>
+   digest-algorithm = token
+~~~
+
+The resource is specified by the effective request URI and any `validator`
+contained in the message. See {#post-not-request-uri} for an example of how
+digest relates to header fields such as `Content-Location` (see [RFC7231]
+Section 3.1.4.2).
+
+The `representation-data-digest` is computed using a single digest-algorithm
+over the entire selected `representation data` of the resource, and then encoded
+in a format specific to the algorithm. For example, the `sha-256`
+digest-algorithm uses base64 encoding 
+
+~~~
+   Digest: sha-256=4REjxQ4yrqUVicfSKYNO/cF9zNj5ANbzgDZt3/h3Qxo=
+~~~
+
+A list of digest-algorithm values registered with IANA is presented in
+{{algorithms}}.
+
+A Digest header field MAY contain multiple `representation-data-digest` values.
+This could, for example, be useful for responses expected to reside in caches
+shared by users with different browsers.
+
+A recipient MAY ignore any or all of the representation-data-digests in a Digest
+header field. This allows the recipient to chose which digest-algorithm(s) to
+use for validation instead of verifying every received
+representation-data-digest.
+
+A sender MAY send a representation-data-digest using a digest-algorithm without
+knowing whether the recipient supports the digest-algorithm, or even knowing
+that the recipient will ignore it.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -216,6 +216,158 @@ interpreted as described in [RFC7230] and [RFC7231].
...
+~~~
+
+In practice, this means that computation of digests is dependent on the HTTP
+message and things that can change it, such as methods or metadata. For
+instance, a message's payload body can be affected by method, e.g. HEAD, or
+mechanisms like Range Requests RFC7233]). While the message body depends on
+`Transfer-Encoding` or other transformations that can be applied by an
+intermediary.
+
+The following examples show how representation metadata, payload transformations
```
ioggstream: I suggest to:

- add L283:L300 to {#resource-representation}
- eventually integrate the description in the examples

And here just say that 

```
Read carefully  {{resource-representation}} before implementing Digest.
The {{examples-unsolicited}} and {{examples-solicited}} sections show how representation metadata, http method, partial representations and other transformations
alter the payload body and in which cases digest is affected.
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -216,6 +216,158 @@ interpreted as described in [RFC7230] and [RFC7231].
...
+~~~
+HTTP/1.1 200 OK
+Content-Type: application/json
+Content-Encoding: br
+Digest: sha-256=4REjxQ4yrqUVicfSKYNO/cF9zNj5ANbzgDZt3/h3Qxo=
+
+iwiAeyJoZWxsbyI6ICJ3b3JsZCJ9Aw==
+~~~
+
+# The Want-Digest Header Field {#want-digest-header}
```
ioggstream: Ok for top-level section, but after §5.

LPardue:
Thanks for the feedback. I think the big question here is about the order of presentation of information. Personally, having come back to read the spec back-to-front with fresh eyes, starting with Resource Representation really didn't help to inform me what Digest Header was. I can imagine I am not alone in this and that it might cause kind individuals who review this document to bounce off. 

The Resource Representation and Digest Algorithm Values sections contain useful content, especially the text in early paragraphs, but the rest of the text boils down to lists of stuff with little normative text. Draft 01 spends 10 pages before it introduces the Digest header - compare that to RFC 3230 which in its entirety is only 10 pages. This is weird given that one premise for adopting the update update is that we can rely on core draft definitions now and things should be simpler.

Looking at contemporary drafts, they tend to bubble up the "meat" of the specification early on. So with the restructure, what I was trying to do is to say "Here's a way to carry an integrity checksum, calculating that checksum is dependent on a bunch on non-obvious stuff; you should read the background and check out the examples later in the document.".

RFC 3230 was written in a different era and we shouldn't follow its structure too much. So while it is logical to present Want-digest first, in practice I'm not convinced that many people care about it

ioggstream:
> starting with Resource Representation really didn't help to inform me what Digest Header was.
> and [reviewers] to bounce off.
> contemporary drafts, [..] bubble up the "meat" of the specification early on. 

Do you prefer an approach like this one https://github.com/httpwg/http-extensions/blob/master/draft-ietf-httpbis-cache-header.md

> The Resource Representation and Digest Algorithm Values sections contain useful content, 
> but [..] with little normative text.

That's true.

> Draft 01 spends 10 pages before it introduces the Digest header 
> - compare that to RFC 3230 which in its entirety is only 10 pages.
> RFC 3230 was written in a different era and we shouldn't follow its structure too much.

:D Yes, RFC3230 structure defined Digest at pg. 9 and 
I agree we can do better. 

> I was trying to [..] say "Here's a way to carry an integrity checksum, 
> calculating that checksum is dependent on a bunch on non-obvious stuff; 
> you should read the background and check out the examples later in the document.".

I think we can achieve the goal with multiple PRs moving one part at a time:
does it sound reasonable to you?

LPardue:
Definitely sounds reasonable, let me see how to pick this apart a bit

ioggstream:
cherry-picking stuff progressively using #1103 and a bunch of patch.

ioggstream:
Fixed with a series of patches described in #1103

PR #1094, opened 2020-03-04
Omitted True dictionary value can have parameters

### LINKED ISSUE #1083: Default dictionary value with parameters
Julian asked in e-mail why we can't have a dictionary with a default True value that is parameterised:

~~~
Example-Field: ;param=on-default-boolean

Parameter values are comparatively simpler, because they can't have any further substructure.

I don't quite get the example. In the spec I see:

 Example-DictHeader: a=?0, b, c=?1; foo=bar

If this would be instead:

 Example-DictHeader: a=?0, b, c; foo=bar

...what would break?
~~~

mnot:
Fixes #1083.


mnot:
Ping @clelland

clelland:
LGTM from here, thanks!

PR #1097, opened 2020-03-04
[client-hints] Editorial feedback

### LINKED ISSUE #1093: CH editorial feedback
Some of this is very nit-picky; feel free to do what you like with it...

## 1. Introduction

> Applications that want to allow the server to optimize content delivery and user experience based on such capabilities have, historically, had to rely on passive identification (e.g., by matching User-Agent (Section 5.5.3 of [RFC7231]) header field against an established database of client signatures), used HTTP cookies [RFC6265] and URL parameters, or use some combination of these and similar mechanisms to enable ad hoc content negotiation.

^ There's something weird going on with tense in this sentence.

> Such techniques are expensive to setup and maintain, are not portable across both applications and servers, and make it hard to reason for both client and server about which data is required and is in use during the negotiation:

"Such techniques are expensive... are not portable... and make it hard..." The third doesn't fit with the first two.

* "Client Hints mitigate the performance concerns..." probably just "Client Hints mitigate performance concerns..." Likewise for "the privacy concerns" below.

* "This document defines the Client Hints infrastructure, a framework that enables servers to opt-in to specific proactive content negotiation features, which will enable them to adapt their content accordingly." --> "This document defines Client Hints, a framework that enables servers to opt-in to specific proactive content negotiation features, adapting their content accordingly."

* `[UA-CH]` as a normative reference probably isn't going to fly, given its status. Why is it normative?

## 2.1. Sending Client Hints

> Clients control which Client Hints are sent in requests, based on their default settings, user configuration, and server preferences.

maybe instead:

> Clients choose what Client Hints to send in a request based on their default settings, user configuration, and server preferences expressed in `Accept-CH`.

## 2.2. Server Processing of Client Hints

> When doing so, and if the resource is cacheable, the server MUST also generate a Vary response header field (Section 7.1.4 of [RFC7231]) to indicate which hints can affect the selected response and whether the selected response is appropriate for a later request.

Be aware that HTTP doesn't _require_ presence of `Vary` when proactive conneg is in use; if the response is truly uncacheable, it isn't necessary.

> Further, depending on the hint used, the server can generate additional response header fields to convey related values to aid client processing.

maybe instead:

> Furthermore, the server can generate additional response header fields (as specified by the hint or hints in use) that convey related values to aid client processing.

## 3.1. The Accept-CH Response Header Field

> The Accept-CH response header field or the equivalent HTML meta element with http-equiv attribute (Section 4.2.5.3 of [HTML]) indicate server support for particular hints indicated in its value.

instead:

> The Accept-CH response header field indicates server support for the hints indicated in its value. This information can also be conveyed by the HTML meta element with the http-equip attribute (Section 4.2.5.3 of [HTML]).

BTW, nothing is said about precedence between the two different mechanisms; separate issue?

> When a client receives an HTTP response advertising support for provided list of Clients Hints, it SHOULD process it as origin ([RFC6454]) opt-in to receive Client Hint header fields advertised in the field-value, for subsequent same-origin requests.

The reference to 6454 is spurious, and it's pretty indirect. Perhaps:

> When a client receives an HTTP response containing Accept-CH, it indicates that the origin wishes to receive the indicated request header fields for subsequent same-origin requests.

~~~
* The opt-in MUST be delivered over a secure transport.
* The opt-in SHOULD be persisted and bound to the origin to enable delivery
   of Client Hints on subsequent requests to the server’s origin, and MUST NOT
  be persisted for an origin that isn’t HTTPS.
~~~

It's kind of weird to have this list just appear. Perhaps turn these into prose?

> For example, based on the Accept-CH example above,

Put "For example" above the example to introduce it.

## 3.2. Interaction with Caches

> When selecting an optimized response based on...

Limits CH to just "optimisation." Probably should just be:

> When selecting a response based on...

yoavweiss:
Closes #1093

### DIFF
--- a/draft-ietf-httpbis-client-hints.md
+++ b/draft-ietf-httpbis-client-hints.md
@@ -121,7 +121,7 @@ Client Hints mitigate performance concerns by assuring that clients will only se
 
 This document defines Client Hints, a framework that enables servers to opt-in to specific proactive content negotiation features, adapting their content accordingly. However, it does not define any specific features that will use that infrastructure. Those features will be defined in their respective specifications.
 
-One example of such a feature is the User Agent Client Hints feature {{!UA-CH}}.
+One example of such a feature is the User Agent Client Hints feature {{UA-CH}}.
 
 ## Notational Conventions
 
--- a/draft-ietf-httpbis-client-hints.md
+++ b/draft-ietf-httpbis-client-hints.md
@@ -67,6 +67,8 @@ normative:
       ins: A. van Kesteren
       name: Anne van Kesteren
       organization: Mozilla
+informative:
+  RFC6265:
   UA-CH:
     target: https://wicg.github.io/ua-client-hints/
     title: User Agent Client Hints
@@ -80,8 +82,6 @@ normative:
       name: Yoav Weiss
       organization: Google
 
-informative:
-  RFC6265:
 
 --- abstract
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-client-hints.md
```diff
@@ -117,11 +117,11 @@ However, traditional proactive content negotiation techniques often mean that cl
...
+Client Hints mitigate performance concerns by assuring that clients will only send the request headers when they're actually going to be used, and privacy concerns of passive fingerprinting by requiring explicit opt-in and disclosure of required headers by the server through the use of the Accept-CH response header.
 
-This document defines the Client Hints infrastructure, a framework that enables servers to opt-in to specific proactive content negotiation features, which will enable them to adapt their content accordingly. However, it does not define any specific features that will use that infrastructure. Those features will be defined in their respective specifications.
+This document defines Client Hints, a framework that enables servers to opt-in to specific proactive content negotiation features, adapting their content accordingly. However, it does not define any specific features that will use that infrastructure. Those features will be defined in their respective specifications.
 
-One example of such a feature is the User Agent Client Hints feature {{UA-CH}}.
+One example of such a feature is the User Agent Client Hints feature {{!UA-CH}}.
```
mnot: ! makes it a normative ref

yoavweiss: Oops. So your comment was to remove the reference altogether? Or can I keep it as informative?

mnot: Informative is probably OK.

yoavweiss: Done!

mnot:
LGTM! thx

Issue #1085, opened 2020-03-03
Serialize vs serialise


clelland:
The Structured Headers spec is inconsistent in its use of UK vs US forms of "seriali{sz}e".

Quick count:

serialization: 3
serialize: 3
serialized: 1
serializer: 3
serializing: 40

serialisation: 25
serialise: 0
serialised: 3
serialiser: 1
serialising: 5

We should pick one -- the RFC style guide doesn't have an opinion (https://tools.ietf.org/html/rfc7322#section-3.1 says either is fine, as long as the document is internally consistent).

I'm not sure if there was previous discussion on this when parameteri**s**ed lists (v10) became lists with parameteri**z**ed members (v11), but that term doesn't exist in the document anymore.

Current count is 7 instances of `/parameteris.*/` in the document, and 0 for `/parameteriz.*/`

mnot:
RFCs standardise on US English (blech). 

This is one of the things that I was planning to let the RFC Editor sort out (it's always good to give them something to do ;) -- but will take a PR if you feel inclined.

clelland:
Sure, why not :)

MikeBishop:
Not quite....  RFC 7322 says:

    The RFC publication language is English.  Spelling may be either
    American or British, as long as an individual document is internally
    consistent.  Where both American and British English spelling are
    used within a document or cluster of documents, the text will be
    modified to be consistent with American English spelling.

clelland:
That's what I read (and referenced in https://github.com/httpwg/http-extensions/issues/1085#issue-574637459) but I figured that @mnot knows more of the processes than I do :)

mnot:
IME they're pretty aggressive about deciding that it's a "mix" and therefore US wins :)

PR #1098, opened 2020-03-04
Standardi{sz}e on US English spelling

clelland:
Changes:
  behaviour -> behavior (and related words)
  parameterised -> parameterized
  serialise -> serialize (and related words)
  unrecognised -> unrecognized

Closes: #1085


mnot:
Thanks!

clelland:
> Looks alright, with two caveats: 1. Ugh; and 2. I haven't checked myself so I'm assuming you got all of them. But more fixes is more good, so :+1:

1. Half-agreed -- Canadian English is somewhere in the middle, so I had a hard time getting rid of the "u" in "behaviour", though I don't mind the zed's. :smile:

2. I grepped the spec against a list of some 1700 words with spelling differences, and those were all of the ones which were caught. I changed them all, except those in the changelog, in case they had historical significance. (And that'll be dropped in the RFC in any case)

PR #1104, opened 2020-03-05
sh: fix overlong ABNF lines

reschke:

PR #1108, opened 2020-03-05
sh: Remove non-ASCII chars

clelland:


clelland:
This just fixes a couple of non-ASCII U+2212 MINUS-SIGN characters that crept in with #1104.

reschke:
Actually, they've been there before. What #1104 broke was the encoding.

clelland:
Ahh -- I thought that was GitHub just exposing the UTF8-ness in the diff when I looked at it.

I'll leave it to editors to decide whether non-ASCII is appropriate in the markdown version of the spec; I'm sure that the IETF has a process for dealing with it when it needs to be a text RFC.

reschke:
No, this really should be an ASCII minus.

LPardue:
https://mailarchive.ietf.org/arch/msg/rfc-interest/hWOHbgqeEj0vRRoj-qsFcI10_jQ/

Issue #1103, opened 2020-03-05
Digest editorial reordering


ioggstream:
## I expect
@LPardue WDYT about this roadmap? I'd address each one in a separate PR.
If you agree, I can cherry-pick the first ones from your PR so you won't have to do it twice :)

Digest Headers to be reordered:

- [x] move the "deprecated algorithms" security consideration directly in {{digest-agorithms}} #1109
- [x] merge the {{digest-algorithm-encoding-examples}} in its parent section {{representation-digest}} #1110
- [x] move Digest Header and Want-Digest header outside Header Field Specifications #1111 #1116
- [x] move {{representation-digest}} under {{digest-header}} #1117
- [x] move {{digest-algorithms}} down #1112
- [x] move {{resource-representation}} examples to Annex B #1122 
- [ ] ensure people start reading from  {{representation-digest}} #1123 
- [ ] refactor Digest to be more direct #1124 

Optional:

- [ ] integrate examples from #1091 into {{resource representation}}

LPardue:
That sounds like a good plan, thanks. Please action it and I'll review promptly.

This is a set of editorial changes that I think would add value to the specification readability. My jumbo PR (#1091) had some problems, so lets focus on landing the easy-to-agree items and not try to rush everything through before the submission deadline if we are uncomfortable. We can continue working on the thornier editorials even after the submission deadline.

ioggstream:
Here's a first bunch of PRs. PTAL :) thanks for all the support!

ioggstream:
@LPardue I'm trying to refactor and I have this proposal:

https://httpwg.org/http-extensions/ioggstream-1103-test/draft-ietf-httpbis-digest-headers.html

we need rewording but just look at the sections ordering and let me know

LPardue:
This section outline looks good. I approved, #1122 so let merge it. What's next?

ioggstream:
> This section outline looks good. I approved, #1122 so let's merge it. What's next?

Now we have #1123 and #1124 which should end the refactoring.

LPardue:
All the PRs look good, even the optional one. Thanks for your work on this.

Let's get them merged, then take a once-over review to catch any basic errors that might have crept in. Then one of us can cut the release.

PR #1109, opened 2020-03-05
Refactor deprecated algorithms. See #1103.

ioggstream:
## This PR

Moves deprecated algorithms in the iana table.

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -337,8 +337,13 @@ The BNF for "parameter" is as is used in [RFC7230]. All digest-algorithm values
 are case-insensitive.
 
 The Internet Assigned Numbers Authority (IANA) acts as a registry for
-digest-algorithm values.  The registry contains the following tokens.
+digest-algorithm values.
+The registry contains the tokens listed below.
 
+Some algorithms, although registered, have since been found vulnerable:
+the MD5 algorithm MUST NOT be used due to collision attacks [CMU-836068]
+and the SHA algorithm is NOT RECOMMENDED due
+to collision attacks [IACR-2019-459].
 
   SHA-256:
   :  
@@ -407,7 +412,7 @@ To allow sender and recipient to provide a checksum which is independent from
 
 
   ID-SHA-256:
-  :
+  :  
     * Description: The sha-256 digest of the representation-data of the resource when no
       content coding is applied (eg. `Content-Encoding: identity`)
     * Reference: [RFC6234], [RFC4648], this document.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -337,25 +337,29 @@ The BNF for "parameter" is as is used in [RFC7230]. All digest-algorithm values
 are case-insensitive.
 
 The Internet Assigned Numbers Authority (IANA) acts as a registry for
-digest-algorithm values.  The registry contains the following tokens.
+The registry contains the tokens listed below.
```
LPardue: seem to have lost digest-algorithm value and the sentence is broken.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -337,25 +337,29 @@ The BNF for "parameter" is as is used in [RFC7230]. All digest-algorithm values
...
-  :  
+  :
     * Description: The SHA-256 algorithm [RFC6234].  The output of
       this algorithm is encoded using the base64 encoding [RFC4648].
     * Reference: [RFC6234], [RFC4648], this document.
     * Status: standard
 
   SHA-512:
-  :  
+  :
```
LPardue: The trailing whitespace is actually needed after all of these colons in order to make the list line up all nicely (I discovered this yesterday). I think that requiring trailing whitespace is a smell that something is broken but I don't know how to fix the root cause.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -337,25 +337,29 @@ The BNF for "parameter" is as is used in [RFC7230]. All digest-algorithm values
 are case-insensitive.
 
 The Internet Assigned Numbers Authority (IANA) acts as a registry for
-digest-algorithm values.  The registry contains the following tokens.
+The registry contains the tokens listed below.
 
+Some algorithms, although registered, have since been proven to be vulnerable:
+the MD5 algorithm MUST NOT be used as it has been found vulnerable to collision
+attacks [CMU-836068], the SHA algorithm is NOT RECOMMENDED as it has been found
+vulnerable to collision attacks [IACR-2019-459].
```
LPardue: how about this?
```suggestion
Some algorithms, although registered, have since been found vulnerable:
the MD5 algorithm MUST NOT be used due to collision attacks
[CMU-836068] and the SHA algorithm is NOT RECOMMENDED due to
collision attacks [IACR-2019-459].
```

draft-ietf-httpbis-digest-headers.md
ioggstream: Yep. The editor removes them every time and I've to `vim` to avoid that.

PR #1112, opened 2020-03-05
Move algos down. See #1103.

ioggstream:
## This PR

Moves Digest Algorithms down


ioggstream:
I verified the formatting here https://httpwg.org/http-extensions/ioggstream-1103-move-digest-algorithms-down/draft-ietf-httpbis-digest-headers.html#algorithms

PR #1117, opened 2020-03-06
Move digest computation under Digest. See #1103

ioggstream:
## This PR

moves the computation part under Digest﻿

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -324,32 +324,6 @@ Location: /authors/123
 {"id": "123", "author": "Camilleri"}
 ~~~
 
-## Computing Digests of Representation Data {#representation-digest}
-
-A representation digest is the value of the output of a digest algorithm,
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -398,6 +372,32 @@ Two examples of its use are
...
 
+## Computing Digests of Representation Data {#representation-digest}
+
+A representation digest is the value of the output of a digest algorithm,
+together with an indication of the algorithm used (and any parameters).
+
+~~~
+   representation-data-digest = digest-algorithm "="
+                                <encoded digest output>
+~~~
```
LPardue: I think there is an opportunity to merge this art with the art for the definition of Digest Header. That way it becomes clear exactly what composes the Digest header, rather than needing to flick back and forth.

ioggstream: Yes, that's in another PR :D

PR #1123, opened 2020-03-06
Introduce resource digest #1103

ioggstream:
## This PR

Introduces properly Representation Digest, which is the core concept of the spec.

The original draft name was in fact `Representation Digest`.

See #1103

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -216,10 +216,10 @@ interpreted as described in [RFC7230] and [RFC7231].
 The definition "validator" in this document is to be interpreted as described in
 Section 7.2 of [RFC7231].
 
-# The Digest Header Field {#digest-header}
+# Representation Digest {#representation-digest}
 
-To avoid inconsistencies, an integrity mechanism for HTTP messages should
-decouple the checksum calculation from:
+The representation digest is an integrity mechanism for HTTP resources that
+decouples the checksum calculation from:
 
 - the payload body - which may be altered by mechanism like Range Requests
   [RFC7233] or the method (eg. HEAD);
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -216,31 +216,58 @@ interpreted as described in [RFC7230] and [RFC7231].
...
+~~~
+
+can be:
+
+- fully contained in the message body,
+- partially-contained in the message body,
+- or not at all contained in the message body.
+
 {{resource-representation}} contains many examples
 showing the HTTP semantics impact on the message and payload body.
```
LPardue: This still reads a bit disjointed to me, so how about something like the below, although it maybe has some errors

" The representation digest is an integrity mechanism for HTTP resources which uses a checksum that is calculated independently of the payload body and message body. It is calculated using the representation data (see [RFC7231]), can be fully or partially contained in the message body, or not contained at all:

~~~
   representation-data := Content-Encoding( Content-Type( bits ) )
~~~

It is prudent to consider the effect HTTP semantics in relation to resources. For example, the message body can be affected by range-requests or different methods such as HEAD, while the payload body is dependent on Transfer encoding or other transformations. {{resource-representation}} contains several examples to help illustrate the effects."

ioggstream: Sounds good, I'll update the PR tomorrow! Thanks being so patient with a non-native speaker!

draft-ietf-httpbis-digest-headers.md
```diff
@@ -218,29 +218,19 @@ Section 7.2 of [RFC7231].
...
 
-can be:
-
-- fully contained in the message body,
-- partially-contained in the message body,
-- or not at all contained in the message body.
-
-{{resource-representation}} contains many examples
-showing the HTTP semantics impact on the message and payload body.
+It is prudent to consider the effect of the HTTP semantics on the messages;
```
```suggestion
This takes into account the effect of the HTTP semantics on the messages;
```
@LPardue We are not giving implementors the freedom to be imprudent.

PR #1128, opened 2020-03-10
Some suggestions for forward compatibility

martinthomson:
Some of this is rewording.

There are two new paragraphs talking about extensions: one deals with dictionaries, the other deals
with extensions to a SH definition.

I also removed the "starts with 'h'" example for greasing.  We know that X- lead to all sorts of
mess, so choosing a bad example here could be just as bad.  I'm happy to split that out or be talked
into backing it out, of course.

### DIFF
--- a/draft-ietf-httpbis-header-structure.md
+++ b/draft-ietf-httpbis-header-structure.md
@@ -143,13 +143,17 @@ To specify a HTTP field as a Structured Field, its authors needs to:
 
 Typically, this means that a field definition will specify the top-level type -- List, Dictionary or Item -- and then define its allowable types, and constraints upon them. For example, a header defined as a List might have all Integer members, or a mix of types; a header defined as an Item might allow only Strings, and additionally only strings beginning with the letter "Q". Likewise, Inner Lists are only valid when a field definition explicitly allows them.
 
-When parsing fails, the field is ignored (see {{text-parse}}); in most situations, violating field-specific constraints should have the same effect. Thus, if a header is defined as an Item and required to be an Integer, but a String is received, it will by default be ignored. If the field requires different error handling, this should be explicitly specified.
+When parsing fails, the entire field is ignored (see {{text-parse}}); in most situations, violating field-specific constraints should have the same effect. Thus, if a header is defined as an Item and required to be an Integer, but a String is received, the field will by default be ignored. If the field requires different error handling, this should be explicitly specified.
 
-However, both Items and Inner Lists allow parameters as an extensibility mechanism; this means that values can later be extended to accommodate more information, if need be. As a result, field specifications are discouraged from defining the presence of an unrecognized Parameter as an error condition.
+Both Items and Inner Lists allow parameters as an extensibility mechanism; this means that values can later be extended to accommodate more information, if need be. To preserve forward compatibility, field specifications are discouraged from defining the presence of an unrecognized Parameter as an error condition.
 
-To help assure that this extensibility is available in the future, and to encourage consumers to use a complete parser implementation, a field definition can specify that "grease" Parameters be added by senders. For example, a specification could stipulate that all Parameters beginning with the letter "h" are reserved for this use, and then encourage them to be sent on some portion of requests. This helps to discourage recipients from writing a parser that does not account for Parameters.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-header-structure.md
```diff
@@ -143,13 +143,17 @@ To specify a HTTP field as a Structured Field, its authors needs to:
...
 
-Note that a field definition cannot relax the requirements of this specification because doing so would preclude handling by generic software; they can only add additional constraints (for example, on the numeric range of Integers and Decimals, the format of Strings and Tokens, the types allowed in a Dictionary's values, or the number of Items in a List). Likewise, field definitions can only use this specification for the entire field value, not a portion thereof.
+Specifications that use Dictionaries can allow for forward compatibility by only defining type and semantics for specific dictionary keys. Such a specification should identify the type corresponding to each key and require that the presence, value, and type associated with unknown keys be ignored. A later specification can set specific constraints on types and semantics for any keys that it defines.
```
mnot: I agree with the sentiment, but I think this better expresses it:

Specifications that use Dictionaries can also allow for forward compatibility by requiring that the presence of -- as well as value and type associated with -- unknown members be ignored. Later specifications can then add additional members, specifying constraints on them as appropriate.

draft-ietf-httpbis-header-structure.md
```diff
@@ -143,13 +143,17 @@ To specify a HTTP field as a Structured Field, its authors needs to:
...
 
-To help assure that this extensibility is available in the future, and to encourage consumers to use a complete parser implementation, a field definition can specify that "grease" Parameters be added by senders. For example, a specification could stipulate that all Parameters beginning with the letter "h" are reserved for this use, and then encourage them to be sent on some portion of requests. This helps to discourage recipients from writing a parser that does not account for Parameters.
+To further assure that this extensibility is available in the future, and to encourage consumers to use a complete parser implementation, a field definition can specify that "grease" Parameters be added by senders. A specification could stipulate that all Parameters that fit a set pattern are reserved for this use and then encourage them to be sent on some portion of requests. This helps to discourage recipients from writing a parser that does not account for Parameters.
```
mnot: s/set/defined/, perhaps?

draft-ietf-httpbis-header-structure.md
```diff
@@ -143,13 +143,17 @@ To specify a HTTP field as a Structured Field, its authors needs to:
...
 
-Note that a field definition cannot relax the requirements of this specification because doing so would preclude handling by generic software; they can only add additional constraints (for example, on the numeric range of Integers and Decimals, the format of Strings and Tokens, the types allowed in a Dictionary's values, or the number of Items in a List). Likewise, field definitions can only use this specification for the entire field value, not a portion thereof.
+Specifications that use Dictionaries can allow for forward compatibility by only defining type and semantics for specific dictionary keys. Such a specification should identify the type corresponding to each key and require that the presence, value, and type associated with unknown keys be ignored. A later specification can set specific constraints on types and semantics for any keys that it defines.
+
+An extension to a field specification can require that an entire field be ignored by a recipient that understands the extension if constraints on the value of keys or parameters it defines are not met.
```
mnot: Just to align with language elsewhere:

An extension to a structured field can then require that an entire field value be ignored by a recipient that understands the extension if constraints on the value it defines are not met.

draft-ietf-httpbis-header-structure.md
martinthomson:
```suggestion
Specifications that use Dictionaries can also allow for forward compatibility by requiring that the presence of -- as well as value and type associated with -- unknown members be ignored. Later specifications can then add additional members, specifying constraints on them as appropriate.
```

draft-ietf-httpbis-header-structure.md
martinthomson:
```suggestion
To further assure that this extensibility is available in the future, and to encourage consumers to use a complete parser implementation, a field definition can specify that "grease" Parameters be added by senders. A specification could stipulate that all Parameters that fit a defined pattern are reserved for this use and then encourage them to be sent on some portion of requests. This helps to discourage recipients from writing a parser that does not account for Parameters.
```

draft-ietf-httpbis-header-structure.md
martinthomson:
```suggestion
An extension to a structured field can then require that an entire field value be ignored by a recipient that understands the extension if constraints on the value it defines are not met.
```

PR #1131, opened 2020-03-10
[Client Hints] Remove unused references

yoavweiss:


yoavweiss:
Landing as this change is not substantive

PR #1140, opened 2020-03-12
Reference new HTTP core semenatics and use field terminology

### LINKED ISSUE #1137: Digest is a message field in the new parlance
The phrase "header field" is no longer the accepted form.  The generic form is "message field".

This matters more for Digest, which is a good candidate for a trailer.

LPardue:
The PR is intended to fix #1137, which asks us to use updated terminology in HTTP core documents.

I've attempted to update all references of RFC 7230 to draft-ietf-httpbis-semantics. A reference to RFC7231 remains but is demoted to informative.

The change converts `Digest` and `Want-Digest` into `fields` meaning they can appear in header or trailer field sections. As noted there is a clear case for Digest trailers, there seems no reason to restrict Want-Digest to request header fields but I can be easily persuaded.

Remaining references to `header fields` relate to externally defined headers that are coined in draft-ietf-httpbis-semantics.

Finally, the IANA registry handling is also updated.

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -25,6 +25,7 @@ author:
     email: lucaspardue.24.7@gmail.com
 
 normative:
+  I-D.ietf-httpbis-semantics:
   RFC1321:
   RFC3174:
   RFC1950:
@@ -35,9 +36,6 @@ normative:
   RFC4648:
   RFC5234:
   RFC6234:
-  RFC7230:
-  RFC7231:
-  RFC7233:
   RFC7405:
   RFC8174:
   UNIX:
@@ -71,6 +69,7 @@ normative:
 
 informative:
   RFC2818:
+  RFC7231:
   RFC7396:
   SRI:
     title: "Subresource Integrity"
@@ -86,13 +85,13 @@ informative:
 
 --- abstract
 
-This document defines the Digest and Want-Digest header fields for HTTP, thus
-allowing client and server to negotiate an integrity checksum of the exchanged
-resource representation data.
+This document defines the HTTP Digest and Want-Digest fields, thus allowing
+client and server to negotiate an integrity checksum of the exchanged resource
+representation data.
 
 This document obsoletes RFC 3230. It replaces the term "instance" with
-"representation", which makes it consistent  with the HTTP Semantic and Context
-defined in RFC 7231.
+"representation", which makes it consistent with the HTTP Semantic and Context
+defined in draft-ietf-httpbis-semantics.
 
 
 --- note_Note_to_Readers
@@ -129,7 +128,7 @@ be combined with other mechanisms that protect representation-metadata, such as
 digital signatures, in order to protect the desired parts of an HTTP exchange in
 whole or in part.
 
-## A Brief History of Integrity Header Fields {#history}
+## A Brief History of HTTP Integrity Fields {#history}
 
 The Content-MD5 header field was originally introduced to provide integrity, but
 HTTP/1.1 ([RFC7231], Appendix B) obsoleted it:
@@ -138,19 +137,20 @@ HTTP/1.1 ([RFC7231], Appendix B) obsoleted it:
   > inconsistently implemented with respect to partial responses.
 
 [RFC3230] provided a more flexible solution introducing the concept of
-"instance", and the header fields `Digest` and `Want-Digest`.
+"instance", and the fields `Digest` and `Want-Digest`.
 
 ## This Proposal
 
-The concept of `selected representation` defined in [RFC7231] made [RFC3230]
-definitions inconsistent with the current standard. A refresh was then required.
+The concept of `selected representation` defined in Section 6 of
+[I-D.ietf-httpbis-semantics] makes [RFC3230] definitions inconsistent with
+current HTTP semantics. This document updates the `Digest` and `Want-Digest`
+field definitions to align with [I-D.ietf-httpbis-semantics] concepts.
 
-This document updates the `Digest` and `Want-Digest` header field definitions to
-align with [RFC7231] concepts.
-
-This approach can be easily adapted to use-cases where the transferred data does
-require some sort of manipulation to be considered a representation or conveys a
-partial representation of a resource (eg. Range Requests [RFC7233]).
+Basing `Digest` on selected representation makes it more straightforward to
+apply it to use-cases where the transferred data does require some sort of
+manipulation to be considered a representation, or conveys a partial
+representation of a resource eg. Range Requests (see Section 8.3 of
+[I-D.ietf-httpbis-semantics]).
 
 Changes are semantically compatible with existing implementations and better
 cover both the request and response cases.
@@ -159,9 +159,9 @@ The value of `Digest` is calculated on selected representation, which is tied to
 the value contained in any `Content-Encoding` or `Content-Type` header fields.
 Therefore, a given resource may have multiple different digest values.
 
-To allow both parties to exchange a Digest of a representation with [no content
-codings](https://tools.ietf.org/html/rfc7231#section-3.1.2.1) two more
-algorithms are added (`ID-SHA-256` and `ID-SHA-512`).
+To allow both parties to exchange a Digest of a representation with no content
+codings (see Section 6.1.2 of [I-D.ietf-httpbis-semantics]) two more algorithms
+are added (`ID-SHA-256` and `ID-SHA-512`).
 
 ## Goals
 
@@ -176,15 +176,15 @@ The goals of this proposal are:
 
 The goals do not include:
 
-  HTTP Message integrity:
+  HTTP message integrity:
   : The digest mechanism described here does not cover the full HTTP message
     nor its semantic, as representation metadata are not included in the
     checksum.
 
-  Header field integrity:
+  HTTP field integrity:
   : The digest mechanisms described here cover only representation and selected
     representation data, and do not protect the integrity of associated
-    representation metadata or other message header fields.
+    representation metadata or other message fields.
 
   Authentication:
   : The digest mechanisms described here are not meant to support authentication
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -138,19 +137,20 @@ HTTP/1.1 ([RFC7231], Appendix B) obsoleted it:
...
 [RFC3230] provided a more flexible solution introducing the concept of
-"instance", and the header fields `Digest` and `Want-Digest`.
+"instance", and the fields `Digest` and `Want-Digest`.
 
 ## This Proposal
 
-The concept of `selected representation` defined in [RFC7231] made [RFC3230]
-definitions inconsistent with the current standard. A refresh was then required.
+The concept of `selected representation` defined in Section 6 of
+[I-D.ietf-httpbis-semantics] makes [RFC3230] definitions inconsistent with
```
martinthomson: I would use `{{!SEMANTICS=I-D.ietf-httpbis-semantics}}`.  But that's just me.  I think that you can do something with the reference above to get a nicer symbolic reference as well, but I forget the syntax.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -138,19 +137,20 @@ HTTP/1.1 ([RFC7231], Appendix B) obsoleted it:
...
+current HTTP semantics. This document updates the `Digest` and `Want-Digest`
+field definitions to align with [I-D.ietf-httpbis-semantics] concepts.
 
-This document updates the `Digest` and `Want-Digest` header field definitions to
-align with [RFC7231] concepts.
-
-This approach can be easily adapted to use-cases where the transferred data does
-require some sort of manipulation to be considered a representation or conveys a
-partial representation of a resource (eg. Range Requests [RFC7233]).
+Basing `Digest` on selected representation makes it more straightforward to
```
martinthomson:
```suggestion
Basing `Digest` on the selected representation makes it more straightforward to
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -293,11 +293,10 @@ Two examples of its use are
 ~~~
 
 
-# The Want-Digest Header Field {#want-digest-header}
+# The Want-Digest Field {#want-digest-field}
```
martinthomson:
```suggestion
# The Want-Digest Field {#want-digest}
```

LPardue:
I intentionally left out changing the title for fear of bike shedding. But now I realise failing to change makes this a non-atomic commit. So let me fix it.

martinthomson:
We all need to respect editor discretion in this.  Maybe don't call it "Hashing Stuff" and we'll be good.

draft-ietf-httpbis-digest-headers.md
LPardue: I take this on board. One problem I encountered was how to reference the core document in the abstract, which can't be a real reference othereise it triggers idnits IIUC

I inclined to modify reference style separately in a single pass PR..

LPardue:
I asked explicitly about the title because this document was called "Resource Digests" before adoption. And we were asked to rename it to "Digest headers", which was fine. I do see the humor in yet another change, but I'd like to get it right and avoid another one!

draft-ietf-httpbis-digest-headers.md
```diff
@@ -86,13 +85,13 @@ informative:
...
-resource representation data.
+This document defines the HTTP Digest and Want-Digest fields, thus allowing
+client and server to negotiate an integrity checksum of the exchanged resource
+representation data.
 
 This document obsoletes RFC 3230. It replaces the term "instance" with
-"representation", which makes it consistent  with the HTTP Semantic and Context
-defined in RFC 7231.
+"representation", which makes it consistent with the HTTP Semantic and Context
+defined in draft-ietf-httpbis-semantics.
```
```suggestion
"representation", which makes it consistent with the HTTP Semantic and Context.
```
Referencing the draft or RFC does not add anything here.

draft-ietf-httpbis-digest-headers.md
ioggstream: Once we close this PR we can merge Martin's suggestion in a separate PR or in the final commit

draft-ietf-httpbis-digest-headers.md
```diff
@@ -138,19 +137,20 @@ HTTP/1.1 ([RFC7231], Appendix B) obsoleted it:
...
+current HTTP semantics. This document updates the `Digest` and `Want-Digest`
+field definitions to align with [I-D.ietf-httpbis-semantics] concepts.
 
-This document updates the `Digest` and `Want-Digest` header field definitions to
-align with [RFC7231] concepts.
-
-This approach can be easily adapted to use-cases where the transferred data does
-require some sort of manipulation to be considered a representation or conveys a
-partial representation of a resource (eg. Range Requests [RFC7233]).
+Basing `Digest` on the selected representation makes it more straightforward to
```
```suggestion
Basing `Digest` on the selected representation makes it straightforward to
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -250,11 +251,11 @@ The example below shows the  `sha-256` digest-algorithm which uses base64 encodi
...
 
-# The Digest Header Field {#digest-header}
+# The Digest Field {#digest-field}
 
-The Digest header field contains a list of one or more representation digest values
-as defined in {{representation-digest}}.
-It can be used in both request and response.
+The Digest field contains a list of one or more representation digest values as
+defined in {{representation-digest}}. It can be used in both request and
+response.
```
```suggestion
The Digest field contains a list of one or more representation digest values
as defined in {{representation-digest}}.
It can be used in both request and response.
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -263,20 +264,19 @@ It can be used in both request and response.
 The resource is specified by the effective request URI and any `validator`
 contained in the message.
 
-The relationship between Content-Location (see [RFC7231] Section 3.1.4.2)
-and Digest is demonstrated in {{post-not-request-uri}}.
-A comprehensive set of examples showing the impacts of
-representation metadata, payload transformations and HTTP methods on digest
-is provided in {{examples-unsolicited}} and {{examples-solicited}}.
+The relationship between Content-Location (see Section 6.2.5 of
+[I-D.ietf-httpbis-semantics]) and Digest is demonstrated in
```
ioggstream: Like above, the CRLF at line 268 doesn't break the line in HTML nor in the TXT
so I won't reflow that.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -263,20 +264,19 @@ It can be used in both request and response.
...
+{{post-not-request-uri}}. A comprehensive set of examples showing the impacts of
+representation metadata, payload transformations and HTTP methods on Digest is
+provided in {{examples-unsolicited}} and {{examples-solicited}}.
 
-A Digest header field MAY contain multiple representation-data-digest values.
-This could be useful for responses expected to reside in caches shared by users
-with different browsers, for example.
+A Digest field MAY contain multiple representation-data-digest values. This
+could be useful for responses expected to reside in caches shared by users with
+different browsers, for example.
```
```suggestion
A Digest field MAY contain multiple representation-data-digest values.
This could be useful for responses expected to reside in caches shared by users
with different browsers, for example.
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -263,20 +264,19 @@ It can be used in both request and response.
...
 
 A recipient MAY ignore any or all of the representation-data-digests in a Digest
-header field. This allows the recipient to choose which digest-algorithm(s) to
-use for validation instead of verifying every received
-representation-data-digest.
+field. This allows the recipient to choose which digest-algorithm(s) to use for
+validation instead of verifying every received representation-data-digest.
```
```suggestion
field. This allows the recipient to choose which digest-algorithm(s) to
use for validation instead of verifying every received representation-data-digest.
```
Like above, minimize patch size.

draft-ietf-httpbis-digest-headers.md
ioggstream: Let's do what the other docs do.

draft-ietf-httpbis-digest-headers.md
LPardue: I disagree. The main purpose of this document is to update Digest to align with semantics, and IMO we should be abundantly clear exactly which version of semantic we are aligning to.

draft-ietf-httpbis-digest-headers.md
LPardue: I'm reflowing paragraphs all over the doc in this PR. This one has bad source formatting, so why not change it?

draft-ietf-httpbis-digest-headers.md
LPardue: This is one paragraph. Having arbitrary line breaks in a contiguous block of text is counter intuitive, why would we want that?.

Text reflow in my editor applies to paragraphs, so if you want to maintain arbitrary line breaks that makes my life way harder and I just don't see the benefit.

draft-ietf-httpbis-digest-headers.md
LPardue: Again, I reject this notion sorry. Maintaining arbitrary line breaks makes further editing of the document harder. I don't care about the patch size, these are very simple changes and in this case the reflow is a direct result of a change in the PR.

draft-ietf-httpbis-digest-headers.md
ioggstream: I generally try to isolate document lines that may be changed together - eg 'It can be used in request and response' - because I thought it simplifies editing.

If it doesn't and instead you find that makes things harder it's fine to let the editor reflow :)

ioggstream:
> I asked explicitly about the title because this document was called "Resource Digests"

IMHO `Representation Digests` is a suitable name, though it's not explicative.

Probably the same applies to many other specs referencing header names in the title. I think we can postpone this without great damages. I'd focus on the other issues instead.

LPardue:
I don't want this PR to bitrot by getting stuck behind a title bikeshed. Therefore I'm merging as is and I'll take the blame for the non-atomic change. We'll get the renaming fixed up soon anyhow.

Issue #1081, opened 2020-02-29
Cookie 5.5 empty cookie name


essen:
Step 4 says:

```
Serialize the cookie-list into a cookie-string by processing each cookie in the cookie-list in order:
1. Output the cookie’s name, the %x3D (“=”) character, and the cookie’s value.
2. If there is an unprocessed cookie in the cookie-list, output the characters %x3B and %x20 (“; “).
```

However looking at WPT tests (for example [1] / [2]) I can see that when we have `Set-Cookie: foo` (empty name with value `foo`) the expectation is to have `Cookie: foo` back, so perhaps these instructions should include a special case when the cookie name is empty.

[1] https://github.com/web-platform-tests/wpt/blob/master/cookies/http-state/resources/test-files/0004-test
[2] https://github.com/web-platform-tests/wpt/blob/master/cookies/http-state/resources/test-files/0004-expected

PR #1144, opened 2020-03-31
[RFC2665bis] Fix nameless/valueless serialization.

mikewest:
Nameless cookies serialize as `value`. Valueless cookies serialize as
`name=`. No, this isn't what we would have chosen if we were designing
cookies from scratch, but it's Chromium and Firefox's shipping behavior,
and it's reasonable to just accept.

This specification change is covered by 0021 and 0022 in the WPT
repository (see
https://wpt.fyi/results/cookies/http-state/general-tests.html?label=master&label=experimental&aligned).

Closes https://github.com/httpwg/http-extensions/issues/1081.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1608,10 +1608,12 @@ compute the cookie-string from a cookie store and a request-uri:
 4.  Serialize the cookie-list into a cookie-string by processing each cookie
     in the cookie-list in order:
 
-    1.  Output the cookie's name, the %x3D ("=") character, and the cookie's
-        value.
+    1.  If the cookies' name is not empty, output the cookie's name followed by
+        the %x3D ("=") character.
 
-    2.  If there is an unprocessed cookie in the cookie-list, output the
+    2.  If the cookies' value is not empty, output the cookie's value.
+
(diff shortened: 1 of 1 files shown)


mikewest:
WDYT, @essen / @johnwilander ?

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1608,10 +1608,12 @@ compute the cookie-string from a cookie store and a request-uri:
 4.  Serialize the cookie-list into a cookie-string by processing each cookie
     in the cookie-list in order:
 
-    1.  Output the cookie's name, the %x3D ("=") character, and the cookie's
-        value.
+    1.  If the cookies' name is not empty, output the cookie's name followed by
+        the %x3D ("=") character.
 
-    2.  If there is an unprocessed cookie in the cookie-list, output the
+    2.  If the cookies' value is not empty, output the cookie's value.
```
essen: Maybe just `2.  Output the cookie's value.` would be enough because if it's empty then nothing is output anyway.

mikewest: Thanks! I think I vaguely prefer being explicit about how an empty value would be serialized (e.g. not at all)... probably doesn't make much of a difference, but keeping it the same as the 1 above seems reasonable to me.

Issue #1119, opened 2020-03-06
rfc6265bis: definition of OWS differs from HTTP core specs


reschke:
Optimally, they'd be the same.

Otherwise, renaming might be a good idea in order to avoid confusion.

Issue #1074, opened 2020-02-24
[rfc6265bis] Cookie parser - cookie name and value with spaces


bakulf:
Similar to #1073, by spec, cookie name and value should not contain spaces, but modern browsers do allow spaces. We even have WPTs.

A similar issue is the trimming of cookie-name and cookie-value. By spec: "SetCookie: foo = bar" is invalid.

What I suggest is to write the BFN in this way (untested, unverified):

```
 cookie-pair       = cookie-name *[SP] "=" *[SP] cookie-value
 cookie-name       = 1*cookie-octet
 cookie-value      = *cookie-octet / ( DQUOTE *cookie-octet DQUOTE )
 cookie-octet      =  %20 / %x21 / %x23-2B / %x2D-3A / %x3C-5B / %x5D-7E
```

PR #1145, opened 2020-03-31
[RFC6265bis] Fix ABNF.

mikewest:
This patch incorporates two ABNF changes:

1.  Closes https://github.com/httpwg/http-extensions/issues/1074 by
    adjusting the `cookie-pair` grammar to include optional spaces.
    We test this behavior in a few distinct WPTs, including
    chromium0014, name0018, value0006.

2.  Closes https://github.com/httpwg/http-extensions/issues/1119
    by removing OWS from the spec, as user agents generally follow
    the "SHOULD" in the current text to limit themselves to a single
    SP rather than serializing with OWS. It's reasonable to lock that
    in, and it also happens to match all the `Cookie:` header tests
    we have.

Thanks to both @reschke and @bakulf for paying attention to this.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -283,7 +283,8 @@ Appendix B.1: ALPHA (letters), CR (carriage return), CRLF (CR LF), CTLs
 CHAR (any {{USASCII}} character), VCHAR (any visible {{USASCII}} character),
 and WSP (whitespace).
 
-The OWS (optional whitespace) rule is defined in Section 3.2.3 of {{RFC7230}}.
+The OWS (optional whitespace) and BWS (bad whitespace) rules are defined in
+Section 3.2.3 of {{RFC7230}}.
 
 ## Terminology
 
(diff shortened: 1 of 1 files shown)


mikewest:
@reschke and @bakulf: Would you mind taking a look at this patch?

mikewest:
Adding @johnwilander as well.

mikewest:
I'm planning on merging this tomorrow unless someone objects or has suggestions. :)

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -466,14 +455,13 @@ grammar:
 ~~~ abnf
 set-cookie-header = "Set-Cookie:" SP set-cookie-string
 set-cookie-string = cookie-pair *( ";" SP cookie-av )
-cookie-pair       = cookie-name "=" cookie-value
-cookie-name       = token
+cookie-pair       = cookie-name *[WSP] "=" *[WSP] cookie-value
+cookie-name       = 1*cookie-octet
```
bakulf, line 459: This is not enough.
What about these:

1. `Set-Cookie: a=b ;   sameSite=strict`
2. `SetCookie: a=b ; sameSite = strict`
3. `SameSite:  a=b; sameSite=lax ; path = /`

reschke: *[WSP]

is the same as

    *WSP

mikewest: Thanks to you both. I've replaced WSP with OWS (I'd mistakenly assumed RFC7230's OWS contained a wider character set than WSP), and added more flexibility around delimiters, in line with the WPT suite's tests.

WDYT?

mikewest:
> if you made sure that the new places for OWS are actually working in practice

I believe these are covered by the parsing tests in WPT (https://github.com/web-platform-tests/wpt/tree/master/cookies/http-state/resources/test-files)

reschke:
...alternative would be "BWS", if you want to allow but discourage.

mikewest:
Ah. That might be a reasonable idea. Does `BWS` have any normative text associated with it?

reschke:
> The BWS rule is used where the grammar allows optional whitespace only for historical reasons. A sender MUST NOT generate BWS in messages. A recipient MUST parse for such bad whitespace and remove it before interpreting the protocol element.

https://greenbytes.de/tech/webdav/rfc7230.html#rfc.section.3.2.3.p.4

mikewest:
Merging this. Happy to take suggestions for further cleanup as we go.

reschke (commented):
FWIW, when I approved this, I was focusing on the WS definition from the point of view of the HTTP spec.

Obviously changing the allowed set of character names is problematic.

draft-ietf-httpbis-rfc6265bis.md
reinwowa47-ship-it: Y

Issue #1059, opened 2020-02-13
Cookie 5.4 storage model confusion


essen:
https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#storage-model

Step 8 says:

```
If the cookie-attribute-list contains an attribute with an attribute-name
of “Secure”, set the cookie’s secure-only-flag to true. Otherwise,
set the cookie’s secure-only-flag to false.
```

But then Step 12 says:

```
If the cookie’s secure-only-flag is not set, [...]
```

It's impossible to reach step 12 with the `secure-only-flag` not set because it's always set in step 8.

I believe a better wording would be `If the cookie's secure-only-flag is false,`.

PR #1146, opened 2020-03-31
[RFC6265bis] Fix 'secure-only-flag' processing.

mikewest:
The 'secure-only-flag' is never "not set". It is either "true" or
"false". This patch fixes two places where we should have been checking
for falsity.

Closes https://github.com/httpwg/http-extensions/issues/1059.


mikewest:
@essen / @johnwilander WDYT?

essen:
Good. I should have sent a PR. :-)

Issue #1143, opened 2020-03-17
Digest in requests with no payload body


ioggstream:
## I expect

An example of using Digest in a request with no payload body.
There is a misconceptions that Digest does not apply to request with no payload body. 

As it's not a special case https://github.com/httpwg/http-extensions/issues/849#issuecomment-558096111 for sha-256 we can just use 

```
In [120]: sha256(b'').hexdigest()                                                                                                                                                                                                            
Out[120]: 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'
```

Other algorithms are free to define a different computation, and the examples will only be tied to sha-256

PR #1155, opened 2020-04-15
Fix: #1143. Example of an empty payload body.

ioggstream:
## This PR

- adds an example of a request with no payload body (eg. HEAD)
- mention an use-case in the Signature section.

Fixes: #1143

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -554,13 +554,17 @@ Digest: sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
 
 ## Server Returns No Representation Data
 
+Requests without a payload body can still send a Digest field
+applying the digest algorithm to an empty representation.
+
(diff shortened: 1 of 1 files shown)


ioggstream:
@martinthomson do you think this is consistent with content-coding like aes128gcm?

draft-ietf-httpbis-digest-headers.md
```diff
@@ -999,6 +1003,9 @@ the transport layer that protects HTTP fields.
 A `Digest` field using NOT RECOMMENDED digest-algorithms SHOULD NOT be used in
 signatures.
 
+Protecting with a signature the `Digest` field of a request which has no payload body
+ensures that an eventual payload has not been stripped off nor been added.
```
LPardue, line 1007: I don't quite follow what this is trying to say. To rephrase it based on my understanding, is it this?

"Using signatures to protect the `Digest` of an empty representation allows receiving endpoints to detect if an eventual payload has been stripped or added."

ioggstream: done.

PR #1160, opened 2020-04-16
[client hints] AD review comments

yoavweiss:
A set of changes following comments as part of the AD review process.


yoavweiss:
@mnot - could you take a look?

PR #1165, opened 2020-04-30
Mention mid-stream trailers in the FAQ.

ioggstream:
## This PR

mentions mid-stream trailers in the FAQ. They cannot be treated easily in this spec
as they are an hypothesis for now. Future digest-algos may consider them.

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1295,8 +1295,10 @@ Location: /authors/123
    While
    [mid-stream trailers](https://github.com/httpwg/http-core/issues/313#issuecomment-584389706)
    are interesting, since this specification is a rewrite of [RFC3230] we do not
-   think we should face that. As a first thought, future works on digest-algorithms
-   may consider mid-stream trailers though.
+   think we should face that. As a first thought, nothing in this document 
+   precludes future work that would find a use for mid-stream trailers, for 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -1290,6 +1290,14 @@ Location: /authors/123
    to send a checksum of a resource representation with no content codings
    applied.
 
+8. What about mid-stream trailers?
+
+   While
+   [mid-stream trailers](https://github.com/httpwg/http-core/issues/313#issuecomment-584389706)
+   are interesting, being this specification a rewrite of [RFC3230] we do not
```
LPardue, line 1297:
```suggestion
   are interesting, since this specification is a rewrite of [RFC3230] we do not
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1290,6 +1290,14 @@ Location: /authors/123
...
+   think we should face that. As a first thought, future works on digest-algorithms
+   may consider mid-stream trailers though.
```
LPardue, line 1299:
```suggestion
   think we should face that. As a first thought, nothing in this document 
   precludes future work that would find a use for mid-stream trailers, for 
   example an incremental digest-algorithm. A document defining such a
   digest-algorithm is best positioned to describe how it is used.
```

Issue #1096, opened 2020-03-04
Multiple PRIORITY_UPDATE frame types instead of internal bitfield 


LPardue:
This issue channels the original issue, https://github.com/kazuho/draft-kazuho-httpbis-priority/issues/120, opened by @ianswett but reflects the final state of discussion.

The PRIORITY_UPDATE frame is inspired by the HTTP/3 PRIORITY frame (RIP) and it reserves a byte to act as a bit-field that indicates the target of signal: a request or a push. The frame layout spends a byte for a single bit and is also inconsistent with the layout of other HTTP/3 frames, which makes things mildly more difficult for implementations.

Ian suggested we use different frame type codepoints to signal the target of the PRIORITY_UPDATE frame. There is already precendence for this and Kazuho noted that 

> In IETF 106, we had a similar discussion IIRC with regard to the QUIC datagram extension, and the WG seemed to favor allowing extensions to reserve a couple of adjacent frame types instead of using bits in the payload. Based on that, I am incliend to changing our design to use two frame types values.

The use of a frame for the initial priority is dependent on other issues such as the headers vs frame debate. If we needed a frame for that purpose, we probably want three frame types and the earliest code point with 3 spaces is 0xF, thus:

* Initial Priority - 0xF - one and only one sent on request stream, contains priority-value
*  Priority Update - 0x10 - reprioritize request stream, sent on control stream, contains ID and priority-value
*    Priority Update - 0x11 - reprioritize push stream, sent on control stream, contains ID and priority-value

In order to make progress independent of the discussion on initial priority frame, we could shift the PRIORITY_UPDATE frame codepoint from 0xF to 0x10, but reserve 0xF for now.

Issue #1079, opened 2020-02-27
Can server send PRIORITY_UPDATE?


LPardue:
_Migrating this issue originally reported by @dikhonov - https://github.com/kazuho/draft-kazuho-httpbis-priority/issues/121#issue-551513184_

Section 5 discusses how client can re-prioritize using the _PRIORITY_UPDATE_ frame and Section 6 discusses how the server can override priority using the `priority` header. It seems, then, that the server cannot send the _PRIORITY_UPDATE_ frame, but it is not forbidden in the draft.

My reply on the original issue was 

> Thanks for raising the issue. This is a gap in the spec and will need filling. The answer depends a little on the direction that the design takes in the WG; the salient question, I think, is: _what is the collaboration story for reprioritization?_
> 
> Once there is a answer, we can lock down some of the normative text in the TODO placeholder
> 
> ```
> TODO: add more description of how to handle things like receiving PRIORITY_UPDATE 
> on wrong stream, a PRIORITY_UPDATE with an invalid ID, etc.wro
> ```

PR #1167, opened 2020-04-30
PRIORITY_UPDATE can be used for initial priority, split H3 frames into two types

LPardue:
This attempts to address #1096 and #1079. It might help towards overcoming the headers vs frames debate.

I made the call to restrict the PRIORITY_UPDATE frame to only be sent by clients.

It replaces ASCII art with new QUIC-style frame definitions.

### DIFF
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -309,6 +309,29 @@ properties of an HTTP request it receives, the server is expected to control the
 cacheability or the applicability of the cached response, by using header fields
 that control the caching behavior (e.g., Cache-Control, Vary).
 
+# Signalling Initial Priority in Frames
+
+Endpoints might prefer to signal the initial priority of a response using a
+frame rather than a header field. This document specifies a new PRIORITY_UPDATE
+frame for HTTP/2 ({{!RFC7540}}) and HTTP/3 ({{!I-D.ietf-quic-http}}). It carries
+priority parameters and references the target of the prioritization based on a
+version-specific identifier; in HTTP/2 this is the Stream ID, in HTTP/3 this is
+either the Stream ID or Push ID.
+
+In HTTP/2 the frame is sent on stream zero and in HTTP/3 it is sent on the client
+control stream ({{!I-D.ietf-quic-http}}, Section 6.2.1). This allows the
+PRIORITY_UPDATE to be sent before the stream it references is created, and
+avoids having to extend the protocol semantics to support continued updates
+during the stream lifetime; see {{reprioritization}}.
+
+Unlike the header field, the PRIORITY_UPDATE frame is a hop-by-hop signal.
+
+Having two format to carry the initial priority signal brings the dilemma of
+choice. A sender can use either signal and sending both is not prohibited. An
+endpoint that receives both signals can process them in any way it chooses but
+MUST NOT treat this as error. Endpoints are advised that sending conflicting
+initial priority signals may lead to sub-optimal prioritization effects.
+
 # Reprioritization
 
 After a client sends a request, it may be beneficial to change the priority of
@@ -324,24 +347,14 @@ transitions to a state that prevents the client from sending additional
 frames on the stream. Therefore, a client cannot reprioritize a response by
 using the Priority header field.  Modifying this behavior would require a
 semantic change to the protocol, but this is avoided by restricting the
-stream on which a PRIORITY_UPDATE frame can be sent. In HTTP/2 the frame
-is on stream zero and in HTTP/3 it is sent on the client control stream
-({{!I-D.ietf-quic-http}}, Section 6.2.1).
-
-This document specifies a new PRIORITY_UPDATE frame type for HTTP/2
-({{!RFC7540}}) and HTTP/3 ({{!I-D.ietf-quic-http}}) which enables
-reprioritization. It carries updated priority parameters and references the
-target of the reprioritization based on a version-specific identifier; in
-HTTP/2 this is the Stream ID, in HTTP/3 this is either the Stream ID or Push ID.
-
-Unlike the header field, the reprioritization frame is a hop-by-hop signal.
+stream on which a PRIORITY_UPDATE frame can be sent.
 
 ## HTTP/2 PRIORITY_UPDATE Frame {#h2-update-frame}
 
-The HTTP/2 PRIORITY_UPDATE frame (type=0xF)  is used by clients for
-reprioritization. It carries the stream ID of the response that is being
-reprioritized, and the updated priority in ASCII text, using the same
-representation as that of the Priority header field value.
+The HTTP/2 PRIORITY_UPDATE frame (type=0xF) is used by clients to signal the
+initial priority of a response, or to reprioritize a response or push stream. It
+carries the stream ID of the response and the priority in ASCII text, using the
+same representation as that of the Priority header field value.
 
 The Stream Identifier field ({{!RFC7540}}, Section 4.1) in the PRIORITY_UPDATE
 frame header MUST be zero (0x0). Receiving a PRIORITY_UPDATE frame with a field
@@ -375,21 +388,32 @@ Stream ID:
 Priority Field Value:
 : The priority update value in ASCII text, encoded using Structured Headers.
 
-The HTTP/2 PRIORITY_UPDATE frame MUST NOT be sent prior to opening the
-stream.  If a PRIORITY_UPDATE is received prior to the stream being opened,
-it MAY be treated as a connection error of type PROTOCOL_ERROR.
+The PRIORITY_UPDATE frame MAY be sent before the stream that it references has
+been created. The Stream ID MUST be within the stream limit. If a server
+receives a PRIORITY_UPDATE for a Stream ID that is beyond the stream limits,
+this MUST be treated as a connection error of type PROTOCOL_ERROR.
+PRIORITY_UPDATE frames received before the request or response has started
+SHOULD be buffered until the stream is opened and applied immediately after the
+request message has been processed. Holding PRIORITY_UPDATES consumes extra
+state on the peer, although the size of the state is bounded by stream limits.
+There is no bound on the number of PRIORITY_UPDATEs that can be sent, so an
+endpoint SHOULD store only the most recently received frame.
 
 If a PRIORITY_UPDATE frame is received with a Prioritized Stream ID of 0x0, the
 recipient MUST respond with a connection error of type PROTOCOL_ERROR.
 
+If a client receives a PRIORITY_UPDATE frame, it MUST respond with a connection
+error of type PROTOCOL_ERROR.
+
 ## HTTP/3 PRIORITY_UPDATE Frame {#h3-update-frame}
 
-The HTTP/3 PRIORITY_UPDATE frame (type=0x10 or 0x11) is used by clients for
-reprioritization. It carries the identifer of the element that is being
-reprioritized, and the updated priority in ASCII text, using the same
-representation as that of the Priority header field value. PRIORITY_UPDATE with
-a frame type of 0x10 is used to reprioritize request streams, PRIORITY_UPDATE
-with a frame time of 0x11 is used to reprioritize push streams.
+The HTTP/3 PRIORITY_UPDATE frame (type=0xF or 0x10) is used by clients to
+signal the initial priority of a response, or to reprioritize a response or push
+stream. It carries the identifer of the element that is being prioritized, and
+the updated priority in ASCII text, using the same representation as that of the
+Priority header field value. PRIORITY_UPDATE with a frame type of 0xF is used
+for request streams, PRIORITY_UPDATE with a frame time of 0x11 is used for push
+streams.
 
 The PRIORITY_UPDATE frame MUST be sent on the client control stream
 ({{!I-D.ietf-quic-http}}, Section 6.2.1). Receiving a PRIORITY_UPDATE frame on a
@@ -398,7 +422,7 @@ error of type H3_FRAME_UNEXPECTED.
 
 ~~~ drawing
 HTTP/3 PRIORITY_UPDATE Frame {
-  Type (i) = 0x10..0x11,
+  Type (i) = 0xF..0x10,
   Length (i),
   Prioritized Element ID (i),
   Priority Field Value (..),
@@ -414,28 +438,27 @@ Prioritized Element ID:
 Priority Field Value:
 : The priority update value in ASCII text, encoded using Structured Headers.
 
-The HTTP/3 PRIORITY_UPDATE frame MUST NOT be sent with an invalid identifier,
-including before the request stream has been opened or before a promised
-request has been received.  If a server receives a PRIORITY_UPDATE specifying
-a push ID that has not been promised, it SHOULD be treated as a connection
-error of type H3_ID_ERROR.
+The PRIORITY_UPDATE frame MAY be sent before the request stream that it
+references has been created. The Stream ID MUST be within the client-initiated
+bidirectional stream limit. If a server receives a PRIORITY_UPDATE for a Stream
+ID that is beyond the stream limits, this MUST be treated as a connection error
+of type H3_ID_ERROR. PRIORITY_UPDATE frames received before the request or
+response has started SHOULD be buffered until the stream is opened and applied
+immediately after the request message has been processed. Holding
+PRIORITY_UPDATES consumes extra state on the peer, although the size of the
+state is bounded by bidirectional stream limits. There is no bound on the number
+of PRIORITY_UPDATEs that can be sent, so an endpoint SHOULD store only the most
+recently received frame.
 
-Because the HTTP/3 PRIORITY_UPDATE frame is sent on the control stream and there
-are no ordering guarantees between streams, a client that reprioritizes a
-request before receiving the response data might cause the server to receive a
-PRIORITY_UPDATE for an unknown request. If the request stream ID is within the
-client-initiated bidirectional stream limits, the PRIORITY_UPDATE frame SHOULD
-be buffered until the stream is opened and applied immediately after the request
-message has been processed. Holding PRIORITY_UPDATES consumes extra state on the
-peer, although the size of the state is bounded by bidirectional stream limits.
-There is no bound on the number of PRIORITY_UPDATES that can be sent, so an
-endpoint SHOULD store only the most recently received frame.
+If a server receives a PRIORITY_UPDATE specifying a Push ID that has not been
+promised, it SHOULD be treated as a connection error of type H3_ID_ERROR.
 
 If a server receives a PRIORITY_UPDATE for a Stream ID that is not a request
 stream, this MUST be treated as a connection error of type H3_ID_ERROR.
 
-If a server receives a PRIORITY_UPDATE for a Stream ID that is beyond the stream
-limits, this MUST be treated as a connection error of type H3_ID_ERROR.
+If a client receives a PRIORITY_UPDATE frame, it MUST respond with a connection
+error of type H3_FRAME_UNEXPECTED.
+
 
 # Merging Client- and Server-Driven Parameters {#merging}
 
@@ -637,7 +660,7 @@ Frame Type:
 : PRIORITY_UPDATE
 
 Code:
-: 0x10
+: 0xF
 
 Specification:
 : This document
@@ -649,7 +672,7 @@ Frame Type:
 : PRIORITY_UPDATE
 
 Code:
-: 0x10 and 0x11
+: 0xF and 0x10
 
 Specification:
 : This document
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -357,18 +357,15 @@ frame header MUST be zero (0x0). Receiving a PRIORITY_UPDATE frame with a field
 of any other value MUST be treated as a connection error of type PROTOCOL_ERROR.
 
 ~~~ drawing
-HTTP/2 PRIORITY_UPDATE Frame {
-  Length (24),
-  Type (8) = 0x10,
-  Flags (8),
-  Reserved (1),
-  Stream Identifier (31) = 0,
-  R (1),
-  Prioritized Stream ID (31),
-  Priority Field Value (..),
-}
+  0                   1                   2                   3
+  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+ +---------------------------------------------------------------+
+ |R|                        Stream ID (31)                       |
+ +---------------------------------------------------------------+
+ |                   Priority Field Value (*)                  ...
+ +---------------------------------------------------------------+
 ~~~
-{: #fig-h2-reprioritization-frame title="HTTP/2 PRIORITY_UPDATE Frame"}
+{: #fig-h2-reprioritization-frame title="HTTP/2 PRIORITY_UPDATE Frame Payload"}
 
 The PRIORITY_UPDATE frame payload has the following fields:
 
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -328,9 +328,9 @@ during the stream lifetime; see {{reprioritization}}.
 Unlike the Priority header field, the PRIORITY_UPDATE frame is a hop-by-hop
 signal. Clients that do not want to expose an end-to-end signal MAY omit the
 header field but send the PRIORITY_UPDATE frame at an early moment assuming that
-it will reach the server as early as the request message it applies to. An
-endpoint that receives both signals can process them in any way it chooses but
-MUST NOT treat this as error.
+it will reach the server as early as the request message it applies to. The
+signal carries by a PRIORITY_UPDATE frame overrides that carried by the header,
+even when the frame was received before the request headers.
 
 # Reprioritization
 
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -438,26 +438,31 @@ Prioritized Element ID:
 Priority Field Value:
 : The priority update value in ASCII text, encoded using Structured Headers.
 
-The PRIORITY_UPDATE frame MAY be sent before the request stream that it
-references has been created. The Stream ID MUST be within the client-initiated
-bidirectional stream limit. If a server receives a PRIORITY_UPDATE for a Stream
-ID that is beyond the stream limits, this SHOULD be treated as a connection error
-of type H3_ID_ERROR. PRIORITY_UPDATE frames received before the request or
-response has started SHOULD be buffered until the stream is opened and applied
-immediately after the request message has been processed. Holding
-PRIORITY_UPDATES consumes extra state on the peer, although the size of the
+The request-stream variant of PRIORITY_UPDATE (type=0x10) MUST reference a
+request stream. If a server receives a PRIORITY_UPDATE (type=0x10)for a Stream
+ID that is not a request stream, this MUST be treated as a connection error of
+type H3_ID_ERROR. PRIORITY_UPDATE (type=0x10) MAY be sent before the request
+stream that it references has been created. The Stream ID MUST be within the
+client-initiated bidirectional stream limit. If a server receives a
+PRIORITY_UPDATE (type=0x10) with a Stream ID that is beyond the stream limits,
+this SHOULD be treated as a connection error of type H3_ID_ERROR.
+
+Request-stream variant PRIORITY_UPDATEs (type=0x10) received before the request
+or response has started SHOULD be buffered until the stream is opened and
+applied immediately after the request message has been processed. Holding
+PRIORITY_UPDATEs consumes extra state on the peer, although the size of the
 state is bounded by bidirectional stream limits. There is no bound on the number
 of PRIORITY_UPDATEs that can be sent, so an endpoint SHOULD store only the most
 recently received frame.
 
-If a server receives a PRIORITY_UPDATE specifying a Push ID that has not been
-promised, it SHOULD be treated as a connection error of type H3_ID_ERROR.
+The push-stream variant PRIORITY_UPDATE (type=0x11) MUST reference a promised
+push stream. If a server receives a PRIORITY_UPDATE (type=0x11) with Push ID
+that is beyond the push limit or has not been promised, this MUST be treated as
+a connection error of type H3_ID_ERROR.
 
-If a server receives a PRIORITY_UPDATE for a Stream ID that is not a request
-stream, this MUST be treated as a connection error of type H3_ID_ERROR.
-
-If a client receives a PRIORITY_UPDATE frame, it MUST respond with a connection
-error of type H3_FRAME_UNEXPECTED.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-priority.md
```diff
@@ -309,6 +309,29 @@ properties of an HTTP request it receives, the server is expected to control the
...
+
+In HTTP/2 the frame is sent on stream zero and in HTTP/3 it is sent on the client
+control stream ({{!I-D.ietf-quic-http}}, Section 6.2.1). This allows the
+PRIORITY_UPDATE to be sent before the stream it references is created, and
+avoids having to extend the protocol semantics to support continued updates
+during the stream lifetime; see {{reprioritization}}.
+
+Unlike the header field, the PRIORITY_UPDATE frame is a hop-by-hop signal.
+
+Having two format to carry the initial priority signal brings the dilemma of
```
ianswett, line 329:
```suggestion
Having two mechanisms to carry the initial priority signal brings the dilemma of
```

draft-ietf-httpbis-priority.md
```diff
@@ -324,37 +347,33 @@ transitions to a state that prevents the client from sending additional
...
+The HTTP/2 PRIORITY_UPDATE frame (type=0xF) is used by clients to signal the
+initial priority of a response, or to reprioritize a response or push stream. It
+carries the stream ID of the response and the priority in ASCII text, using the
+same representation as that of the Priority header field value.
 
 The Stream Identifier field ({{!RFC7540}}, Section 4.1) in the PRIORITY_UPDATE
-frame header MUST be zero (0x0).
+frame header MUST be zero (0x0). Receiving a PRIORITY_UPDATE frame with a field
+of any other value MUST be treated as a connection error of type
+PROTOCOL_ERROR.
```
ianswett, line 362:
```suggestion
of any other value MUST be treated as a connection error of type PROTOCOL_ERROR.
```

draft-ietf-httpbis-priority.md
royfielding: +1

draft-ietf-httpbis-priority.md
```diff
@@ -324,37 +347,33 @@ transitions to a state that prevents the client from sending additional
...
-({{!I-D.ietf-quic-http}}, Section 6.2.1).
-
-This document specifies a new PRIORITY_UPDATE frame type for HTTP/2
-({{!RFC7540}}) and HTTP/3 ({{!I-D.ietf-quic-http}}) which enables
-reprioritization. It carries updated priority parameters and references the
-target of the reprioritization based on a version-specific identifier; in
-HTTP/2 this is the Stream ID, in HTTP/3 this is either the Stream ID or Push ID.
-
-Unlike the header field, the reprioritization frame is a hop-by-hop signal.
+stream on which a PRIORITY_UPDATE frame can be sent.
```
royfielding, line 350: Should be more like

```
  Therefore, a client cannot reprioritize a response by
  sending additional metadata (e.g., trailer fields) on the request stream.
  We avoid interoperability failure by restricting PRIORITY_UPDATE
  frames to the control stream, with each such frame explicitly identifying
  the stream to which the new priority applies.
```

LPardue: Thanks. I tweaked the suggestion slightly and made cbcdc60cf5a92. PTAL.

LPardue:
> Wouldn't it be easier to just define h2's stream zero as the control stream for h2 (at top of document)? 

In my head that's what I call it. However, @kazuho has been quite particular with correcting me back use stream zero terminology in the past. 

@kazuho WDYT about Roy's suggestion to define stream zero as `control` in this document?

draft-ietf-httpbis-priority.md
```diff
@@ -369,70 +388,77 @@ Stream ID:
...
 
-The HTTP/3 PRIORITY_UPDATE frame (type=0xF) carries the identifier of the
-element that is being reprioritized, and the updated priority in ASCII text,
-using the same representation as that of the Priority header field value.
+The HTTP/3 PRIORITY_UPDATE frame (type=0xF or 0x10) is used by clients to
+signal the initial priority of a response, or to reprioritize a response or push
+stream. It carries the identifer of the element that is being prioritized, and
+the updated priority in ASCII text, using the same representation as that of the
+Priority header field value. PRIORITY_UPDATE with a frame type of 0xF is used
+for request streams, PRIORITY_UPDATE with a frame time of 0x11 is used for push
```
kazuho, line 415: 0x10?

draft-ietf-httpbis-priority.md
```diff
@@ -309,6 +309,29 @@ properties of an HTTP request it receives, the server is expected to control the
...
 that control the caching behavior (e.g., Cache-Control, Vary).
 
+# Signalling Initial Priority in Frames
+
+Endpoints might prefer to signal the initial priority of a response using a
+frame rather than a header field. This document specifies a new PRIORITY_UPDATE
+frame for HTTP/2 ({{!RFC7540}}) and HTTP/3 ({{!I-D.ietf-quic-http}}). It carries
+priority parameters and references the target of the prioritization based on a
+version-specific identifier; in HTTP/2 this is the Stream ID, in HTTP/3 this is
+either the Stream ID or Push ID.
```
kazuho, line 319: How about stating that endpoints that do not want to expose an end-to-end signal MAY omit the header field, but send the PRIORITY_UPDATE frame at an early moment assuming that it would reach the server as early as the request headers? Would could go on to explain that in such case, all the responses that the server would send would have the priority in the frame being applied.

The benefit of phrasing that way is that we can get rid of the last paragraph of this section.

draft-ietf-httpbis-priority.md
```diff
@@ -320,41 +343,37 @@ browser would send a reprioritization frame with the priority field value
...
+  Flags (8),
+  Reserved (1),
+  Stream Identifier (31),
+  R (1),
+  Prioritized Stream ID (31),
+  Priority Field Value (..),
+}
 ~~~
-{: #fig-h2-reprioritization-frame title="HTTP/2 PRIORITY_UPDATE Frame Payload"}
+{: #fig-h2-reprioritization-frame title="HTTP/2 PRIORITY_UPDATE Frame"}
```
kazuho, line 376: Seeing the definition of the frame _headers_ for H2 was surprising to me, though I understand that it reflects the change to the HTTP/3 draft. Maybe I should ask if the change in HTTP/3 is a good thing.

kazuho:
@LPardue 
> WDYT about Roy's suggestion to define stream zero as control in this document?

I'd be fine with that, assuming that we'd have the definition somewhere (as @royfielding suggests).

draft-ietf-httpbis-priority.md
LPardue: As a personal opinion I prefer to see the frame in whole because it avoids having to cross reference to the frame headers, which are defined in some other document. 

But I can drop them if people want. I see that as an editorial change so perhaps we can avoid blocking this and fix up later if needs be.

kazuho: I agree that this is an editorial issue, and that we need to ask the WG to resolve (because it is about the convention covering all H2 extensions, not just this one).

MikeBishop: FWIW, ALTSVC is the only extension to have reached RFC status previously.  It [does not include](https://tools.ietf.org/html/rfc7838#section-4) the frame header; only the payload.

LPardue: You forgot ORIGIN frame https://tools.ietf.org/html/rfc8336#section-2.1 which also specifies the frame payload.

Now that H2 and H3 have different frame format definitions, I think extensions applicable to both will hit this problem. I'd like to figure what people think is the most acceptable way to present things.

LPardue: One option would be to write documents that update RFC 7838 and RFC 8336 to use the new format and define the H3 frame at the same time...

draft-ietf-httpbis-priority.md
LPardue: I've tried to incorporate this suggestion in https://github.com/httpwg/http-extensions/pull/1167/commits/f5ac92a045a3074e503c93d304a90799e387ab07 but found it trickier to write than I first thought it would be, so it ended up a little different in the end. Please take a look and add some further suggestions if I'm not hitting your mark.

LPardue:
This PR changes the HTTP/3 frame format in a breaking way, which is unfortunate because we don't have strong versioning in the extension. It's feasible that interop between endpoints that expect the frame could suffer because they parse them differently. 

One idea is to link the priorities draft version to the HTTP/3 draft version. For example, agree that priorities-01 would be used with h3-28. However, the drafts will progress at different rates and we might get stuck.

@Kazuho made a suggestion that would avoid this problem. We we pick a large frame type ID that  we change for each version of the priorities draft. The final draft that is ready for RFC can then switch back the "nice" lower values. I'll make a commit that actions this proposal in order to get review feedback.

LPardue:
https://github.com/httpwg/http-extensions/pull/1167/commits/127f1490fb0654c4676d43f7eda76d084a02466b adds "big numbers" for the H3 frame types but leaves the H2 frame as `0xF`, mainly because H2's space is only 8 bit and I don't want to pollute it. Maybe others also want an experimental value there but IMO we can progress without it.

ianswett:
FWIW, I'm ok with changing the frame type value, but in our code it's just as easy(or easier) to switch by draft version.

LPardue:
> FWIW, I'm ok with changing the frame type value, but in our code it's just as easy(or easier) to switch by draft version.

My concern is that we'll see an uncoordinated staggered rollout of new HTTP/3 endpoints, and that clients talk something different than the server expects, causing the frame parsing to fail and generating a connection error.

ianswett:
Our servers will likely support draft 25, 27 and 28 at once.

Is the concern that servers will support multiple drafts at once, and clients will typically only support a single version, but server connections won't be aware of what version they're speaking?

I suspect this is a case of different implementations having very different ways of supporting multiple versions of QUIC at once?

LPardue:
I hope implementation know what draft they are speaking! I'm more worried about needed to rev the priority draft version multiple times; if we couple to draft 28 we get stuck unless there is a draft 29. I anticipate priorities could iterate more quickly than the specification and more quickly than people might be able to deploy different versions.

Using a unique big number for frame type would even allow implementations to retrofit it to draft 27, which might be more feasible to implement and test in the short term.

ianswett:
Thanks for clarifying @LPardue, that makes sense.  I was a bit confused about how one might not know what version they're speaking.

draft-ietf-httpbis-priority.md
MikeBishop: True, I try to block that one out.

They have different frame format definitions, sure; the question is whether the payload needs to vary between H2/H3.  I think the real things you need to specify are:

- What stream it's on (part of the header in H2, intrinsic property in H3)
  - Control stream vs. stream zero language plays into this
- Flags if used in H2; since H3 doesn't have flags, this might lead to a payload difference
- Payload

The simplest path for dual-version extensions is not to use flags and only define the payload.

draft-ietf-httpbis-priority.md
```diff
@@ -309,6 +312,26 @@ properties of an HTTP request it receives, the server is expected to control the
...
+PRIORITY_UPDATE to be sent before the stream it references is created, and
+avoids having to extend the protocol semantics to support continued updates
+during the stream lifetime; see {{reprioritization}}.
+
+Unlike the Priority header field, the PRIORITY_UPDATE frame is a hop-by-hop
+signal. Clients that do not want to expose an end-to-end signal MAY omit the
+header field but send the PRIORITY_UPDATE frame at an early moment assuming that
+it will reach the server as early as the request message it applies to. An
+endpoint that receives both signals can process them in any way it chooses but
+MUST NOT treat this as error.
```
kazuho, line 333: How about changing the last sentence to something like: _The signal carried by a PRIORITY_UPDATE frame overrides that carried by the header, even when the frame was received before the request headers._

Note that this PR recommends such a behavior; it states that a server SHOULD buffer and apply the content of PRIORITY_UPDATE frame that precedes the header field. For good reasons.

Consider the case of an H3 client sending a HTTP request with a Priority header field, then changing its mind and sending a PRIORITY_UPDATE frame. As H3 does not guarantee delivery order across streams, the server might receive the frame before the header. I'd assume that we'd want servers to apply the value carried by the frame. Even when the protocol is H2, I am not sure if all the H2 stacks have the guarantee that the information across multiple streams are sent in the order they are issued, or that they are provided to the application in the order they are received.

draft-ietf-httpbis-priority.md
```diff
@@ -309,6 +312,26 @@ properties of an HTTP request it receives, the server is expected to control the
...
+
+PRIORITY_UPDATE can be sent by clients on the control stream. This allows the
+PRIORITY_UPDATE to be sent before the stream it references is created, and
+avoids having to extend the protocol semantics to support continued updates
+during the stream lifetime; see {{reprioritization}}.
+
+Unlike the Priority header field, the PRIORITY_UPDATE frame is a hop-by-hop
+signal. Clients that do not want to expose an end-to-end signal MAY omit the
+header field but send the PRIORITY_UPDATE frame at an early moment assuming that
+it will reach the server as early as the request message it applies to. The
```
martinthomson, line 331: The important "when" here is "before prioritization information is not useful to the server".  That might be later than when the server receives the request.

draft-ietf-httpbis-priority.md
```diff
@@ -369,70 +388,77 @@ Stream ID:
 Priority Field Value:
 : The priority update value in ASCII text, encoded using Structured Headers.
 
-The HTTP/2 PRIORITY_UPDATE frame MUST NOT be sent prior to opening the
-stream.  If a PRIORITY_UPDATE is received prior to the stream being opened,
-it MAY be treated as a connection error of type PROTOCOL_ERROR.
+The PRIORITY_UPDATE frame MAY be sent before the stream that it references has
+been created. The Stream ID MUST be within the stream limit. If a server
+receives a PRIORITY_UPDATE for a Stream ID that is beyond the stream limits,
+this MUST be treated as a connection error of type PROTOCOL_ERROR.
```
martinthomson, line 394: SHOULD.  I don't think that I could guarantee that this could be enforced.

```suggestion
this SHOULD be treated as a connection error of type PROTOCOL_ERROR.
```

Well, I guess that this is the h2 variant, but still.

draft-ietf-httpbis-priority.md
```diff
@@ -369,70 +388,77 @@ Stream ID:
...
-a PRIORITY_UPDATE for an unknown request. If the request stream ID is within
-bidirectional stream limits, the PRIORITY_UPDATE frame SHOULD be buffered
-until the stream is opened and applied immediately after the request message
-has been processed. Holding PRIORITY_UPDATES consumes extra state on the peer,
-although the size of the state is bounded by bidirectional stream limits. There
-is no bound on the number of PRIORITY_UPDATES that can be sent, so an
-endpoint SHOULD store only the most recently received frame.
+The PRIORITY_UPDATE frame MAY be sent before the request stream that it
+references has been created. The Stream ID MUST be within the client-initiated
+bidirectional stream limit. If a server receives a PRIORITY_UPDATE for a Stream
```
martinthomson, line 443: This needs to be qualified; if the frame contains a Push ID, then this doesn't apply.

draft-ietf-httpbis-priority.md
```diff
@@ -369,70 +388,77 @@ Stream ID:
...
+ID that is beyond the stream limits, this MUST be treated as a connection error
```
martinthomson, line 444:
```suggestion
ID that is beyond the stream limits, this SHOULD be treated as a connection error
```

draft-ietf-httpbis-priority.md
```diff
@@ -369,70 +388,77 @@ Stream ID:
...
+of type H3_ID_ERROR. PRIORITY_UPDATE frames received before the request or
+response has started SHOULD be buffered until the stream is opened and applied
+immediately after the request message has been processed. Holding
+PRIORITY_UPDATES consumes extra state on the peer, although the size of the
+state is bounded by bidirectional stream limits. There is no bound on the number
+of PRIORITY_UPDATEs that can be sent, so an endpoint SHOULD store only the most
+recently received frame.
+
+If a server receives a PRIORITY_UPDATE specifying a Push ID that has not been
+promised, it SHOULD be treated as a connection error of type H3_ID_ERROR.
```
martinthomson, line 454: This one could be a MUST.

draft-ietf-httpbis-priority.md
```diff
@@ -646,7 +672,7 @@ Frame Type:
 : PRIORITY_UPDATE
 
 Code:
-: 0xF
+: 0xF and 0x10
```
martinthomson, line 675: I would skip a number here and let people use the low bit of the type as a flag if they want.  Otherwise, this is a tiny bit more awkward.

draft-ietf-httpbis-priority.md
```diff
@@ -320,41 +343,37 @@ browser would send a reprioritization frame with the priority field value
...
- |R|                        Stream ID (31)                       |
- +---------------------------------------------------------------+
- |                   Priority Field Value (*)                  ...
- +---------------------------------------------------------------+
+HTTP/2 PRIORITY_UPDATE Frame {
+  Length (24),
+  Type (8) = 0xF,
+  Flags (8),
+  Reserved (1),
+  Stream Identifier (31),
```
martinthomson, line 370:
```suggestion
  Stream Identifier (31) = 0,
```

draft-ietf-httpbis-priority.md
```diff
@@ -309,6 +312,26 @@ properties of an HTTP request it receives, the server is expected to control the
...
+identifier; in HTTP/2 this is the Stream ID, in HTTP/3 this is either the Stream
+ID or Push ID.
+
+PRIORITY_UPDATE can be sent by clients on the control stream. This allows the
+PRIORITY_UPDATE to be sent before the stream it references is created, and
+avoids having to extend the protocol semantics to support continued updates
+during the stream lifetime; see {{reprioritization}}.
+
+Unlike the Priority header field, the PRIORITY_UPDATE frame is a hop-by-hop
+signal. Clients that do not want to expose an end-to-end signal MAY omit the
```
martinthomson, line 329: I just have to say that this "do not want" statement assumes a great deal without a whole lot of support.

draft-ietf-httpbis-priority.md
LPardue: hmm yeah, the current text is actually a bit nasty because it doesn't restrict PRIORITY_UPDATE to be within MAX_PUSH_ID either. Combining those factors means the client has a bit too much sway over the amount of state a server needs to hold. Let me fix both things.

draft-ietf-httpbis-priority.md
LPardue: paging @kazuho who suggested this text

LPardue:
We also discussed this PR during the May 2020 virtual interim and the conversation segued into a discussion about reprioritization which is related but ultimately is tracked separately in #1021.

As editor, I believe this PR adds changes that better describe the protocol with the design it had when adopted by the WG. And therefore it can be merged independent of the WG finding consensus on reprioritization. 

I've received a lot of feedback on this PR (thanks) and believe I've addressed all but one comment. I aim to resolve that one soon and then plan to merge this PR unless I hear some pushback.

draft-ietf-httpbis-priority.md
LPardue: should be fixed in https://github.com/httpwg/http-extensions/pull/1167/commits/ca71a499138f646c99fc1c57a72277f2db4fc1e0

draft-ietf-httpbis-priority.md
```diff
@@ -369,70 +383,82 @@ Stream ID:
...
-PRIORITY_UPDATE on wrong stream, a PRIORITY_UPDATE with an invalid ID, etc.
+If a client receives a PRIORITY_UPDATE frame, it MUST respond with a connection
+error of type PROTOCOL_ERROR.
 
 ## HTTP/3 PRIORITY_UPDATE Frame {#h3-update-frame}
 
-The HTTP/3 PRIORITY_UPDATE frame (type=0xF) carries the identifier of the
-element that is being reprioritized, and the updated priority in ASCII text,
-using the same representation as that of the Priority header field value.
+The HTTP/3 PRIORITY_UPDATE frame (type=0x1CCB8BBF1F0700 or 0x1CCB8BBF1F0701) is
```
guoye-zhang, line 405: Is it 0x1CCB8BBF1F0700/0x1CCB8BBF1F0701 or 0x10/0x11?

draft-ietf-httpbis-priority.md
LPardue: Taking the feedback received here and elsewhere, I'm reverted the diagram back to our traditional format.

draft-ietf-httpbis-priority.md
LPardue: In HTTP/3, the value should be the long 0x1CCB8BBF1F0700/0x1CCB8BBF1F0701, I'll fix this up.

LPardue:
This issue got tied up with the discussion about reprioritization. Now that the chairs have declared consensus on the matter, I'm giving fair warning that I'm aiming to land this change in the next day or two. I'll remind folks that this has had pretty substantial review already and I'm not looking for relitigation.

I've just pushed two commits: 1) revert the H2 frame diagram format, we can tackle this later if we care. 2) fix an inconsistency in frame type values report by @guoye-zhang.

There is an open question about a servers ability to understand a safe limit of buffering, which might be subject to implementation API. I'd like to spin that off into it's issue so we can make progress here.

LPardue:
FYI I had some feedback that the frame type values were a bit cumbersome so I've shortened them to 0xF0700 and 0xF0701. See https://github.com/httpwg/http-extensions/pull/1167/commits/b007e937c7c4572fe4bf2437ed38eb8a360696aa

draft-ietf-httpbis-priority.md
```diff
@@ -319,31 +322,39 @@ references the new JavaScript file, while the prefetch is in progress, the
...
+ID or Push ID. Unlike the Priority header field, the PRIORITY_UPDATE frame is a
+hop-by-hop signal.
+
+PRIORITY_UPDATE frames are sent by clients on the control stream, and therefore
+can be sent even after the send-side of the request stream is being closed. This
+also allows the PRIORITY_UPDATE frame to be sent as early as the stream it
+references is created. Depending on the transmission logic of the endpoints and
+on the network condition in case of HTTP/3, servers might receive a
+PRIORITY_UPDATE frame that references a request stream that is yet to be opened.
+Furthermore, clients might omit the priority request header, using
```
kazuho, line 338:
```suggestion
Furthermore, clients might omit the priority request header field, using
```

draft-ietf-httpbis-priority.md
```diff
@@ -319,31 +322,39 @@ references the new JavaScript file, while the prefetch is in progress, the
...
+PRIORITY_UPDATE frames to indicate both the initial priority and the updated
+priority.
+
+When a server receives a PRIORITY_UPDATE frame referring to a client-initiated
+request that is yet to be opened, the server buffers the priorities being
+carried by the received frame and applies them once the request is being opened.
+The signal carried by a PRIORITY_UPDATE frame overrides that carried by the
+header, even when the frame was received before the request headers.
```
kazuho, line 346:
```suggestion
header field, even when the frame was received before the request headers.
```

draft-ietf-httpbis-priority.md
```diff
@@ -369,70 +380,82 @@ Stream ID:
...
+Stream ID that is not a request stream, this MUST be treated as a connection
+error of type H3_ID_ERROR. PRIORITY_UPDATE (type=0xF0700) MAY be sent before the
+request stream that it references has been created. The Stream ID MUST be within
+the client-initiated bidirectional stream limit. If a server receives a
+PRIORITY_UPDATE (type=0xF0700) with a Stream ID that is beyond the stream
+limits, this SHOULD be treated as a connection error of type H3_ID_ERROR.
+
+Request-stream variant PRIORITY_UPDATEs (type=0xF0700) received before the
+request or response has started SHOULD be buffered until the stream is opened
+and applied immediately after the request message has been processed. Holding
```
kazuho, line 444: This is about existing text, and we can defer the decision, but I think it might be better to say "SHOULD be buffered and be applied once the stream is opened."

The problem with current text is that it does deal with servers sending responses _while_ they receive requests (i.e. early response).

draft-ietf-httpbis-priority.md
```diff
@@ -369,70 +380,82 @@ Stream ID:
...
-The HTTP/2 PRIORITY_UPDATE frame MUST NOT be sent prior to opening the
-stream.  If a PRIORITY_UPDATE is received prior to the stream being opened,
-it MAY be treated as a connection error of type PROTOCOL_ERROR.
+The PRIORITY_UPDATE frame MAY be sent before the stream that it references has
+been created. The Stream ID MUST be within the stream limit. If a server
+receives a PRIORITY_UPDATE for a Stream ID that is beyond the stream limits,
+this SHOULD be treated as a connection error of type PROTOCOL_ERROR.
+PRIORITY_UPDATE frames received before the request or response has started
+SHOULD be buffered until the stream is opened and applied immediately after the
+request message has been processed. Holding PRIORITY_UPDATES consumes extra
```
kazuho, line 389:
```suggestion
request message has been processed. Holding PRIORITY_UPDATEs consumes extra
```

kazuho (approved):
@LPardue Thank you for all the work. LGTM, just editorial suggestions.

draft-ietf-httpbis-priority.md
LPardue: Agreed that is better.

LPardue: At first I agrred but I've thought about a downside. So I've created https://github.com/httpwg/http-extensions/issues/1260 where we can discuss separately.

draft-ietf-httpbis-priority.md
```diff
@@ -369,70 +380,82 @@ Stream ID:
...
+and applied immediately after the request message has been processed. Holding
+PRIORITY_UPDATE frames consumes extra state on the peer, although the size of
+the state is bounded by bidirectional stream limits. There is no bound on the
+number of PRIORITY_UPDATE frames that can be sent, so an endpoint SHOULD store
+only the most recently received frame.
+
+The push-stream variant PRIORITY_UPDATE (type=0xF0701) MUST reference a promised
+push stream. If a server receives a PRIORITY_UPDATE (type=0xF0701) with Push ID
+that is beyond the push limit or has not been promised, this MUST be treated as
+a connection error of type H3_ID_ERROR.
```
MikeBishop, line 453:
```suggestion
push stream. If a server receives a PRIORITY_UPDATE (type=0xF0701) with a Push ID
that is greater than the maximum Push ID or which has not yet been promised, this
MUST be treated as a connection error of type H3_ID_ERROR.
```

draft-ietf-httpbis-priority.md
```diff
@@ -369,70 +380,82 @@ Stream ID:
...
+
+PRIORITY_UPDATE frames of either type are only sent by a client. If a client
```
MikeBishop, line 455:
```suggestion
PRIORITY_UPDATE frames of either type are only sent by clients. If a client
```

MikeBishop (commented):
All of these are nits except for the "within the stream limit" conversation on the HTTP/2 piece.  Feel free to split that to a separate issue if needed.

draft-ietf-httpbis-priority.md
```diff
@@ -319,31 +322,39 @@ references the new JavaScript file, while the prefetch is in progress, the
...
+references is created. Depending on the transmission logic of the endpoints and
+on the network condition in case of HTTP/3, servers might receive a
+PRIORITY_UPDATE frame that references a request stream that is yet to be opened.
+Furthermore, clients might omit the priority request header field, using
+PRIORITY_UPDATE frames to indicate both the initial priority and the updated
+priority.
+
+When a server receives a PRIORITY_UPDATE frame referring to a client-initiated
+request that is yet to be opened, the server buffers the priorities being
+carried by the received frame and applies them once the request is being opened.
```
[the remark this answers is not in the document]
LPardue: I agree that "is being opened" is ugly. However, I think substituting it for arrives in touching on https://github.com/httpwg/http-extensions/issues/1260

For consistency with other text in this PR, how about I use "processed" here and we can change it again in the resolution to #1260?

draft-ietf-httpbis-priority.md
```diff
@@ -369,70 +380,82 @@ Stream ID:
...
+SHOULD be buffered until the stream is opened and applied immediately after the
+request message has been processed. Holding PRIORITY_UPDATE frames consumes
+extra state on the peer, although the size of the state is bounded by stream
+limits. There is no bound on the number of PRIORITY_UPDATE frames that can be
+sent, so an endpoint SHOULD store only the most recently received frame.
 
-TODO: add more description of how to handle things like receiving
-PRIORITY_UPDATE on wrong stream, a PRIORITY_UPDATE with an invalid ID, etc.
+If a PRIORITY_UPDATE frame is received with a Stream ID of 0x0, the recipient
+MUST respond with a connection error of type PROTOCOL_ERROR.
```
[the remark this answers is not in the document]
LPardue: Yes. I spotted this when used the HTTP/3 format to describe the full HTTP/2 frame (not just the payload). In that version I called it Prioritized Stream ID to avoid the ambiguation. I'd lost that change by reverting. Thanks for spotting it.

draft-ietf-httpbis-priority.md
```diff
@@ -369,70 +380,82 @@ Stream ID:
 Priority Field Value:
 : The priority update value in ASCII text, encoded using Structured Headers.
 
-The HTTP/2 PRIORITY_UPDATE frame MUST NOT be sent prior to opening the
-stream.  If a PRIORITY_UPDATE is received prior to the stream being opened,
-it MAY be treated as a connection error of type PROTOCOL_ERROR.
+The PRIORITY_UPDATE frame MAY be sent before the stream that it references has
+been created. The Stream ID MUST be within the stream limit. If a server
```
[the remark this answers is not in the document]
LPardue: This is an important observation, thanks for catching it. I think this is a side effect of applying H3's rules naively rather than an intentional action. I'm open to phrasing this differently but I fear that the scope of the problem is actually a bit large. We probably need a little bit more thought about what it means to receive PRIORITY_UPDATES for H2 idle or closed streams. Especially given the amount of text RFC 7540 spends on describing this.

Perhaps the best action is to excise the entirety of this paragraph and create an issue to track this?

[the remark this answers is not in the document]
MikeBishop: I'd also be okay keeping it, provided there's an issue to figure out the right language in 7540-land.

draft-ietf-httpbis-priority.md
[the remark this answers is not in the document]
MikeBishop: WFM.

draft-ietf-httpbis-priority.md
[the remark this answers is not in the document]
MikeBishop: Split out as #1261.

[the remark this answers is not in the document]
LPardue: thanks! In that case I'm intending to keep the text

PR #1168, opened 2020-04-30
Initial priority frame

LPardue:
This is a follow up to the PRIORITY_UPDATE tweaks in https://github.com/httpwg/http-extensions/pull/1167

In this PR we acknowledge that some people would like to use a frame to signal the initial priority, so we define a new frame INITIAL_PRIORITY that is sent on the request or push stream.

I can't recall if we thought this would work for HTTP/2, so I've only defined the frame in HTTP/3. Hoping to resolve that either in this PR or in later discussion.

Something to look out for, compared to the PRIORITY_UPDATE frame, I've allowed this to be sent by either endpoint, so that it mirrors the flexibility of the header.

Since having two ways to do something means that there will inevitable be an endpoint that chooses to use both signals at the same time, I've added some text to discuss it.

### DIFF
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -309,6 +309,49 @@ properties of an HTTP request it receives, the server is expected to control the
 cacheability or the applicability of the cached response, by using header fields
 that control the caching behavior (e.g., Cache-Control, Vary).
 
+# Signalling Initial Priority in Frames
+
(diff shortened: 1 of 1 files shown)


LPardue:
I overlooked the discussion on https://lists.w3.org/Archives/Public/ietf-http-wg/2020JanMar/0171.html which implied we'd be happy enough just using a PRIORITY_UPDATE frame on the control stream.

Closing this PR.

PR #1169, opened 2020-05-05
Correct indentation in ser-dictionary

### LINKED ISSUE #1164: Dictionary serialisation algo problem
@annevk noticed that [dictionary serialisation](https://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html#ser-dictionary) is messed up.

I _think_ the fix is to indent the top-level 2-5 -- but I want to have a good look at it, compare to implementation, and I'd like others to verify, as I'm wary of making changes at this late stage.

Blame seems to be [this commit](https://github.com/httpwg/http-extensions/commit/4f3252bbc73844559bf7bc8891424de08531f7da), which I vaguely remember as being... surprising. That will teach me to slow down when something weird happens with Git...

mnot:
Fixes #1164

### DIFF
--- a/draft-ietf-httpbis-header-structure.md
+++ b/draft-ietf-httpbis-header-structure.md
@@ -991,7 +991,7 @@ _RFC Editor: Please remove this section before publication._
 
 ## Since draft-ietf-httpbis-header-structure-18
 
-_tbd_
+* Fix indentation in Dictionary serialisation (#1164).
 
 ## Since draft-ietf-httpbis-header-structure-17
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-header-structure.md
```diff
@@ -586,16 +586,16 @@ Given an ordered Dictionary as input_dictionary (each member having a member_nam
...
-      1. Append "," to output.
-      2. Append a single SP to output.
-6. Return output.
+   2. If member_value is Boolean true:
+       1. Append the result of running Serializing Parameters ({{ser-params}}) with parameters to output.
+   3. Otherwise:
+       1. Append "=" to output.
+       2. If member_value is an array, append the result of running Serializing an Inner List ({{ser-innerlist}}) with (member_value, parameters) to output.
+       3. Otherwise, append the result of running Serializing an Item ({{ser-item}}) with (member_value, parameters) to output.
+   4. If more members remain in input_dictionary:
```
annevk, line 595: This is a pre-existing issue, but remain is a bit weird as "for each" is typically not an operation that immediately consumes, or maybe it is in languages I'm not familiar with?

mnot: There's language elsewhere that says that the functions consume input.

PR #1172, opened 2020-05-18
Default urgency is 3

LPardue:
When we discussed removing the semantics of urgency levels, we also
thought it would be a good idea to change the default from 1 to 3.

We agreed to do this and landed a change that removed the semantic
text but unfortunately forgot to also change the default.

This change simple fixes the oversight.


LPardue:
See https://github.com/httpwg/http-extensions/issues/1026#issuecomment-582776695 for background

Issue #998, opened 2019-12-04
Tabs?


mnot:
Right now we use OWS, and #983 suggests moving from SP to RWS for inner list separation.

Should we allow tabs in SH? Error on them? 

If we want to allow them, 983 makes sense, as it's more consistent. However, we'll need to test for tabs...

phluid61:
What's the reason to allow tabs? Compatibility with current headers? Do actual tab characters show up in the wild that often?

mnot:
We had them in there because we were reusing OWS, not purposefully. AFAICT they don't appear in the wild much at all. So +1 for #1002.

PR #1173, opened 2020-05-21
Allow Lists and Dictionaries to be delimited by OWS

mnot:
From Magnus's feedback

### DIFF
--- a/draft-ietf-httpbis-header-structure.md
+++ b/draft-ietf-httpbis-header-structure.md
@@ -117,7 +117,7 @@ Note that as a result of this strictness, if a field is appended to by multiple 
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}} when, and only when, they appear in all capitals, as shown here.
 
-This document uses algorithms to specify parsing and serialization behaviors, and the Augmented Backus-Naur Form (ABNF) notation of {{!RFC5234}} to illustrate expected syntax in HTTP header fields. In doing so, it uses the VCHAR, SP, DIGIT, ALPHA and DQUOTE rules from {{!RFC5234}}. It also includes the tchar rule from {{!RFC7230}}.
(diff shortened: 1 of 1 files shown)


mnot:
also @clelland PTAL

kaduk:
This does not instill much of a panicked reaction in me, which is probably a good sign.
I trust it is intentional to not apply a similar treatment to inner lists (which also has a prose description I had commented on previously) on the grounds of them not resembling existing top-level header field structures?  That's certainly defensible, though I have to wonder whether someone will simply assume that it must be the same (not that we have much hope of changing such a person's behavior).

phluid61:
How does this vibe with #998? I know we spent a bit of time on it, but it's long enough ago that I can't recall detail.

phluid61:
Ah, right, I think I remember. Doing it this way means inner-list is different:

~~~
inner-list = "(" *SP [ sf-item *( 1*SP sf-item ) *SP ] ")"
             *parameter
~~~

The original motivation to remove tabs was because the above was going to be

~~~
  "(" OWS [ sf-item *( RWS sf-item ) OWS ] ")"
~~~

but there was consternation about RWS. Most of the discussion is in #983

mnot:
@phluid61 #988 was where my mind went first too, but that seems to be just about in inner lists (which makes sense). I think this is separate; intermediaries that don't know about the semantics and syntactic limitations of a given field shouldn't be meddling inside it (which inner list delimitation is).

clelland:
@reschke, I think so -- that's the only reason to allow this. Existing servers may already legitimately be using TAB when combining multiple field lines. And I suppose it makes sense, given the recent discussion about not breaking up individual list or dictionary members over multiple lines, that we don't need to replace *SP with OWS everywhere.

I'm okay with this change -- does it warrant a note in the prose somewhere that this special case only exists to accommodate field concatenation?

> Note that it says "comma an optional whitespace", not "OWS". Depending on how you read it, it might include other whitespace characters. I believe this needs to be fixed in the Semantics spec.

Let's hope we don't need to add VTAB or FF to this, too :)

mnot:
@reschke I don't know that that's necessary; we already [define whitespace](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#whitespace) pretty clearly.

In SF, I looked at changing "whitespace", but in each case the ABNF and algorithms make it clear what we're talking about, so I don't see a strong need to disambiguate.

reschke:
I'm ok with the change over here; that said, the Semantics spec could be clearer. Opened https://github.com/httpwg/http-core/issues/370.

mnot:
@bsdphk and I have been talking about this, and have come to the conclusion that this change went too far. 

Specifically, steps 1 and 5 in "Parsing Structured Fields" should just be SP, not OWS, because there isn't any risk of an intermediary inserting TAB before the first field line or after the last field line; it's only between List and Dictionary members that we need to be concerned.

Reverting that part of the change also makes writing the test cases considerably simpler.

See 6842793e.

PR #1199, opened 2020-05-26
digest-headers: remove non-ASCII characters from prose

reschke:
(inserted with https://github.com/httpwg/http-extensions/commit/4178324c374fd1e75acda09df6d7b2afb232de07)


ioggstream:
Thanks for the editorial work @reschke !

PR #1212, opened 2020-06-01
Fix __Host prefix example wording

danielgustafsson:
Add what seems to be a missing word in the __Host example text.


danielgustafsson:
Thanks!

PR #1213, opened 2020-06-01
Fix: #850. Support for parameters.

### LINKED ISSUE #850: Clarify digest-algorithm syntax with parameters (or drop parameters).

ioggstream:
## This PR

defines a way to pass parameters to be used for digest computation.

## It's done

Taking into account the fact that RFC3230 references parameter in the definition of `instance-digest` https://tools.ietf.org/html/rfc3230#section-4.2

Considering that being `digest-algorithm` just a token,  we can't add parameters to it.

```
This second example shows an hypotetical `bikeshed` digest algorithm using
two parameters: `bs` and `offset`.

~~~ example
    bikeshed=8E9q0okq; bs=8; offset=2
~~~
```

## Note

fixes #850

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -253,7 +253,7 @@ The example below shows the  `sha-256` digest-algorithm which uses base64 encodi
    sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
 ~~~
 
-This second example shows an hypotetical `bikeshed` digest algorithm using
+This second example shows a hypothetical `bikeshed` digest algorithm using
 two parameters: `bs` and `offset`.
 
 ~~~ example
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -209,6 +209,9 @@ This document uses the Augmented BNF defined in [RFC5234] and updated by
 [RFC7405] along with the "#rule" extension defined in Section 4 of
 {{SEMANTICS}}.
 
+The definition "parameter" in this document is to be interpreted as described in
+Section 4.4.1.4 of {{SEMANTICS}}.
+
 The definitions "representation", "selected representation", "representation
 data", "representation metadata", and "payload body" in this document are to be
 interpreted as described in {{SEMANTICS}}.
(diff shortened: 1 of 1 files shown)


ioggstream:
This could work for passing parameters for mi-sha256 too. WDYT?

ioggstream:
@reschke is this syntax supported in structured-fields?

draft-ietf-httpbis-digest-headers.md
```diff
@@ -250,6 +253,13 @@ The example below shows the  `sha-256` digest-algorithm which uses base64 encodi
    sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
 ~~~
 
+This second example shows an hypotetical `bikeshed` digest algorithm using
```
LPardue, line 256:
```suggestion
This second example shows a hypothetical `bikeshed` digest algorithm using
```

LPardue:
Thanks for this. I didn't notice before that Instance digest defined parameters as part of the digest-algorithm. As you point out, since digest-algorithm is defined as token, it seems wrong to say it can have a parameter. This proposal makes it clear that the parameters are part of the representation-data-digest and I support this. 

I'd be keen to get input from others on if they believe this is the intention of Digest parameters - do we know of any parameters used in the wild?

I'm less concerned about the structured fields question since we are just documenting a legacy HTTP field.

reschke:
I believe this really needs more motivation that the fact that RFC 3230 mentioned it in one sentence, and failed to define it properly.

What's the use case?

ioggstream:
@LPardue @reschke  I don't know of any parameter in the wild, but removing that will deviate from RFC 3230.

If we do that, we should explicit:

- that we will deprecate that feature (eg. because parameters can make things less secure);
- that parameters are forbidden;
- that if you want to use a parameterized algorithm you should encode their values  in the digest-value and take care of the security considerations inside the definition of your digest-algorithm;

~~~
bikeshed=bs=8|offset=2|8E9q0okq
~~~

mnot:
Discussion should probably be on the issue, not the PR...

ioggstream:
Closed in favor of #1259

PR #1223, opened 2020-06-30
"identity" is not a valid field value for "Content-Encoding"

reschke:
(it only exists in Accept-Encoding)

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -402,13 +402,13 @@ To allow sender and recipient to provide a checksum which is independent from
   {: vspace="0"}
   ID-SHA-512
   : * Description: The sha-512 digest of the representation-data of the resource when no
-    content coding is applied (eg. `Content-Encoding: identity`)
+    content coding is applied
     * Reference: [RFC6234], [RFC4648], this document.
     * Status: standard
 
   ID-SHA-256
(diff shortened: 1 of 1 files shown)


ioggstream:
Hi Julian, thanks for your review! 

Reading https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#field.content-encoding I understand that the `content-coding` allowed values are defined her https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.6.1.2.p.5

Reading the `Accept-Encoding` it seems that `identity` shouldn't be on that list, but iiuc it's there.
Ideas?

LPardue:
not that it is a canonical source of reference but the MDN article is also a bit weird wrt identity https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding

reschke:
It's in the registry because it has a special meaning in Accept-Encoding, and thus can not ever be defined as actual encoding.

And yes, the comment in the registry table could be clearer.

ioggstream:
@reschke can we "fix" the Semantics document before changing Digest?

reschke:
I don't see why one is a precondition for the other...

ioggstream:
If we are sure that there's an issue in semantics, it's ok. LGTM

Issue #1234, opened 2020-07-23
RFC 6265bis: Specify that empty name and empty value cookies should be ignored in 5.4


DCtheTall:
Currently according to the spec, any cookies set with a `Set-Cookie` header should be ignored if their name *and* value are empty. However, this check is not applied in section 5.4 of the spec, and as a result the CookieStore had a [bug](https://github.com/WICG/cookie-store/issues/149) allowed for sites to work around this check.

To me, this is a symptom of a bigger issue: 6265bis treats non-HTTP cookies as an afterthought, [e.g. not specifying how non-HTTP cookies should be accessed other than to ignore `HttpOnly` cookies](https://github.com/httpwg/http-extensions/issues/1233).

I think it's worth opening a separate issue just to track whether we add the empty-name-empty-value check to 5.4 of 6265bis, and if there are other checks on the `Set-Cookie` header in 5.3 that are implied in 5.4 but not made explicit.

DCtheTall:
CC @mikewest

mikewest:
As you've noticed, the document currently assumes that 5.4 is only going to be exercised after 5.3. I think this particular issue is distinct from the issue you raise in the other bug, as the only non-HTTP APIs that existed before the async API you're working through used cookie strings, which were parsed just as the header was.

Given that new mechanism, it does make sense to move "If both the name string and the value string are empty, ignore the set-cookie-string entirely." from 5.3 to 5.4. Skimming through, I think that's the only restriction that causes cookies to be ignored at the parsing stage.

PR #1236, opened 2020-07-27
Move empty cookie-name and empty cookie-value check from 5.3 to 5.4

DCtheTall:
Fixes [#1234](https://github.com/httpwg/http-extensions/issues/1234).

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -2159,7 +2159,8 @@ The "Cookie Attribute Registry" will be updated with the registrations below:
 ## draft-ietf-httpbis-rfc6265bis-07
 
 *  Moved instruction to ignore cookies with empty cookie-name and cookie-value
-   from section 5.3 to 5.4.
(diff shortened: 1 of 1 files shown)


DCtheTall:
> a changelog entry in a new -07 section

Done

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -2156,6 +2156,11 @@ The "Cookie Attribute Registry" will be updated with the registrations below:
    an informative reference:
    <https://github.com/httpwg/http-extensions/issues/1159>.
 
+## draft-ietf-httpbis-rfc6265bis-07
+
+*  Moved instruction to ignore cookies with empty cookie-name and cookie-value
+   from section 5.3 to 5.4.
```
mikewest, line 2162:
```suggestion
*  Moved instruction to ignore cookies with empty cookie-name and cookie-value
   from {{set-cookie}} to {{storage-model}} to ensure that they apply to cookies
   created without parsing a cookie string.
```

DCtheTall: Done.

PR #1239, opened 2020-08-04
Collapse TLS errors down to two.

### LINKED ISSUE #1227: TLS [Untrusted, Expired] Peer Certificate
Looking at what to map x509 errors (as returned by [openssl](https://www.openssl.org/docs/manmaster/man3/X509_STORE_CTX_get_error.html) into, the most obvious one is `tls_untrusted_peer_certificate`. 

However, there's also `tls_expired_peer_certificate`, which is one of these errors. It seems odd to have these two, but not any of the many other verification errors.

At first glance, I suspect it may be better to have a single `tls_peer_certificate_verification` error, and let the `detail` convey the reason - whether it's expiration or something else.

mnot:
Fixes #1227.

### DIFF
--- a/draft-ietf-httpbis-proxy-status.md
+++ b/draft-ietf-httpbis-proxy-status.md
@@ -352,14 +352,19 @@ This section lists the Proxy Error Types defined by this document. See {{registe
 * Extra Parameters: None.
 * Recommended HTTP status code: 504
 
-### TLS Error
+### TLS Protocol Error
 
-* Name: tls_error
-* Description: The intermediary encountered a TLS error when communicating with the next hop.
+* Name: tls_protocol_error
+* Description: The intermediary encountered a TLS error when communicating with the next hop, either during handshake or afterwards.
 * Extra Parameters: None.
 * Recommended HTTP status code: 502
 
-Details of the error should be conveyed in the details parameter.
+### TLS Certificate Error
+
+* Name: tls_certificate_error
+* Description: The intermediary encountered an error when verifying the certificate presented by the next hop.
+* Extra Parameters: None.
+* Recommended HTTP status code: 502
 
 ### TLS Error Received
 
--- a/draft-ietf-httpbis-proxy-status.md
+++ b/draft-ietf-httpbis-proxy-status.md
@@ -366,9 +366,9 @@ This section lists the Proxy Error Types defined by this document. See {{registe
 * Extra Parameters: None.
 * Recommended HTTP status code: 502
 
-### TLS Error Received
+### TLS Alert Received
 
-* Name: tls_error_received
+* Name: tls_alert_received
 * Description: The intermediary received a TLS alert from the next hop.
 * Extra Parameters:
   - alert_message: a sf-token containing the applicable description string from the TLS Alerts registry.
--- a/draft-ietf-httpbis-proxy-status.md
+++ b/draft-ietf-httpbis-proxy-status.md
@@ -361,7 +361,6 @@ This section lists the Proxy Error Types defined by this document. See {{registe
 
 Details of the error should be conveyed in the details parameter.
 
-
 ### TLS Error Received
 
 * Name: tls_error_received
--- a/draft-ietf-httpbis-proxy-status.md
+++ b/draft-ietf-httpbis-proxy-status.md
@@ -352,56 +352,20 @@ This section lists the Proxy Error Types defined by this document. See {{registe
 * Extra Parameters: None.
 * Recommended HTTP status code: 504
 
-### TLS Handshake Error
-
-* Name: tls_handshake_error
-* Description: The intermediary encountered an error during TLS handshake with the next hop.
-* Extra Parameters:
-  - alert_message: a sf-token containing the applicable description string from the TLS Alerts registry.
-* Recommended HTTP status code: 502
-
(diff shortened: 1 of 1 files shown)


mnot:
PTAL @PiotrSikora @sleevi @davidben

draft-ietf-httpbis-proxy-status.md
```diff
@@ -352,56 +352,19 @@ This section lists the Proxy Error Types defined by this document. See {{registe
...
-* Name: tls_unexpected_peer_certificate
-* Description: The intermediary received an unexpected peer certificate (e.g., SPKI doesn't match) during the TLS handshake with the next hop.
-* Extra Parameters:
-  - identity: a sf-string containing a comma-separated list of Subject Alternative Names from the certificate received from the next hop.
-  - sha256: a sf-string containing the hex-encoded SHA-256 of the certificate received from the next hop.
-  - spki: a sf-string containing the base64-encoded SHA-256 of the Subject Public Key Info (SPKI) from the certificate received from the next hop.
-* Recommended HTTP status code: 502
-
-### TLS Missing Proxy Certificate
+Details of the error should be conveyed in the details parameter.
```
sleevi, line 362: So in looking in context, this error stands out as being the only one that explicitly mentions `details`, which might cause some confusion. For example, `http_protocol_error` doesn't mention `details`, yet you use so as part of the illustrative example.

For what it's worth, you *could* split this into two: `tls_protocol_error` and `tls_certificate_error`. The former indicates something went wrong with the TLS protocol (which could include both handshake and ongoing use), while the latter indicates an error while verifying the certificate (for which `details` is useful). Similar to `http_protocol_error`, it serves as both the fallback (e.g. if the proxy doesn't expose certificate errors distinct from TLS errors and treats them just as handshake errors) and as a distinct state (errors sending that may not result in a peer alert - e.g. if it receives a malformed message *from* the peer)

mnot: Thanks, that's helpful. PTAL.

sleevi:
Looks reasonable to me! 👍

mnot:
@PiotrSikora PTAL; I'll merge tomorrow if I don't hear anything.

mnot:
@PiotrSikora how do you distinguish between a proxy certificate and the certificate used for the next hop?

draft-ietf-httpbis-proxy-status.md
```diff
@@ -352,56 +352,24 @@ This section lists the Proxy Error Types defined by this document. See {{registe
...
-* Extra Parameters:
-  - alert_message: a sf-token containing the applicable description string from the TLS Alerts registry.
-* Recommended HTTP status code: 502
-
-### TLS Untrusted Peer Certificate
-
-* Name: tls_untrusted_peer_certificate
-* Description: The intermediary received an untrusted peer certificate during TLS handshake with the next hop.
+* Name: tls_protocol_error
+* Description: The intermediary encountered a TLS error when communicating with the next hop, either during handshake or afterwards.
```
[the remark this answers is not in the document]
mnot: Possibly, depending on how the implementation works. Note that we have an open issue regarding guidance for generating appropriate error codes.

PiotrSikora:
> @PiotrSikora how do you distinguish between a proxy certificate and the certificate used for the next hop?

By "proxy certificate" I mean the certificate sent by the CDN (when acting as a client) to the origin server for the purpose of mutual authentication.

In contrast, `tls_certificate_error` would apply (at least in my mind) only to the certificate presented by the origin server.

mnot:
So, I think the argument is that `missing_proxy_certificate` is either a `tls_error_received` (i.e., the server you're connecting to gave you an error to the effect that you need to present a client certificate), or a `proxy_configuration_error` (i.e., you realised you need to present a certificate before connecting, but don't have one) depending on exactly how the error occurred. Is that sufficient?

PiotrSikora:
As someone who spent countless hours debugging connection issues to various origin servers, my main concern is that `tls_certificate_error` alone doesn't provide enough information as to why the connection failed, it only narrows it down to a TLS certificate issue, but for anything definitive that you would want to share with the customer, you'd still need to go and dig further into logs, or try to replicate the issue. Yes, I know that you could put more information in the "details" field, but I'm not a big fan of implementation-specific free-form text.

Having said that, I'm not going to oppose if you prefer to have only those 3 generic fields.

draft-ietf-httpbis-proxy-status.md
```diff
@@ -352,56 +352,24 @@ This section lists the Proxy Error Types defined by this document. See {{registe
...
-* Description: The next hop rejected the client certificate provided by the intermediary during TLS handshake.
-* Extra Parameters: None.
-* Recommended HTTP status code: 500
-
-### TLS Error
+### TLS Error Received
 
-* Name: tls_error
-* Description: The intermediary encountered a TLS error when communicating with the next hop.
+* Name: tls_error_received
```
PiotrSikora, line 371: Nit: `tls_alert_received` would be technically more correct.

mnot: Done.

mnot:
Thanks. As Ryan points out, the problem is that implementations have different terminology and granularity of error reporting, and it's not the place of this specification to impose that on them.

Issue #1238, opened 2020-07-31
Clearing Client Hint preferences with browser history seems odd


amtunlimited:
> User agents MUST clear persisted opt-in preferences when any one of site data, browsing history, browsing cache, cookies, or similar, are cleared.

Browsing history seems to stand out to me here. There are a couple of pedantic questions (does this mean full history, or back beyond a certain amount of time?) and doesn't seem to fit with precedent. In general, I don't expect much other information to be cleared when removing history information, but that might very well just be me.

davidben:
I agree the text seems a bit overprescriptive here. I could imagine UAs that want to clear it with browsing history and UAs that don't, depending on what they're trying to accomplish. It's a question of which "please forget this" use cases the browser is aiming to meet with each option.

Perhaps the UA has an option whose aim is "please make this site forget about me". Then the UA should clear client hint preferences, site data, and cache, but may or may not decide history is separate. (Assuming the history is not web-visible!) Or perhaps the aim is "I don't want these links to appear in address bar completions". Then perhaps the UA will clear just history but not any site data because the user didn't ask to be logged out. Or perhaps the UA wants to combine these two into one option. Or perhaps the option is for developers, to clear particular pieces of cache to diagnose some issue. Then maybe the client hint cache should be cleared or maybe the option is just about the HTTP cache.

Ideally the text would capture enough that UAs can make the correct decision for each of their features. (And, since it has privacy implications when UAs forget, probably explicitly call out clearing this in "please make this site forget about me" cases like clearing cookies and site data.)

reschke:
FWIW, the spec has been approved early July and is in the RFC Editor queue (see https://datatracker.ietf.org/doc/draft-ietf-httpbis-client-hints/). Changes at this stage can only be done with a lot of care.

amtunlimited:
@reschke I was unaware, thank you.

@davidben I guess the question is is the issue worth poking? I don't think it'd consider it a blocker for publishing, but I'm not super experienced in the standards processes, where I suppose things are less "tweakable" after-the-fact

davidben:
Ah yeah, the timing is awkward. At the same time, I don't think it does anyone any good if the document does not match what people intend to implement. Since browsing history is, well, a browser notion, I suppose the main question is what browser implementors would do.

[...]

PR #1243, opened 2020-08-12
Remove requirement to clear Client Hint preferences on browser history clear

amtunlimited:
Fix: #1238 

@yoavweiss 
@igrigorik 
@davidben


amtunlimited:
Adjusted scope of change because the Client Hints proposal is further along in the process than I realized

yoavweiss:
This LGTM based on the discussion in #1238

@mnot - thoughts on landing this at this point? Note that https://github.com/httpwg/http-extensions/pull/1282 means we'd probably need to republish a draft anyway.

mnot:
I think you can do both of these changes in AUTH48, but you should send an e-mail to the list giving folks a heads-up first. Note that because the draft is now at the RFC Editor, publishing a new revision probably isn't going to help.

reschke:
Am 06.10.2020 um 05:15 schrieb Mark Nottingham:
> I think you can do both of these changes in AUTH48, but you should send
> an e-mail to the list giving folks a heads-up first. Note that because
> the draft is now at the RFC Editor, publishing a new revision probably
> isn't going to help.
> ...

FWIW, and IMHO, publishing a new version is more transparent, and *will*
work if the RPC hasn't started editing yet, and if the AD is ok with it.

Best regards, Julian

amtunlimited:
I'm going to go ahead and close this.

Issue #970, opened 2019-10-31
Computing Digest in requests


ioggstream:
## Uniform digest request semantic

While discussing #923 @reschke made the following consideration

> I just don't think that it would be a good idea to vary the semantics based on the request method. 

In this issue we will gather some examples and considerations so to investigate whether  
digest in requests should be computed on the request body.

LPardue:
Does this need fixing, or can we close with no action?

ioggstream:
I'll add a FAQ and close.

ioggstream:
IMHO: we should find a better name for this issue.

- the digest semantic is always the same: digest is computed on the selected resource representation data;
- digest computation depends on method and headers, because they alter the meaning of the payload body (eg. with partial representations).

ioggstream:
Me and Julian @reschke exchanged some email but the point is still open.

Julian suggests that in requests, Digest should always be computed on the payload-data, independently of the request semantic (eg. even if the request contains a partial representation). This would make it easy its implementation.

I understand this position, but I think that it is not aligned with RFC3230 which introduces Digest because Content-MD5 definition was inconsistent respect to partial representations. Note that this I-D defines `representation-data-digest` which is potentially unrelated to the `payload-data`.

If HTTP forbids partial representations in requests though, we will get @reschke behavior without the need of changing the specification; otherwise we really need @mnot / chair support.

ioggstream:
@LPardue about a new method accepting Content-Range in requests  https://github.com/httpwg/http-core/pull/620/files

LPardue:
Discussion at the last interim seemed to indicate that a "Content-Digest" would go some way to unblocking this issue, a PR will come later this week.

ioggstream:
fixed via #1543

PR #1248, opened 2020-08-17
Fix: #970. Digest semantic depends on method.

ioggstream:
## This PR

Clarifies why digest semantic on requests depends on the underlying method semantic and can't be generalized on all methods.

The clarification is added in the FAQ.

Fixes: #970

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1352,6 +1352,17 @@ Location: /authors/123
    example an incremental digest-algorithm. A document defining such a
    digest-algorithm is best positioned to describe how it is used.
 
+9. Why don't we say that in requests, Digest is always computed on the payload body?
+
+   This [discussion is not trivial](https://github.com/httpwg/http-extensions/pull/923)
+   because while in both POST and PATCH the representation-digest is computed on the payload body,
+   this is because of the semantics of those methods. For example, an hypothetical method where
+   the request body contains a partial representation would behave differently, and Digest should
+   be computed on the complete representation data. An example of such an hypothetical use case
+   would be a resumable upload mechanism implemented with PUT + Content-Range like
+   [this one](https://cloud.google.com/storage/docs/performing-resumable-uploads#resume-upload).
+
+
 # Acknowledgements
 {:numbered="false"}
 The vast majority of this document is inherited from [RFC3230], so thanks
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1355,13 +1355,16 @@ Location: /authors/123
 
 9. Why don't we say that in requests, Digest is always computed on the payload body?
 
-   This [discussion is not trivial](https://github.com/httpwg/http-extensions/pull/923)
-   because while in both POST and PATCH the representation-digest is computed on the payload body,
-   this is because of the semantics of those methods. For example, an hypothetical method where
-   the request body contains a partial representation would behave differently, and Digest should
-   be computed on the complete representation data. An example of such an hypothetical use case
-   would be a resumable upload mechanism implemented with PUT + Content-Range like
-   [this one](https://cloud.google.com/storage/docs/performing-resumable-uploads#resume-upload).
+   Digest is computed on the selected representation-data (see {{representation-digest}}
+   that depends on the semantics;
+   the method is the primary source of request semantics (see Section 8 of {{SEMANTICS}})
+   that might be further specialized by header fields.
+   If the request body contains a partial representation, computing the representation-digest
+   on the payload body mimics Content-MD5 behavior and introduces a
+   discrepancy between request and responses.
+   In POST and PATCH the representation-digest is computed on the payload body
+   because of the semantics of those methods.
+   [Discussion is still open](https://github.com/httpwg/http-extensions/pull/923)
 
 
 # Acknowledgements


draft-ietf-httpbis-digest-headers.md
```diff
@@ -1352,6 +1352,17 @@ Location: /authors/123
...
 
+9. Why don't we say that in requests, Digest is always computed on the payload body?
+
+   This [discussion is not trivial](https://github.com/httpwg/http-extensions/pull/923)
+   because while in both POST and PATCH the representation-digest is computed on the payload body,
+   this is because of the semantics of those methods. For example, an hypothetical method where
+   the request body contains a partial representation would behave differently, and Digest should
+   be computed on the complete representation data. An example of such an hypothetical use case
+   would be a resumable upload mechanism implemented with PUT + Content-Range like
+   [this one](https://cloud.google.com/storage/docs/performing-resumable-uploads#resume-upload).
```
LPardue, line 1363:
```suggestion
   This [discussion is not trivial](https://github.com/httpwg/http-extensions/pull/923).
   When using POST and PATCH the representation-digest is computed on the payload body but
   that this is because of the semantics of those methods. However, other method semantics can vary.
   For example, a hypothetical method where the request body contains a partial representation
   would behave differently, and it's Digest would be computed on the complete representation data.
   An example of such an hypothetical use case would be a resumable upload mechanism 
   implemented with PUT + Content-Range, such as
   [this one](https://cloud.google.com/storage/docs/performing-resumable-uploads#resume-upload).
```

just some editorializing

reschke: No. 

The Digest shoult not depend on the method. And giving an example of something that the HTTP spec explicitly forbids (PUT + Range) doesn't make it better.

reschke:
> Clarifies why digest semantic on requests depends on the underlying method semantic and can't be generalized on all methods.

To me, that is the opposite of what needs to be done. The semantics of "Digest" in a request should always be the same.

If would be good to understand *why* you want to make Digest depend on the method.

draft-ietf-httpbis-digest-headers.md
ioggstream: > No.

Can you provide some normative references in SEMANTICS which forbids to create a method or a mechanism conveying a partial representation? It would be really useful :) 

> [..] HTTP spec explicitly forbids (PUT + Range)

let it be an `UPDATE` method which conveys a partial representation instead of a patch-document. 
Probably the proliferation of different ways of updating resources via partial representations and the (mis)use of Content-MD5 by goog, amz & co requires some attention from us.

> Digest should not depend on the method

Digest does not depend directly on the method, but on the SEMANTICS (which is impacted by the method).
imho stating that Digest just depends on the payload body is not in line with RFC3230 spirit.

ioggstream:
> No.
> Digest should not depend on the method
> The semantics of "Digest" in a request should always be the same.

Can you provide some normative references in SEMANTICS which forbids to create a method or a mechanism conveying a partial representation? It would be really useful :) 

> [..] HTTP spec explicitly forbids (PUT + Range)

let it be an `UPDATE` method which conveys a partial representation instead of a patch-document. 
Probably the proliferation of different ways of updating resources via partial representations and the (mis)use of Content-MD5 by goog, amz & co requires some attention from us.

> why you want to make Digest depend on the method

Digest does not depend directly on the method, but on the SEMANTICS (which may depend on the method).
imho stating that Digest in request just depends on the payload body is not in line with RFC3230 spirit.

reschke:
> Can you provide some normative references in SEMANTICS which forbids to create a method or a mechanism conveying a partial representation in a request? It would be really useful :)

Of course a method can carry a partial representation. What I'm saying is that even when it does, a Digest request header field would *still* reflect the contents of the payload, in this case the partial payload.

> Digest does not depend directly on the method, but on the SEMANTICS (which may depend on the method).
imho stating that Digest in request just depends on the payload body is not in line with RFC3230 spirit.

But a recipient has no idea about the "semantics" of new methods, thus the behavior would essentially be undefined. How is that useful?

Can wou please provide a *concrete* example where you would want the Digest request header field *not* to refer to the payload, whatever it is?

(FWIW, the "spirit" of RFC 3230 is really irrelevant here unless you can demonstrate that it's actually useful and deployable; revising specs includes cleaning them up)

LPardue:
There seems to be some similarity to the discussion on CONNECT-UDP at https://github.com/httpwg/http-core/issues/436. What I mean by this is that special-casing things based on methods can be difficult, and that examples are based on legacy behaviours, which we'd rather not encourage any further.

If I understand Julien's point (and I might not!), then a spec that has text which says "you can do anything" doesn't help the implementers of Digest today. If there is no actionable advice, then the text has dubious value. Therefore it is better to leave this unwritten and allow someone with an actual case to come along later and describe what is required.

reschke:
Trying to clarify.

The spec needs to state what the Digest header field means on *any* request and response. We can't leave that open.

(And, FWIW, I haven't seen *any* argument in favor of special-casing methods - if there is one, we should discuss it!)

LPardue:
> The spec needs to state what the Digest header field means on any request and response. We can't leave that open.

I agree. It is up to some future proposal to make the argument for special case behaviour.

ioggstream:
I think the issue name is misleading, as:

- digest semantic is always the same: digest is computed on the selected resource representation data;
- digest computation depends on method;

Stating that on requests, digest should always be computed on the payload body effectively *alters* digest semantic depending on request and response.

reschke:
> Stating that on requests, digest should always be computed on the payload body effectively alters digest semantic depending on request and response.

So what do you think the digest *shoud* refer to in request?

mnot:
Can this PR (and branch) be deleted?

ioggstream:
Closed in favor of #1543

Issue #968, opened 2019-10-31
A client requests an unsupported Digest, the server MAY reply with a 400


LPardue:
I think this example demonstrates the under specification of the Want-Digest header. Does it work as 

1. A discovery mechanism that allows the client to probe the server for the set of digests it supports (client and server specifically)
2. A unilateral advertisement of the algorithms supported by the sending endpoint

But to try and understand if it is 1) or 2) I see nothing in RFC 3230 that prescribes behaviour like this but I may well have missed it. Either way I don't think we should be putting normative text (MAY reply) in the section header. 

So lets decouple things.

I'm assuming in this example case that we operate under the model that server is allowed to respond with a 4xx under any circumstance. But specifically in this case the server is unwilling to accept requests with `sha` algorithm.

Then next it advertises the list of algorithms it can generate? In which case, a header names Want-Digest is not very inutuitive. It seems to act asymmetrically, while we are expending a lot of effort to understand and describe how Digest can be used in Request and Response.

ioggstream:
> I don't think we should be putting normative text (MAY reply) in the section header.

agree, will fix it after IETF106.

> I see nothing in RFC 3230 that prescribes behaviour like this

it's not a prescribed behavior, as rfc3230 is not very rigid.
imho it's a possibility which we have in rfc3230 that we may use to implement some sort of negotiation. 

the alternative would be that the server responds with a sha-256 digest 
that the client won't be able to process, as cited in rfc3230#4.3.2

> A sender MAY send an instance-digest using a digest-algorithm without
   knowing whether the recipient supports the digest-algorithm, or even
   knowing that the recipient will ignore it

LPardue:
> imho it's a possibility which we have in rfc3230 that we may use to implement some sort of negotiation.

Our job with this draft is to update the status quo. So we need to be really careful not create new modes or capabilities, especially if they end up being somehow incompatible with how people actually use Want-Digest today. So my problem is that this example is quite suggestive of a protocol design aspect but has no useful text. And trying to write up negotiation is likely to hit barriers or just be ignored in extant deployments.  

I'd be happier with a separate example that shows a 4xx response that includes the digest of the error body (i.e. the returned error page). I'd be happy with a stand alone example that shows a request with Want-Digest and a response that includes no Digest and no Want-Digest  because of the algorithm mismatch. 

Given the timeframes lets punt this to IETF 106.

LPardue:
This also relates to 

> 6. What changes in `Want-Digest`?

[...]

PR #1249, opened 2020-08-18
Fix: #968. Clarifications on Want-Digest.

ioggstream:
## This PR

clarifies that no normative parts were added to Want-Digest, only examples.

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1133,6 +1133,19 @@ registry:
   SHOULD NOT be used.
 * Status: obsoleted
 
+## Obsolete "contentMD5" token in Digest Algorithm {#iana-contentMD5}
+
+This memo adds the "contentMD" token in the [HTTP Digest Algorithm
+Values](https://www.iana.org/assignments/http-dig-alg/http-dig-alg.xhtml)
+registry:
+
+* Digest Algorithm: contentMD5
+* Description: Section 5 of [RFC3230] defined the "contentMD5" token to be used only in Want-Digest.
+  This token is obsoleted and MUST NOT be used.
+* Reference: {{iana-contentMD5}} of this document, Section 5 of [RFC3230].
+* Status: obsoleted
+
+
 ## The "ID-SHA-256" Digest Algorithm {#iana-ID-SHA-256}
 
 This memo registers the "ID-SHA-256" digest algorithm in the [HTTP Digest
@@ -1443,3 +1456,4 @@ _RFC Editor: Please remove this section before publication._
 * Identity is not a valid field value for content-encoding #1223
 * Mention trailers #1157
 * Reference httpbis-semantics #1156
+* Add contentMD5 as an obsoleted digest-algorithm #1249
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -459,7 +459,7 @@ representation of the patched resource.
 `Digest` usage with PATCH is thus very similar to the POST one, but with the
 resource's own semantic partly implied by the method and by the patch document.
 
-# Deprecate Negotiation of Content-MD5
+# Deprecate Negotiation of Content-MD5 {#deprecate-contentMD5}
 
 This RFC deprecates the negotiation of Content-MD5 as it has been obsoleted by
 [RFC7231].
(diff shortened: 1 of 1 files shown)


LPardue:
Hmm I think the issue might have conflated two things. 1) We are deprecated use of the MD5 algorithm in digests (including want-digest with `md5` 2) We are deprecating the use of content-MD5 as described in https://tools.ietf.org/html/rfc3230#section-5. I think we should expand the `deprecate-contentMD5` section a little to explicitly state that want-digest with `contentMD5` means. It's easy to confuse these things with the current text.

ioggstream:
> I think we should expand the `deprecate-contentMD5` section a little
> to explicitly state what want-digest with `contentMD5` means in this draft i.e. is it ignored?

To me the implementor is free to ignore or error on  contentMD5: OTOH it's an unregistered algorithm.
We could just reserve that word to avoid that to be registered, but I don't like this kind of entries lingering around (eg. see https://github.com/httpwg/http-core/issues/388 )

re-reading the draft, I found the message quite clear: forget md5 and all that relates to it.

reschke:
> We could just reserve that word to avoid that to be registered, but I don't like this kind of entries lingering around 

Having the value in the registry makes a lot of sense, because that's what people are supposed to look up. It should just make clear that it's deprecated.

>  (eg. see httpwg/http-core#388 )

That's IMHO a bit misleading as "identity" is not a real content-coding.

LPardue:
My interpretation is different than Roberto's; RFC3230 registers the value `contentMD5` for use in `want-digest`, and only `want-digest`.

```
   First, we add to the set of digest-algorithm values (in section
   4.1.1) the token "contentMD5", with the provision that this digest-
   algorithm MUST NOT be used in a Digest header field.
```

then

```
  The presence of the "contentMD5" digest-algorithm with a non-zero
   qvalue in a Want-Digest header field indicates that the sender wishes
   to receive a Content-MD5 header on messages associated with the
   Request-URI.
```

so a client sending `want-digest: contentMD5;q=1` is asking for the server to provide a `Content-MD5` in the response, independent of the `Digest`. That `contentMD5` is missing from any registry seems like an oversight.

In the meantime, RFC 7231 has deprecated Content-MD5 and says:

```
   The Content-MD5 header field has been removed because it was
   inconsistently implemented with respect to partial responses.
```

So the right thing to do, I think, is to register this as an obsolete digest-algorithm. Whether that needs to be a `SHOULD NOT be used in Want-Digest and MUST NOT be used in Digest`, or a MUST NOT in both (probably unenforceable) I don't know.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -459,7 +459,7 @@ representation of the patched resource.
 `Digest` usage with PATCH is thus very similar to the POST one, but with the
 resource's own semantic partly implied by the method and by the patch document.
 
-# Deprecate Negotiation of Content-MD5
+# Deprecate Negotiation of Content-MD5 {#deprecate-contentMD5}
 
 This RFC deprecates the negotiation of Content-MD5 as it has been obsoleted by
 [RFC7231].
```
ioggstream, line 465:
```suggestion
[RFC7231].
The `contentMD5` token defined in Section 5 of [RFC3230] MUST NOT be used as a digest-algorithm.
```

ioggstream:
1- I'll register the algorithm in the table and mark it as Obsolete
2- PTAL https://github.com/httpwg/http-extensions/pull/1249/files#r472865789

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1332,7 +1346,11 @@ Location: /authors/123
 
 6. What changes in `Want-Digest`?
 
-   We allow to use the `Want-Digest` in responses to advertise the supported
+   The contentMD5 token defined in Section 5 of [RFC3230] is deprecated by {{deprecate-contentMD5}}.
+
+   To clarify that `Digest` and `Want-`Digest can be used in both requests and responses
```
ioggstream, line 1351:
```suggestion
   To clarify that `Digest` and `Want-Digest` can be used in both requests and responses
```

ioggstream:
> So the right thing to do, I think, is to register this as an obsolete digest-algorithm. Whether that needs to be a SHOULD NOT be used in Want-Digest and MUST NOT be used in Digest, or a MUST NOT in both (probably unenforceable) I don't know.

Done :)

LPardue (approved):

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1133,6 +1134,19 @@ registry:
   SHOULD NOT be used.
 * Status: obsoleted
 
+## Obsolete "contentMD5" token in Digest Algorithm {#iana-contentMD5}
+
+This memo adds the "contentMD" token in the [HTTP Digest Algorithm
```
ioggstream, line 1139:
```suggestion
This memo adds the "contentMD5" token in the [HTTP Digest Algorithm
```

PR #1253, opened 2020-08-26
Fix: #1211. Prefer lowercase digest-algorithm names.

### LINKED ISSUE #1211: digest-algorithms SHOULD be lowercase (wrt structured headers)
## I suggest

that digest-algorithms `SHOULD` be lowercase

## Note

digest-algorithms are currently case insensitive.
We won't break specs suggesting to express them in lowercase.
This could make it easy a future transition to S-H

ioggstream:
## This PR

- use lowercase for digest-algoritms in the doc
- case is unchanged when referring to the actual algorithms (eg. SHA-1)
- prefer lowercase for digest-algorithms
- update IANA table to use lowercase digest-algorithms

## Note

digest-algorithms remain case insensitive

Fixes #1211 Discussion here https://lists.w3.org/Archives/Public/ietf-http-wg/2020JulSep/0119.html

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -335,7 +335,8 @@ some digest-algorithms, one or more parameters can be supplied.
 ~~~
 
 The BNF for "parameter" is defined in Section 5.4.1.4 of
-{{SEMANTICS}}. All digest-algorithm values are case-insensitive but SHOULD be used in lowercase.
+{{SEMANTICS}}. All digest-algorithm values are case-insensitive
+but the lower case is preferred.
 
 The Internet Assigned Numbers Authority (IANA) acts as a registry for
 digest-algorithm values.
@@ -1382,7 +1383,7 @@ Location: /authors/123
    to send a checksum of a resource representation with no content codings
    applied.
    To simplify a future transition to Structured Fields {{?I-D.ietf-httpbis-header-structure}}
-   we introduced that digest-algorithms, though case-insensitive, SHOULD be expressed in lowercase.
+   we suggest to use lowercase for digest-algorithms.
 
 8. What about mid-stream trailers?
 
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -341,7 +341,6 @@ The Internet Assigned Numbers Authority (IANA) acts as a registry for
 digest-algorithm values.
 The registry contains the tokens listed below.
 
-<<<<<<< HEAD
 Some digest-algorithms, although registered, rely on vulnerable algorithms:
 the "md5" digest-algorithm MUST NOT be used due to collision attacks [CMU-836068]
 and the "sha" digest-algorithm MUST NOT be used due
@@ -365,7 +364,7 @@ to collision attacks [IACR-2020-014].
   : * Description: The MD5 algorithm, as specified in [RFC1321].
       The output of this algorithm is encoded using the
       base64 encoding  [RFC4648].
-      The "md5" algorithm MUST NOT be used as it's now vulnerable
+      This digest-algorithm MUST NOT be used as it's now vulnerable
       to collision attacks [CMU-836068].
     * Reference: [RFC1321], [RFC4648], this document.
     * Status: deprecated
@@ -373,7 +372,7 @@ to collision attacks [IACR-2020-014].
   sha
   : * Description:  The SHA-1 algorithm [RFC3174].  The output of this
       algorithm is encoded using the base64 encoding  [RFC4648].
-      The "sha" algorithm MUST NOT be used as it's now vulnerable
+      This digest-algorithm MUST NOT be used as it's now vulnerable
       to collision attacks [IACR-2020-014].
     * Reference: [RFC3174], [RFC6234], [RFC4648], this document.
     * Status: deprecated
@@ -399,7 +398,7 @@ to collision attacks [IACR-2020-014].
     * Status: standard
 
 To allow sender and recipient to provide a checksum which is independent from
-`Content-Encoding`, the following additional algorithms are defined:
+`Content-Encoding`, the following additional digest-algorithms are defined:
 
   {: vspace="0"}
   id-sha-512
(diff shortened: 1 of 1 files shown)


ioggstream:
Hi @manger could you please take a look at this PR?

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1080,40 +1080,51 @@ registry. The allowed values for the "Status" fields are described below.
...
 Values](https://www.iana.org/assignments/http-dig-alg/http-dig-alg.xhtml)
 registry:
 
-* Digest Algorithm: MD5
+* Digest Algorithm: md5
 * Description: As specified in {{algorithms}}.
 * Status: As specified in {{algorithms}}.
 
-## Update "CRC32C" Digest Algorithm {#iana-CRC32C}
+## Update "UNIXsum" and "UNIXcksum" Digest Algorithm {#iana-unixsum-unixcksum}
```
ioggstream, line 1093: @LPardue iiuc we forgot to mention the update of UNIX*sum

draft-ietf-httpbis-digest-headers.md
```diff
@@ -335,49 +335,49 @@ some algorithms, one or more parameters can be supplied.
 ~~~
 
 The BNF for "parameter" is defined in Section 5.4.1.4 of
-{{SEMANTICS}}. All digest-algorithm values are case-insensitive.
+{{SEMANTICS}}. All digest-algorithm values are case-insensitive but SHOULD be used in lowercase.
```
reschke, line 339: TBH, I'm not sure how a SHOULD is justified here. This is not using SF syntax, so why does it matter?

ioggstream: Which is a sensible way to signal we want to transition to lowercase and in future releases of Digest we may want to switch to SF ?

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1053,7 +1053,7 @@ the whole payload before sending a message (eg. see {{?I-D.thomson-http-mice}}).
 
 Digest may expose information details of encrypted payload when the checksum
 is computed on the unecrypted data.
```
manger, line 1055: typo: unencrypted

manger: Looks ok.
Given the labels are being changed to lowercase it might be good to add to the note on the IANA page https://www.iana.org/assignments/http-dig-alg/http-dig-alg.xhtml: "Digest algorithm names are case-insensitive"

ioggstream: fixed on master, thanks!

ioggstream: @manger imho as algorithms are case-insensitive since RFC3230 you can already add the note to the page: do you need any action on my side for that?

draft-ietf-httpbis-digest-headers.md
```diff
@@ -335,49 +335,49 @@ some algorithms, one or more parameters can be supplied.
...
 
 The Internet Assigned Numbers Authority (IANA) acts as a registry for
 digest-algorithm values.
 The registry contains the tokens listed below.
 
 Some algorithms, although registered, have since been found vulnerable:
-the MD5 algorithm MUST NOT be used due to collision attacks [CMU-836068]
-and the SHA algorithm MUST NOT be used due
+the "md5" algorithm MUST NOT be used due to collision attacks [CMU-836068]
+and the "sha" algorithm MUST NOT be used due
```
LPardue, line 346: Unfortunately due to the overloading of algorithm and the way this section is structured, the lowercasing of these values seems inconsistent with your stated aims. A few ways to fix this come to mind:

1) Unwind the change and leave them as uppercase
2) 

```
the "md5" token MUST NOT be used due to collision attacks in the MD5 algorithm [link], the "sha" token MUST NOT be used due to collision attacks [link]
```

3) 
```
the MD5 algorithm (token "md5") MUST NOT be used ...
```

I have a slight preference for option 1.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1080,40 +1080,51 @@ registry. The allowed values for the "Status" fields are described below.
...
 * Status: As specified in {{algorithms}}.
 
-## Update "CRC32C" Digest Algorithm {#iana-CRC32C}
+## Update "UNIXsum" and "UNIXcksum" Digest Algorithm {#iana-unixsum-unixcksum}
+
+This memo updates the "UNIXsum" and "UNIXcksum" digest algorithms in the [HTTP Digest Algorithm
+Values](https://www.iana.org/assignments/http-dig-alg/http-dig-alg.xhtml)
+registry:
+
+* Digest Algorithm: As specified in {{algorithms}}.
```
LPardue, line 1099: For consistency, probably just easier to do these as separate items, even if it means the text is more verbose.

LPardue:
I made one comment on what I thought was an inconsistency in use of uppercase or not. After further reading I think there are other instances where the editorial rule - uppercase for actual algorithm, lowercase for digest-algorithm (or token?), is hard to rationalise about because the document seems to interchangeably use "algorithm", "digest algorithm" and "digest-algorithm". In order to land this change I think we need to solve that editorial problem. Part of what would help that is to state our algorithm editorial rules in the terminology.

ioggstream:
> state our algorithm editorial rules in the terminology

+1

draft-ietf-httpbis-digest-headers.md
ioggstream: I interpreted the following as digest-algorithms because the sentence starts with

> Some algorithms, although registered

Rewording like below should fix imho

> Some digest-algorithms, although registered

OTOH to reference algorithm names instead of tokens, we should use SHA-1

```suggestion
the MD5 algorithm MUST NOT be used due to collision attacks [CMU-836068]
and the SHA-1 algorithm MUST NOT be used due
```

draft-ietf-httpbis-digest-headers.md
ioggstream: @reschke I used `preferred`. WDYT?

draft-ietf-httpbis-digest-headers.md
```diff
@@ -332,49 +335,50 @@ some digest-algorithms, one or more parameters can be supplied.
...
     * Reference: [RFC6234], [RFC4648], this document.
     * Status: standard
 
-  MD5
+  md5
   : * Description: The MD5 algorithm, as specified in [RFC1321].
       The output of this algorithm is encoded using the
       base64 encoding  [RFC4648].
-      The MD5 algorithm MUST NOT be used as it's now vulnerable
+      This digest-algorithm MUST NOT be used as it's now vulnerable
```
ioggstream, line 368: Avoid repetitions.

LPardue (approved):

PR #1255, opened 2020-08-31
Always use digest-algorithm (with dash). See #1211.

ioggstream:
## This PR

- is preliminary to #1253 

Consistently use `digest-algorithm` instead of `digest algorithm`.  In case we need to further focus on the value, we can say `digest-algorithm value` or `digest-algorithm token`.

Looking at "content-coding" vs "content coding" in https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html I think that this choice makes the document more readable.

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -159,7 +159,7 @@ the value contained in any `Content-Encoding` or `Content-Type` header fields.
 Therefore, a given resource may have multiple different digest values.
 
 To allow both parties to exchange a Digest of a representation with no content
-codings (see Section 7.1.2 of {{SEMANTICS}}) two more algorithms
+codings (see Section 7.1.2 of {{SEMANTICS}}) two more digest-algorithms
 are added ("id-sha-256" and "id-sha-512").
 
 ## Goals
@@ -169,7 +169,7 @@ The goals of this proposal are:
    1. Digest coverage for either the resource's `representation data` or
       `selected representation data` communicated via HTTP.
 
-   2. Support for multiple digest algorithms.
+   2. Support for multiple digest-algorithms.
 
    3. Negotiation of the use of digests.
 
@@ -242,7 +242,7 @@ together with an indication of the algorithm used (and any parameters)
                                 <encoded digest output>
 ~~~
 
-The checksum is computed using one of the `digest-algorithms` listed in {{algorithms}}
+The checksum is computed using one of the digest-algorithms listed in {{algorithms}}
 and then encoded in the associated format.
 
 The example below shows the  "sha-256" digest-algorithm which uses base64 encoding.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -338,9 +338,9 @@ The Internet Assigned Numbers Authority (IANA) acts as a registry for
 digest-algorithm values.
 The registry contains the tokens listed below.
 
-Some algorithms, although registered, have since been found vulnerable:
+Some digest-algorithms, although registered, rely on vulnerable algorithms:
 the MD5 algorithm MUST NOT be used due to collision attacks [CMU-836068]
-and the SHA algorithm MUST NOT be used due
+and the SHA-1 algorithm MUST NOT be used due
```
ioggstream, line 343: If we want to reference the crypto algorithm, SHA-1 is the right name.

Generally, with this statement we explicitly forbid using SHA-1 and MD5 based digest-algorithms, even if not listed here (that is: a future  multihash digest-algorithm with https://github.com/multiformats/multicodec/blob/master/table.csv#L71)

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1079,7 +1079,7 @@ registry. The allowed values for the "Status" fields are described below.
 
 ## Deprecate "MD5" Digest Algorithm {#iana-MD5}
 
-This memo updates the "MD5" digest algorithm in the [HTTP Digest Algorithm
+This memo updates the "MD5" digest-algorithm in the [HTTP Digest Algorithm
```
ioggstream, line 1082: @LPardue here and below, we could replace `digest-algorithm` with `entry`

draft-ietf-httpbis-digest-headers.md
```diff
@@ -338,9 +338,9 @@ The Internet Assigned Numbers Authority (IANA) acts as a registry for
 digest-algorithm values.
 The registry contains the tokens listed below.
 
-Some algorithms, although registered, have since been found vulnerable:
-the "MD5" algorithm MUST NOT be used due to collision attacks [CMU-836068]
-and the "SHA" algorithm MUST NOT be used due
+Some digest-algorithms, although registered, rely on vulnerable algorithms:
+the "MD5" digest-algorithm MUST NOT be used due to collision attacks [CMU-836068]
+and the "SHA" digest-algorithm MUST NOT be used due
```
LPardue, line 343: Based on your other PR, I think this would be lower case md5 and sha?

draft-ietf-httpbis-digest-headers.md
LPardue: Based on your other PR, lowercase again?

draft-ietf-httpbis-digest-headers.md
ioggstream: I split the PRs:

- #1256  lowercased only the *new* digest algorithms because that doesn't update anything;
- this will consistently use digest-algorithm
- the last PR will lowercase existing algorithms name.

LPardue: Makes sense, please proceed 🙂

LPardue (approved):

draft-ietf-httpbis-digest-headers.md
ioggstream: will be lowercased by the other PR but not in this one.

PR #1259, opened 2020-09-07
Fix: #850. Obsolete parameters.

ioggstream:
## This PR

- obsoletes digest parameters

See #850

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -466,9 +466,18 @@ The `contentMD5` token defined in Section 5 of [RFC3230] MUST NOT be used as a d
 
 # Obsolete Digest Header Field Parameters {#obsolete-parameters}
 
-This RFC obsoletes the usage of parameters with `Digest` introduced in
+This document obsoletes the usage of parameters with `Digest` introduced in
 Section 4.1.1 and 4.2 of [RFC3230] because this feature has not been widely deployed
 and complicates field-value processing.
+
+Field parameters provided a common way to attach additional information
+to a representation-data-digest,
+but if they are used as an input to validate the checksum, an attacker could alter them to steer
+the validation behavior.
+
+A digest-algorithm can still be parameterized defining its own way to encode parameters into the
+representation-data-digest in such a way as to mitigate security risks related to its computation.
+
 
 # Relationship to Subresource Integrity (SRI)
 
@@ -1065,13 +1074,7 @@ in conjuction with the encrypted content-coding {{?RFC8188}}.
 
 ## Algorithm Agility
 
-Algorithm agility is achieved establishing the IANA Digest Algorithm Values registry,
-and adding it the "Status" field making it possible to deprecate and obsolete insecure algorithms.
-
-The now obsoleted [RFC3230] allowed passing parameters to digest-algorithms:
-if those parameters are used to verify the checksum, there will be enough space for bad or insecure
-implementations. If a digest-algorithm want to support parameters, it can encode them into the
-representation-data-digest.
+...
 
 # IANA Considerations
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -1059,7 +1065,13 @@ in conjuction with the encrypted content-coding {{?RFC8188}}.
...
 ## Algorithm Agility
 
-...
+Algorithm agility is achieved establishing the IANA Digest Algorithm Values registry,
+and adding it the "Status" field making it possible to deprecate and obsolete insecure algorithms.
+
+The now obsoleted [RFC3230] allowed passing parameters to digest-algorithms:
+if those parameters are used to verify the checksum, there will be enough space for bad or insecure
+implementations. If a digest-algorithm want to support parameters, it can encode them into the
+representation-data-digest.
```
LPardue, line 1074: These paragraphs are confusing under the section title. I think it would be clearer to 

1) Take what's in the first paragraph and combine it with what's in section 12.2 and 12.3. I suggest we do that in a different PR

2) Rename the section to something like "Paramaterising digest-algorithms"

3) Take the second paragraph and explain a little more why the old means was "bad or insecure". For example, what could have gone wrong? That makes the problem statement clearer so that the solution of embedding parameters make more sense to anyone considering what to do.

ioggstream: Sure, I'll move 
```
Algorithm agility is achieved establishing the IANA Digest Algorithm Values registry,
and adding it the "Status" field making it possible to deprecate and obsolete insecure algorithms.
```
to another PR!

ioggstream: I propose to move all this section in {{obsolete-parameters}} as removing parameters we do not have security considerations.

ioggstream: About combining 12.2 and 12.3, I re-read rfc 8126 and iiuc those sections should remain unchanged ("keep iana considerations for iana")... Let's discuss it in https://github.com/httpwg/http-extensions/pull/1284

draft-ietf-httpbis-digest-headers.md
```diff
@@ -278,6 +278,8 @@ A recipient MAY ignore any or all of the representation-data-digests in a Digest
 field. This allows the recipient to choose which digest-algorithm(s) to use for
 validation instead of verifying every received representation-data-digest.
 
+A recipient MUST ignore a representation-data-digest containing parameters.
```
mnot, line 281: What does this _mean_? `representation-data-digest` above does not contain syntax for `parameters`, so how does a parser know if parameters are present?

LPardue: My comment was stuck pending. I think mnot and I have the same thought

ioggstream: Thanks @mnot @LPardue ! 

The confusion comes from the fact that the old spec allowed for parameters(even if they were not presenti in the syntax) while this one does not.

I wanted to provide guidance in how to manage a Digest header sent by an old implementation: which is the correct way to address this issue?

Hyp 1: add the parameter definition to the ABNF and explicit that they SHOULD|MUST be ignored, while the checksum should be processed

Hyp 2: add the parameter definition to the ABNF and explicit that if a `representation-data-digest` contains parameter *only this checksum should be ignored*

Hyp 3: the presence of a parameter in any `representation-data-digest` will invalidate the syntax of the whole Digest value.


Thanks for your support, R.

ioggstream, line 281:
```suggestion
```

Removed this line. I'll open another issue for providing guidance in the parameter transition.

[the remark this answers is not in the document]
ioggstream: Line removed :)

ioggstream:
@LPardue I removed the wrong line. If you PTAL we can merge and cutoff

LPardue (approved):
Let's do it!

PR #1283, opened 2020-10-07
Add definition and semantics of applied parameter.

roger-on-github:
Define an 'applied' parameter for Priority HTTP Header Field (responses only). The presence of this parameter indicates that a Priority HTTP Header Field was recognized in the request. Its value indicates that the requested prioritization was applied to the scheduling of the response (or not).

This parameter will be useful to clients whose optimal behavior depends on whether prioritization controls are available. For example, consider a media client that plays a live stream by successively downloading and displaying short media segments as they are produced. The standard model is to assume a connection whose bandwidth is just slightly above the media bit rate. If each media segment is 400ms, the RTT is 100ms, and the client is playing 3 segments behind the live edge, overall performance can be improved by pipelining successive segment requests to amortize the RTT cost -- but only if the remote end does not send responses in parallel (incremental=0). If responses are sent in parallel then the next segment to play will share bandwidth with subsequent response(s) and take longer than 400ms to arrive, causing playback to stall. A client that requests pipelining and discovers that it is not available can then issue future requests in serial, without pipelining them.

### DIFF
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -212,7 +212,7 @@ the Priority header field.
 In both cases, the set of priority parameters is encoded as a Structured Fields
 Dictionary ({{!STRUCTURED-FIELDS}}).
 
-This document defines the urgency(`u`) and incremental(`i`) parameters. When
+This document defines the urgency(`u`), incremental(`i`) and applied ('ap') parameters. When
 receiving an HTTP request that does not carry these priority parameters, a
 server SHOULD act as if their default values were specified. Note that handling
 of omitted parameters is different when processing an HTTP response; see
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-priority.md
```diff
@@ -568,6 +587,22 @@ to do so is an implementation decision. For example, a server might
 pre-emptively send responses of a particular incremental type based on other
 information such as content size.
 
+A server or intermediary that recognizes a Priority HTTP Header Field in a
+request and which successfully applies that priority to its scheduling MUST
+ensure that the response contains a Priority HTTP Header Field
+with the same parameters and an applied parameter with a value of true.
```
martinthomson, line 593: What if the parameter is not understood?

What if only some parameters were respected?

What if the response carries different priority information intentionally?

roger-on-github: If the parameter is not understood then it cannot be applied, by definition, so applied would be false. Similarly if only some parameters were respected, or if the response intentionally applied different values.

roger-on-github: FYI, there's a discussion of this going on in the Issues section, https://github.com/httpwg/http-extensions/issues/1274

LPardue:
Closing this since the linked issues has been closed with no action.

Issue #1250, opened 2020-08-18
Digest editorial refresh before WGLC


ioggstream:
## Tasks

- [x] algorithm registration #1533 
- [x] digest-algorithm lowercase #1253 
- [x] deprecate parameters and wait for feedback :) #1259 @LPardue 
- [x] Algorithm agility #1284 
- [x] sha-1 deprecated o obsoleted? #1292 
- [x] detail encrypted quirks eg with content-coding #1285 
- [x] clean up Representation Digest https://github.com/httpwg/http-extensions/pull/937#discussion_r472018411
- [x] remove reference to "validator fields" if they are not used anymore
- [x] consistent usage of backticks #1255 

```
`Digest` vs Digest vs "Digest"
`digest-algorithms` vs digest-algoritms vs "digest-algorithms"
```

- [x] consistent usage of "-" #1256  

```
digest-algorithms vs digest algorithms

```

- [x] consistent usage of lowercase digest-algorithms with double quote "sha-256" in text #1255

- [x] close #1139 and #1138  about old algorithms unless somebody comes with a proposal. @LPardue

ioggstream:
@LPardue do you have any news on #1139 and #1138 about old algorithms?

LPardue:
I'll propose some new text shortly

ioggstream:
* Abstract: 'thus' reads oddly; try a different word?
* [x] Abstract: It's sufficient to say that this document obsoletes 3230; you don't need to go into why in the abstract.
* [x] 1. Introduction: 'core' is an informal term; recommend just "HTTP does not define...", or dropping the first sentence.
* [x] 1. Introduction: Use of 'resource' is incorrect; I think you're looking for 'representation' (throughout). I.e., it is meaningless to say that you protect the integrity of a resource, because a resource is an ephemeral concept.
* [ ] 1. Introduction: Second sentence of second para is very long, and very passive.
* [x] 1. Introduction: 'representation-data' should drop the hyphen.

ioggstream:
[ ] * 1. Introduction: Typically, the introduction contains a 'map' to the rest of the specification, laying out the logical progression of concepts introduced. I can see that 1.2. does that partially, but it needs to be completed and up-front.
[ ] * 1.1. Is this section necessary? If so, is it necessary for it to be a separate section?
[x] * 1.2. 'This Proposal' is a very odd name for a specification section
[x] * 1.2. The first para is difficult to read; suggest starting with 'This document updates the...'
[...]

PR #1284, opened 2020-10-07
Achieving algorithm agility. See #1250

ioggstream:
## This PR

Describes how to achieve algorithm agility:

- add Status to IANA table;
- use Want-Digest to negotiate newer algos
- use multiple digest values while transitioning to new algos (see https://tools.ietf.org/html/rfc7696#section-2.3 but I didn't cite it as I don't think it is worth unless we want to add MANDATORY algorithms)

## Note:

I didn't merge this section with IANA Considerations because of [Keep IANA Considerations for IANA]( https://tools.ietf.org/html/rfc8126#section-1.1)

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1075,7 +1075,7 @@ to the algorithm used to compute the checksum based on the type and status of th
 in which the checksum algorithm is defined.
 
 The entries associated to a checksum algorithm which becomes insecure or otherwise undesirable
-can be marked as "deprecated",
+are marked as "deprecated",
 while the ones associated to a standardized algorithm without known problems can be marked
 as "standard".
 
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -275,9 +275,6 @@ For example, a server may provide representation-data-digest values using differ
 allowing it to support a population of clients with different evolving capabilities;
 this is particularly useful in support of transitioning away
 from weaker algorithms should the need arise (see {{algorithm-agility}}).
-Another use case for sending multiple checksums is
-for responses expected to reside in caches shared by users with
-different browsers.
 
 A recipient MAY ignore any or all of the representation-data-digests in a Digest
 field. This allows the recipient to choose which digest-algorithm(s) to use for
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1083,7 +1083,7 @@ An endpoint might have a preference for algorithms,
 such as preferring "standard" algorithms over "deprecated" ones.
 Transition from weak algorithms is supported
 by negotiation of digest-algorithm using `Want-Digest` (see {{want-digest}})
-or by sending multiple representation-data-digest values for the receiver to choose.
+or by sending multiple representation-data-digest values from which the receiver chooses.
 Endpoints are advised that sending multiple values consumes resources,
 which may be wasted if the receiver ignores them (see {{digest}}).
 
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1085,7 +1085,7 @@ Transition from weak algorithms is supported
 by negotiation of digest-algorithm using `Want-Digest` (see {{want-digest}})
 or by sending multiple representation-data-digest values for the receiver to choose.
 Endpoints are advised that sending multiple values consumes resources,
-which may be wasted if the receiver ignores them (see {{digest}})."
+which may be wasted if the receiver ignores them (see {{digest}}).
 
 # IANA Considerations
 
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1068,7 +1068,7 @@ The security properties of digest-algorithms are not fixed.
 Algorithm agility is achieved establishing the IANA Digest Algorithm Values
 registry (see {{iana-digest-algorithm-registry}}) and Section 2.1 of {{?RFC7696}}).
 
-To help endpoints make secure algorithm choices,
+To help endpoints understand weaker algorithms from stonger ones,
 this document adds to the IANA Digest Algorithm Values registry
 a new "Status" field containing the most-recent appraisal of the digest-algorithm according
 to the algorithm used to compute the checksum based on the type and status of the primary document
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1070,9 +1070,7 @@ digest-algorithm from the IANA Digest Algorithm Values registy in {{iana-digest-
 
 To help endpoints understand weaker algorithms from stonger ones,
 this document adds to the IANA Digest Algorithm Values registry
-a new "Status" field containing the most-recent appraisal of the digest-algorithm according
-to the algorithm used to compute the checksum based on the type and status of the primary document
-in which the checksum algorithm is defined.
+a new "Status" field containing the most-recent appraisal of the digest-algorithm.
 
 The entries associated to a checksum algorithm which becomes insecure or otherwise undesirable
 are marked as "deprecated",
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -270,11 +270,14 @@ The relationship between `Content-Location` (see Section 7.8 of
 representation metadata, payload transformations and HTTP methods on Digest is
 provided in {{examples-unsolicited}} and {{examples-solicited}}.
 
-A `Digest` field MAY contain multiple representation-data-digest values. This
-could be useful
-to transition from weak algorithms or
+A `Digest` field MAY contain multiple representation-data-digest values.
+For example, a server may provide representation-data-digest values using different algorithms,
+allowing it to support a population of clients with different evolving capabilities;
+this is particularly useful in support of transitioning away
+from weaker algorithms should the need arise (see {{algorithm-agility}}).
+Another use case for sending multiple checksums is
 for responses expected to reside in caches shared by users with
-different browsers, for example.
+different browsers.
 
 A recipient MAY ignore any or all of the representation-data-digests in a Digest
 field. This allows the recipient to choose which digest-algorithm(s) to use for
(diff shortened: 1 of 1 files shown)


jricher:
This only addresses the algorithm used in digesting the message body, not the algorithm used to sign content.

LPardue:
No problem @jricher :) This is one of the joys of a shared repo. Labels help but aren't a complete solution.

ioggstream:
> Sorry, ignore that -- I thought this was a comment on the signatures draft, my mistake.

@jricher still you are still in time to  provide some feedback on Digest :D we hope to release it soon!

@LPardue PTAL :P

draft-ietf-httpbis-digest-headers.md
```diff
@@ -271,7 +271,9 @@ representation metadata, payload transformations and HTTP methods on Digest is
 provided in {{examples-unsolicited}} and {{examples-solicited}}.
 
 A `Digest` field MAY contain multiple representation-data-digest values. This
-could be useful for responses expected to reside in caches shared by users with
+could be useful
+to transition from weak algorithms or
+for responses expected to reside in caches shared by users with
 different browsers, for example.
```
LPardue, line 277: If we're going to expand this we can probably write this better. The first thing that spring into my head is something like

A `Digest` field MAY contain multiple representation-data-digest values. For example, a server may provide representation-data-digest values using different algorithms, allowing it to support a population of clients with different evolving capabilities.

LPardue: and optionally tack on

"This is particularly useful in support of transitioning away from weaker algorithms should the need arise; see {{algorithm-agiity}}"

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1059,7 +1061,16 @@ in conjuction with the encrypted content-coding {{?RFC8188}}.
...
+Algorithm agility is achieved establishing the IANA Digest Algorithm Values registry,
+and adding it the "Status" field making it possible to deprecate and obsolete insecure algorithms.
+
+Transition from weak algorithms can be achieved
+negotiating the most suitable digest-algorithm using `Want-Digest` (see {{want-digest}})
+or sending multiple representation-data-digest, thus allowing the receiver to select the best
+possible digest-algorithm.
+Even in this case, it is good practice to avoid sending non-necessary
+representation-data-digest values because the receiver could either ignore them (see {{digest}})
+or uselessly consume resources to evaluate all of them.
```
LPardue, line 1073: Now that there is text, I can see the bones of what you're trying to say. But I think we can rewrite this to be a bit clearer. Here's strawperson text:

"The security properties of digest algorithms are not fixed. It is in an endpoint's best interest to support and use the most secure algorithms that are at its disposal. However, there is little that this specification can do to mandate behaviour. To help endpoints make better algorithm choices a new "Status" field has been added to the IANA Digest Algorithm Values registry (see Section xyz), which contains the most-recent appraisal of the algorithm. Algorithms that have been found insecure or otherwise undesirable can be marked as "deprecated" or "obsolete". Algorithms with no known problems can be marked as "Standard", while new algorithms can be marked "Experimental".

An endpoint might have a preference for algorithms, such as preferring "standard" algorithms over "deprecated" ones. Algorithm agility is supported by negotiation of digest-algorithm using ``Want-Digest` (see {{want-digest}})` or by sending multiple representation-data-digest values for the receiver to choose. Endpoints are advised that sending multiple values consumes resources, which may be wasted if the receiver ignores them (see {{digest}})."

ioggstream:
@LPardue I tried to integrate your text, which improves significantly the description. Thanks!

If we want to shorten the text, we could replace some parts referencing RFC7696 and pointing the reader there (explanation of algorithm agility, transition from weaker algorithms, ...)

WDYT?

draft-ietf-httpbis-digest-headers.md
```diff
@@ -270,9 +270,14 @@ The relationship between `Content-Location` (see Section 7.8 of
...
-could be useful for responses expected to reside in caches shared by users with
-different browsers, for example.
+A `Digest` field MAY contain multiple representation-data-digest values.
+For example, a server may provide representation-data-digest values using different algorithms,
+allowing it to support a population of clients with different evolving capabilities;
+this is particularly useful in support of transitioning away
+from weaker algorithms should the need arise (see {{algorithm-agility}}).
+Another use case for sending multiple checksums is
+for responses expected to reside in caches shared by users with
+different browsers.
```
LPardue, line 280: This sentence seems to me to just repeat the notion of evolving capability already introduced in the paragraph.

I suspect this text, originally in RFC 3230, was attempting to address transparent caching proxies. I'm not convinced we need to keep it in that form given how many HTTP deployments work. Maybe @mnot has an opinion.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1059,11 +1064,32 @@ in conjuction with the encrypted content-coding {{?RFC8188}}.
 
 ## Algorithm Agility
 
-...
+The security properties of digest-algorithms are not fixed.
+Algorithm agility is achieved establishing the IANA Digest Algorithm Values
+registry (see {{iana-digest-algorithm-registry}}) and Section 2.1 of {{?RFC7696}}).
```
LPardue, line 1069: How about,

> Algorithm Agility (see {{?RFC7696}}) is achieved by providing implementations flexibility in their choice of digest-algorithm from the IANA Digest Algorithm Values registy in {{iana-digest-algorithm-registry}}.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1059,11 +1064,32 @@ in conjuction with the encrypted content-coding {{?RFC8188}}.
...
+
+To help endpoints make secure algorithm choices,
```
LPardue, line 1071: Let's not say secure, because algorithms will inevitably be broken. How about 

```suggestion
To help endpoints understand weaker algorithms from stonger ones,
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1059,11 +1064,32 @@ in conjuction with the encrypted content-coding {{?RFC8188}}.
...
+this document adds to the IANA Digest Algorithm Values registry
+a new "Status" field containing the most-recent appraisal of the digest-algorithm according
+to the algorithm used to compute the checksum based on the type and status of the primary document
+in which the checksum algorithm is defined.
```
LPardue, line 1075: The remember of the paragraph is hard to parse and I don't think it is strictly true. I'd like to think that we can deprecate algorithms without the need of coordination with the original defining document. The reason some algorithm becomes deprecated or obsolete could because it is a dead. I think it would be simpler to just remove everything after "... appraisal of the digest-algorithm."

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1059,11 +1064,32 @@ in conjuction with the encrypted content-coding {{?RFC8188}}.
...
+
+The entries associated to a checksum algorithm which becomes insecure or otherwise undesirable
+can be marked as "deprecated",
```
LPardue, line 1078:
```suggestion
are marked as "deprecated",
```

let's use the active voice.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1059,11 +1064,32 @@ in conjuction with the encrypted content-coding {{?RFC8188}}.
...
+while the ones associated to a standardized algorithm without known problems can be marked
```
LPardue, line 1079:
```suggestion
while the ones associated to a standardized algorithm without known problems are marked
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1059,11 +1064,32 @@ in conjuction with the encrypted content-coding {{?RFC8188}}.
...
+as "standard".
+
+An endpoint might have a preference for algorithms,
+such as preferring "standard" algorithms over "deprecated" ones.
+Transition from weak algorithms is supported
+by negotiation of digest-algorithm using `Want-Digest` (see {{want-digest}})
+or by sending multiple representation-data-digest values for the receiver to choose.
```
LPardue, line 1086:
```suggestion
or by sending multiple representation-data-digest values from which the receiver chooses.
```

LPardue (changes requested):
Ahh I'd overlooked RFC 7696 sorry. I think what's here is almost right, and I've made some minor suggestions.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1059,11 +1064,32 @@ in conjuction with the encrypted content-coding {{?RFC8188}}.
...
+Endpoints are advised that sending multiple values consumes resources,
+which may be wasted if the receiver ignores them (see {{digest}})."
```
martinthomson, line 1088:
```suggestion
which may be wasted if the receiver ignores them (see {{digest}}).
```

draft-ietf-httpbis-digest-headers.md
ioggstream, line 1069:
```suggestion
Algorithm Agility (see {{?RFC7696}}) is achieved by providing implementations flexibility in their choice of
digest-algorithm from the IANA Digest Algorithm Values registy in {{iana-digest-algorithm-registry}}.
```

ioggstream: applied!

draft-ietf-httpbis-digest-headers.md
ioggstream: Done

draft-ietf-httpbis-digest-headers.md
ioggstream: In our spec  https://github.com/httpwg/http-extensions/blame/master/draft-ietf-httpbis-digest-headers.md#L1079:1082
we reused this part of RFC3864 

      Specify "standard", "experimental", "informational", "historic",
      "obsoleted", or some other appropriate value according to the type
      and status of the primary document in which it is defined.  For
      non-IETF specifications, those formally approved by other
      standards bodies should be labelled as "standard"; others may be
      "informational" or "deprecated" depending on the reason for
      registration.

which is more relaxed than the present. I will relax our statement and ask to reviewers for feedback so that we can land this PR.

draft-ietf-httpbis-digest-headers.md
ioggstream: @LPardue Are you proposing this?

```suggestion
```

draft-ietf-httpbis-digest-headers.md
ioggstream:
```suggestion
this document adds to the IANA Digest Algorithm Values registry
a new "Status" field containing the most-recent appraisal of the digest-algorithm.
```

draft-ietf-httpbis-digest-headers.md
LPardue: Yes.

ioggstream:
Reorganized the algorithm status here https://github.com/httpwg/http-extensions/pull/1284/commits/d389ea9b0e25de45f4f27885d511dc91f4221631# @LPardue

LPardue:
looks good. Let's land that sentence deletion and merge this.

PR #1291, opened 2020-10-16
Add changelog since -03

ioggstream:
## This PR

Adds changelog since -03


LPardue:
I think we missed the parameter changes?

Also I just realised that the draft order is ascending, folks typically do it in descending order so the most recent changes are easier to see

ioggstream:
1. The parameter didn't land yet :( do you have a solution for Mark's point? https://github.com/httpwg/http-extensions/pull/1259#discussion_r501403171
2. ok :)

pes2020kk (commented):
ios 14

ioggstream:
@LPardue seems spam: should we report it to gh?

Issue #1311, opened 2020-10-24
remove redundant definitions after transitioning to structured fields


ioggstream:
## I expect

- [x] to reference sf-decimal and sf-integer. It is probably more readable to define the params first, and then reference them in the content-identifier section
- [x] use SF reference, use Unixtime reference #1313

msporny:
+1, agree that we should refer to sf-integer. Where would we refer to sf-decimal in the spec?

ioggstream:
> +1, agree that we should refer to sf-integer. Where would we refer to sf-decimal in the spec?

https://github.com/httpwg/http-extensions/pull/1312/files#diff-1d57bca6223a0fee3ef29148c2550c0f862e72f67a56cc7f57b5a72fbd8320e3R262

Issue #1305, opened 2020-10-23
signatures: datatypes of (created) and (expires)


llorllale:
The draft defines an odd asymmetry between the data types of `(created)` and `(expires)` (respectively: integer string and decimal string). The non-normative examples all show `(expires)` as an integer string. Suggest we have them both as integer strings as:

* it simplifies implementations
* if omission of nanoseconds is OK on `(created)` (yes, imo), then it should be OK as well for `(expires)` imo

richanna:
If you are signing a request being sent over a high throughput link between systems with good clock synchronization, 1000ms is way more lifetime than necessary. That said, I'm hard pressed to think of a scenario that would be mitigated by an expiration time of +200ms but not by an expiration time of +1000ms.

msporny:
Agree, created/expires should be an integer representing time from epoch with 1000ms resolution.

jricher:
These fields have been removed and redefined as parameters.

PR #1312, opened 2020-10-24
Remove redundant definitions. See #1311

ioggstream:
## This PR

- uses sf-integer and sf-decimal to avoid re-defining Integer String and Decimal String

## Note

This PR does not aim at deciding the precision of the associated fields, but only to delegate their definition to Structured Fields.
*created and *expires precision can be discussed in another PR :)


draft-ietf-httpbis-message-signatures.md
```diff
@@ -499,11 +493,13 @@ alg
...
+: RECOMMENDED. The `created` parameter is an `sf-integer` containing the signature's Creation Time,
+  expressed as the canonicalized value of the `*created` content identifier, as defined in Section 2.  If not specified, the signature's Creation Time is undefined.  This parameter is useful when signers are not capable of controlling the Date HTTP Header such as when operating in certain web browser environments.
 
 expires
 : 
-: OPTIONAL. The `expires` parameter is a Decimal containing the signature's Expiration Time, expressed as the canonicalized value of the `*expires` content identifier, as defined in Section 2.  If the signature does not have an Expiration Time, this parameter MUST be omitted.  If not specified, the signature's Expiration Time is undefined.
+: OPTIONAL. The `expires` parameter is a `sf-decimal` containing the signature's Expiration Time,
```
msporny, line 501:
```suggestion
: OPTIONAL. The `expires` parameter is a `sf-integer` containing the signature's Expiration Time,
```

I believe this was a decimal initially because I wanted to support millisecond and nanosecond precision... the use case was high frequency trading. In time, it became evident that if people want that sort of resolution, they can create their own X-Header (or the like... High-Precision-Expiration: 234.3342).

msporny (changes requested):

draft-ietf-httpbis-message-signatures.md
ioggstream: @msporny thanks! we are discussing create/expires here https://github.com/httpwg/http-extensions/pull/1280

In this PR I just wanted to get rid of duplicate definitions leaving the semantic discussion to other issues.

llorllale: #1305 

> I believe this was a decimal initially because I wanted to support millisecond and nanosecond precision... the use case was high frequency trading. In time, it became evident that if people want that sort of resolution, they can create their own X-Header (or the like... High-Precision-Expiration: 234.3342).

Thanks for sharing the historic reasons behind this!

llorllale (approved):

draft-ietf-httpbis-message-signatures.md
richanna: A non-standard `High-Precision-Expiration` header field doesn't enable sub-second expiration precision. The signature's expiration time is defined by the `*expires` parameter. Profiling specs can't change that, they can only add additional expiration semantics on top of that. We should not encourage that though, as it introduces confusion – which expiration rules apply? Do they both have to pass, or just one? etc.

If we want to eliminate decimals, let's express both `*created` and `*expires` in milliseconds.

martinthomson: I'm amazed.  Is the implication here that signers and verifiers have clocks synchronized with sub-second precision?

ioggstream: @martinthomson 

1-  I agree with your consideration
2- In this PR I just wanted to get rid of duplicate definitions leaving the semantic discussion to other issues. Specifically, there are other parts defining expires as a decimal. 
3- We should - in other PRs - discuss the precision of *created and *expires which imho should be expressed with `seconds` precision, but I don't think should be discussed in this PR :D 





 this PR is mostly an editorial one which just removes redundant definitions. 

Once we agree we can delegate definitions

msporny: @martinthomson wrote:
> I'm amazed. Is the implication here that signers and verifiers have clocks synchronized with sub-second precision?

Yes, this is true in high-frequency trading systems (for example)... some of which have highly tuned HTTP stacks and are looking at QUIC + HTTP/3 with interest. The legal implications here are that, in some of those systems, that you're supposed to execute orders as you receive them and sometimes the digital signature time matters (from a contractual basis). Here's what a couple of hundred milliseconds in a high-frequency trading system looked like back in 2013 (when we started on this spec):

https://youtu.be/B_k_elbBz8c?t=103

Things move much more quickly now, which is why we might want to drop this use case and get them to create and sign over their own header that has nanosecond (or better) precision. 

All that to say, we could probably back off on the millisecond precision and go to second-based precision if we wanted to support constrained devices that may not be able to do 64-bit integers?

Also, agree with @ioggstream -- all of this is for another PR.

msporny: @richanna wrote:
> A non-standard High-Precision-Expiration header field doesn't enable sub-second expiration precision. The signature's expiration time is defined by the *expires parameter. Profiling specs can't change that, they can only add additional expiration semantics on top of that. We should not encourage that though, as it introduces confusion – which expiration rules apply? Do they both have to pass, or just one?

Hmm, I wish things worked like that. :)

For example, the high frequency traders won't create a profiling spec, they'll do whatever they want to with this spec as it's base. Notably, they won't sign over `expires` (since it's optional) and will instead add their own header with nanosecond precision, sign over that, and use the standard signature verification thing PLUS their own verification algorithm that takes their header into account.

That's not necessarily a bad thing, per se... and I admit that this is a corner case... but we should be careful to not prevent this sort of usage of the specification.

richanna: > Is the implication here that signers and verifiers have clocks synchronized with sub-second precision?

Yes. In addition to @msporny's high-frequency trading example, this level of clock synchronization is not uncommon within distributed systems.

> For example, the high frequency traders won't create a profiling spec, they'll do whatever they want to with this spec as it's base. Notably, they won't sign over expires (since it's optional) and will instead add their own header with nanosecond precision, sign over that, and use the standard signature verification thing PLUS their own verification algorithm that takes their header into account.

Yes, and that results in a non-standard, non-interoperable implementation. That's what I want to avoid by including support for sub-second expiration times in the spec.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -258,15 +250,17 @@ The following table shows example canonicalized values for different content ide
 
 The signature's Creation Time ({{signature-metadata}}) is identified by the `*created` identifier.
 
-Its canonicalized value is an Integer String containing the signature's Creation Time expressed as the number of seconds since the Epoch, as defined in [Section 4.16](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_16) of {{POSIX.1}}.
+Its canonicalized value is the `sf-integer` value of the `created` parameter (see {{params}})
+containing the signature's Creation Time expressed as the number of seconds since the Epoch, as defined in [Section 4.16](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_16) of {{POSIX.1}}.
```
richanna, line 254: The definition of `created` refers to the canonicalized value, so referring to `created` creates a circular reference. Small change to fix that:

```suggestion
Its canonicalized value is an `sf-integer` containing the signature's Creation Time expressed as the number of seconds since the Epoch, as defined in [Section 4.16](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_16) of {{POSIX.1}}.
```

draft-ietf-httpbis-message-signatures.md
```diff
@@ -258,15 +250,17 @@ The following table shows example canonicalized values for different content ide
...
 
 > The use of seconds since the Epoch to canonicalize a timestamp simplifies processing and avoids timezone management required by specifications such as [RFC3339].
 
 ## Signature Expiration Time
 
 The signature's Expiration Time ({{signature-metadata}}) is identified by the `*expires` identifier.
 
-Its canonicalized value is a Decimal String containing the signature's Expiration Time expressed as the number of seconds since the Epoch, as defined in [Section 4.16](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_16) of {{POSIX.1}}.
+Its canonicalized value is the `sf-decimal` value of the `expires` parameter (see {{params}})
+containing the signature's Expiration Time expressed as the number of seconds since the Epoch, as defined in [Section 4.16](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_16) of {{POSIX.1}}.
```
richanna, line 263: Removing circular reference as mentioned above.

```suggestion
Its canonicalized value is an `sf-decimal` containing the signature's Expiration Time expressed as the number of seconds since the Epoch, as defined in [Section 4.16](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_16) of {{POSIX.1}}.
```

richanna (approved):

Issue #1185, opened 2020-05-25
Max values, precision for Integer String and Decimal String not defined


richanna:
The definitions for Integer String and Decimal String do not specify a maximum value.  The definition for Decimal String (used to provide sub-second precision for Expiration Time) does not define minimum or maximum precision requirements.  It should set a reasonable requirement here (e.g., MUST support up to 3 decimal places and no more).

Better yet, are there existing definitions we can reference?

ioggstream:
Maybe S-H draft can help.

imho we should just not mandate for subsecond precision and allow both `int` and `float`

ioggstream:
@richanna we are using structured fields: can we close? :)

jricher:
Draft no longer defines data types but uses structured field definitions.

PR #1313, opened 2020-10-24
Reuse existing definitions for Unix time and Structured fields.

ioggstream:
## This PR

reuses existing definitions for S-F and Unix time.


llorllale:
Doesn't #1312 make this PR redundant? (or regressive, really, because #1312 references the exact definitions to use - `sf-integer` and `sf-decimal` from structured headers)

ioggstream:
This probably fixes #1185

Issue #1190, opened 2020-05-25
Remove algorithm-specific rules for content identifiers


richanna:
The rules that restrict when the signer can or must include certain identifiers appear to be related to the pseudo-revving of the Cavage draft that happened when the `hs2019` algorithm was introduced.  We should drop these rules, as it can be expected that anyone implementing this draft will support all content identifiers.

PR #1314, opened 2020-10-25
Fix: #1190. Remove algorithm specific rules for content identifiers.

ioggstream:
## This PR

removes algorithm specific rules for content identifiers

## Notes

Current layout is not backward compatible with http-signatures, so it's safe to do it.

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -377,8 +377,8 @@ The following sections describe each of these steps in detail.
 5. The signer creates an ordered list of content identifiers representing the message content and signature metadata to be covered by the signature, and assigns this list as the signature's Covered Content.
    * Each identifier MUST be one of those defined in Section 2.
    * This list MUST NOT be empty, as this would result in creating a signature over the empty string.
-   * If the signature's Algorithm name does not start with rsa, hmac, or ecdsa, signers SHOULD include `*created` and `*request-target` in the list.
-   * If the signature's Algorithm starts with rsa, hmac, or ecdsa, signers SHOULD include `date` and `*request-target` in the list.
+   * Signers SHOULD include `*created` and `*request-target` in the list.
+   * Signers SHOULD include `date` and `*request-target` in the list.
    * Further guidance on what to include in this list and in what order is out of scope for this document.  However, the list order is significant and once established for a given signature it MUST be preserved for that signature.
 
 
@@ -412,10 +412,6 @@ The following table presents a non-normative example of metadata values that a s
 ### Create the Signature Input {#canonicalization}
 
 The Signature Input is a US-ASCII string containing the content that will be signed.  To create it, the signer concatenates together entries for each identifier in the signature's Covered Content in the order it occurs in the list, with each entry separated by a newline `"\n"`.  An identifier's entry is a US-ASCII string consisting of the lowercased identifier followed with a colon `":"`, a space `" "`, and the identifier's canonicalized value (described below).
-
-If Covered Content contains `*created` and the signature's Creation Time is undefined or the signature's Algorithm name starts with `rsa`, `hmac`, or `ecdsa` an implementation MUST produce an error.
-
-If Covered Content contains `*expires` and the signature does not have an Expiration Time or the signature's Algorithm name starts with `rsa`, `hmac`, or `ecdsa` an implementation MUST produce an error.
 
 If Covered Content contains an identifier for a header field that is not present or malformed in the message, the implementation MUST produce an error.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-message-signatures.md
```diff
@@ -413,10 +413,6 @@ The following table presents a non-normative example of metadata values that a s
 
 The Signature Input is a US-ASCII string containing the content that will be signed.  To create it, the signer concatenates together entries for each identifier in the signature's Covered Content in the order it occurs in the list, with each entry separated by a newline `"\n"`.  An identifier's entry is a US-ASCII string consisting of the lowercased identifier followed with a colon `":"`, a space `" "`, and the identifier's canonicalized value (described below).
 
-If Covered Content contains `*created` and the signature's Creation Time is undefined or the signature's Algorithm name starts with `rsa`, `hmac`, or `ecdsa` an implementation MUST produce an error.
```
llorllale, line 416: Should not remove this entirely. Implementations should still produce an error if the corresponding header for a signature parameter is not found.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -413,10 +413,6 @@ The following table presents a non-normative example of metadata values that a s
...
-
-If Covered Content contains `*expires` and the signature does not have an Expiration Time or the signature's Algorithm name starts with `rsa`, `hmac`, or `ecdsa` an implementation MUST produce an error.
```
llorllale, line 418: Same here - should error if not found

llorllale (changes requested):
Concerned about the complete removal of language around how implementations must behave if the corresponding header of a covered content identifier is not found.

draft-ietf-httpbis-message-signatures.md
jricher: Is the suggestion that this constraint be applied regardless of algorithm name?

I completely agree that this "algorithm name starts with" language is hugely problematic.

ioggstream: @llorllale @jricher right! Both of you are correct! Fixing...

draft-ietf-httpbis-message-signatures.md
```diff
@@ -413,10 +413,6 @@ The following table presents a non-normative example of metadata values that a s
 
 The Signature Input is a US-ASCII string containing the content that will be signed.  To create it, the signer concatenates together entries for each identifier in the signature's Covered Content in the order it occurs in the list, with each entry separated by a newline `"\n"`.  An identifier's entry is a US-ASCII string consisting of the lowercased identifier followed with a colon `":"`, a space `" "`, and the identifier's canonicalized value (described below).
```
ioggstream, line 415:
```suggestion

If Covered Content contains `*created` and the signature's Creation Time is undefined an implementation MUST produce an error.
```

draft-ietf-httpbis-message-signatures.md
llorllale: @jricher 

> Is the suggestion that this constraint be applied regardless of algorithm name?

Yes.

ioggstream: Fixed!

llorllale (approved):

draft-ietf-httpbis-message-signatures.md
```diff
@@ -377,8 +377,8 @@ The following sections describe each of these steps in detail.
 5. The signer creates an ordered list of content identifiers representing the message content and signature metadata to be covered by the signature, and assigns this list as the signature's Covered Content.
    * Each identifier MUST be one of those defined in Section 2.
    * This list MUST NOT be empty, as this would result in creating a signature over the empty string.
-   * If the signature's Algorithm name does not start with rsa, hmac, or ecdsa, signers SHOULD include `*created` and `*request-target` in the list.
-   * If the signature's Algorithm starts with rsa, hmac, or ecdsa, signers SHOULD include `date` and `*request-target` in the list.
+   * Signers SHOULD include `*created` and `*request-target` in the list.
+   * Signers SHOULD include `date` and `*request-target` in the list.
```
richanna, line 381:
```suggestion
```

1. Signers may not have control over the `Date` header. (pre-signed messages, XmlHttpRequest callers, etc.)
2. The `*request-target` guidance is redundant with the line above.

draft-ietf-httpbis-message-signatures.md
ioggstream: thx @llorllale, I fixed it below: does it work for you?

ioggstream (commented):
@richanna @llorllale merged your suggestion, thanks and sorry for the delay!

draft-ietf-httpbis-message-signatures.md
```diff
@@ -377,8 +377,7 @@ The following sections describe each of these steps in detail.
 5. The signer creates an ordered list of content identifiers representing the message content and signature metadata to be covered by the signature, and assigns this list as the signature's Covered Content.
    * Each identifier MUST be one of those defined in Section 2.
    * This list MUST NOT be empty, as this would result in creating a signature over the empty string.
-   * If the signature's Algorithm name does not start with rsa, hmac, or ecdsa, signers SHOULD include `*created` and `*request-target` in the list.
-   * If the signature's Algorithm starts with rsa, hmac, or ecdsa, signers SHOULD include `date` and `*request-target` in the list.
+   * Signers SHOULD include `*created` and `*request-target` in the list.
```
jricher, line 380: Suggest we make this more generic, especially since it's all advice (SHOULD) and not requirement:

```suggestion
   * Signers SHOULD include `*request-target` in the list.
   * Signers SHOULD include a date stamp, such as the `date` header or the `*created` field in the list.
```

Issue #1315, opened 2020-10-25
Removing subheaders content identifiers


ioggstream:
## Discussion

At the last http-wg interim meeting we discussed the opportunity to have sub-headers content identifiers, and it was noted that they are subject to a variety of problems.

I suggest to:

- remove them in the current draft
- eventually refining the proposal in a new PR with the support of the ML

I would be great to get more information about the use cases of this features and share some thoughts with the wg.

msporny:
Can you please explain what happened in the HTTP WG interim meeting for those of us that could not attend? Exactly what is the problem and could you provide a concrete example of the problem?

ioggstream:
Hi @msporny, hope you're fine :)

Here are the [interim minutes](https://httpwg.org/wg-materials/interim-20-10/minutes.html). I don't know if there are some recordings.

One of the problem is the following: 

1- if Signature references a sub-header, the verifier must know the syntax of the header
2- the header will thus be parsed before validating the signature
3- it does not require to send a valid signature to force the verifier to parse (split, token, indexing, ...) whatever header
4- the verifier might not know the header syntax: at this point it can either guess or fail
5- if they guess, the sender can steer the way the signature is processed

@martinthomson  explained better than me a generalization of this issue.

richanna:
> 1- if Signature references a sub-header, the verifier must know the syntax of the header
> 2- the header will thus be parsed before validating the signature

The verifier only needs to know how to parse a Structured Field. It doesn't need to know anything about what the contents mean, or what is or isn't actually permitted by the field's specification. It does not need to interpret the field's value, or act on it in any way.

> 3- it does not require to send a valid signature to force the verifier to parse (split, token, indexing, ...) whatever header

I'm not seeing the concern here. The only thing I can think of is that a malicious signer could try to blow up the verifier by signing a single key out of an absurdly large header field value. But that's a vulnerability in the Structured Field parser implementation, not this draft.

> 4- the verifier might not know the header syntax: at this point it can either guess or fail
> 5- if they guess, the sender can steer the way the signature is processed

See above; the verifier does not need to know anything about the header field, it only needs to know how to parse Structured Fields. We can make it more explicit that a signer is only permitted to use these with Structured Fields, but even if they used it with a non-SF header field, it might not provide the protection they want, but it wouldn't pose an availability risk to the verifier.

ioggstream:
> The verifier only needs to know how to parse a Structured Field

Not all fields are S-F.

> a malicious signer ..

[...]

PR #1317, opened 2020-10-25
Fix: #1315. Remove subheader covered content.

ioggstream:
## This PR

removes sub-header covered content


jricher:
This PR no longer fits against the current draft, which also has less dependence on the subheader content for basic functionality.

jricher:
Obsoleted by #1529

PR #1320, opened 2020-10-27
Validate http headers in CH

mnot:
(FYI, if you're too far into AUTH48 don't worry about merging this, although the RFC Editor might want those sections annotated, and if so, the whitespace should be removed from the beginning of the lines)

### DIFF
--- a/draft-ietf-httpbis-client-hints.md
+++ b/draft-ietf-httpbis-client-hints.md
@@ -155,8 +155,8 @@ Accept-CH is a Structured Header {{!I-D.ietf-httpbis-header-structure}}. Its val
 
 For example:
 
-~~~ example
-  Accept-CH: Sec-CH-Example, Sec-CH-Example-2
+~~~ http-message
+Accept-CH: Sec-CH-Example, Sec-CH-Example-2
 ~~~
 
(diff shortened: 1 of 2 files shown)


yoavweiss:
Thanks! I manually marked the examples as `http-message` as part of the AUTH48 process. Might be late for the sf.json changes though

mnot:
No worries about `sf.json` - it's just a hint file. If you still can remove the leading whitespace that would be cool, but it's not the end of the world if you don't.

yoavweiss:
> If you still can remove the leading whitespace that would be cool, but it's not the end of the world if you don't.

Too late for that, I believe.

reschke:
No.

<https://www.rfc-editor.org/auth48/C414> says it's waiting for author approval. By definition, it is not too late.

yoavweiss:
OK. I sent out the revised XML yesterday. Should I simply send another revised version that also includes these changes?

mnot:
Yes, they'll incorporate (they're very nice that way). Please make sure you remove the leading whitespace as per this PR.

yoavweiss:
Will do, thanks! :)

yoavweiss (approved):
LGTM

PR #1324, opened 2020-10-28
Introduce "Schemeful Same-Site" cookies.

miketaylr:
Opening for discussion, per the last interim group meeting.

This should correspond to https://tools.ietf.org/html/draft-west-cookie-incrementalism-01#section-3.3

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -938,26 +938,30 @@ following conditions holds:
 
 ## "Same-site" and "cross-site" Requests  {#same-site-requests}
 
-A request is "same-site" if its target's URI's origin's registrable domain
-is an exact match for the request's client's "site for cookies", or if the
+Two origins, A and B, are considered same-site if the following algorithm 
+returns true:
+
+1.  If A and B are both scheme/host/port triples:
+
+    1.  If A's scheme does not equal B's scheme, return false.
+
+    2.  Let hostA be A's host, and hostB be B's host.
+
+    3.  If hostA equals hostB and hostA's registrable domain is null, return 
+        true.
+
+    4.  If hostA's registrable domain equals hostB's registrable domain and is 
+        non-null, return true.
+
+2.  If A and B are both the same globally unique identifier, return true.
+
+3.  Return false.
+
+Note: The port component of the origins is not considered.
+
+A request is "same-site" if its target's URI's origin
+is same-site with the request's client's "site for cookies", or if the
 request has no client. The request is otherwise "cross-site".
-
-For a given request ("request"), the following algorithm returns `same-site` or
-`cross-site`:
-
-1.  If `request`'s client is `null`, return `same-site`.
-
-    Note that this is the case for navigation triggered by the user directly
-    (e.g. by typing directly into a user agent's address bar).
-
-2.  Let `site` be `request`'s client's "site for cookies" (as defined in the
-    following sections).
-
-3.  Let `target` be the registrable domain of `request`'s current url.
-
-4.  If `site` is an exact match for `target`, return `same-site`.
-
-5.  Return `cross-site`.
 
 The request's client's "site for cookies" is calculated depending upon its
 client's type, as described in the following subsections:
@@ -972,18 +976,18 @@ most likely believes themselves to be interacting. We'll label this domain the
 "top-level site".
 
 For a document displayed in a top-level browsing context, we can stop here: the
-document's "site for cookies" is the top-level site.
+document's "site for cookies" is top-level site's origin.
 
 For documents which are displayed in nested browsing contexts, we need to audit
 the origins of each of a document's ancestor browsing contexts' active documents
 in order to account for the "multiple-nested scenarios" described in Section 4
-of {{RFC7034}}. A document's "site for cookies" is the top-level site if and
-only if the document and each of its ancestor documents' origins have the same
-registrable domain as the top-level site. Otherwise its "site for cookies" is
-the empty string.
+of {{RFC7034}}. A document's "site for cookies" is the top-level site's origin 
+if and only if the document and each of its ancestor documents' origins have the
+same origin as the top-level site. Otherwise its "site for cookies" is an origin
+set to a globally unique identifier.
 
 Given a Document (`document`), the following algorithm returns its "site for
-cookies" (either a registrable domain, or the empty string):
+cookies":
 
 1.  Let `top-document` be the active document in `document`'s browsing context's
     top-level browsing context.
@@ -1000,10 +1004,10 @@ cookies" (either a registrable domain, or the empty string):
     1.  Let `origin` be the origin of `item`'s URI if `item`'s sandboxed origin
         browsing context flag is set, and `item`'s origin otherwise.
 
-    2.  If `origin`'s host's registrable domain is not an exact match for
-        `top-origin`'s host's registrable domain, return the empty string.
+    2.  If `origin` is not same-site to `top-origin`, return an origin set to a 
+        globally unique identifier.
 
-5.  Return `top-origin`'s host's registrable domain.
+5.  Return `top-origin`.
 
 ### Worker-based requests {#worker-requests}
 
@@ -1025,21 +1029,22 @@ document's "site for cookies".
 
 Shared workers may be bound to multiple documents at once. As it is quite
 possible for those documents to have distinct "site for cookie" values, the
-worker's "site for cookies" will be the empty string in cases where the values
-diverge, and the shared value in cases where the values agree.
+worker's "site for cookies" will be an origin set to a globally unique 
+identifier in cases where the values diverge, and the shared origin in cases 
+where the values agree.
 
 Given a WorkerGlobalScope (`worker`), the following algorithm returns its "site
-for cookies" (either a registrable domain, or the empty string):
+for cookies":
 
-1.  Let `site` be `worker`'s origin's host's registrable domain.
+1.  Let `site` be `worker`'s origin.
 
 2.  For each `document` in `worker`'s Documents:
 
     1.  Let `document-site` be `document`'s "site for cookies" (as defined
         in {{document-requests}}).
 
-    2.  If `document-site` is not an exact match for `site`, return the empty
-        string.
+    2.  If `document-site` is cross-site with `site`, return an origin
+        set to a globally unique identifier.
 
 3.  Return `site`.
 
@@ -1056,13 +1061,13 @@ request, and its "site for cookies" will be those defined in
 
 Requests which are initiated by the Service Worker itself (via a direct call to
 `fetch()`, for instance), on the other hand, will have a client which is a
-ServiceWorkerGlobalScope. Its "site for cookies" will be the registrable domain
-of the Service Worker's URI.
+ServiceWorkerGlobalScope. Its "site for cookies" will be the Service Worker's 
+URI's origin.
 
 Given a ServiceWorkerGlobalScope (`worker`), the following algorithm returns its
-"site for cookies" (either a registrable domain, or the empty string):
+"site for cookies":
 
-1.  Return `worker`'s origin's host's registrable domain.
+1.  Return `worker`'s origin.
 
 ## The Set-Cookie Header {#set-cookie}
 
@@ -1445,9 +1450,8 @@ user agent MUST process the cookie as follows:
 15. If the cookie's `same-site-flag` is not "None":
 
     1.  If the cookie was received from a "non-HTTP" API, and the API was called
-        from a context whose "site for cookies" is not an exact match for
-        request-uri's host's registrable domain, then abort these steps and
-        ignore the newly created cookie entirely.
+        from a context whose "site for cookies" is not same-site, then abort 
+        these steps and ignore the newly created cookie entirely.
 
     2.  If the cookie was received from a "same-site" request (as defined in
         {{same-site-requests}}), skip the remaining substeps and continue
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -941,7 +941,9 @@ following conditions holds:
 Two origins, A and B, are considered same-site if the following algorithm
 returns true:
 
-1.  If A and B are both scheme/host/port triples:
+1.  If A and B are both the same globally unique identifier, return true.
+
+2.  If A and B are both scheme/host/port triples:
 
     1.  If A's scheme does not equal B's scheme, return false.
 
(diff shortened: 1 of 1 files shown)


miketaylr:
Can you please review, @chlily1?

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -939,26 +939,30 @@ following conditions holds:
...
+3.  Return false.
 
-5.  Return `cross-site`.
+Note: The port component of the origins is not considered.
+
+A request is "same-site" if its target's URI's origin
+is same-site with the request's client's "site for cookies", or if the
+request has no client. The request is otherwise "cross-site".
 
 The request's client's "site for cookies" is calculated depending upon its
```
chlily1, line 967: Maybe spell out explicitly that the site for cookies is an origin.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -973,18 +977,18 @@ most likely believes themselves to be interacting. We'll label this domain the
...
 For documents which are displayed in nested browsing contexts, we need to audit
 the origins of each of a document's ancestor browsing contexts' active documents
 in order to account for the "multiple-nested scenarios" described in Section 4
-of {{RFC7034}}. A document's "site for cookies" is the top-level site if and
-only if the document and each of its ancestor documents' origins have the same
-registrable domain as the top-level site. Otherwise its "site for cookies" is
-the empty string.
+of {{RFC7034}}. A document's "site for cookies" is the top-level site's origin 
+if and only if the document and each of its ancestor documents' origins have the
+same origin as the top-level site. Otherwise its "site for cookies" is an origin
```
chlily1, line 987: We still want to only require the ancestors to share a (scheme, registrable domain) with the top-level site. Requiring the whole chain of ancestors to be same-origin is too strict.

I think this should say something like

> A document's "site for cookies" is the top-level site's origin if and only if the top-level site's origin is same-site with the document's origin and with each of the document's ancestor documents' origins.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1001,10 +1005,10 @@ cookies" (either a registrable domain, or the empty string):
     1.  Let `origin` be the origin of `item`'s URI if `item`'s sandboxed origin
         browsing context flag is set, and `item`'s origin otherwise.
 
-    2.  If `origin`'s host's registrable domain is not an exact match for
-        `top-origin`'s host's registrable domain, return the empty string.
+    2.  If `origin` is not same-site to `top-origin`, return an origin set to a
```
chlily1, line 1008: I dunno, is it "same-site to" or "same-site with"?

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1026,21 +1030,22 @@ document's "site for cookies".
 
 Shared workers may be bound to multiple documents at once. As it is quite
 possible for those documents to have distinct "site for cookie" values, the
```
chlily1, line 1032: nit: preexisting typo, but mind fixing this? "site for cookies"

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1026,21 +1030,22 @@ document's "site for cookies".
...
-worker's "site for cookies" will be the empty string in cases where the values
-diverge, and the shared value in cases where the values agree.
+worker's "site for cookies" will be an origin set to a globally unique 
+identifier in cases where the values diverge, and the shared origin in cases 
+where the values agree.
```
chlily1, line 1035: Previously, this would be the shared site (registered domain), whereas now they need to all be exactly the same origin. (The algorithm below seems fine though.)

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1026,21 +1030,22 @@ document's "site for cookies".
...
+1.  Let `site` be `worker`'s origin.
 
 2.  For each `document` in `worker`'s Documents:
 
     1.  Let `document-site` be `document`'s "site for cookies" (as defined
         in {{document-requests}}).
 
-    2.  If `document-site` is not an exact match for `site`, return the empty
-        string.
+    2.  If `document-site` is cross-site with `site`, return an origin
```
chlily1, line 1047: Don't think we actually ever defined what it means for two origins to be cross-site, so maybe say "not same-site with". (Or define cross-site on line 942)

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1440,9 +1445,8 @@ user agent MUST process the cookie as follows:
 15. If the cookie's `same-site-flag` is not "None":
 
     1.  If the cookie was received from a "non-HTTP" API, and the API was called
-        from a context whose "site for cookies" is not an exact match for
-        request-uri's host's registrable domain, then abort these steps and
-        ignore the newly created cookie entirely.
+        from a context whose "site for cookies" is not same-site, then abort
```
chlily1, line 1448: same-site with what? (I don't actually know how best to phrase this...)

I think we need to resolve the broader question of how to define SameSite behavior for non-HTTP APIs.

draft-ietf-httpbis-rfc6265bis.md
miketaylr: The next sentence refers to determining the "site for cookies" depending on the type (documents & dedicated workers: origin, shared workers: maybe origin, maybe unique identifier, service worker: service worker's URI origin). 

So I think we're fine leaving it as is.

miketaylr: Actually, I see what you mean now.

draft-ietf-httpbis-rfc6265bis.md
miketaylr: Let's go with "with", since that's used elsewhere in the document.

draft-ietf-httpbis-rfc6265bis.md
miketaylr: Do you have any concerns with this text as-is, or just pointing out the difference?

chlily1: Sorry for being unclear. I was pointing out that this doesn't match the actual algorithm below. I think we can just replace "where the values diverge" with "where the values are not all same-site with the worker's origin", and replace "the shared origin" with "the worker's origin", in cases where the values are all same-site with it.

miketaylr: Ah, now I see. Thanks!

draft-ietf-httpbis-rfc6265bis.md
miketaylr: > I think we need to resolve the broader question of how to define SameSite behavior for non-HTTP APIs.

Yeahhhhhhhh.

Until then I think what we're trying to capture is something like:

> "the origin of the calling API's "site for cookies" is same-site with the request-uri's origin?

We could say something like that, or possibly:

> If the cookie was received from a "non-HTTP" API, and the API was not called from a first-party context, then abort these steps and ignore the newly created cookie entirely.

"first-party context" *is* used in the RFC elsewhere, but it's not defined. It is in https://tools.ietf.org/html/draft-west-first-party-cookies-04#section-2.1... 🤔 

It seems like a lot of machinery to bring in just to use it here, unless we want to take that route for the further non-HTTP API route.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -939,26 +939,30 @@ following conditions holds:
...
 
-A request is "same-site" if its target's URI's origin's registrable domain
-is an exact match for the request's client's "site for cookies", or if the
-request has no client. The request is otherwise "cross-site".
+Two origins, A and B, are considered same-site if the following algorithm
+returns true:
 
-For a given request ("request"), the following algorithm returns `same-site` or
-`cross-site`:
+1.  If A and B are both scheme/host/port triples:
```
mikewest, line 945: Can we use the same "[tuple origin](https://html.spec.whatwg.org/#concept-origin-tuple)" terminology as HTML here?

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -939,26 +939,30 @@ following conditions holds:
...
-    following sections).
+    3.  If hostA equals hostB and hostA's registrable domain is null, return
+        true.
 
-3.  Let `target` be the registrable domain of `request`'s current url.
+    4.  If hostA's registrable domain equals hostB's registrable domain and is
+        non-null, return true.
 
-4.  If `site` is an exact match for `target`, return `same-site`.
+2.  If A and B are both the same globally unique identifier, return true.
```
mikewest, line 957: It seems reasonable to me to match HTML's ordering from "[same-site](https://html.spec.whatwg.org/multipage/origin.html#same-site)" and "[schemelessly same-site](https://html.spec.whatwg.org/multipage/origin.html#schemelessly-same-site)" here (e.g. exit early for "the same opaque origin", rather than processing tuple origins first).

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -973,18 +977,18 @@ most likely believes themselves to be interacting. We'll label this domain the
 "top-level site".
 
 For a document displayed in a top-level browsing context, we can stop here: the
-document's "site for cookies" is the top-level site.
+document's "site for cookies" is top-level site's origin.
```
mikewest, line 980: Above (on new line 976), we define "top-level site" as the registrable domain of the URI displayed to the user, which means that calculating its "origin" isn't really a thing you can do (no scheme, no port, etc). I'd suggest that we need to change that term entirely given the new usage. You don't actually want it to be a "site" at all, you now want something like the origin of top-level document. Perhaps doing same-site comparisons against the "top-level origin" would be reasonable?

mikewest (commented):
A few small comments. Also, please add a changelog entry.

draft-ietf-httpbis-rfc6265bis.md
sbingler: Hi Mike,
Has there been any changes on the IETF/WHATWG term differences? This same point came up during the PR into cookie inc. https://github.com/mikewest/cookie-incrementalism/pull/3#discussion_r390675247

mikewest: Ha! Ok. I still agree with myself from earlier in the year. :)

Ping @mnot @tfpauly. What would y'all like the working group's policy to be vis a vis these terms? I'd like for us to end up using the same terms with the same definition regardless of venue.

draft-ietf-httpbis-rfc6265bis.md
sbingler: Seems reasonable to me too. I can't remember why I originally had ordered like this.

draft-ietf-httpbis-rfc6265bis.md
reschke: The new core specs have their definition of "origin". See <https://greenbytes.de/tech/webdav/draft-ietf-httpbis-semantics-12.html#origin>.

reschke: Do HTML's and "URL"'s align? And do they differ from what HTTP says?

mikewest: >  "URL"

Hey there, scare quotes! :) You know this, but I'm referring to https://url.spec.whatwg.org/#origin for the latter, and https://html.spec.whatwg.org/multipage/origin.html#concept-origin for the former. URL's definition of a "URL's origin" depends upon HTML's definition of an opaque origin, and fulfills HTML's definition of a tuple origin.

> And do they differ from what HTTP says?

They do a few things that https://greenbytes.de/tech/webdav/draft-ietf-httpbis-semantics-12.html#origin doesn't, most notably giving guidance for scheme-based rules beyond `http` and `https` (`blob` and `data` are particularly relevant), and expanding upon RFC6465's "globally unique identifier" concept. Cookies need to have consistent rules for both.

miketaylr:
Thanks for the review comments! I'm going to let this bake for a few days (and let some folks respond), and eat a lot of pumpkin pie in the meantime. Looking to pick it up early next week.

draft-ietf-httpbis-rfc6265bis.md
mikewest: Ping @mnot and @tfpauly for the conversation here and below ("URL", etc).

mikewest:
I'd suggest that we try to get this landed, and then spin off a new -07. WDYT?

miketaylr:
> I'd suggest that we try to get this landed, and then spin off a new -07. WDYT?

SGTM! My goal is to address feedback before EOW (and I'll file a new follow-up issue to hash out origin vs tuple and we can make changes the next go around).

draft-ietf-httpbis-rfc6265bis.md
miketaylr: > Perhaps doing same-site comparisons against the "top-level origin" would be reasonable?

Yep, makes sense. We already use the top-level browsing context's active document's origin in the "site for cookies" algorithm, so we can define a "top-level origin" as the same thing in the prose here.

draft-ietf-httpbis-rfc6265bis.md
miketaylr: I think we can do this more simply by passing in the "site for cookies" of the context (which I take to mean a document...) and the "site for cookies" of the "top-level origin" into the same-site algorithm.

miketaylr:
PTAL @mikewest @chlily1 

(I'm punting on the "origin tuple" rec -- I filed #1337 for folks to align on that)

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1445,9 +1450,10 @@ user agent MUST process the cookie as follows:
 15. If the cookie's `same-site-flag` is not "None":
 
     1.  If the cookie was received from a "non-HTTP" API, and the API was called
-        from a context whose "site for cookies" is not an exact match for
-        request-uri's host's registrable domain, then abort these steps and
-        ignore the newly created cookie entirely.
+        from a browsing context's active document whose "site for cookies" is
+        not same-site with the top-level browsing context's active document's
+        "site for cookies", then abort these steps and ignore the newly created
```
chlily1, line 1455: Any reason not to say "top-level origin" in place of 'the top-level browsing context's active document's "site for cookies"'?

miketaylr: I was told I got paid by word-count.... 

(I can change it!)

chlily1:
> PTAL @mikewest @chlily1

LGTM! Thanks!

sbingler (changes requested):
I can't seem to comment on the line but 654 isn't correct any longer.

miketaylr:
> I can't seem to comment on the line but 654 isn't correct any longer.

Great catch, let me fix that up.

mikewest:
Thanks for punting the terminology conversation out to a separate bug. LGTM2; I'll land this and spin out an -07 draft.

miketaylr:
Thanks all!

PR #1323, opened 2020-10-28
Require "Secure" for "SameSite=None".

miketaylr:
Opening for discussion, per the last interim group meeting.

This should correspond to https://tools.ietf.org/html/draft-west-cookie-incrementalism-01#section-3.2

PR #1325, opened 2020-10-28
Treat cookies as "SameSite=Lax" by default.

miketaylr:
Opening for discussion, per the last interim group meeting.

This should correspond to https://tools.ietf.org/html/draft-west-cookie-incrementalism-01#section-3.1

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -661,7 +661,7 @@ with same-site requests, and with "cross-site" top-level navigations, as
 described in {{strict-lax}}. If the value is "None", the cookie will be sent
 with same-site and cross-site requests. If the "SameSite" attribute's value is
 something other than these three known keywords, the attribute's value will be
-treated as "None".
+subject to a default enforcement mode that is equivalent to "Lax".
 
 The "SameSite" attribute affects cookie creation as well as delivery. Cookies
 which assert "SameSite=Lax" or "SameSite=Strict" cannot be set in responses to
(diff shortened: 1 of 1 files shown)


miketaylr:
Can you please review, @chlily1?

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1254,10 +1254,10 @@ attribute-name of HttpOnly and an empty attribute-value.
 If the attribute-name case-insensitively matches the string "SameSite", the
 user agent MUST process the cookie-av as follows:
 
-1.  Let `enforcement` be "None".
+1.  Let `enforcement` be "Default".
 
-2.  If cookie-av's attribute-value is a case-insensitive match for "Strict",
-    set `enforcement` to "Strict".
+2.  If cookie-av's attribute-value is a case-insensitive match for "None",
+    set `enforcement` to "None".
```
chlily1, line 1261: We lost "Strict" here! I think you need to insert the "None" step and renumber the subsequent steps, rather than replacing the existing step 2.

miketaylr: Whoops. 😅  Nice catch!

miketaylr:
Thanks for the review! I fixed the bug you found. cc @mikewest if you also want to have a look (or merge).

chlily1:
> Thanks for the review! I fixed the bug you found. cc @mikewest if you also want to have a look (or merge).

Thanks, LGTM

chlily1:
Sorry, one more thing,

Could you please update the non-normative section about SameSite on line 664 to note that if the SameSite attribute value is not specified or is something other than the 3 recognized keywords, it will be subject to a default enforcement mode that is equivalent to Lax?

mikewest:
This looks right to me. Can you also update the changelog in the appendix with this new behavior? Happy to merge it once Lily is happy; we should then be able to spin out an -07 draft since the CfA went through (@mnot to confirm from a process perspective).

miketaylr:
OK, done. For @chlily1 & @mikewest to look at again.  🦃

miketaylr:
Note: this should land before https://github.com/httpwg/http-extensions/pull/1323 (assuming there aren't any further changes).

chlily1:
> OK, done. For @chlily1 & @mikewest to look at again. 🦃

LGTM, thanks!

mikewest:
As per @mnot and @tfpauly's note on the mailing list, there's consensus to adopt the incrementalism draft, so I'll start landing these. Thanks for your patience! :)

PR #1329, opened 2020-11-05
udpated description to reflect draft current status

jricher:


OR13 (approved):

draft-ietf-httpbis-message-signatures.md
```diff
@@ -26,7 +26,7 @@ author:
     region: WA
     code: 98108-1226
     country: United States of America
-
+    role: editor
```
OR13, line 29: assume this means you can merge PRs? : )

jricher: No, this is just a designation within the IETF -- often documents have multiple "authors" but a single designated "editor" to lead the unruly crew. I noticed @richanna 's entry didn't have that flag set so I grabbed that change on the way by since we're updating the framing text.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -80,8 +80,7 @@ This document describes a mechanism for creating, encoding, and verifying digita
...
-
+This draft was originally largely based on draft-cavage-http-signatures-12, but has since diverged. [The community of that I-D](https://github.com/w3c-dvcg/http-signatures/issues?page=2&q=is%3Aissue+is%3Aopen) and the authors have identified several issues with the current text, many of which are noted within the text.  Additionally, the authors have identified a number of new features that are required in order to support additional use cases.
```
mnot, line 83:
```suggestion
This work was originally based on draft-cavage-http-signatures-12, but has since diverged from it, to reflect discussion since adoption by the HTTP Working Group. In particular, it addresses issues that have been identified, and adds features to support new use cases. It is a work-in-progress and not yet suitable for deployment.
```

It's confusing to refer to a different draft's issues list; I'd just do something like this.

ioggstream (approved):
LGTM

Issue #773, opened 2019-02-25
Correct behaviour for a POST-redirect-GET when using SameSite=Lax


pmhsfelix:
In the context of <https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-02>, 
I would like to clarity what is the intended behaviour for the `GET` request in the following scenario, when using a cookie with `SameSite=Lax`:
- `POST` to `https://example.com/path-1` triggered by a cross-site document.
- Response to the `POST` is a redirect status code to `https://example.com/path-2`.
- `GET` to `https://example.com/path-2`.

When using `SameSite=Lax` it is rather clear that the first request (`POST`) will not contain the cookie, since it is using a non-safe method.
However, the behaviour for the `GET` is not completely clear because:
- It is a safe method.
- However, it was triggered by a non-safe cross-site request.

I've tested it with two different browsers and obtained two distinct behaviours:
- Safari (macOS Mojave) - the cookie is not sent on the `GET` request.
- Firefox (`65.0.1`) - the cookie is sent on the `GET` request.

chlily1:
There's now a sentence in the spec clarifying this issue. (Still needs WPT coverage.)

In the above scenario, the Lax cookie should be sent on the `GET` request. I believe Chrome sends the cookie (tested on Chrome 86), and Safari does as well (tested on 13.1.3).

chlily1:
@mnot Since we have clarified this in the spec and have alignment between Chrome, Firefox, and Safari, can we close this issue?

mnot:
Yes (see [our working procedures](https://httpwg.org/CONTRIBUTING.html#resolving-issues) for details).

miketaylr:
@mnot Can @chlily1 (and @englehardt) be added as collaborators to the repo so they can close issues (in accordance with working procedures)?

PR #1331, opened 2020-11-24
RFC 6265bis: SameSite -- clarify request method on redirects

chlily1:
This clarifies the question raised in issue #773 about SameSite cookies on redirects.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1278,7 +1278,9 @@ drop-in mechanism that mitigates the risk of CSRF attacks, developers may set
 the `SameSite` attribute in a "Lax" enforcement mode that carves out an
 exception which sends same-site cookies along with cross-site requests if and
 only if they are top-level navigations which use a "safe" (in the {{RFC7231}}
-sense) HTTP method.
+sense) HTTP method. (Note that a request's method is updated on redirects, such
+that "safe"ness is determined based on the method of the most recent redirect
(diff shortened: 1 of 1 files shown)


chlily1:
@mikewest PTAL, thanks!

reschke:
The request method is only updated for *some* of the redirects, and only if it is POST.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1278,7 +1278,9 @@ drop-in mechanism that mitigates the risk of CSRF attacks, developers may set
 the `SameSite` attribute in a "Lax" enforcement mode that carves out an
 exception which sends same-site cookies along with cross-site requests if and
 only if they are top-level navigations which use a "safe" (in the {{RFC7231}}
-sense) HTTP method.
+sense) HTTP method. (Note that a request's method is updated on redirects, such
+that "safe"ness is determined based on the method of the most recent redirect
+hop.)
```
mikewest, line 1283: How about the following change to incorporate @reschke's suggestion:

```suggestion
sense) HTTP method. (Note that a request's method may be changed from POST
to GET for some redirects (see sections 6.4.2 and 6.4.3 of {{RFC7231}}); in these
cases, a request's "safe"ness is determined based on the method of the current
redirect hop.)
```

mikewest:
The change generally looks fine. I'd suggest that adding tests would be helpful to ensure we guide implementations towards interoperable behavior.

Issue #1221, opened 2020-06-18
duplicate digest-algorithms in field value


ioggstream:
## I expect

To clarify usage of duplicate algorithms, eg. with different parameters

## Instead

It is not specified

## Note

RFC3230 does not forbid duplicate digest-algorithm in field value, because the receiver MAY ignore one or more representation-digest eg:

```
Digest: sha-256=bxxx.., sha-256=axxx...
```

[multihash](https://multiformats.io/multihash/#examples) encodes the algorithm and its parameters inside the string, and this is a use-case for duplicate digest-algorithms. This behavior is not in line with S-F though. 

```
Digest: mh=fafafaf, mh=fbfbfbf
```

In case of digest-algorithms with parameters, this is more complex. The following valid example shows a duplicate digest-algorithm `bs` with two different parameter values.

```
Digest: bs=axxx; p=1,    bs=bxxx; p=2
```

Do we need an example for that?
Which is a suitable way to define it using structured fields/headers?

@LPardue

LPardue:
I think this is tightly coupled with the parameter discussion on #850.

ioggstream:
Forbidding parameters would fix this.

ioggstream:
@msporny do you see any drawback in this? PTAL :)

msporny:
Multihash doesn't have any parameter use cases that I can think of since you're supposed to encode all of those parameters via the multihash header. That is, it's always one multibase encoded multihash-header plus value. Multihash specifically would be fine w/ forbidding parameters.

It does raise the question on what happens if multiple multihash values are expressed. They would probably all be valid... one is a sha-256, one is a blake2b-32, etc. So, there is a strong use case for multiple "duplicate" "mh" digest algorithms (without parameters).

I can't comment on other hash expression schemes as it's not an area with which I'm an expert.

Does that help, @ioggstream?

ioggstream:
@msporny thanks for your feedback!

> Multihash specifically would be fine w/ forbidding parameters.

Ok!

> ... if multiple multihash values are expressed. They would probably all be valid

That is fine with the current Digest definition, but does not fit with structured-fields where every token is an unique keyword.
Seems that a future transition to a S-F scheme has many points to tackle...

ioggstream:
As of now it's ok having duplicates, as that depends on the digest-algorithm. We could at most spend a couple of words on the subject in the security considerations (eg if one receives multiple values for  `sha-256`, ..)

PR #1333, opened 2020-12-01
Fix: 1221. Non-normative hints for duplicate digest-algorithms.

ioggstream:
## This PR

Provides non-normative feedback on duplicate digest-algorithms

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1111,6 +1111,7 @@ A receiver is permitted to ignore any representation-data-digest value,
 so validation of duplicates is left as an implementation decision.
 Endpoints might select all, some or none of the values for checksum comparison and,
 based on the intersection of those results, conditionally pass or fail digest validation.
+
 # IANA Considerations
 
 ## Establish the HTTP Digest Algorithm Values {#iana-digest-algorithm-registry}
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1100,11 +1100,7 @@ which may be wasted if the receiver ignores them (see {{digest}}).
 
 ### Duplicate digest-algorithm in field value
 
-As the receiver is free to ignore any representation-data-digest value (see {{digest}}),
-this is especially true when `Digest` contains duplicate digest-algorithms.
-The semantic of the digest-algorithm could impact on the decision, though.
-
-For example, the receiver of the following field
+An endpoint might receive multiple representation-data-digest values (see {{digest}}) that use the same digest-algorithm with different or identical digest-values. For example:
 
 ~~~ example
 Digest: sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=,
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1114,33 +1114,6 @@ Digest: sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=,
 could consider it suspicious and ignore it, use only one of its values, or both of them.
 
 
-When defining a digest-algorithm that requires to support multiple values
-its encoding could be defined accordingly
-to simplify the work of the receiver.
-
-An hypothetical digest-algorithm based on a parameter could have multiple values
-as illustrated in the following example showing two checksums of the same representation-data.
-
-~~~ example
-representation-data: '{"hello": "world"}'
-shake-128/5: <\x1c?W\xda
-shake-128/10: <\x1c?W\xda\xafA8_\xd6
-~~~
-
-This digest-algorithm could define its encoding accordingly,
-for example concatenating the shake-128 length and the checksum,
-
-~~~ example
-shake-128/5: 5/<\x1c?W\xda
-shake-128/10: 10/<\x1c?W\xda\xafA8_\xd6
-~~~
-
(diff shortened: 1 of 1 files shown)


LPardue:
Thanks for putting this together. However, I'm not sure if all the text really helps address #1221 now that we obsoleted parameters.

An endpoint that receives a digest field with more than one `digest-algorithm` of the same type has a couple of options. Ignore it altogether (no validation); or do validation and pick the first, pick the last, pick them all, etc. Obviously if you have the same `digest-algorithm` and different `digest-values` there is a chance that your pick could be wrong and validation might fail. We already describe a similar case wrt SRI, saying `This document specifies no requirements or guidance for user agents that experience such cases.`. I think what to do is best left as an implementation choice.

I don't see the value in talking about hypotheticals. We removed parameters because of lack evidence of usage, if someone wants to go that route they can design their own solution, which is what we recommend in Section 8 already.

So in summary, I think at most all we can do is document that it is an implementation decision to pick which entries in the possible set of digests should be verified.

ioggstream:
@LPardue I wrote those considerations thinking about multihash, which would use a single digest-algorithm token for different crypto algorithms, but I agree I went too far :) OTOH I'd save the first lines, as I think it's worth notifying them to the implementors.

Could you please `suggest` and edit to the PR, 
eg. removing https://github.com/httpwg/http-extensions/pull/1333/files#diff-de34152d297a9eb29b85b24962949ba2e309478a5c32e1d94fd4c3ee27b4584dR1117-R1142
?

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1098,6 +1098,49 @@ or by sending multiple representation-data-digest values from which the receiver
...
+shake-128/5: 5/<\x1c?W\xda
+shake-128/10: 10/<\x1c?W\xda\xafA8_\xd6
+~~~
+
+and then encoding it using the `sf-list` notation defined in {{?I-D.ietf-httpbis-header-structure}}.
+
+~~~ example
+Digest: shake-128=(:NS88HD9X2g==: :MTAvPBw/V9qvQThf1g==:)
+~~~
+
```
LPardue, line 1143:
```suggestion
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1098,6 +1098,49 @@ or by sending multiple representation-data-digest values from which the receiver
...
+As the receiver is free to ignore any representation-data-digest value (see {{digest}}),
+this is especially true when `Digest` contains duplicate digest-algorithms.
+The semantic of the digest-algorithm could impact on the decision, though.
+
+For example, the receiver of the following field
+
+~~~ example
+Digest: sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=,
+        sha-256=47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=
+~~~
```
LPardue, line 1112:
```suggestion
An endpoint might receive multiple representation-data-digest values (see {{digest}}) that use the same digest-algorithm with different or identical digest-values. For example:

~~~ example
Digest: sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=,
        sha-256=47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=
~~~
```

A receiver is permitted to ignore any representation-data-digest value, so validation of duplicates is left as an implementation decision. Endpoints might select all, some or none of the values for checksum comparison and, based on the intersection of those results, conditionally pass or fail digest validation.

LPardue (commented):
WDYT?

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1098,6 +1098,18 @@ or by sending multiple representation-data-digest values from which the receiver
...
+An endpoint might receive multiple representation-data-digest values (see {{digest}}) that use the same digest-algorithm with different or identical digest-values. For example:
+
+~~~ example
+Digest: sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=,
+        sha-256=47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=
+~~~
+
+could consider it suspicious and ignore it, use only one of its values, or both of them.
+
+
```
ioggstream, line 1112:
```suggestion
A receiver is permitted to ignore any representation-data-digest value,
so validation of duplicates is left as an implementation decision.
Endpoints might select all, some or none of the values for checksum comparison and,
based on the intersection of those results, conditionally pass or fail digest validation.
```

ioggstream: @LPardue is this it?

LPardue: yep

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1098,6 +1098,19 @@ or by sending multiple representation-data-digest values from which the receiver
...
+~~~ example
+Digest: sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=,
+        sha-256=47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=
+~~~
+
+A receiver is permitted to ignore any representation-data-digest value,
+so validation of duplicates is left as an implementation decision.
+Endpoints might select all, some or none of the values for checksum comparison and,
+based on the intersection of those results, conditionally pass or fail digest validation.
```
LPardue, line 1113:
```suggestion
based on the intersection of those results, conditionally pass or fail digest validation.

```

LPardue (approved):
One whitespace nit but otherwise LGTM

Issue #852, opened 2019-07-10
Create a threat model for `Digest` header.


ioggstream:
## I wish

A threat model for `Digest`.

## Questions

- is this useful/required?
- should this go into Security Considerations or in another document?

ioggstream:
# Security issues for HTTP Digest

## Actors

- User
- User-Agent: see RFC7230
- Intermediaries that do not do TLS termination
- Intermediaries that do TLS termination.

- Application: the application actually running the service

## Attacker capabilities

1. Intermediaries can alter HTTP messages

2. Intermediaries that do transport-layer termination can alter HTTP messages

3. The application can erroneously or purposedly alter the message

4. User can forge messages with the same checksum values

## Attacker non-capabilities

# Attacks and mitigations

## Alteration of HTTP message headers

### The attack

An intermediary can alter an HTTP header by chance or purpose.

### Proposed mitigation

Use an end-to-end transport-layer security mechanism.


## Alteration of HTTP representation-metadata headers

### The attack

A TLS-terminator intermediary can alter representation-metadata by chance or purpose.

### Proposed mitigation

Use a signature mechanism that covers both `Digest` and the representation-metadata.

## Send a Digest header with a different payload

### The attack

A TLS-terminator intermediary can replace the representation-data with another
one having the same digest-algorithm.

A malicious User can forge two messages with:

  - different representation-data
  - same Digest header values
  
send an HTTP message with the first representation-data
then pretend he sent the one with the second representation-data.

A malicious Server could implement a similar behavior.

### Proposed mitigation

Use a digest-algorithm that is not subject to collision (eg. sha-256),
or refuse to use digest-algorithms subect to collision.

## Server resource exhaustion for calculating Digest on partial representations

### The attack

A malicious User could exhaust server resources via:

1- reconstruction of the complete representation
2- digest calculation of the complete representation
3- forcing the server to evaluate many different digest-algorithms for a given representation

### Proposed mitigation

If you are exposed to this kind of attack because 
reconstructing complete representations of your 
objects is computationally intensive, avoid using
Digest with mechanisms like Range-Requests and PATCH.

If you support multiple digest-algorithms:

- limit the supported digest-algorithms to the
  ones you really care for;
- establish a policy for validating only the digest
  using the stronger digest-algorithm.

## Confidence in recipient's Digest validation 

### The attack

The recipient may skip Digest header validation by chance or purpose
and still be compliant to the specification: as explained in {#digest-header} 
[...]

PR #1335, opened 2020-12-02
Fix: #852. Mention resource consumption.

ioggstream:
## This PR

includes security considerations from #852

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1176,6 +1176,12 @@ so validation of duplicates is left as an implementation decision.
 Endpoints might select all, some or none of the values for checksum comparison and,
 based on the intersection of those results, conditionally pass or fail digest validation.
 
+## Resource exhaustion
+
+`Digest` validation consumes computational resources.
+In order to avoid resource exhaustion, implementations can restrict
+validation of the algorithm types, number of validations, or the size of content.
+
 # IANA Considerations
 
 ## Establish the HTTP Digest Algorithm Values Registry {#iana-digest-algorithm-registry}


draft-ietf-httpbis-digest-headers.md
```diff
@@ -1102,6 +1102,18 @@ or by sending multiple representation-data-digest values from which the receiver
 Endpoints are advised that sending multiple values consumes resources,
 which may be wasted if the receiver ignores them (see {{digest}}).
 
+## Resource exhaustion
+
+`Digest` validation consumes computational resources.
+To mitigate the risk of resource exhaustion, it is advisable to:
+
+- limit the set of digest-algorithms to the ones that
+  will be effectively used;
```
LPardue, line 1111: who "limits" what's "used"? Is this a intended to say a receiver should limit the set of algorithms that it will validate for any given Digest field, or are you saying something different?

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1102,6 +1102,18 @@ or by sending multiple representation-data-digest values from which the receiver
...
+- establish policies to use only the stronger digest-algorithm
+  provided;
```
LPardue, line 1113: If I'm trying to avoid resource consumption I might prefer a weaker algorithm...

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1102,6 +1102,18 @@ or by sending multiple representation-data-digest values from which the receiver
...
+- take adequate precautions when validating the checksum
+  on messages conveying partial representations  (see Section 5 of {{PATCH}}).
+
```
LPardue, line 1116: Is it worth mentioning that validation might be proportional to resource size, and therefore restrictions on size can be a mitigation?

draft-ietf-httpbis-digest-headers.md
ioggstream: Example: if I just expect to use "sha512", I could configure the validation middleware to ignore other values.

draft-ietf-httpbis-digest-headers.md
ioggstream: my point is about wasting resources to process unneeded checksums eg.  which are purposedly added to waste receiver's cpu time.

draft-ietf-httpbis-digest-headers.md
ioggstream: I'd write something on large payloads, probably MICE contains something on that... I'll have a look.

ioggstream:
@LPardue probably those considerations are redundant: implementers know that computing digest has a cost. Do you think it's worth spending more time to land this PR or we can close it?

LPardue:
I'd go with a simpler statement such as

```
`Digest` validation consumes computational resources. In order to avoid resource exhaustion, implementations can restrict validation of the algorithm types, number of validations, or the size of content.
```

or say nothing.

ioggstream:
@LPardue this PR contains your proposal.

LPardue (approved):
Thank, this works for me. Let's merge and we can iterate later if anyone thinks of something to add.

PR #1339, opened 2020-12-08
Remove leftover paragraph about unknown SameSite values

FiloSottile:
It looks like this was forgotten in #1325, when unknown values were switched from None to Default.

/cc @mikewest this caused us some confusion over in Go, if you have time I'd appreciate hearing what you think we should do in golang/go#36990. Thank you!


mikewest (approved):
Thanks! LGTM.

PR #1341, opened 2020-12-15
digest-headers: ref http core -13

reschke:
also changes "payload body" to "payload data"


ioggstream:
Thanks for your great work!

PR #1347, opened 2020-12-18
add priority field

LPardue:
I think this was our decided path forward, I still had some garbage in my local work copy see https://github.com/httpwg/http-extensions/pull/1344/files#r546142029

### DIFF
--- a/sf.json
+++ b/sf.json
@@ -5,6 +5,7 @@
     "variants": "dict",
     "signature": "dict",
     "signature-input": "dict",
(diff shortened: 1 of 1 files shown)


mnot:
Why is this labeled `meta`?

LPardue:
This is a PR for your json file, it relates to prioirites but is not a PR to the priorities spec. I labelled it as meta to get your attention, which seems to have worked.

mnot:
So it did... so it did.

Issue #889, opened 2019-08-08
SameSite cookies: Cross-site top-level navigation followed by redirect back to the first party


chlily1:
(Notation: => is a navigation, -> is a redirect, "A" is the first party site which has SameSite cookies set on it.)

Currently, top-level A=>B->A results in sending both Strict and Lax cookies on the redirect back to A, as reflected in this WPT testcase: https://github.com/web-platform-tests/wpt/blob/master/cookies/samesite/form-get-blank.html#L58

This should ideally send Lax (but not Strict) cookies, because the analogous top-level navigation scenario A=>B=>A would result in sending only Lax but not Strict cookies to A on the last navigation.

kiding:
I'd like to weigh in on resolving this issue since the browser implementations have diverged. 

Following the A=>B->A example, the Chrome/Chromium test cases (implemented by @chlily1, exported to web-platform-tests) assume both A's are "strictly same-site." (See [`wpt@2379191`](https://github.com/web-platform-tests/wpt/blob/2379191282384315dfa83973551badbc997ce683/cookies/samesite/form-get-blank.https.html#L58))

On the Safari/WebKit side, the test cases assume both A's are _not strict_. (See [`webkit@d615092`](https://github.com/WebKit/webkit/blob/d61509279f6864b7c81e2219f87c3d1130549c32/LayoutTests/http/tests/cookies/same-site/fetch-after-top-level-cross-origin-redirect-expected.txt#L7)) This behavior (which I personally prefer) is consistent with the A=>B=>A scenario. Firefox/Gecko behaves like WebKit. (See [`gecko-dev@15cc802`](https://github.com/mozilla/gecko-dev/blob/15cc8023f7861e4bb56cd43f179d5c45937d7c19/dom/security/test/general/test_same_site_cookies_redirect.html#L38-L43)) There even is a 2-year-old [bug 1456408](https://bugzilla.mozilla.org/show_bug.cgi?id=1456408) that's basically the same as this issue.

FYI, A simple demo page to test with is available at https://brass-tremendous-marmoset.glitch.me.

englehardt:
There are actually three conditions for which we'd want to consider these types of redirects:

1. User types B.com into the address bar which redirects to A.com. (request's client is null, so visit to A.com is considered strictly same-site)
2. User clicks a link on A.com navigating to B.com which then redirects back to A.com. (request's initiator is A.com, so final visit to A.com is considered strictly same-site)
[...]

Issue #1337, opened 2020-12-04
[6265bis] Align on HTML terminology for origins


miketaylr:
It seems desirable to align with HTML and use [tuple origin](https://html.spec.whatwg.org/multipage/origin.html#concept-origin-tuple) instead of "scheme/host/port triple" (from https://tools.ietf.org/html/rfc6454).

See https://github.com/httpwg/http-extensions/pull/1324/files#r530145898 and https://greenbytes.de/tech/webdav/draft-ietf-httpbis-semantics-12.html#origin for related discussion.

cf. https://github.com/mikewest/cookie-incrementalism/pull/3#discussion_r390775633

(edit: updated to point to multi-page version of HTML, apologies for linking the single-page one!)

reschke:
Unless there's a problem with it, it might be better to use http-core's definition (*if* there is a problem with it, now would be a good time to raise an issue)

royfielding:
The tuple origin concept is an extension (or limitation) of permissions associated with same origin beyond what is considered the origin. HTTP core only defines origin, which is consistent with the current text in WHATWG html, and would still be consistent if you refer to the html tuple concept in Cookies.

miketaylr:
I think it would also be a win for clarity in 6265bis if in place of `globally unique identifier` we used HTML's [opaque origin](https://html.spec.whatwg.org/multipage/origin.html#concept-origin-opaque). AFAICT, http-core doesn't have a mechanism defined for this.

miketaylr:
It sounds like there are no objections, so I will write a PR (once #1348 is merged, to avoid conflicts for some of the usage of "globally unique identifier".

englehardt:
> It sounds like there are no objections, so I will write a PR (once #1348 is merged, to avoid conflicts for some of the usage of "globally unique identifier".

SGTM

PR #1348, opened 2020-12-22
Consider redirects when defining same-site

englehardt:
Addresses #889.

Note that this change also simplifies the definition by referencing the [fetch definition](https://html.spec.whatwg.org/multipage/origin.html#same-site) of when two origins are considered same-site. The text is the current version is effectively a copy-paste of that definition.


englehardt:
PTAL @chlily1 @miketaylr @annevk

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -939,30 +939,17 @@ following conditions holds:
 
 ## "Same-site" and "cross-site" Requests  {#same-site-requests}
 
-Two origins, A and B, are considered same-site if the following algorithm
-returns true:
+Two origins are same-site if they satisfy the same-site criteria defined in
```
chlily1, line 942: This seems to be defined in html, not fetch (and there it's spelled "same site" without a hyphen). https://html.spec.whatwg.org/multipage/origin.html#same-site

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -939,30 +939,17 @@ following conditions holds:
...
+{{FETCH}}. A request is "same-site" if the following criteria are true:
 
-1.  If A and B are both the same globally unique identifier, return true.
+1.  The request is not the result of a cross-site redirect. That is,
+    the origin of every url in the request's url list is same-site with the
+    request's current url's origin.
```
chlily1, line 947: Should "current url" be separately called out as a term that's defined in fetch?

draft-ietf-httpbis-rfc6265bis.md
englehardt: Good catch! I've updated the text here to fix both the reference and spelling issue.

draft-ietf-httpbis-rfc6265bis.md
englehardt: "current url" is defined in [here](https://github.com/httpwg/http-extensions/blame/master/draft-ietf-httpbis-rfc6265bis.md#L330), which I think is sufficient?

chlily1: Oops, I missed that we already had it there. Thanks!

chlily1:
Thanks, this LGTM!

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -939,30 +939,17 @@ following conditions holds:
...
+{{HTML}}. A request is "same-site" if the following criteria are true:
 
-1.  If A and B are both the same globally unique identifier, return true.
+1.  The request is not the result of a cross-site redirect. That is,
+    the origin of every url in the request's url list is same-site with the
+    request's current url's origin.
 
-2.  If A and B are both scheme/host/port triples:
+2.  The target's URI's origin is same-site with the request's client's "site
+    for cookies" (which is an origin), or if the request has no client.
```
miketaylr, line 950: Trying to understand what it means to not have a client, https://fetch.spec.whatwg.org/#concept-request-client

Should this be phrased `request client is null` instead of `request has no client`? Or does it mean something else?

reschke (commented):
That seems very misleading to me. If you add a date, you'll also need a URI that point to that very version.

englehardt:
> > If you add a date, you'll also need a URI that point to that very version.
> 
> You can grab a snapshot using this fancy button:
> 
> Screen Shot 2021-01-25 at 9 29 52 AM

I looked a bit more at the process here, and linking to a permalink in a spec is discouraged by HTML. See https://whatwg.org/working-mode#anchors. Specifically `However, other standards organizations are discouraged from referencing these snapshots, as they generally contain contain known issues that have been fixed in the Living Standard, and can mislead implementers and web developers.`.

Following that process: I've filed https://github.com/whatwg/html/issues/6329, and will reference that anchor using a non-snapshot link & date as a separate reference from the rest of these HTML terms the spec already references.

draft-ietf-httpbis-rfc6265bis.md
englehardt: I don't think we should update this to `request client is null` because there are requests from user agents that actually don't have a notion of a client.

I was debating whether we need to tweak the rest of the definition when reading with that lens, but I think the cross-site redirect restriction (bullet point 1) should still apply to a request from a user agent that doesn't have a client.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -89,6 +89,13 @@ normative:
       ins: D. Denicola
       name: Domenic Denicola
       organization: Google, Inc.
+  SAMESITE:
+    target: https://html.spec.whatwg.org/#same-site
+    title: HTML - Living Standard
+    date: 2021-01-26
```
reschke, line 95:
```suggestion
```

reschke (changes requested):

draft-ietf-httpbis-rfc6265bis.md
miketaylr: @chlily1 made a good point I think (offline), can we say something like ~"is null or has no client"? That sounds good to me.

draft-ietf-httpbis-rfc6265bis.md
englehardt: @reschke Our hope here is to capture the context of the time when the HTML spec was referenced, akin to how one would use `Accessed: <date>` when citing a web reference in other contexts. There is some precedence for this in other recent IETF documents: https://tools.ietf.org/html/rfc8833#ref-HTML5, but I also see documents that leave the date off (e.g., https://raw.githubusercontent.com/httpwg/http-extensions/master/draft-ietf-httpbis-bcp56bis.md).

Is the prevailing norm that no dates are included in references to living standards? If so, I won't object to dropping it.

englehardt: We reached out to folks from the RFC Production Center (RPC) for guidance on how to reference HTML. Their guidance is: `We recommend including a date if the intent is to refer to a specific version in time or if the text refers to a specific section within a document.  If the goal is for readers to refer the most current version of the document at any given time, we still suggest including the date but it is not required.`. This reference clearly meets the first use case, and as such we should keep a date here.

draft-ietf-httpbis-rfc6265bis.md
englehardt: SGTM. I've added that in

miketaylr: looks good, thanks

miketaylr (approved):

draft-ietf-httpbis-rfc6265bis.md
miketaylr: @reschke thoughts on @englehardt's last comment? If you're happy with it, can you update your review (to not request changes)? Thanks.

miketaylr:
@englehardt it would be good to also add a changelog entry here.

englehardt:
@chlily1 Would you mind taking another look at this post-merge with #1384? I've simplified the language in the same-site section regarding how we calculate same-site for requests which are reload navigations resulting from UI. Previously you had:

>  whether that request is same-site or cross-site should be evaluated based on the client of the request that originally navigated to the reloaded page, rather than the client of the request itself.

But consider the example of a user that types in B.com in the address bar (i.e., no client), B.com HTTP redirects to A.com, and then the user manually refreshes A.com. The B.com to A.com redirect makes the initial load cross-site, and thus we'd expect the user-initiated reload to still be cross-site. But if we followed the quoted text above, we'd use the client of the initial load (i.e., null) and conclude the new request should be same-site. I've instead opted to use the criteria you added later in the draft:

> Requests which are the result of a reload navigation triggered through a user interface element are same-site if the reloaded document was originally navigated to via a same-site request.

englehardt:
I've also made a terminology change, moving from `target's URI's origin` to `request's current url's origin` for the third criteria. This aligns with what I've added in the first criteria, and actually the original term `target` was never defined in the draft and isn't used elsewhere.

sbingler (approved):

chlily1:
> @chlily1 Would you mind taking another look at this post-merge with #1384? I've simplified the language in the same-site section regarding how we calculate same-site for requests which are reload navigations resulting from UI. Previously you had:
> 
> > whether that request is same-site or cross-site should be evaluated based on the client of the request that originally navigated to the reloaded page, rather than the client of the request itself.
> 
> But consider the example of a user that types in B.com in the address bar (i.e., no client), B.com HTTP redirects to A.com, and then the user manually refreshes A.com. The B.com to A.com redirect makes the initial load cross-site, and thus we'd expect the user-initiated reload to still be cross-site. But if we followed the quoted text above, we'd use the client of the initial load (i.e., null) and conclude the new request should be same-site. I've instead opted to use the criteria you added later in the draft:
> 
> > Requests which are the result of a reload navigation triggered through a user interface element are same-site if the reloaded document was originally navigated to via a same-site request.

Nice catch, thanks! Lgtm.

draft-ietf-httpbis-rfc6265bis.md
englehardt: I'm going to merge this now since we're blocking a PR for #1337. Feel free to file a follow-up issue if there's more to discuss!

Issue #1349, opened 2020-12-23
Digest: problems reproducing some §10 examples


bc-pi:
I admit to struggling to fully grasp the meaning of HTTP representation-data and/or selected representation as it relates to which actual bytes to use as input to the digest function (I was trying to write down how to use the Digest value as part of a simple digital signature scheme for HTTP messages FWIW). 

In an effort to better understand, I went to work through the examples in [Section 10](https://www.ietf.org/archive/id/draft-ietf-httpbis-digest-headers-04.html#name-examples-of-unsolicited-dig) but ran into some things that didn't seem right. 

Specifically for the [response in sec 10.8](https://www.ietf.org/archive/id/draft-ietf-httpbis-digest-headers-04.html#section-10.8-6), with some luck and experimentation I was able to reproduce the digest value of `0o/WKwSfnmIoSlop2LV/ISaBDth05IeW27zzNMUh5l8=` from an input of `{"status": "created", "id": "123", "ts": 1569327729, "instance": "/books/123"}` as the body or representation-data. However, the JSON in the body of that example is semantically equivalent to that but formatted differently and consequently has a different digest value:
```
HTTP/1.1 201 Created
Content-Type: application/json
Digest: id-sha-256=0o/WKwSfnmIoSlop2LV/ISaBDth05IeW27zzNMUh5l8=
Location: /books/123

{
  "status": "created",
  "id": "123",
  "ts": 1569327729,
  "instance": "/books/123"
}
```
Similarly for the [response in sec 10.10](https://www.ietf.org/archive/id/draft-ietf-httpbis-digest-headers-04.html#section-10.10-6) the digest value in the example was reproducible from `{"title": "Not Found", "detail": "Cannot PATCH a non-existent resource", "status": 404}` but the formatting of the JSON in the example differs:
```
HTTP/1.1 404 Not Found
Content-Type: application/problem+json
Digest: sha-256=UJSojgEzqUe4UoHzmNl5d2xkmrW3BOdmvsvWu1uFeu0=

{
  "title": "Not Found",
  "detail": "Cannot PATCH a non-existent resource",
  "status": 404
}
```

I don't think I'm dong anything wrong here. But please correct me, if so. 

I realize there are line length limitations to the textual RFC style output that need to be considered but I do think having examples that readers can easily and reliably follow and reproduce is worthwhile.

LPardue:
Thanks for reporting this @bc-pi ! Its quite possible that these got broken along the way due to rewrap. 

[...]

PR #1350, opened 2020-12-28
Fix: #1349. Include spaces in digest computation.

ioggstream:
## This PR

- fixes examples where Digest was not computed on the actual bytes in the examples
- warns about Digest computation in the examples.

## Note

@bc-pi could you please have a look and check if this works for you?
@LPardue where can we provide editorial conventions used in the examples?

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -550,6 +550,15 @@ The following examples demonstrate interactions where a server responds with a
 `Digest` field even though the client did not solicit one using
 `Want-Digest`.
 
+Some examples include JSON objects in the payload data.
+For presentation purposes, objects that fit completely within the line-length limits
+are presented on a single line using compact notation with no leading space.
+Objects that would exceed line-length limits are presented across multiple lines
+(one line per key-value pair) with 2 spaced of leading indentation.
+
+`Digest` is media-type agnostic
+and does not provide canonicalization algorithms for specific formats.
+Examples of `Digest` are calculated inclusive of any space.
 
 ## Server Returns Full Representation Data {#example-full-representation}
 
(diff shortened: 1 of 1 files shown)


bc-pi:
Yup, I was able to reproduce the same digest values.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1473,17 +1479,33 @@ using SHA algorithms for a range of encodings. Note that these are formatted as
 base64. This function could be adapted to other algorithms and should take into
 account their specific formatting rules.
 
+While the JSON objects in the examples are consistently formatted
+using the compact notation (no indentation) for single-key entries
+and 2-space indentation for multi-key entries,
+`Digest` is media-type agnostic and does not provide canonicalization
+algorithms for specific formats.
+
```
LPardue, line 1487: Since the appendix will get removed, I think it would help to put text to this effect in the opening paragraph of Section 10 and 11. 

I suggest a slight rewording of the text for extra clarity: 

"Some examples include JSON objects in the payload data. For presentation purposes,  objects that fit completely within the line-length limits are presented on a single line using compact notation with no leading whitespace. Objects that would exceed line-length limits are presented across multiple lines (one line per key-value pair) with 2 spaced of leading indentation. `Digest` is media-type agnostic and does not provide canonicalization algorithms for specific formats. Examples of Digest are calculated inclusive of any whitespace."

ioggstream: Reasonable, I'll do it;  but I'd avoid repeating things. Should we move §10 and §11 under a common "Example" section?

LPardue: That's what we had before and it makes the table of contents too hard to read.

ioggstream: Ok, @LPardue if you like the current setup approve/merge

ioggstream:
@LPardue PTAL

LPardue (approved):

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1494,14 +1529,13 @@ print("Identity | sha256 |", digest(item))
 print("Encoding | digest-algorithm | digest-value")
 print("Brotli | sha256 |", digest(item, encoding=brotli.compress))
 # Encoding | digest-algorithm | digest-value
-# Brotli , sha256 4REjxQ4yrqUVicfSKYNO/cF9zNj5ANbzgDZt3/h3Qxo=
-
+# Brotli | sha256 | 4REjxQ4yrqUVicfSKYNO/cF9zNj5ANbzgDZt3/h3Qxo=
```
bc-pi, line 1532: Shouldn't `# Identity | sha256 | 4REjxQ4yrqUVicfSKYNO/cF9zNj5ANbzgDZt3/h3Qxo=` a few lines above be `# Identity | sha256 | X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=` ?

ioggstream: done, thx

PR #1354, opened 2020-12-29
digest: fix SRI section ref

### LINKED ISSUE #1353: digest-headers: cites Section 3.4.5 of SRI but that section does not exist

LPardue:
fixes #1353

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -509,15 +509,6 @@ The SRI `integrity` attribute contains a cryptographic hash algorithm and digest
 value which is similar to `representation-data-digest` (see
 {{representation-digest}}). The major differences are in serialization format.
 
-The SRI digest value is calculated over the representation data of a resource
-without any content-coding applied.
-Section 3.3.5 of [SRI] describes the benefit of not applying content-codings:
-the SRI `integrity` attribute can contain multiple algorithm-value
-pairs where each applies to a different representation data,
-thus protecting different representations of a given resource. However, this is different
-from the design of representation digests, where multiple `Digest` field-values
-all protect the same representation.
-
 SRI does not specify handling of partial representation data (e.g. Range
 requests). In contrast, this document specifies handling in terms that are fully
 compatible with core HTTP concepts (an example is provided in
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -508,7 +508,7 @@ value which is similar to `representation-data-digest` (see
 
 The SRI digest value is calculated over the identity encoding of the resource,
 not the selected representation (as specified for `representation-data-digest`
-in this document). Section 3.4.5 of [SRI] describes the benefit of the identity
+in this document). Section 3.5 of [SRI] describes the benefit of the identity
 approach - the SRI `integrity` attribute can contain multiple algorithm-value
 pairs where each applies to a different identity encoded payload. This allows
 for protection of distinct resources sharing a URL. However, this is a contrast


reschke (changes requested):
But that section doesn't even "identity" or content codings; it seems more context is needed.

LPardue:
Ha, Touche! The actual piece I want to reference is **section 3.3.5** - https://www.w3.org/TR/2016/REC-SRI-20160623/#does-response-match-metadatalist

This says:

```
This algorithm allows the user agent to accept multiple, valid strong hash functions. For example, a developer might write a script element such as:
Example 7

<script src="https://example.com/example-framework.js"
        integrity="sha384-Li9vy3DqF8tnTXuiaAJuML3ky+er10rcgNR/VqsVpcw+ThHmYcwiB1pbOxEbzJr7
                   sha384-+/M6kredJcxdsqkczBUjMLvqyHb1K/JThDXWsBVxMEeZHEaMKEOEct339VItX1zB"
        crossorigin="anonymous"></script>

which would allow the user agent to accept two different content payloads, one of which matches the first SHA384 hash value and the other matches the second SHA384 hash value.
```

Given this, do you think the current Digest text and section reference is enough, or is there more that I could say here?

reschke:
That section talks about different representations in general, not specifically about content codings, right? As such I find that piece of text still very confusing.

LPardue:
This PR needs work to address Julian's points. I intend to get to this in the near future.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -509,13 +509,13 @@ The SRI `integrity` attribute contains a cryptographic hash algorithm and digest
...
 {{representation-digest}}). The major differences are in serialization format.
 
-The SRI digest value is calculated over the identity encoding of the resource,
-not the selected representation (as specified for `representation-data-digest`
-in this document). Section 3.4.5 of [SRI] describes the benefit of the identity
-approach - the SRI `integrity` attribute can contain multiple algorithm-value
-pairs where each applies to a different identity encoded payload. This allows
-for protection of distinct resources sharing a URL. However, this is a contrast
-to the design of representation digests, where multiple `Digest` field-values
+The SRI digest value is calculated over the representation data of a resource
```
ioggstream, line 512: Fix 1: The correct terminology is `without any content-coding applied`, as "identity" is not a content-coding. https://github.com/httpwg/http-core/pull/389

draft-ietf-httpbis-digest-headers.md
```diff
@@ -509,13 +509,13 @@ The SRI `integrity` attribute contains a cryptographic hash algorithm and digest
...
+without any content-coding applied.
+Section 3.3.5 of [SRI] describes the benefit of not applying content-codings:
```
ioggstream, line 514: Fix 2: reference the section from https://github.com/httpwg/http-extensions/pull/1354#issuecomment-752270667

draft-ietf-httpbis-digest-headers.md
```diff
@@ -509,13 +509,13 @@ The SRI `integrity` attribute contains a cryptographic hash algorithm and digest
...
+the SRI `integrity` attribute can contain multiple algorithm-value
+pairs where each applies to a different representation data,
+thus protecting different representations of a given resource. However, this is different
```
ioggstream, line 517: Fix 3: if it's the same url, then it's the same resource, which may have different representation => different representation data (eg. content-language, content-encoding, ...)

ioggstream:
@LPardue reading https://github.com/w3c/webappsec-subresource-integrity/blob/master/index.bs#L353 I am still not sure whether `SRI` *always* ignore content-codings... it seems at a minimum confusing to me.

ioggstream:
@reschke I read SRI and it seems unclear whether it is content-coding aware or not (see comment above https://github.com/httpwg/http-extensions/pull/1354#issuecomment-760888084).

I tried to address all your concerns on SRI here. Feel free to amend, otherwise.

reschke (approved):

Issue #1357, opened 2020-12-30
digest-headers: confusing section "Server Returns No Representation Data"


reschke:
In 10.2 Server Returns No Representation Data:

- what does the Digest request header field on the HEAD request refer to, and what is the recipient supposed to do with it?
- there's only one digest in the response, but the prose talks about two

ioggstream:
> In 10.2 Server Returns No Representation Data:
> 
> * what does the Digest request header field on the HEAD request refer to, 
Requests without payload data can still send a `Digest` field applying the digest-algorithm to an empty representation.

> and what is the recipient supposed to do with it?
The recipient can validate that the digest-value matches the one of an empty representation or not.

> * there's only one digest in the response, but the prose talks about two
True. this is confusing, should reference the previous example.:

```
the id-sha-256 digest value in the response is the same of the sha-256 digest value of the response in {{example-full-representation}}
```

PR #1376, opened 2021-01-11
Fix: #1357. Clarify Digest+HEAD.

ioggstream:
## This PR

Clarifies `Server Returns No Representation Data`.

1- `Digest` + HEAD is used to get a resource checksum
2- no Digest in request for clarity
3- only use sha-256 to focus on the actual example.

Fix: #1357

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -251,6 +251,11 @@ together with an indication of the algorithm used:
                                 <encoded digest output>
 ~~~
 
+When a message has no representation data
+it is still possible to assert that no representation data was sent
+computing the representation digest on an empty string
+(see {{usage-in-signatures}}).
+
 The checksum is computed using one of the digest-algorithms listed in {{algorithms}}
 and then encoded in the associated format.
 
@@ -590,22 +595,18 @@ Digest: sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
 
 ## Server Returns No Representation Data
 
-Requests without content can still send a `Digest` field
-applying the digest-algorithm to an empty representation.
+In this example, a HEAD request is used to retrieve the checksum
+of a resource.
 
 The response `Digest` field-value is calculated over the JSON object
 `{"hello": "world"}`, which is not shown because there is no payload
 data.
-
-In this example there is no content coding applied, so the "sha-256" and the "id-sha-256"
-digest-values in the response would be the same.
 
 Request:
 
 ~~~ http-message
 HEAD /items/123 HTTP/1.1
 Host: foo.example
-Digest: sha-256=47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=
 
 ~~~
 
@@ -614,7 +615,7 @@ Response:
 ~~~ http-message
 HTTP/1.1 200 OK
 Content-Type: application/json
-Digest: id-sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
+Digest: sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
 
 ~~~
 


reschke (changes requested):
"Requests without payload data can still send a `Digest` field" -- unless I'm missing something it's the *response* which has the Digest header field, no?

ioggstream:
Digest can be sent in requests and response

> "Requests without payload data can still send a `Digest` field" -- unless I'm missing something it's the _response_ which has the Digest header field, no?

Nope,  Digest can be sent in requests too. 

When used in conjuction with signatures, the client that want to certify that no-payload was attached will:

1. compute the Digest of an empty string
2. sign that Digest header with some signature mechanism
3. attach the signature and digest to the request
4. send the request with no payload data

The server will then:
1. verify the signature according to the chosen mechanism
2. if no payload is provided, the Digest is the one of an empty string
3. eventually store the signature to prove that the request had no payload

Clearly the signature mechanism should provide enough information to define the semantic of the request (eg. the method and all relevant headers).

HTH, R.

reschke:
Well.

What *exactly* does "Digest" in a HEAD request refer to, and what is a recipient supposed to do with it?

ioggstream:
> So the Digest is for the empty request payload. Maybe it would be good to state this in the spec?

+1

draft-ietf-httpbis-digest-headers.md
```diff
@@ -586,16 +586,19 @@ Digest: sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
 
 ## Server Returns No Representation Data
 
+In this example, a HEAD request is used to retrieve the checksum
+of a resource.
```
LPardue, line 590: I'd move this to after the next paragraph (e.g. after L596 as it presently is in the PR).

draft-ietf-httpbis-digest-headers.md
```diff
@@ -586,16 +586,19 @@ Digest: sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
...
+
 Requests without payload data can still send a `Digest` field
 applying the digest-algorithm to an empty representation.
+This can be useful in conjunction with signatures
+to ensure that the sender did not convey any payload
+(see {{usage-in-signatures}}).
 
 The response `Digest` field-value is calculated over the JSON object
```
LPardue, line 598: Per Julian's comment, we can also state "the request `Digest` field-value is calculated over ..."

LPardue (changes requested):

draft-ietf-httpbis-digest-headers.md
ioggstream: :heavy_check_mark:

draft-ietf-httpbis-digest-headers.md
```diff
@@ -250,6 +250,11 @@ together with an indication of the algorithm used:
                                 <encoded digest output>
 ~~~
 
+When a message has no representation data (eg. a HEAD request)
+it is still possible to assert that no representation data was sent
+computing the representation digest on an empty string
```
LPardue, line 255:
```suggestion
by computing the representation digest on an empty string
```

LPardue (commented):
LGTM now, modulo one nit

LPardue (approved):

ioggstream:
@reschke PTAL :)

reschke:
I'll repeat what I said earlier: "That said, this is IMHO a very misleading example, as the HTTP spec is very clear that GET/HEAD request bodies ("A payload within a HEAD request message has no defined semantics; sending a payload body on a HEAD request might cause some existing implementations to reject the request."). It would probably better to have this example on a request that actually can take meaningful payload..."

LPardue:
In order to make progress here, I suggest we simplify the example to remove the Request digest. That keeps the example focused on it's name "Server Returns No Representation Data".

ioggstream:
@reschke removed Digest from HEAD.

ioggstream:
@reschke PTAL ;)

ioggstream:
@reschke PTAL :)

reschke (approved):

ioggstream:
YAY :rocket:

PR #1386, opened 2021-01-31
nits for Structured Field Values parsing algo

Jxck:
align for other sentences for make algorithm clear.


mnot:
Sorry, this is already at the RFC Editor and has final approval; too late.

Issue #1392, opened 2021-02-04
digest-headers: refer to 6151/6194 for md5/sha1 deprecation


seanturner:
In s12.2, should refer (informatively) to [RFCs 6151](https://datatracker.ietf.org/doc/rfc6151/) and [6194](https://datatracker.ietf.org/doc/rfc6194/) for IETF references that deprecate the use of md5 and sha1, respectively.

LPardue:
Thanks for the pointers, sgtm

PR #1397, opened 2021-02-04
Fix: #1392. Reference RFC 6515 and 6194 for deprecating md5 and sha1.

ioggstream:
## This PR

- references RFC6515 and 6194 for deprecating md5 and sha1

## Note

Since there are RFCs, we could even skip explaining that those algorithms are vulnerable to collisions.

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -349,10 +349,13 @@ The Internet Assigned Numbers Authority (IANA) acts as a registry for
 digest-algorithm values.
 The registry contains the tokens listed below.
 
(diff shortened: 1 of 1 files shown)


LPardue (commented):
Most of these changes seem good. However, I'm not entirely convinced about the change of `RFC3230` to `DIGEST` - it obfuscates the document that we are obsoleting, overloading it with a term that is heavily used through the document. Could we split that off into a different PR for further discussion?

ioggstream:
cc: @seanturner

seanturner (approved):
You beat me to the PR!

LPardue (approved):

ioggstream:
@seanturner feel free to PR for the rest of the algorithms :P

PR #1398, opened 2021-02-04
Update 5843 and replace normative references with brace notation.

ioggstream:
## This PR

- replaces references with brace notation
- updates [5843 - Additional Hash Algorithms for HTTP Instance Digests](https://tools.ietf.org/html/rfc5843)  <-  @reschke is that correct?


LPardue (commented):
I don't think this will work, the majority of the text touched in the PR is what needs to appear verbatim in the registry - https://www.iana.org/assignments/http-dig-alg/http-dig-alg.xhtml

I think its just easier to refer to RFCs

PR #1400, opened 2021-02-08
Allow connection_terminated to be emitted for mid-stream errors.

PiotrSikora:


mnot:
So, what's the difference to `http_response_incomplete`, then? Should we just remove `http_response_incomplete` perhaps?

PiotrSikora:
> So, what's the difference to `http_response_incomplete`, then? Should we just remove `http_response_incomplete` perhaps?

`connection_terminated` can be also sent if the connection is terminated before or when sending HTTP request.

`http_response_incomplete` is more generic and can be sent in response to other events as well (e.g. read/write timeouts).

But more importantly, either can be sent in response to the same event (i.e. connection terminated when receiving HTTP response), depending on the existing proxy implementation or various policies, i.e. "proxy should be as specific as it wants to be" from your last email.

PR #1402, opened 2021-02-08
Style.

PiotrSikora:


draft-ietf-httpbis-proxy-status.md
```diff
@@ -99,7 +99,7 @@ Parameters on each member convey additional information about that intermediary'
...
+Proxy-Status MAY be sent in HTTP trailers. For example, if an intermediary is streaming a response and the upstream connection suddenly terminates, Proxy-Status can be appended only to the trailers of the outgoing message, since the headers have already been sent. As with all trailers, it might be silently discarded along the path to the user agent, so this SHOULD NOT be done unless it is not possible to send it in headers, and an intermediary MUST NOT send Proxy-Status as a trailer field unless it has also sent a corresponding Proxy-Status header field in the same message, so that the trailer value's ordering relative to other intermediaries is preserved.
```
mnot, line 102:
```suggestion
Proxy-Status MAY be sent in HTTP trailers. For example, if an intermediary is streaming a response and the upstream connection suddenly terminates, Proxy-Status can be appended to the trailers of the outgoing message, since the headers have already been sent. As with all trailers, it might be silently discarded along the path to the user agent, so this SHOULD NOT be done unless it is not possible to send it in headers, and an intermediary MUST NOT send Proxy-Status as a trailer field unless it has also sent a corresponding Proxy-Status header field in the same message, so that the trailer value's ordering relative to other intermediaries is preserved.
```
'only' seems inappropriate here, since we subsequently advise that it needs to be sent as a header to be included in trailers.

mnot (changes requested):

draft-ietf-httpbis-proxy-status.md
PiotrSikora: I accepted this change, but how am I misreading it? The sentence says: 

> [...], Proxy-Status can be **only** appended to the trailers of the outgoing message, **since the headers have already been sent**.

so we're referring to the case where headers have been already sent, forwarding of the response body has been interrupted, and the only way that we can send Proxy-Status is in the trailers. How is it inappropriate?

I'm not a native speaker, but every time I read it without "only", it just felt wrong.

mnot: Ah, never mind, you're correct; I misread. I'll merge this and reinstate it.

PR #1411, opened 2021-02-18
Update signature input methods and requirements

jricher:
Update the signature input method to always require the signature parameters to be covered.

Redefine all fields like `@request-target` as specialty input fields (non-headers) and allow extension by registry.

﻿Define signature parameters using structured field semantics.

Define canonicalization rules more directly.

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -252,7 +252,27 @@ The following table shows example canonicalized values for different content ide
 {: title="Non-normative examples of list prefix canonicalization."}
 
 
-## Signature Creation Time
+## Specialty Content Fields
+
+Content not found in an HTTP header can be included in the signature base string by defining a content identifier and the canonicalization method for its content.
+
+Specialty content identifiers MUST start with the asterisk `*` character. This specification defines the following specialty content identifiers:
+
+*created
+: The signature creation time. {{content-created}}
+
+*expires
+: The signature expiration time. {{content-expires}}
+
+*request-target
+: The target request endpoint. {{content-request-target}}
+
+*signature-params
+: The signature metadata parameters. {{content-signature-params}}
+
+Additional specialty content identifiers MAY be defined and registered in the HTTP Signatures Specialty Content Identifier Registry. {{content-registry}}
+
+### Signature Creation Time {#content-created}
 
 The signature's Creation Time ({{signature-metadata}}) is identified by the `*created` identifier.
 
@@ -260,13 +280,13 @@ Its canonicalized value is an `sf-integer` containing the signature's Creation T
 
 > The use of seconds since the Epoch to canonicalize a timestamp simplifies processing and avoids timezone management required by specifications such as [RFC3339].
 
-## Signature Expiration Time
+### Signature Expiration Time {#content-expires}
 
 The signature's Expiration Time ({{signature-metadata}}) is identified by the `*expires` identifier.
 
 Its canonicalized value is an `sf-decimal` containing the signature's Expiration Time expressed in "Unix time".
 
-## Target Endpoint
+### Target Endpoint {#content-request-target}
 
 The request target endpoint, consisting of the request method and the path and query of the effective request URI, is identified by the `*request-target` identifier.
 
@@ -276,7 +296,7 @@ Its value is canonicalized as follows:
 2. Append a space " ".
 3. Append the path and query of the request target of the message, formatted according to the rules defined for the :path pseudo-header in {{!HTTP2=RFC7540}}, Section 8.1.2.3.  The resulting string is the canonicalized value.
 
-### Canonicalization Examples
+#### Canonicalization Examples
 
 The following table contains non-normative example HTTP messages and their canonicalized `*request-target` values.
 
@@ -322,6 +342,14 @@ Host: server.example.com</sourcecode></td>
         </tr>
     </tbody>
 </table>
+
+### Signature Parameters {#content-signature-params}
+
+The signature input ({{signature-metadata}}) is identified by the `*signature-params` identifier.
+
+Its canonicalized value is an the list of all signing metadata used to create the Signature-Input header value representing this signature. {{params}}
+
+\[\[ Editor's Note: Can we just re-use the structured headers serialization method here to get a consistent serialization? \]\]
 
 # HTTP Message Signatures {#message-signatures}
 
@@ -411,7 +439,9 @@ If Covered Content contains `*created` and the signature's Creation Time is unde
 
 If Covered Content contains `*expires` and the signature does not have an Expiration Time an implementation MUST produce an error.
 
-If Covered Content contains an identifier for a header field that is not present or malformed in the message, the implementation MUST produce an error.
+The Signature Input MUST include the signature metadata specialty field `*signature-params` as the last entry in the covered content. {{content-signature-params}}
+
+If Covered Content contains an identifier for a header field that is malformed or is not present in the message, the implementation MUST produce an error.
 
 If Covered Content contains an identifier for a Dictionary member that references a header field that is not present, is malformed in the message, or is not a Dictionary Structured Field, the implementation MUST produce an error. If the header field value does not contain the specified member, the implementation MUST produce an error.
 
@@ -430,6 +460,7 @@ x-example: Example header with some whitespace.
 x-dictionary: b=2
 x-dictionary: a=1
 x-list: (a, b, c)
+*signature-params: (*request-target *created host date cache-control x-empty-header x-example x-dictionary:a x-dictionary:b x-list); keyid="test-key-a"; alg=hs2019; created=1402170695; expires=1402170995
 ~~~
 {: title="Non-normative example Signature Input" artwork-name="example-sig-input" #example-sig-input}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-message-signatures.md
```diff
@@ -744,6 +790,7 @@ The Signature Input is:
 ~~~
 *created: 1402170695
 *request-target: post /foo?param=value&pet=dog
+*signature-params: (*created *request-target); keyid="test-key-a"; created=1402170695
```
richanna, line 793:
```suggestion
*Signature-Input: mySignatureId=(*request-target); keyid="test-key-a"; created=1402170695
```

draft-ietf-httpbis-message-signatures.md
```diff
@@ -323,6 +346,32 @@ Host: server.example.com</sourcecode></td>
...
+Note that an HTTP message could contain multiple signatures, but only the signature parameters used for the current signature are included.
+
+#### Canonicalization Examples
+
+Given the following signature parameters:
+
+|Property|Value|
+|--- |--- |
+|Algorithm|`hs2019`|
+|Covered Content|`@request-target`, `host`, `date`, `cache-contol`, `x-emptyheader`, `x-example`, `x-dictionary;key=b`, `x-dictionary;key=a`, `x-list;prefix=3`|
```
fimbault, line 364: small typo (cache-contol -> cache-control)

draft-ietf-httpbis-message-signatures.md
```diff
@@ -161,21 +160,41 @@ Additionally, some examples use '\\' line wrapping for long values that contain
...
+Some content within HTTP messages can undergo transformations that change the bitwise value without altering meaning of the content (for example, the merging together of header fields with the same name).  Message content must therefore be canonicalized before it is signed, to ensure that a signature can be verified despite such intermediary transformations. This document defines rules for each content identifier that transform the identifier's associated content into such a canonical form.
 
-The following sections define content identifiers, their associated content, and their canonicalization rules.
+Content identifiers are defined using production grammar defined by {{!RFC8941}} section 4.
+The content identifier is an `sf-string` value. The content identifier
+type MAY define parameters which are included using the `parameters` rule.
```
ioggstream, line 169: Where is the `parameters` rule defined?

draft-ietf-httpbis-message-signatures.md
```diff
@@ -161,21 +160,41 @@ Additionally, some examples use '\\' line wrapping for long values that contain
...
-An HTTP header field value is canonicalized as follows:
+Unless overridden by additional parameters and rules, the HTTP header field value MUST be canonicalized with the following steps:
 
 1. Create an ordered list of the field values of each instance of the header field in the message, in the order that they occur (or will occur) in the message.
 2. Strip leading and trailing whitespace from each item in the list.
-3. Concatenate the list items together, with a comma "," and space " " between each item. The resulting string is the canonicalized value.
+3. Concatenate the list items together, with a comma "," and space " " between each item. 
+
+The resulting string is the canonicalized value.
+
```
ioggstream, line 190: TODO: add example

draft-ietf-httpbis-message-signatures.md
```diff
@@ -197,17 +216,19 @@ The following table shows example canonicalized values for header fields, given
...
+|`"server"`|www.example.com|
+|`"x-empty-header"`||
+|`"x-obs-fold-header"`|Obsolete line folding.|
+|`"x-ows-header"`|Leading and trailing whitespace.|
 {: title="Non-normative examples of header field canonicalization."}
 
 ## Dictionary Structured Field Members
 
-An individual member in the value of a Dictionary Structured Field is identified by the lowercased field name, followed by a semicolon `":"`, followed by the member name. An individual member in the value of a Dictionary Structured Field is canonicalized by applying the serialization algorithm described in Section 4.1.2 of {{!StructuredFields=I-D.ietf-httpbis-header-structure}} on a Dictionary containing only that member.
+An individual member in the value of a Dictionary Structured Field is identified by using the parameter `key` on the content identifier for the header. The value of this parameter is a the key being identified, without any parameters present on that key in the original dictionary.
```
ioggstream, line 229: @jricher as discussed during IETF-109 this feature is still very problematic.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -244,46 +267,46 @@ The following table shows example canonicalized values for different content ide
...
+|`"x-list-a";prefix=6`|(a, b, c, d, e, f)|
+|`"x-list-b";prefix=0`|()|
 {: title="Non-normative examples of list prefix canonicalization."}
 
 
-## Signature Creation Time
+## Specialty Content Fields
 
-The signature's Creation Time ({{signature-metadata}}) is identified by the `*created` identifier.
+Content not found in an HTTP header can be included in the signature base string by defining a content identifier and the canonicalization method for its content.
```
ioggstream, line 280: It is not clear whether content found in an HTTP header can be further included in the signature.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -244,46 +267,46 @@ The following table shows example canonicalized values for different content ide
...
 
-Its canonicalized value is an `sf-integer` containing the signature's Creation Time expressed in "Unix time".
+To differentiate speciality content identifiers from HTTP headers, specialty content identifiers MUST start with the "at" `@` character. This specification defines the following specialty content identifiers:
```
ioggstream, line 282: Use `specialty` for consistency

draft-ietf-httpbis-message-signatures.md
```diff
@@ -333,27 +382,45 @@ HTTP Message Signatures have metadata properties that provide information regard
...
-: A timestamp representing the point in time that the signature was generated. Sub-second precision is not supported. A signature's Creation Time MAY be undefined, indicating that it is unknown.
-
-Covered Content:
-: An ordered list of content identifiers (Section 2) that indicates the metadata and message content that is covered by the signature. The order of identifiers in this list affects signature generation and verification, and therefore MUST be preserved.
+: A timestamp representing the point in time that the signature was generated, represented as an integer. Sub-second precision is not supported. A signature's Creation Time MAY be undefined, indicating that it is unknown. `sf-integer`
 
 Expiration Time:
-: A timestamp representing the point in time at which the signature expires. An expired signature always fails verification. A signature's Expiration Time MAY be undefined, indicating that the signature does not expire.
+: A timestamp representing the point in time at which the signature expires, represented as an integer. An expired signature always fails verification. A signature's Expiration Time MAY be undefined, indicating that the signature does not expire.
```
ioggstream, line 391: signature always expire ;) better would be to say that it's unknown.

draft-ietf-httpbis-message-signatures.md
jricher: I don't necessarily disagree but I wanted to at least make it consistent with the rest of the syntax. An upside is that we don't NEED this functionality in the same way that we used to (with the new `@signature-params` speciality field), and it's also more straightfoward to apply than it was before.

draft-ietf-httpbis-message-signatures.md
jricher: Content found in an HTTP header can be included in the signature by adding the header to the signed content, which is described in its own section.

draft-ietf-httpbis-message-signatures.md
jricher: "speciality" is a typo on my part that I didn't notice! It actually took me a few reads to find the problem here. :)

draft-ietf-httpbis-message-signatures.md
jricher: I don't think that's actually the case. A signature may be perpetual depending on the use case. The amount of time that's passed might limit how much I'm willing to trust it, though. Especially for something like an HTTP message request!

draft-ietf-httpbis-message-signatures.md
jricher: It's defined in RFC8491: https://www.rfc-editor.org/rfc/rfc8941.html#section-3.1.2-4

We need to have a more formal import of the ABNF rules in this document, but that hasn't been done yet.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -161,21 +160,41 @@ Additionally, some examples use '\\' line wrapping for long values that contain
...
+In order to allow signers and verifiers to establish which content is covered by a signature, this document defines content identifiers for data items covered by an HTTP Message Signature.
 
-Some content within HTTP messages may undergo transformations that change the bitwise value without altering meaning of the content (for example, the merging together of header fields with the same name).  Message content must therefore be canonicalized before it is signed, to ensure that a signature can be verified despite such innocuous transformations.  This document defines rules for each content identifier that transform the identifier's associated content into such a canonical form.
+Some content within HTTP messages can undergo transformations that change the bitwise value without altering meaning of the content (for example, the merging together of header fields with the same name).  Message content must therefore be canonicalized before it is signed, to ensure that a signature can be verified despite such intermediary transformations. This document defines rules for each content identifier that transform the identifier's associated content into such a canonical form.
```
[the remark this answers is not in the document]
jricher: Thanks, I missed that non-normative must when I made the other editorial changes here. 

The fact that some parties want to have evidence of transformation of some things is exactly why you specify what NEEDS to be signed when applying this specification. The start of that language was included in #1419

PR #1420, opened 2021-03-04
RFC 6265bis: Specify handling of CTL characters in set-cookie-string

chlily1:
This adds explicit steps to handle control characters in set-cookie-strings.


chlily1:
@englehardt PTAL, thanks.

I added a step to the Storage-model algorithm because I realized that cookies can be created without parsing a set-cookie-line (e.g. through the CookieStore API).

englehardt (approved):
This LGTM. Thanks @chlily1!

I've added some wpts for this in https://github.com/web-platform-tests/wpt/pull/28290. Firefox fails when CR is present, Chrome/Edge when NUL is present, and Safari when other CTL characters are present. But I think the new language here is the right thing to target.

chlily1:
Thanks for the review, and thank you for adding the tests!

PR #1425, opened 2021-03-08
RFC6265bis: Update Top-Level Navigation & Widgets SameSite example

sbingler:
The Top-Level Navigation SameSite example has an out of date recommendation to drop the SameSite attribute for cross-site cookie access. Instead, SameSite=None should be used.

The widgets example doesn't differentiate between None, Lax, and Strict SameSite.

& Some URL touch ups to meet the schemeful definition of a site.


draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1951,32 +1951,32 @@ deal reasonably well with top-level navigations.
...
 relies on not one, but two cookies: one conceptually granting "read" access,
-another granting "write" access. The latter could be marked as `SameSite`, and
-its absence would prompt a reauthentication step before executing any
-non-idempotent action. The former could drop the `SameSite` attribute entirely,
-or choose the "Lax" version of enforcement, in order to allow users access to
-data via top-level navigation.
+another granting "write" access. The latter could be marked as `SameSite=Strict`,
+and its absence would prompt a reauthentication step before executing any
+non-idempotent action. The former could be marked as `SameSite=Lax`, or
+`SameSite=None`, in order to allow users access to data via top-level navigation.
```
chlily1, line 1966: nit: the clause "in order to allow users access to data via top-level navigation" should go with SameSite=Lax, so maybe move it to be closer. And add some more rationale for why SameSite=None might be used, e.g.

``` suggestion
non-idempotent action. The former could be marked as `SameSite=Lax`, in
order to allow users access to data via top-level navigation, or
`SameSite=None`, to permit access in all contexts (including cross-site
embedded contexts).
```

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1951,32 +1951,32 @@ deal reasonably well with top-level navigations.
...
 
-The `SameSite` attribute is inappropriate for some important use-cases. In
-particular, note that content intended for embedding in a cross-site contexts
-(social networking widgets or commenting services, for instance) will not have
-access to same-site cookies. Cookies may be required for requests triggered in
-these cross-site contexts in order to provide seamless functionality that relies
-on a user's state.
+The `Lax` and `Strict` values for the `SameSite` attribute are inappropriate
+for some important use-cases. In particular, note that content intended for
+embedding in a cross-site contexts (social networking widgets or commenting
```
chlily1, line 1972: nit: typo (preexisting): contexts -> context (or remove the "a")

chlily1 (changes requested):
Thanks for updating this!

chlily1 (approved):
Thanks!

Issue #1422, opened 2021-03-05
RFC6265bis Spec does not match Lax-Allowing-Unsafe behavior implemented in Chrome and Firefox


sbingler:
Both Chrome and Firefox (perhaps others) implement the [Lax-Allowing-Unsafe](https://tools.ietf.org/html/draft-west-cookie-incrementalism-01#section-3.1.1) behavior specified in cookie incrementalism, but RFC6265bis does not specify this behavior.

I believe it should.

chlily1:
Addressed by #1435

PR #1435, opened 2021-03-16
RFC 6265bis: Add Lax-allowing-unsafe SameSite enforcement mode

chlily1:
Addresses #1422

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1358,16 +1358,16 @@ If the user agent uses "Lax-allowing-unsafe" enforcement, it MUST apply the
 following modification to the retrieval algorithm defined in
 {{retrieval-algorithm}}:
 
-Replace the condition in the final bullet point of step 1 of the retrieval
+Replace the condition in the penultimate bullet point of step 1 of the retrieval
 algorithm reading
 
-     * The HTTP request's method is "safe".
+     * The HTTP request associated with the retrieval uses a "safe" method.
 
 with
 
      * At least one of the following is true:
 
-       1.  The HTTP request's method is "safe".
+       1.  The HTTP request associated with the retrieval uses a "safe" method.
 
        2.  The cookie's same-site-flag is "Default" and the amount of time
            elapsed since the cookie's creation-time is at most a duration of the
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -876,7 +876,7 @@ found-year) are initially "not set".
     the minute-value, and the second-value, respectively. If no such date
     exists, abort these steps and fail to parse the cookie-date.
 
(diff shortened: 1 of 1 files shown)


chlily1:
@englehardt PTAL.

Ended up adapting the text from Cookie Incrementalism into two sections, one with a focus on the behavior (with a monkey patch to the algorithm) and the other more focusing on the justification for it.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1303,6 +1303,45 @@ When possible, developers should use a session management mechanism such as
...
+`SameSite` enforcement mode, may be set for top-level navigations, regardless of
+HTTP request method, as specified in {{storage-model}}.)
+
+"Lax-allowing-unsafe" is not a distinct value of the `SameSite` attribute.
+Rather, user agents MAY apply "Lax-allowing-unsafe" enforcement only to cookies
+that did not explicitly specify a `SameSite` attribute (i.e., those whose
+same-site-flag was set to "Default" by default). If a user agent applies
+"Lax-allowing-unsafe" enforcement, the user agent SHOULD restrict
+"Lax-allowing-unsafe" enforcement to cookies whose age is less than a duration
+of the user agent’s choosing (2 minutes seems reasonable).
```
englehardt, line 1324: I think the wording here is a bit open ended. How about:

"To limit the scope of this compatibility mode, user agents which apply "Lax-allowing-unsafe" SHOULD restrict the enforcement to cookies which were created recently. Deployment experience has shown that a cookie age of 2 minutes or less to be a reasonable limit."

englehardt (approved):
Overall this LGTM. Having this as a monkey patch seems like the best solution. I've added two suggestions for alternate wording.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -2041,6 +2080,33 @@ reload navigation triggered through the user interface may replay the original
...
 cross-site, like the request that initially navigated to the page.
 
+### Top-level requests with "unsafe" methods {#unsafe-top-level-requests}
+
+The "Lax" enforcement mode described in {{strict-lax}} allows a cookie to be
+sent with a cross-site HTTP request if and only if it is a top-level navigation
+with a "safe" HTTP method. Implementation experience shows that this is
+difficult to apply as the default behavior, as some sites may rely on cookies
+not explicitly specifying a `SameSite` attribute being included on top-level
+cross-site requests with "unsafe" HTTP methods (as was the case in a previous
```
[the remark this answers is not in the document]
chlily1: Done.

draft-ietf-httpbis-rfc6265bis.md
chlily1: Done.

chlily1:
Thanks for the review!
This is dependent on 6265-non-http-apis so I'll merge it after that one lands.

PR #1442, opened 2021-04-05
bcp56bis: fix weird link format

reschke:


mnot:
I think this is a matter of taste, so I'm inclined to leave it as is. Could you please update your PRs to make the style consistent and back out these changes?

reschke:
Consistent with what?

There are two suggested forms of references ("Section x of [FOO]" and "[FOO], Section x"). Both have their legitimate uses.

What doesn't work for me is using the second form inside a sentence without any delimitation, because it introduces a comma in the sentence structure where it does not be belong.

PR #1445, opened 2021-04-05
bcp56bis: make link to HTTP auth more specific

reschke:


draft-ietf-httpbis-bcp56bis.md
```diff
@@ -407,7 +407,7 @@ As noted in {{?I-D.ietf-httpbis-semantics}}, a user agent is allowed to automati
...
+Generally, when a redirected request is made, its header fields are copied from the original request's. However, they can be modified by various mechanisms; e.g., sent Authorization ({{?I-D.ietf-httpbis-semantics, Section 11}}) and Cookie ({{?I-D.ietf-httpbis-rfc6265bis}}) header fields will change if the origin (and sometimes path) of the request changes. An application using HTTP should specify if any request header fields that it defines need to be modified or removed upon a redirect; however, this behaviour cannot be relied upon, since a generic client (like a browser) will be unaware of such requirements.
```
mnot, line 410:
```suggestion
Generally, when a redirected request is made, its header fields are copied from the original request's. However, they can be modified by various mechanisms; e.g., sent Authorization ({{I-D.ietf-httpbis-semantics, Section 11}}) and Cookie ({{?I-D.ietf-httpbis-rfc6265bis}}) header fields will change if the origin (and sometimes path) of the request changes. An application using HTTP should specify if any request header fields that it defines need to be modified or removed upon a redirect; however, this behaviour cannot be relied upon, since a generic client (like a browser) will be unaware of such requirements.
```

mnot (changes requested):
Currently, the inline section referencing in kramdown-rfc2629 doesn't support indicators for reference status.

Issue #1363, opened 2021-01-02
digest-headers: support vs SRI


reschke:
"There is a chance that a user agent supporting both mechanisms may find one validates successfully while the other fails. This document specifies no requirements or guidance for user agents that experience such cases."

This makes it sound almost as if it could be ok to ignore an error on one level if there's no error on the other level. I think the spec should say that a user agent supporting both mechanisms must implement all requirements of both specs, no matter what.

LPardue:
Thanks Julian. I agree and will make a PR to tighten up the language.

ioggstream:
@reschke  should we really specify what a SRI implementor should do? 

RFC3230 explicitly states that https://tools.ietf.org/html/rfc3230#section-4.3.2

>   A recipient MAY ignore any or all of the instance-digests in a Digest
>   header field.

and we are not changing this.

PR #1457, opened 2021-04-13
Fix: #1363. Non-normative options in conjuction with SRI.

ioggstream:
## This PR

- clarifies that you can always ignore `Digest`
- you can ignore `Digest` when using SRI
- if you validate both, it is nice to provide useful errors

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -547,10 +547,13 @@ not important. However, a user agent supporting both could benefit from
 performing representation digest validation first because it does not always
 require a conversion into identity encoding.
 
-There is a chance that a user agent supporting both mechanisms may find one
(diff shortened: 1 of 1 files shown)


reschke (approved):
I'm not happy with validation being optional; but that's a separate problem.

With that, I believe this PR is an improvement.

ioggstream:
> I'm not happy with validation being optional

Me neither, but this comes from RFC3230.

reschke:
> Me neither, but this comes from RFC3230.

Well. This is a standards track doc developed in an IETF WG. If we want to change things, we can.

reschke (commented):
That doesn't seem to be producing a list in the actual Internet Draft; see https://httpwg.org/http-extensions/draft-ietf-httpbis-digest-headers.html#section-9.1 - maybe an indentation problem.

Issue #1366, opened 2021-01-02
digest-headers: general feedback on Section 6 [POST, PATCH]


reschke:
> POST and PATCH requests can appear to convey partial representations but are semantically acting on resources. The enclosed representation, including its metadata, refers to that action.

Aren't all HTTP requests "acting on resources"?

> In these requests the representation digest MUST be computed on the representation-data of that action. This is the only possible choice because representation digest requires complete representation metadata (see Section 2).

IMHO, that should be the case for *any* requests (but we already have a ticket for that).

> In responses,
> 
>    if the representation describes the status of the request, Digest MUST be computed on the enclosed representation (see Section 10.8 );
>    if there is a referenced resource Digest MUST be computed on the selected representation of the referenced resource even if that is different from the target resource. That might or might not result in computing Digest on the enclosed representation.
> The latter case might be done according to the HTTP semantics of the given method, for example using the Content-Location header field. In contrast, the Location header field does not affect Digest because it is not representation metadata.

- [x] This is true, but "might be done" is a bit vague. In general, this is the way it is done.

> 6.1. Digest and PATCH

Why do we need this if 6 is about POST and PATCH already?

>In PATCH requests, the representation digest MUST be computed on the patch document because the representation metadata refers to the patch document and not to the target resource (see Section 2 of [PATCH]).

How is this different from POST?

> In PATCH responses, the representation digest MUST be computed on the selected representation of the patched resource.

- [x] No. General rules should apply here. There is no special case for PATCH.

> Digest usage with PATCH is thus very similar to POST, but with the resource's own semantic partly implied by the method and by the patch document.

- [x] Nope.

ioggstream:
Seems we need further discussion here :) We just tried to address POST and PATCH before generic methods, but probably we need to discuss this even further.

ioggstream:
@reschke says:

>> In PATCH responses, the representation digest MUST be computed on the selected representation of the patched resource.

>No. General rules should apply here. There is no special case for PATCH.

Agree, we don't need to be normative here. Just ensure there's an example.

ioggstream:
Mark's feedback.

> * 6. Would this be better titled 'Using Digest in State-Changing Requests'?
ok


> * 6. The first paragraph reads oddly; are we just saying that state-changing requests like POST and PATCH contain a representation that affects the state of the resource?

Yes, but I we wanted to clarify that  (differently from PUT)  the content does not necessarily describe the target resource.

[...]

PR #1459, opened 2021-04-13
Improve POST and PATCH section. See #1366.

ioggstream:
## This PR

improves POST and PATCH section considering some of the hints provided in #1366 

- the fact that PATCH responses require digest to be computed on the representation-data of the patched document is implied by the rest of the document;


mnot:
Can this PR (and branch) be deleted?

ioggstream:
Fixed via the introduction of content-digest #1543

Issue #1302, opened 2020-10-22
rfc6265bis: ABNF should just define field *values* (not the whole field line)


reschke:

miketaylr:
@reschke can you provide some more context here? I don't follow. Thanks.

reschke:
From <https://greenbytes.de/tech/webdav/draft-ietf-httpbis-rfc6265bis-07.html#abnf-syntax>:

```
set-cookie-header = "Set-Cookie:" SP BWS set-cookie-string
set-cookie-string = BWS cookie-pair *( BWS ";" OWS cookie-av )
```

This defines ABNF for a field *line*. This is bad, because field line parsing is generic, and furthermore, there are no field lines in HTTP/2 or HTTP/3.

Just define the ABNF for the field value, consistent with the specs we wrote since ~2014.

chlily1:
Closing, addressed by #1469

PR #1469, opened 2021-04-21
RFC 6265bis: Update ABNF to define field values

chlily1:
This adjusts the ABNF for Set-Cookie and Cookie to define field values, not the whole field line, addressing #1302.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -469,8 +469,7 @@ the user agent.
 
 ### Syntax {#abnf-syntax}
 
-Informally, the Set-Cookie response header contains the header name
-"Set-Cookie" followed by a ":" and a cookie. Each cookie begins with a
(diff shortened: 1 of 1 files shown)


chlily1:
@reschke @englehardt PTAL.

I decided to just refer to set-cookie-string and cookie-string (even though the extra layer of indirection is a bit redundant) because there are still references to those throughout the text and it seemed simpler to keep those.

reschke (commented):
Works for me, but I would prefer to get rid of essentially unused ABNF productions.

I also note that the paragraph above ("Syntax") needs work; it should only talk about the syntax of the field value, and not discuss ": etc.

chlily1:
> I also note that the paragraph above ("Syntax") needs work; it should only talk about the syntax of the field value, and not discuss ": etc.

Thanks for pointing that out! I've updated that as well.

reschke (approved):

englehardt (approved):

PR #1472, opened 2021-04-23
Minor editorial fixes

ioggstream:
## This PR

- content-type / media type
- header / header field


reschke:
This should include updating the core spec references; I will update this PR accordingly unless you get to that before me.

reschke (approved):

PR #1480, opened 2021-04-26
Mark's review on section 5. See #1250

ioggstream:
# This PR

- [x] * 5. I think you should come out and say that digest-algorithm values MUST be compared in a case-insensitive fashion.
- [x]  5. 'The IANA acts a registry' reads oddly; it maintains the registry. Also, saying 'the registry contains the tokens listed below' is misleading; these are initial registrations, which can be modified and added to after the RFC is published.
- [x] * 5. There should not be RFC2219-MUSTs in IANA registry text. 
- [x] This can be addressed by having a requirement similar to "Deprecated digest algorithms MUST NOT be used."

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -347,15 +347,15 @@ Digest-algorithm values are used to indicate a specific digest computation.
    digest-algorithm = token
 ~~~
 
-All digest-algorithm values are case-insensitive
-but lower case is preferred.
+All digest-algorithm token values are case-insensitive
+but lower case is preferred;
+digest-algorithm token values MUST be compared in a case-insensitive fashion.
 
-The Internet Assigned Numbers Authority (IANA) acts as a registry for
+The Internet Assigned Numbers Authority (IANA) maintains a registry for
 digest-algorithm values.
-The registry contains the tokens listed below.
+The registry is initialized with the tokens listed below.
 
-Some digest-algorithms, although registered, rely on vulnerable algorithms
-and MUST not be used:
+Deprecated digest algorithms MUST not be used:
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -347,15 +347,15 @@ Digest-algorithm values are used to indicate a specific digest computation.
...
 
-The Internet Assigned Numbers Authority (IANA) acts as a registry for
+The Internet Assigned Numbers Authority (IANA) maintains a registry for
 digest-algorithm values.
-The registry contains the tokens listed below.
+The registry is initialized with the tokens listed below.
 
-Some digest-algorithms, although registered, rely on vulnerable algorithms
-and MUST not be used:
+Deprecated digest algorithms MUST not be used:
```
LPardue, line 358:
```suggestion
Deprecated digest algorithms MUST NOT be used:
```

LPardue (approved):
lgtm but keen to see mnot's approval too

ioggstream:
- [ ]   5. Also, is the entire checksum value case-insensitive, or just the token identifying the value?

@mnot the checksum value is defined by the digest-algorithm, which defines its case.

- [ ]  5- Also, these might be better placed in IANA Considerations.

@mnot can we just drop any mention of IANA in this section? I like it.


- [ ] * 5. The ability to have a digest algorithm that ignores content-codings is something of a surprise after reading section 2; it needs to be at least mentioned up there.


@mnot it is stated in the introduction https://github.com/httpwg/http-extensions/blame/main/draft-ietf-httpbis-digest-headers.md#L170 . how would you specify it?

mnot (approved):

PR #1477, opened 2021-04-26
digest: editorial over abstract and intro

LPardue:
This change is done while ignoring the fact that we might want a bigger document reshuffle when bringing in a "content-digest" header. But it helps illustrate how the abstract / intro could be editorialized to address some off-list feedback.

PR #1481, opened 2021-04-26
Mark's review: section 1. See #1250

ioggstream:
- [ ] * 1. Introduction: Typically, the introduction contains a 'map' to the rest of the specification, laying out the logical progression of concepts introduced. I can see that 1.2. does that partially, but it needs to be completed and up-front.
- [ ] * 1.1. Is this section necessary? If so, is it necessary for it to be a separate section?
- [x] * 1.2. 'This Proposal' is a very odd name for a specification section
- [x] * 1.2. The first para is difficult to read; suggest starting with 'This document updates the...'
- [x] * 1.2. Use of eg. isn't appropriate; spell out 'for example'
- [x] * 1.3. proposal -> specification (or document)
- [ ] * 1.3. Can we move from lists to prose here, or at least a single type of list?

See https://github.com/httpwg/http-extensions/pull/new/ioggstream-mark-review-3

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -141,20 +141,10 @@ HTTP/1.1 ([RFC7231], Appendix B) obsoleted it:
 [RFC3230] provided a more flexible solution introducing the concept of
 "instance", and the fields `Digest` and `Want-Digest`.
 
-## This Proposal
+## This Document
 
-The concept of `selected representation` defined in Section 3.2 of
-{{!SEMANTICS=I-D.ietf-httpbis-semantics}} makes [RFC3230] definitions inconsistent with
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -141,20 +141,10 @@ HTTP/1.1 ([RFC7231], Appendix B) obsoleted it:
 [RFC3230] provided a more flexible solution introducing the concept of
 "instance", and the fields `Digest` and `Want-Digest`.
 
-## This Proposal
+## This Document
```
LPardue, line 144: This is still a weird section header :smile: 

 https://github.com/httpwg/http-extensions/pull/1477 tackles the problem differently

ioggstream:
Closed in favor of #1477

PR #1512, opened 2021-05-12
Fix layout of table 4

macedogm:
The layout of table 4 is currently broken in the txt and html versions of the spec.

Screenshot 2021-05-13 at 01 25 57

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -326,48 +326,16 @@ Its value is canonicalized as follows:
 
 The following table contains non-normative example HTTP messages and their canonicalized `@request-target` values.
 
-<table>
-    <name>Non-normative examples of <tt>@request-target</tt> canonicalization.</name>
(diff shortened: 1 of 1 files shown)


martinthomson:
I believe that any complaint should be taken to https://trac.tools.ietf.org/tools/xml2rfc/trac/newticket which is where the fault lies.  Perhaps this draft is being a little ambitious in its use of features that are poorly implemented, but this change eliminates semantics that are (it seems) important.

macedogm:
@martinthomson Thanks for pointing me to the right direction. It was my first contribution. I thought that the spec was written in markdown format directly, since other tables are also using md syntax. But the issue, if it's really an issue, might be in the generator instead.

jricher:
This underlying table doesn't use markdown tables because it has too much internal structure. We are looking at a better way to represent these examples in a way that doesn't break the RFC rendering process. :)

Closing this PR but #1530 has been filed to track the problem and its eventual solution.

PR #1529, opened 2021-06-04
Remove List Prefixes

jricher:
Remove the list prefix parameter operation on structured headers. This was previously added to allow for signing of a subset of combined headers, but since the combined header value cannot be considered a structured header value, the prefix functionality did not ultimately serve this purpose.

However, common cases such as adding additional `Via` headers to a request could still be covered by a different mechanism.

Note that dictionary key indexing is still allowed and facilitates multi-signature use cases.
﻿


reschke:
> This was previously added to allow for signing of a subset of combined headers, _but since the combined header value cannot be considered a structured header value_, the prefix functionality did not ultimately serve this purpose.

(emphasis added by me)

That sounds as if a subset of a list is not a valid list? How so?

Issue #1542, opened 2021-06-14
Add Content-Digest field


ioggstream:
## I expect

- [x] content-digest field
- [x] want-content-digest field

## Notes

`Content-Digest` has the same syntax and algorithms of `Digest`.
It is  computed on the actual content bits sent in the request or response.

1. in the general case, eg. a simple GET, Digest and Content-Digest
will convey the same values.

2. When partial representations are conveyed (eg. range-requests),
Digest will contain
the checksum of the whole representation-data, while Content-Digest
will be computed
on the actual exchanged bits.

3. For HEAD responses, Digest will convey the checksum of the selected
representation data, while
Content-Digest, since there's no content, will convey the checksum of
an empty string.

PR #1543, opened 2021-06-14
Fix: #1542. Add Content-Digest.

ioggstream:
## This PR

- [x] adds Content-Digest
- [x] adds Want-Content-Digest
- [x] add examples
- [ ] do we need a further rationale on the discussion?

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -88,7 +88,8 @@ informative:
 
 --- abstract
 
-This document defines the HTTP Digest and Want-Digest fields, which allows
+This document defines the HTTP Digest, Want-Digest,
+Content-Digest and Want-Content-Digest fields, which allow
 client and server to negotiate an integrity checksum of the exchanged resource
 representation data.
 
@@ -135,6 +136,9 @@ This document describes Digest integrity for HTTP and is structured as follows:
   digests,
 - {{digest}} defines the Digest request and response header and trailer field,
 - {{want-digest}} defines the Want-Digest request and response header and
+  trailer field,
+- {{content-digest}} defines the Content-Digest request and response header and trailer field,
+- {{want-content-digest}} defines the Want-Content-Digest request and response header and
   trailer field,
 - {{algorithms}} and {{deprecate-contentMD5}} describe algorithms and their
   relation to Digest,
@@ -189,7 +193,10 @@ HTTP messages or fields. However, it can be combined with other mechanisms that
 protect representation metadata, such as digital signatures, in order to protect
 the phases of an HTTP exchange in whole or in part.
 
-`Digest` does not define means for authentication, authorization or privacy.
+To support use cases where a simple checksum of the content bytes is required,
+this document introduces the `Content-Digest` request and response field.
+
+This specification does not define means for authentication, authorization or privacy.
 
 
 ## Notational Conventions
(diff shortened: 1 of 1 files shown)


LPardue:
I think this is a good start. I took a stab at updating the concept overview to better draw out the differences between Digest and Content-Digest - see https://github.com/httpwg/http-extensions/compare/ioggstream-1542...lucas/content-digest-plusplus?expand=1

Think we should wait for WG feedback before landing this PR into main branch though.

ioggstream:
@LPardue integrated your suggestion in the PR, thanks!

LPardue:
Thanks. (And thanks for putting together the bulk of this pr!)

I think the changes to overview set an example of the kinds of changes we probably need to make throughout the document. But for now what's here is indicative enough of the Content-Digest change to let people assess it does solve their issues.

ioggstream:
After we get consensus on the doc's content, I'm even ok to rewrite everything from scratch :DDD

draft-ietf-httpbis-digest-headers.md
```diff
@@ -318,6 +327,79 @@ Want-Digest: sha-256
 Want-Digest: sha-512;q=0.3, sha-256;q=1, unixsum;q=0
 ~~~
 
+
+# The Content-Digest Field {#content-digest}
```
mnot, line 331: So, strictly speaking, this field name is appropriate according to the current terminology in Semantics. 

What triggered my unease was that the `Content-` prefix is used on several headers that _don't_ refer to the actual content of the message (using the current Semantics definition) -- they refer to the _representation_. See <https://github.com/httpwg/http-core/issues/878>.

I think it's safe to ship this as `Content-Digest` as long as it's being super-precise about what it's referring to.

I could also see flipping `Content-Digest`'s semantics with `Digest`, to align with most (but not all) other uses of the `Content-` prefix. I'm not convinced that doing so is necessary, or even preferable; just mentioning it as something to consider.

ioggstream: Good insight Mark, we'll provide an example of Content-Digest vs Digest in the same message.

About flipping, I think that Content-Digest, being similar to Content-md5 seems to me a more intuitive name.

LPardue: Flipping names seems bad to me. Prefer the way we suggested

b---c: Wouldn't flipping the names be problematic with respect to compatiblity with existing implementations of RFC3230's Digest?

draft-ietf-httpbis-digest-headers.md
```diff
@@ -148,18 +152,38 @@ This document describes Digest integrity for HTTP and is structured as follows:
...
-representation of a resource, for example Range Requests (see Section 14.2 of
-{{SEMANTICS}}).
+field, see {{digest}}. At a high level the value contains a checksum, computed
+over `selected representation data` (Section 3.2;
+{{!SEMANTICS=I-D.ietf-httpbis-semantics}}), that the recipient can use to
+validate integrity. Basing `Digest` on the selected representation makes it
+straightforward to apply it to use-cases where the transferred data requires
+some sort of manipulation to be considered a representation, or conveys a
+partial representation of a resource, for example Range Requests (see Section
+14.2 of {{SEMANTICS}}).
```
MikeBishop, line 162:
```suggestion
some sort of manipulation to be considered a representation or conveys a
partial representation of a resource, such as Range Requests (see Section
14.2 of {{SEMANTICS}}).
```
There are only two items here, so you don't need a comma before "or"; "for example" would need a comma after it as well, and that would be a lot of commas.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -148,18 +152,38 @@ This document describes Digest integrity for HTTP and is structured as follows:
...
+
+To support use cases where a simple checksum of the content bytes is required,
```
MikeBishop, line 164:
```suggestion
To support use-cases where a simple checksum of the content bytes is required,
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -148,18 +152,38 @@ This document describes Digest integrity for HTTP and is structured as follows:
...
+this document introduces the `Content-Digest` request and response field, see
```
MikeBishop, line 165:
```suggestion
this document introduces the `Content-Digest` request and response field; see
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -318,6 +327,79 @@ Want-Digest: sha-256
 Want-Digest: sha-512;q=0.3, sha-256;q=1, unixsum;q=0
 ~~~
 
+
+# The Content-Digest Field {#content-digest}
+
+The `Content-Digest` field contains a comma-separated list of one or more content digest
+values.
+A content digest value is computed applying a digest-algorithm to the actual message content
```
MikeBishop, line 335:
```suggestion
A content digest value is computed by applying a digest-algorithm to the actual message content
```

MikeBishop (approved):
As noted at the interim, this looks like a worthwhile direction.  Minor editorial nits while reading through, but no disagreements with the proposal.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -318,6 +327,79 @@ Want-Digest: sha-256
 Want-Digest: sha-512;q=0.3, sha-256;q=1, unixsum;q=0
 ~~~
 
+
+# The Content-Digest Field {#content-digest}
+
+The `Content-Digest` field contains a comma-separated list of one or more content digest
```
jricher, line 333: Since we're defining a new header here, can we just use structured fields for it entirely? So it'd be a dictionary with the field names as algorithms and the values as binaries.

LPardue: Excellent question. I like the idea but think that it would make it awkward because the algorithms define their encoding. So you'd have to carve out some exceptions and differences, which might catch implementers out.

ioggstream: I wanted to switch to SF, but after a long discussion it was suggested to retain the current grammar,
even because SF do not support uppercase dictionary keys: this is a show stopper wrt being backward compatible.


Moreover  I don't agree wrt using different parsers  for `C-D` and  `D`: it is valuable that the same parser can process both fields.

ioggstream:
@LPardue do we need examples?

ioggstream:
@LPardue should we create a separate example section for Content-Digest or just add some examples in parallel with Digest?

draft-ietf-httpbis-digest-headers.md
```diff
@@ -318,6 +332,79 @@ Want-Digest: sha-256
...
+similarly to `Digest` (see {{digest}})
+
+~~~ http-message
+Content-Digest: sha-256=4REjxQ4yrqUVicfSKYNO/cF9zNj5ANbzgDZt3/h3Qxo=,
+                id-sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
+~~~
+
+A recipient MAY ignore any or all of the content-digests in a Content-Digest
+field. This allows the recipient to choose which digest-algorithm(s) to use for
+validation instead of verifying every received representation-data-digest.
```
LPardue, line 367:
```suggestion
validation instead of verifying every received content-digest.
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -318,6 +332,79 @@ Want-Digest: sha-256
...
+while streaming the content.
+
+# The Want-Content-Digest Field {#want-content-digest}
+
+The `Want-Content-Digest` field indicates the sender's desire to receive a content
+digest on messages associated with the request URI and representation metadata.
+It can be used in both requests and responses.
+
+~~~
+   Want-Content-Digest = 1#want-digest-value
```
LPardue, line 388: Given how similar Want-Digest and Want-Content-Digest are, and that they share the `want-digest-value` we can probably editorialize this a bit. But in order not to block this PR, we can do that work in a follow up.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1290,6 +1313,28 @@ Status:  permanent
 
 Specification document(s):  {{digest}} of this document
 
+## Want-Content-Digest Field Registration
+
+This section registers the `Want-Digest` field in the "Hypertext Transfer
```
LPardue, line 1318:
```suggestion
This section registers the `Want-Content-Digest` field in the "Hypertext Transfer
```

draft-ietf-httpbis-digest-headers.md
ioggstream: filed https://github.com/httpwg/http-extensions/issues/1632

LPardue (approved):

Issue #1340, opened 2020-12-09
[6265bis] Clarification on cookie size limits


miketaylr:
https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-07#section-6.1 states:
>    Practical user agent implementations have limits on the number and
   size of cookies that they can store.  General-use user agents SHOULD
   provide each of the following minimum capabilities:
>      *   At least 4096 bytes per cookie (as measured by the sum of the
      length of the cookie's name, value, and attributes).

IMHO, the prose here is a little confusing. It's clearly talking about limits on size, but then says the "minimum capability should be at least 4096 bytes per cookie". That I think can be cleaned up.

In addition, it's worthwhile looking into the current situation of interop on these limits. @chlily1 pointed me at http://www.ruslog.com/tools/cookies.html which has some interesting (older) research on the topic.

Other related links: 

Chrome considers name, =, value, and attributes when checking the 4096 cap:
https://source.chromium.org/chromium/chromium/src/+/master:net/cookies/parsed_cookie.cc;drc=b35da1f38c91cbc8a3ec0f934b747360fa9a0f73;bpv=1;bpt=1;l=133

Firefox only looks at name and value when checking the 4096 cap, and not attributes.
https://searchfox.org/mozilla-central/rev/ffdb6a4d934b3f5294f18cf0e1df618109ccb36b/netwerk/cookie/CookieCommons.cpp#193-197

But, Firefox enforces a limit on the path attribute:
https://searchfox.org/mozilla-central/rev/ffdb6a4d934b3f5294f18cf0e1df618109ccb36b/netwerk/cookie/CookieCommons.cpp#189-191

So neither Firefox nor Chrome are strictly following 6.1 here (but yes, it's a SHOULD).

Need to look at what Safari and other clients do.

miketaylr:
I forgot to note that interop here seems desirable to prevent using these differences as a browser fingerprinting mechanism.

karlcow:
@miketaylr 

WebKit

https://github.com/WebKit/WebKit/blob/88278b55563e5ccdc0b3419c6c391c3becc19e40/Source/WebCore/platform/network/curl/CookieUtil.cpp#L143-L144

and MAX_COOKIE_LINE is `5000`

https://github.com/WebKit/WebKit/blob/88278b55563e5ccdc0b3419c6c391c3becc19e40/Source/WebCore/platform/network/curl/CookieUtil.cpp#L37-L49

[...]

PR #1563, opened 2021-06-22
Issue #1340 - Update cookie size requirements

miketaylr:
@chlily1 do you think we should move these requirements into the Set-Cookie parsing algorithm and make them `MUST NOT` instead of `SHOULD NOT`? Or fine to leave as-is?

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1759,13 +1759,21 @@ following minimum capabilities:
 
 *   At least 3000 cookies total.
 
+User agents MAY limit the maximum number of cookies they store, and may evict
+any cookie at any time (whether at the request of the user or due to
+implementation limitations).
+
+Note that a limit on the maximum number of cookies also limits the total size of
+the stored cookies, due to the length limits which MUST be enforced in
+{{set-cookie}}.
+
 Servers SHOULD use as few and as small cookies as possible to avoid reaching
 these implementation limits and to minimize network bandwidth due to the
 Cookie header field being included in every request.
 
 Servers SHOULD gracefully degrade if the user agent fails to return one or more
-cookies in the Cookie header field because the user agent might evict any cookie at
-any time on orders from the user.
+cookies in the Cookie header field because the user agent might evict any cookie
+at any time.
 
 ## Application Programming Interfaces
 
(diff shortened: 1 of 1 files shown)


miketaylr:
(It feels easier to write tests against MUST vs SHOULD)

miketaylr:
Actually, yeah. Let's do that. I'll update this PR.

miketaylr:
PTAL @chlily1

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1155,7 +1155,10 @@ parse a set-cookie-string:
 5.  Remove any leading or trailing WSP characters from the name string and the
     value string.
 
-6.  The cookie-name is the name string, and the cookie-value is the value string.
+6.  If the sum of cookie-name and cookie-value is more than 4096 bytes, abort
```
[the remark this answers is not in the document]
miketaylr: Oops, good catch.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1749,9 +1755,6 @@ Practical user agent implementations have limits on the number and size of
 cookies that they can store. General-use user agents SHOULD provide each of the
```
chlily1, line 1755: Just to be explicit, maybe add a sentence mentioning that the user agent MAY limit the max number of cookies they store, and may evict any cookie at any time (whether on orders from the user or due to implementation limitations). And also explicitly say that a limit on the max number of cookies also limits the total size of the stored cookies, due to the length limits which must be enforced in the parsing step.

chlily1 (commented):
Algorithm changes look good. Had a suggestion for the "Limits" section for clarity.

draft-ietf-httpbis-rfc6265bis.md
miketaylr: good suggestion, thx.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1749,13 +1755,18 @@ Practical user agent implementations have limits on the number and size of
...
 
-*   At least 4096 bytes per cookie (as measured by the sum of the length of the
-    cookie's name, value, and attributes).
-
 *   At least 50 cookies per domain.
 
 *   At least 3000 cookies total.
 
+User agents MAY limit the maximum number of cookies they store, and may evict
+any cookie at any time (whether at the request of the user or due to
```
chlily1, line 1763: nit: can you delete the "on orders from the user" on line 1776 since it now mentions that here?

miketaylr: done.

chlily1:
Thanks! LGTM.

PR #1565, opened 2021-06-29
added new specialty content fields, fix request-target

jricher:
This adds several new specialty identifiers for requests and responses to cover the method, authority, scheme, origin, path, query, and status code values of the HTTP message.

It also edits the `request-target` identifier's associated value to remove the verb and the string construction rules and instead align with HTTP Semantics.

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -49,6 +49,7 @@ author:
 
 normative:
     RFC2104:
+    RFC3986:
     FIPS186-4:
         target: https://csrc.nist.gov/publications/detail/fips/186/4/final
         title: Digital Signature Standard (DSS)
@@ -57,6 +58,12 @@ normative:
         target: https://pubs.opengroup.org/onlinepubs/9699919799/
         title: The Open Group Base Specifications Issue 7, 2018 edition
         date: 2018
+    SEMANTICS: I-D.ietf-httpbis-semantics
+    MESSAGING: I-D.ietf-httpbis-messaging
+    HTMLURL:
+        target: https://url.spec.whatwg.org/
+        title: URL (Living Standard)
+        date: 2021
 
 informative:
     RFC6234:
@@ -139,9 +146,9 @@ Additionally, all changes to content not covered by the signature are considered
 The terms "HTTP message", "HTTP request", "HTTP response",
 `absolute-form`, `absolute-path`, "effective request URI",
 "gateway", "header field", "intermediary", `request-target`,
-"sender", and "recipient" are used as defined in {{!MESSAGING=RFC7230}}.
+"sender", and "recipient" are used as defined in {{MESSAGING}}.
 
-The term "method" is to be interpreted as defined in Section 4 of {{!SEMANTICS=RFC7231}}.
+The term "method" is to be interpreted as defined in Section 4 of {{SEMANTICS}}.
 
 For brevity, the term "signature" on its own is used in this document to refer to both digital signatures and keyed MACs.  Similarly, the verb "sign" refers to the generation of either a digital signature or keyed MAC over a given input string.  The qualified term "digital signature" refers specifically to the output of an asymmetric cryptographic signing operation.
 
@@ -177,7 +184,7 @@ This document contains non-normative examples of partial and complete HTTP messa
 
 HTTP Message Signatures are designed to be a general-purpose security mechanism applicable in a wide variety of circumstances and applications. In order to properly and safely apply HTTP Message Signatures, an application or profile of this specification MUST specify all of the following items:
 
-- The set of [content identifiers](#covered-content) that are expected and required. For example, an authorization protocol could mandate that the `Authorization` header be covered to protect the authorization credentials and mandate the signature parameters contain a `created` parameter, while an API expecting HTTP message bodies could require the `Digest` header to be present and covered. 
+- The set of [content identifiers](#covered-content) that are expected and required. For example, an authorization protocol could mandate that the `Authorization` header be covered to protect the authorization credentials and mandate the signature parameters contain a `created` parameter, while an API expecting HTTP message bodies could require the `Digest` header to be present and covered.
 - A means of retrieving the key material used to verify the signature. An application will usually use the `keyid` parameter of the signature parameters ({{signature-params}}) and define rules for resolving a key from there, though the appropriate key could be known from other means.
 - A means of determining the signature algorithm used to verify the signature content is appropriate for the key material. For example, the process could use the `alg` parameter of the signature parameters ({{signature-params}}) to state the algorithm explicitly, derive the algorithm from the key material, or use some pre-configured algorithm agreed upon by the signer and verifier.
 - A means of determining that a given key and algorithm presented in the request are appropriate for the request being made. For example, a server expecting only ECDSA signatures should know to reject any RSA signatures, or a server expecting asymmetric cryptography should know to reject any symmetric cryptography.
@@ -193,7 +200,7 @@ In order to allow signers and verifiers to establish which content is covered by
 Some content within HTTP messages can undergo transformations that change the bitwise value without altering meaning of the content (for example, the merging together of header fields with the same name).  Message content must therefore be canonicalized before it is signed, to ensure that a signature can be verified despite such intermediary transformations. This document defines rules for each content identifier that transform the identifier's associated content into such a canonical form.
 
 Content identifiers are defined using production grammar defined by [RFC8941, Section 4](#RFC8941).
-The content identifier is an `sf-string` value. The content identifier
+The content identifier itself is an `sf-string` value. The content identifier
 type MAY define parameters which are included using the `parameters` rule.
 
 ~~~ abnf
@@ -277,7 +284,7 @@ The following table shows example canonicalized values for different content ide
 
 ## Specialty Content Fields {#specialty-content}
 
-Content not found in an HTTP header can be included in the signature base string by defining a content identifier and the canonicalization method for its content. 
+Content not found in an HTTP header can be included in the signature base string by defining a content identifier and the canonicalization method for its content.
 
 To differentiate specialty content identifiers from HTTP headers, specialty content identifiers MUST start with the "at" `@` character. This specification defines the following specialty content identifiers:
 
@@ -288,41 +295,44 @@ To differentiate specialty content identifiers from HTTP headers, specialty cont
 : The method used for a request. ({{content-request-method}})
 
 @target-uri
-: The target URI for a request. ({{content-target-uri}})
+: The full target URI for a request. ({{content-target-uri}})
 
 @authority
-: The authority for a request. ({{content-request-authority}})
+: The authority of the target URI for a request. ({{content-request-authority}})
 
 @scheme
-: The scheme of the requested URI. ({{content-request-scheme}})
+: The scheme of the target URI for a request. ({{content-request-scheme}})
 
 @request-target
 : The request target. ({{content-request-target}})
 
 @path
-: The absolute path portion of the request target. ({{content-request-path}})
+: The absolute path portion of the target URI for a request. ({{content-request-path}})
 
-@query-param
-: The parsed query parameters of the request target. ({{content-request-query}})
+@query
+: The query portion of the target URI for a request. ({{content-request-query}})
 
-@status-code
+@query-params
+: The parsed query parameters of the target URI for a request. ({{content-request-query-params}})
+
+@status
 : The status code for a response. ({{content-status-code}}).
 
 Additional specialty content identifiers MAY be defined and registered in the HTTP Signatures Specialty Content Identifier Registry. ({{content-registry}})
 
 ### Signature Parameters {#signature-params}
 
-HTTP Message Signatures have metadata properties that provide information regarding the signature's generation and/or verification.
+HTTP Message Signatures have metadata properties that provide information regarding the signature's generation and/or verification, such as the list of covered content, a timestamp, identifiers for verification key material, and other utilities.
 
 The signature parameters specialty content is identified by the `@signature-params` identifier.
 
 Its canonicalized value is the serialization of the signature parameters for this signature, including the covered content list with all associated parameters.
 
-* `alg`: The HTTP message signature algorithm from the HTTP Message Signature Algorithm Registry, as an `sf-string` value.
-* `keyid`: The identifier for the key material as an `sf-string` value.
 * `created`: Creation time as an `sf-integer` UNIX timestamp value. Sub-second precision is not supported.
 * `expires`: Expiration time as an `sf-integer` UNIX timestamp value. Sub-second precision is not supported.
 * `nonce`: A random unique value generated for this signature.
+* `alg`: The HTTP message signature algorithm from the HTTP Message Signature Algorithm Registry, as an `sf-string` value.
+* `keyid`: The identifier for the key material as an `sf-string` value.
 
 Additional parameters can be defined in the [HTTP Signature Parameters Registry](#iana-param-contents).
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-message-signatures.md
```diff
@@ -382,7 +343,153 @@ This example shows a canonicalized value for the parameters of a given signature
...
+
+Would result in the following `@method` value:
+
+~~~
+"@method": POST
+~~~
+
+### Authority {#content-request-authority}
+
+The `@authority` identifier refers to the authority component of the target of the HTTP request message. In HTTP 1.1, this is usually conveyed using the `Host` header, while in HTTP 2 and HTTP 3 it is conveyed using the `:authority` pseudo-header. The value is the fully-qualified authority component of the request, comprised of the host and, optionally, port of the request target, as a string.
```
mnot, line 367: It'd probably be good to reference the relevant sections, to avoid ambiguity about what value to use.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -382,7 +343,153 @@ This example shows a canonicalized value for the parameters of a given signature
...
+
+~~~
+OPTIONS * HTTP/1.1
+Host: www.example.com
+~~~
+
+Would result in the following `@request-origin` value:
+
+~~~
+"@request-origin": *
```
mnot, line 426: Would it?

Also, it might be good to talk about what distinguishes `@request-origin` from `@authority`.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -382,7 +343,153 @@ This example shows a canonicalized value for the parameters of a given signature
...
+~~~
+
+### Path {#content-request-path}
+
+The `@path` identifier refers to the target path of the HTTP request message. The value is the absolute path of the request target, with no query parameters and no trailing `?` character.
```
mnot, line 431: Best to use "query component" throughout (and maybe reference RFC3986); they're not always parameterised in the `a=b&c=d` sense.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -382,7 +343,153 @@ This example shows a canonicalized value for the parameters of a given signature
...
+Host: www.example.com
+~~~
+
+Would result in the following `@query` value:
+
+~~~
+"@query": param=value&foo=bar&baz=batman
+~~~
+
+The identifier supports an OPTIONAL `key` parameter containing the name of a single query parameter. When this parameter is present, the value is the normalized key-value pair of the named query parameter, not including any leading `?` characters or separating `&` characters.
```
mnot, line 463: Hmm. This format is specific to HTML, although used in a very wide fashion. At a minimum, we need to mention that this is meaningful when that format is use, with a reference to the appropriate part of url.spec.whatwg.org.

draft-ietf-httpbis-message-signatures.md
jricher: I wasn't sure about this -- it's supposed to be the path+query portion of the request, the bit after the verb on the HTTP/1.1 request line.

draft-ietf-httpbis-message-signatures.md
jricher: Good point, will do.

draft-ietf-httpbis-message-signatures.md
jricher: Good point. It's wide enough that most developers don't make that distinction, but we should for sure reference the WHATWG spec here specifically. Would a different name help that, like `query-parameters`?

draft-ietf-httpbis-message-signatures.md
mnot: Ah. 'origin' is the (scheme, host, port) tuple. If you want path+query, we should invent another term.

draft-ietf-httpbis-message-signatures.md
mnot: Hmm, not sure.

jricher:
@mnot I haven't added references in yet but I tried to move the language in the direction you are suggesting. 

@richanna I added the target-uri field we discussed, please review the text.

draft-ietf-httpbis-message-signatures.md
jricher: Changed the name back to request-target but with largely the same value definition as here.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -275,74 +277,38 @@ The following table shows example canonicalized values for different content ide
...
 @signature-params
 : The signature metadata parameters for this signature. ({{signature-params}})
 
-Additional specialty content identifiers MAY be defined and registered in the HTTP Signatures Specialty Content Identifier Registry. ({{content-registry}})
+@method
+: The method used for a request. ({{content-request-method}})
 
-### Request Target {#content-request-target}
+@target-uri
+: The target URI for a request. ({{content-target-uri}})
```
martinthomson, line 291: absolute?

draft-ietf-httpbis-message-signatures.md
```diff
@@ -275,74 +277,38 @@ The following table shows example canonicalized values for different content ide
...
-    </tbody>
-</table>
+@authority
+: The authority for a request. ({{content-request-authority}})
+
+@scheme
+: The scheme of the requested URI. ({{content-request-scheme}})
+
+@request-target
+: The request target. ({{content-request-target}})
```
martinthomson, line 300: I like that you have a shortcut for the whole URI (request-uri) and the  individual components (scheme, authority, path, query-param), but that leaves `@request-target` as a bit of an odd medium.  Perhaps it's time to deprecate it.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -275,74 +277,38 @@ The following table shows example canonicalized values for different content ide
...
+
+@path
+: The absolute path portion of the request target. ({{content-request-path}})
+
+@query-param
+: The parsed query parameters of the request target. ({{content-request-query}})
+
+@status-code
+: The status code for a response. ({{content-status-code}}).
```
martinthomson, line 309: Save some bytes: "@status" ?

draft-ietf-httpbis-message-signatures.md
martinthomson: You also need to be careful about c14n here.  Is this as it appears in the protocol, verbatim, or are you expected to case-normalize and remove (or add) any explicit default port value?

The document uses structured field canonicalization, but that is incomplete when it comes to important fields like authority, which effectively has its own c14n rules.

This isn't a new issue of course, but the breakdown here makes it more obvious.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -382,7 +348,209 @@ This example shows a canonicalized value for the parameters of a given signature
...
+
+~~~
+POST /path?param=value HTTP/1.1
+Host: www.example.com
+~~~
+
+Would result in the following `@scheme` value:
+
+~~~
+"@scheme": http
```
martinthomson, line 424: See above regarding c14n.  You need to choose a case here because some versions of HTTP don't carry a value.

draft-ietf-httpbis-message-signatures.md
ioggstream: While this PR goes in the right direction, I think that we should limit to resource identification (eg. target-uri).

draft-ietf-httpbis-message-signatures.md
```diff
@@ -382,7 +343,153 @@ This example shows a canonicalized value for the parameters of a given signature
...
+Host: www.example.com
+~~~
+
+Would result in the following `@path` value:
+
+~~~
+"@path": /path
+~~~
+
+### Query Components {#content-request-query}
```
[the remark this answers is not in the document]
jricher: That's why we're proposing them as separate parts - if you want to handle them differently, you can. If you don't, just use the string. It's about giving developers the tools to find the pieces that fit the best for their use case.

draft-ietf-httpbis-message-signatures.md
jricher: Yes, I believe that was the intent here, we can add that to the definition.

draft-ietf-httpbis-message-signatures.md
jricher: Perhaps, the OAuth PoP draft didn't have it. I kept it in here because it's relatively easy to get from an HTTP/1.1 message's request line. We'd basically be saying "anything past the authority part of the URL", really.

draft-ietf-httpbis-message-signatures.md
jricher: @ioggstream We need to consider how people use this in the wild, not just the purity of HTTP semantics. The fact is that people do use HTML query parameters all the time, and the overwhelming majority of developers even consider that to be part of HTTP, not HTML. I do think we need to allow for signing of the query string as-is, but I support signing of query parameters when they exist.

draft-ietf-httpbis-message-signatures.md
jricher: If there are good canonicalization rules we can point to from another document, that would be best to do -- we don't want to invent a new process here if we don't need to.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -382,7 +348,209 @@ This example shows a canonicalized value for the parameters of a given signature
...
+Host: www.example.com
+~~~
+
+Would result in the following `@query` value:
+
+~~~
+"@query": param=value&foo=bar&baz=batman
+~~~
+
+If the request uses HTML form parameters in the query string, this identifier allows addressing of individual query parameters by using an OPTIONAL `key` parameter containing the name of a single query parameter. When this parameter is present, the value is the normalized key-value pair of the named query parameter, separated by an equals sign, not including any leading `?` characters or separating `&` characters.
```
MikeBishop, line 520: I think you need a reference to "HTML form parameters" here.

draft-ietf-httpbis-message-signatures.md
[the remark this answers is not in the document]
MikeBishop: I share the unease.  Are duplicate parameter keys prohibited?  If they aren't, how does the signature pick which instance is being signed?

draft-ietf-httpbis-message-signatures.md
martinthomson: It is less easy to get from HTTP/2 and HTTP/3.  It's somewhat fragile as a consequence of this depending on HTTP/1.1-specific rules for request construction.

draft-ietf-httpbis-message-signatures.md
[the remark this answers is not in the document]
martinthomson: The proposal is to leave the string untouched, so if the query string is composed of key value pairs (like might be generated by an HTML form submission), then the ordering is preserved and any duplicates remain.  I don't have any particular problem with that, with one minor exception:

`https://example.com/path?` is - I believe - a valid URI, but indistinguishable from `https://example.com/path` if the different components are included.  I don't have a way to exploit that, but it suggests that maybe `@query` needs to include the initial `?`.

draft-ietf-httpbis-message-signatures.md
jricher: Good to know about the distinctions. Since each of these identifiers is optional for use we can point out the difficulty in constructing it for HTTP/2 and HTTP/3 in the text, so people working in that space can choose the individual components instead.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -275,74 +277,38 @@ The following table shows example canonicalized values for different content ide
...
+@scheme
+: The scheme of the requested URI. ({{content-request-scheme}})
+
+@request-target
+: The request target. ({{content-request-target}})
+
+@path
+: The absolute path portion of the request target. ({{content-request-path}})
+
+@query-param
```
MikeBishop, line 305: By analogy with the others, how about "@query"?

draft-ietf-httpbis-message-signatures.md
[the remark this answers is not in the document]
MikeBishop: @martinthomson, the concern is about the behavior on line 520:

> If the request uses HTML form parameters in the query string, this identifier allows addressing of individual query parameters by using an OPTIONAL `key` parameter containing the name of a single query parameter. When this parameter is present, the value is the normalized key-value pair of the named query parameter, separated by an equals sign, not including any leading `?` characters or separating `&` characters.

That's not the string untouched, but a part selected by parsing the string.  That's the bit I'm cautious about.

draft-ietf-httpbis-message-signatures.md
jricher: I think that makes sense for the query string as-is. Maybe I was overloading too much and it makes sense to split out the HTML-flavored query parameters portion to its own identifier, in that case.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -382,7 +348,209 @@ This example shows a canonicalized value for the parameters of a given signature
...
+
+~~~
+"@scheme": http
+~~~
+
+If used in a response message, the `@scheme` identifier refers to the associated value of the request that triggered the response message being signed.
+
+### Request Target {#content-request-target}
+
+The `@request-target` identifier refers to the full request target of the HTTP request message. The value of the request target can take different forms, depending on the type of request.
```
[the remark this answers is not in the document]
jricher: You always need to be able to accurately re-create the signing input, so if that's difficult for a given API/application/deployment they wouldn't sign that part. If you have to start guessing for ANYTHING it's a bad idea to sign it. We haven't filled out the security considerations section yet but this is a key point for this protocol.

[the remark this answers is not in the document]
MikeBishop: Right, a signature is useless if the recipient can't verify it reliably, and since the elements aren't signed separately, including any element which can't be reliably reconstructed renders the entire signature meaningless.  I think you're assuming a closed system, where the server won't bother to sign something it knows some of its clients can't verify.  I'm thinking about a more open system where the server won't necessarily know the client capabilities; we should build something where the client is highly likely to be able to verify reliably.

draft-ietf-httpbis-message-signatures.md
yaronf: And what about "@reason-phrase"? RFC 7230 says clients SHOULD ignore it, but they very often don't.

reschke: If they do not ignore it, they are broken already.

jricher: Agree that signing the reason phrase is both fragile and against best HTTP practices.

draft-ietf-httpbis-message-signatures.md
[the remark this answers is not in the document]
jricher: It's feeling to me that the query-untouched and query-as-html-parameters ought to be different identifiers and not overloading a single identifier, which is easy enough to split out. I agree with the concerns about signing individual parameters in a lot of cases, but I also think that the tradeoffs are going to be worth it for a lot of applications.

jricher:
Responded to comments and updated normalization rules, please re-review.

draft-ietf-httpbis-message-signatures.md
jricher: I added warnings about this identifier in its definition below and will incorporate it in the next version for discussion, if the group chooses to remove it we can do so.

draft-ietf-httpbis-message-signatures.md
jricher: These instructions have been added, pointing to the canonicalization rules in SEMANTICS specifically, which reference 3986 in turn.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1043,25 +1279,24 @@ This results in the following `Signature-Input` and `Signature` headers being ad
...
+  cZgLxVwialuH5VnqJS4JN8PHD91XLfkjMscTo4jmVMpFd3iLVe0hqVFl7MDt6TMkw\
+  IyVFnEZ7B/VIQofdShO+C/7MuupCSLVjQz5xA+Zs6Hw+W9ESD/6BuGs6LF1TcKLxW\
+  +5K+2zvDY/Cia34HNpRW5io7Iv9/b7iQ==:
 ~~~
 
-### Header Coverage using rsa-pss-sha512
+### Selective Coverage using rsa-pss-sha512
 
-This example covers all the specified headers in `test-request` except for the body digest header using the `rsa-pss-sha512` algorithm.
+This example covers additional elements in `test-request`, not including the body Digest header using the `rsa-pss-sha512` algorithm.
```
jricher, line 1292:
```suggestion
This example covers additional elements in `test-request`, not including the body Digest header, using the `rsa-pss-sha512` algorithm.
```

jricher:
Folding in this PR to the main document, discussion can continue on specifics of identifiers and their values as needed.

PR #1566, opened 2021-06-30
Add request/response binding

jricher:
This creates a mechanism to cryptographically bind a response to its associated request by letting the responder sign the signature value of the request, and the requester validate the signature based on their original request's signature.﻿

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -622,7 +622,7 @@ When a signed request message results in a signed response message, the `@reques
 `key`
 : Identifies which signature from the response to sign.
 
-The value MUST contain the `sf-binary` representation of the signature value of the referenced request identified by the `key` parameter.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-message-signatures.md
```diff
@@ -384,6 +387,82 @@ This example shows a canonicalized value for the parameters of a given signature
...
+
+This would result in the following unsigned response message:
+
+~~~ http-message
+HTTP/1.1 200 OK
+Date: Tue, 20 Apr 2021 02:07:56 GMT
+Content-Type: application/json
+Content-Length: 68
+
+{"from": "world", "busy": true, "your call is very important to us"}
```
bc-pi, line 427: this here JSON isn't valid

jricher: Do I look like a parser to you? :P

Thanks, I threw the signing test together quickly, and I'll fix that.

jricher: Updated with more correcter JSON.

Issue #1579, opened 2021-07-22
RFC 6265bis: Cookie length constraints only apply to Set-Cookie header parsing


recvfrom:
RFC6265bis mentions the cookie length constraints in relation to parsing the Set-Cookie headers but doesn't in the more general Storage Model section.  From chatting with @chlily1 about this, it'd be desirable to have these constraints be listed in that section as well.  Since there are other ways cookies can come in to the cookie store (Ex: Cookie Store API), having these constraints be enforced when adding cookies from any source to the cookie store would better ensure cookie format consistency.

PR #1581, opened 2021-07-23
RFC 6565bis: Add length limits to storage model algorithm

chlily1:
Addresses #1579

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1152,7 +1152,7 @@ parse a set-cookie-string:
     value string.
 
 5.  If the sum of the lengths of the name string and the value string is more
-    than 4096 bytes, abort these steps and ignore the set-cookie-string entirely.
+    than 4096 octets, abort these steps and ignore the set-cookie-string entirely.
 
 6.  The cookie-name is the name string, and the cookie-value is the value string.
 
@@ -1190,7 +1190,7 @@ parse the unparsed-attributes:
 5.  Remove any leading or trailing WSP characters from the attribute-name
     string and the attribute-value string.
 
-6.  If the attribute-value is longer than 1024 bytes, ignore the cookie-av
+6.  If the attribute-value is longer than 1024 octets, ignore the cookie-av
     string and return to Step 1 of this algorithm.
 
 7.  Process the attribute-name and attribute-value according to the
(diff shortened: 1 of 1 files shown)


chlily1:
@miketaylr @sbingler PTAL

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1406,10 +1406,13 @@ user agent MUST process the cookie as follows:
 3.  If the cookie-name or the cookie-value contains a %x00-1F / %x7F (CTL)
     character, abort these steps and ignore the cookie entirely.
 
-4.  Create a new cookie with name cookie-name, value cookie-value. Set the
+4.  If the sum of the lengths of cookie-name and cookie-value is more than
+    4096 bytes, abort these steps and ignore the cookie entirely.
```
sbingler, line 1410: Recommend using the term "octet(s)" instead of "byte(s)" to match other usages. 
The line under 6.1 should also be updated.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1479,28 +1483,29 @@ user agent MUST process the cookie as follows:
 
     2.  Set the cookie's domain to the canonicalized request-host.
 
-9.  If the cookie-attribute-list contains an attribute with an
+10. If the cookie-attribute-list contains an attribute with an
     attribute-name of "Path", set the cookie's path to attribute-value of
     the last attribute in the cookie-attribute-list with an attribute-name
```
sbingler, line 1488: For clarity I recommend "the last attribute in the cookie-attribute-list with **both** an attribute-name..."

sbingler (changes requested):
Looks pretty good, just a couple comments.

draft-ietf-httpbis-rfc6265bis.md
miketaylr: Plot twist: _both_ reads as less clear to me (but is probably fine 🙈 )

draft-ietf-httpbis-rfc6265bis.md
miketaylr: I don't have strong opinions on bytes vs octets, but if you change it, it's probably worth changing for the other length limits (in parsing).

miketaylr (approved):

draft-ietf-httpbis-rfc6265bis.md
chlily1: Thanks, "octets" is probably better since elsewhere it says
```
NOTE: Despite its name, the cookie-string is actually a sequence of
   octets, not a sequence of characters.
```

draft-ietf-httpbis-rfc6265bis.md
chlily1: They sound about equally clear to me, but I went with "both". Thanks!

sbingler: Glad we covered all our bases then

draft-ietf-httpbis-rfc6265bis.md
sbingler: Thanks, and you can ignore my comment about 6.1 I was looking at an outdated version.

Issue #1210, opened 2020-05-30
Note that cookie serialisation algorithm may produce output that doesn't match cookie-string


gsnedders:
Ideally we'd define server-side parsing of cookie-string (or at least recommend behaviour here for things that don't match cookie-string!), given this does cause problems for some servers (e.g. https://github.com/aio-libs/aiohttp/issues/2683).

Until then we should at least make it clear that the "the algorithm to compute the cookie-string from a cookie store and a request-uri" can create things that don't match cookie-string (e.g., nameless cookies after #1144). There's a number of WPT expected results that don't match cookie-string.

miketaylr:
> Until then we should at least make it clear that the "the algorithm to compute the cookie-string from a cookie store and a request-uri" can create things that don't match cookie-string (e.g., nameless cookies after #1144).

Can you clarify what a fix would look like? Bonus points for a PR. 😅

gsnedders:
> Can you clarify what a fix would look like? Bonus points for a PR. 😅
> 
> (I mean, I can also attempt a PR, just not clear on what the recommended fix would be)

You attempting would probably be as good as me attempting!

miketaylr:
I'm skeptical, however willing to try!

And now I think I actually understand the bug report (and didn't when I commented). @gsnedders is saying that the ABNF in https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-07#section-4.2.1 has a bug, since `cookie-pair` is defined as:

`   cookie-pair       = cookie-name BWS "=" BWS cookie-value`

But nameless and valueless cookies exist.

cf. 
https://github.com/httpwg/http-extensions/issues/1119
https://github.com/httpwg/http-extensions/issues/1074

miketaylr:
I *think* maybe this can be solved by making the following changes:

This is pretty simple to grok:
`cookie-pair = cookie-name cookie-value`

This is supposed to convey "a nameless cookie, OR a valueless cookie (but not both)"
`cookie-name = 1*cookie-octet / (*cookie-octet BWS "=" BWS)`

That should work for the following valid cookie-pairs:

`f;` (nameless)
`=f;` (nameless2)
`f=;` (named cookie with empty `cookie-value`, i.e., valueless)
`f=1` (name + value)

If that's the route we go, then in https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-07#section-4.2.1

cookie-string = cookie-pair *( ";" SP cookie-pair )

That should work.

[...]

PR #1592, opened 2021-08-05
RFC6265bis: Improve handling of nameless cookies when serializing

recvfrom:
Update the cookie serialization logic so that cookies without a name and that have a '=' character in the value aren't serialized in a way that would cause them to be interpreted incorrectly.

See https://github.com/httpwg/http-extensions/issues/1210#issuecomment-893863325 for more details

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1735,12 +1735,15 @@ cookie-string from a given cookie store.
 4. Serialize the cookie-list into a cookie-string by processing each cookie
    in the cookie-list in order:
 
-   1.  If the cookies' name is not empty, output the cookie's name followed by
-       the %x3D ("=") character.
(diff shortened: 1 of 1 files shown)


chlily1 (commented):
Hmm, I think we want some metrics here to evaluate the impact of this change, and ideally some data about how various servers may or may not choke on the new serialization.

Assuming we do decide to go forward this, just have one suggestion.

cc @sbingler @miketaylr

sbingler:
The language looks more clear now.
The conditional prefixing of '=' bothers me but I agree with https://github.com/httpwg/http-extensions/issues/1210#issuecomment-893891032 that always adding the '=' could break even more sites.
The later comment about sites purposefully depending on `Set-Cookie: =foo=bar` as being sent as `foo=bar` is hacky and awful and very likely occurring in the wild.

Anyway, we should for sure get some metrics.

PR #1594, opened 2021-08-06
Add "Accept-Signature" header definition

### LINKED ISSUE #1207: Signature negotiation
## Question

Do we want to support negotiation, eg. [see webpack Accept-Signature](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#name-the-accept-signature-header) ?

jricher:
Adds an "Accept-Signature" header field for a potential verifier to signal to a potential signer that a signature is requested in the next message, either as a response to a request or in the next request following a response. 

Closes #1207


aljones15 (approved):
This is pretty clear. Not sure my opinion counts much here, but it is a good idea.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1048,6 +1048,47 @@ Signature: sig1=:P0wLUszWQjoi54udOtydf9IWTfNhy+r53jGFj9XZuP4uKwxyJo\
 
 The proxy's signature and the client's original signature can be verified independently for the same message, depending on the needs of the application.
 
+# Requesting Signatures
+
+While a signer is free to attach a signature to a request or response without prompting, it is often desirable for a potential verifier to signal that it expects a signature from a potential signer using the `Accept-Signature` header.
+
+The message to which the requested signature is applied is known as the "target message". When the `Accept-Signature` header is sent in an HTTP Request message, the header indicates that the client desires the server to sign the response using the identified parameters. When used in a response, the header indicates that the server desires the client to sign its next request to the server with the identified parameters.
```
mnot, line 1055: Is it just limited to the next request (e.g., after redirection if it occurs in a 3xx) or is it _subsequent_ requests?

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1048,6 +1048,47 @@ Signature: sig1=:P0wLUszWQjoi54udOtydf9IWTfNhy+r53jGFj9XZuP4uKwxyJo\
...
+
+The target message of an `Accept-Signature` MUST include all labeled signatures, each covering the same identified content of the `Accept-Signature` field.
+
+The sender of an `Accept-Signature` header MUST include identifiers that are appropriate for the type of the target message. For example, if the target message is a response, the identifiers can not include the `@status` identifier. 
+
+
+## The Accept-Signature Header {#accept-signature-header}
+
+The `Signature-Input` HTTP header field is a Dictionary Structured Header {{!RFC8941}} containing the metadata for one or more requested message signatures to be generated from content within the target HTTP message. Each member describes a single message signature. The member's name is an identifier that uniquely identifies the requested message signature within the context of the target HTTP message. The member's value is the serialization of the desired covered content of the target message, including any allowed signature metadata parameters, using the serialization process defined in {{signature-params}}.
```
mnot, line 1064:
```suggestion
The `Accept-Signature` HTTP header field is a Dictionary Structured Header {{!RFC8941}} containing the metadata for one or more requested message signatures to be generated from content within the target HTTP message. Each member describes a single message signature. The member's name is an identifier that uniquely identifies the requested message signature within the context of the target HTTP message. The member's value is the serialization of the desired covered content of the target message, including any allowed signature metadata parameters, using the serialization process defined in {{signature-params}}.
```

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1048,6 +1048,47 @@ Signature: sig1=:P0wLUszWQjoi54udOtydf9IWTfNhy+r53jGFj9XZuP4uKwxyJo\
...
+    1. The name of the member is the label of the output signature as specified in {{signature-input-header}}
+    2. Parse the value of the member to obtain the list of covered content identifiers
+    3. Process the requested parameters, such as the signing algorithm and key material. If any requested parameters cannot be fulfilled, or if the requested parameters conflict with those deemed appropriate to the target message, the process fails and returns an error.
+    4. Select any additional parameters necessary for completing the signature
+    5. Create the `Signature-Input` and `Signature` header values and associate them with the label
+3. Optionally create any additional `Signature-Input` and `Signature` values, with unique labels not found in the `Accept-Signature` field
+4. Combine all labeled `Signature-Input` and `Signature` values and attach both headers to the target message
+
+Note that by this process, a signature applied to a target message MUST have the same label, MUST have the same covered content list, and MAY have additional parameters. Also note that the target message MAY include additional signatures not specified by the `Accept-Signature` field.
+
```
mnot, line 1091: It might be good to mention that responses which are the result of negotiation need to either be uncacheable, or carry a `Vary` header field.

mnot (commented):
A couple of comments

draft-ietf-httpbis-message-signatures.md
jricher: That's a good question. My thought was that it was the next request, not all subsequent requests. Unlike authentication, I don't think I'd necessarily expect signatures to be applied all the time for a given service. I'd like to hear more voices on that though because I could be limited by my own experience.

draft-ietf-httpbis-message-signatures.md
jricher: Copypaste, thanks!

draft-ietf-httpbis-message-signatures.md
jricher: What's the rationale for this proposal? I'm not sure I follow.

mnot: See [here](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#field.vary). Basically, if a request header is used to select a response, you need to do one of these things; otherwise a cache could serve the wrong one to subsequent clients.

jricher: Thanks, that makes sense! I'll add text.

jricher:
Updated text based on feedback.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1048,6 +1051,49 @@ Signature: sig1=:P0wLUszWQjoi54udOtydf9IWTfNhy+r53jGFj9XZuP4uKwxyJo\
 
 The proxy's signature and the client's original signature can be verified independently for the same message, depending on the needs of the application.
 
+# Requesting Signatures
+
+While a signer is free to attach a signature to a request or response without prompting, it is often desirable for a potential verifier to signal that it expects a signature from a potential signer using the `Accept-Signature` field.
+
+The message to which the requested signature is applied is known as the "target message". When the `Accept-Signature` field is sent in an HTTP Request message, the field indicates that the client desires the server to sign the response using the identified parameters and the target message is the response to this request. The server can choose to also continue signing future responses to the same client in the same way. Responses that are the result of signature negotiation in this manner MUST either be marked as not cacheable or contain a `Vary` field, in order to prevent a proxy from returning a response with a signature intended for a different request.
```
mnot, line 1058:
```suggestion
The message to which the requested signature is applied is known as the "target message". When the `Accept-Signature` field is sent in an HTTP Request message, the field indicates that the client desires the server to sign the response using the identified parameters and the target message is the response to this request. All responses from resources that support such signature negotiation SHOULD either be uncacheable or contain a `Vary` header field that lists `Accept-Signature`, in order to prevent a cache from returning a response with a signature intended for a different request.
```

It's a SHOULD in the core HTTP specs, so making it a MUST here seems a bit too high of a bar. I deleted 'the server can choose' because it implies that HTTP is a stateful protocol.

jricher: Thanks, this is a good change and I'll add it. Should we make a similar change in the next paragraph, which currently has:

> The client can choose to also continue signing future requests to the same server in the same way.

Or is it more acceptable in this direction because a client has more of an "identity" in mind for a given server it's talking to?

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1048,6 +1051,49 @@ Signature: sig1=:P0wLUszWQjoi54udOtydf9IWTfNhy+r53jGFj9XZuP4uKwxyJo\
...
+
+While a signer is free to attach a signature to a request or response without prompting, it is often desirable for a potential verifier to signal that it expects a signature from a potential signer using the `Accept-Signature` field.
+
+The message to which the requested signature is applied is known as the "target message". When the `Accept-Signature` field is sent in an HTTP Request message, the field indicates that the client desires the server to sign the response using the identified parameters and the target message is the response to this request. All responses from resources that support such signature negotiation SHOULD either be uncacheable or contain a `Vary` header field that lists `Accept-Signature`, in order to prevent a cache from returning a response with a signature intended for a different request.
+
+When the `Accept-Signature` field is used in an HTTP Response message, the field indicates that the server desires the client to sign its next request to the server with the identified parameters, and the target message is the client's next request. The client can choose to also continue signing future requests to the same server in the same way.
```
jricher, line 1060: @mnot is it ok to include the language about a client choosing to sign future requests, or is that too stateful?

mnot: This is fine; the state is associated with the resource, which *is* stateful. It's connection state we have to watch out for.

MikeBishop: Is it the next request specifically, or future requests more broadly?  This seems to presuppose the server knows the clients order of requests.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1048,6 +1051,49 @@ Signature: sig1=:P0wLUszWQjoi54udOtydf9IWTfNhy+r53jGFj9XZuP4uKwxyJo\
...
+
+The target message of an `Accept-Signature` field MUST include all labeled signatures indicated in the `Accept-Header` signature, each covering the same identified components of the `Accept-Signature` field.
```
MikeBishop, line 1062: MUST seems a bit strong here.  Is this signal actually mandatory for interop?  I'd think SHOULD, or even stick to "the server desires" as above.

PR #1599, opened 2021-08-11
Updated Signature Language

jricher:
This PR updates the language used within the signatures draft to align better with HTTP terminology and use cases.

Specifically:

- "covered content" is now "covered components"
- components now consist of a "component identifier" and a "component value"
- the signature parameters data structure is used to leverage metadata instead of the signature-input header value
- signature-input and signature fields are now explicitly listed as either headers or trailers

No syntactic changes to the specification have been made.
﻿


draft-ietf-httpbis-message-signatures.md
```diff
@@ -778,23 +802,25 @@ In order to create a signature, a signer MUST follow the following algorithm:
 
 3. If applicable, the signer sets the signature's expiration time property to the time at which the signature is to expire.
 
-4. The signer creates an ordered list of content identifiers representing the message content and signature metadata to be covered by the signature, and assigns this list as the signature's Covered Content.
-   * Once an order of covered content is chosen, the order MUST NOT change for the life of the signature.
-   * Each covered content identifier MUST either reference an HTTP header in the request message {{http-header}} or reference a specialty content field listed in {{specialty-content}} or its associated registry.
-   * Signers of a request SHOULD include some or all of the control data in the covered content list, such as the `@method`, `@authority`, `@target-uri`, or some combination thereof.
+4. The signer creates the signature parameters, which is an ordered set of component identifiers representing the message components to be covered by the signature, and assigns signature metadata parameters to this value. The value of this is later used as the value of the `Signature-Input` field as described in {{signature-input-header}}.
```
ioggstream, line 805:
```
4. The signer creates the signature parameters, which is ...
```
is the wording correct? Is "signature parameters" a definition?

draft-ietf-httpbis-message-signatures.md
```diff
@@ -991,13 +1022,19 @@ Signature: sig1=:P0wLUszWQjoi54udOtydf9IWTfNhy+r53jGFj9XZuP4uKwxyJo\
   BNFv3r5S9IXf2fYJK+eyW4AiGVMvMcOg==:
 ~~~
 
+It is RECOMMENDED that the `Signature` HTTP field be included as a header, though it is possible to include it as a trailer to facilitate signing a message after its content has been processed by the signer.
```
ioggstream, line 1025: I'd recall - if it's not done elsewhere - that intermediaries can legitimately remove trailers

draft-ietf-httpbis-message-signatures.md
jricher: Yes, it's described here in this sentence.

jricher: Updated language here to be more clear as to what's being created by the signer -- it's a data structure consisting of an order set with a bunch of parameters.

draft-ietf-httpbis-message-signatures.md
jricher: Added language to both places talking about trailers.

Issue #1561, opened 2021-06-21
priorities: align with h2spec bis


LPardue:
There's work over in h2spec bis to tighten up RFC7540 and as a by product better align it with HTTP/3. That will help make the priorities doc self-consistent between sections that speak to HTTP/2 or HTTP/3. For example, https://github.com/httpwg/http2-spec/pull/860 would make frame definition formats consistent, which we can use to define the HTTP/2 PRIORITY_UPDATE frame. 

This issue is for tracking h2spec bis progress and deciding a good point to start some editorial tweaks.

LPardue:
see https://github.com/httpwg/http-extensions/pull/1601

PR #1601, opened 2021-08-13
Reorient the text to the HTTP/2bis spec

LPardue:
This turned out harder than I thought, because HTTP/2 bis makes a point about
deprecating some things, not talking about others and pointing people back to RFC 7540.
So if we want to target that document, without more substantial text changes, we end
up needing to tread across both lines.

This is an effort that attempts to take that approach to conclusion. It replaces
many references to "HTTP/2" with "RFC 7540". But importantly not all. The
remaining references to HTTP/2 are inteded to point to RFC-to-be HTTP/2 bis.

There's an alternative to this PR, which is to remove a lot of the text that needs to
refer to 7540. But I thought it would be interesting to see how this played out.

possibly addressed #1561

### DIFF
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -78,10 +78,10 @@ important for an HTTP server to prioritize the HTTP responses, or the chunks of
 those HTTP responses, that it sends.
 
 RFC 7540 {{?RFC7540}} stream priority allowed a client to send a series of
-priority signals that communicate to the server a "priority tree"; the structure of this tree
-represents the client's preferred relative ordering and weighted distribution of
-the bandwidth among HTTP responses. Servers could use these priority signals as
-input into prioritization decision making.
+priority signals that communicate to the server a "priority tree"; the structure
+of this tree represents the client's preferred relative ordering and weighted
+distribution of the bandwidth among HTTP responses. Servers could use these
+priority signals as input into prioritization decision making.
 
 The design and implementation of RFC 7540 stream priority was observed to have
 shortcomings, explained in {{motivation}}.
@@ -89,11 +89,13 @@ shortcomings, explained in {{motivation}}.
 these stream priority signals.
 
 This document describes an extensible scheme for prioritizing HTTP responses
-that uses absolute values. {#header-field} defines the Priority HTTP header
-field that can be used by both client and server to specify the precedence of
-HTTP responses in a standardized, extensible, protocol-version-independent,
-end-to-end format. {#h2-update-frame} and {#h3-update-frame} define
-version-specific frames for reprioritization. This prioritization scheme and its
+that uses absolute values. {{parameters}} defines priority parameters, which are
+a standardized and extensible format of priority information. {{header-field}}
+defines the Priority HTTP header field that can be used by both client and
+server to exchange parameters in order to specify the precedence of HTTP
+responses in a protocol-version-independent and end-to-end manner.
+{{h2-update-frame}} and {{h3-update-frame}} define version-specific frames that
+carry parameters for reprioritization. This prioritization scheme and its
 signals can act as a substitute for RFC 7540 stream priority.
 
 ## Notational Conventions
@@ -113,6 +115,9 @@ This document uses the variable-length integer encoding from
 
 The term control stream is used to describe the HTTP/2 stream with identifier
 0x0, and HTTP/3 control stream; see {{Section 6.2.1 of !HTTP3=I-D.ietf-quic-http}}.
+
+The term HTTP/2 priority signal is used to describe the priority information
+sent from clients to servers in HTTP/2 frames; see {{Section 5.3.2 of HTTP2}}.
 
 
 # Motivation for Replacing RFC 7540 Priorities {#motivation}
@@ -162,12 +167,13 @@ to manipulate the server's stored prioritization state.
 
 HTTP/2 priority signals are required to be delivered and processed in the order
 they are sent so that the receiver handling is deterministic. Porting HTTP/2
-priority signals to protocols that do not provide this guarantee presents
-challenges. For example, HTTP/3 {{HTTP3}} lacks global ordering, which meant
-that early attempts to port HTTP/2 priority signals required adding more
-information to the signals and more complicated processing. The problems with
-the porting design could not be easily resolved across multiple draft iterations
-and ultimately the feature was entirely dropped before publication.
+priority signals to protocols that do not provide ordering guarantees presents
+challenges. For example, HTTP/3 {{HTTP3}} lacks global ordering across streams
+that would carry priority signals. Early attempts to port HTTP/2 priority
+signals to HTTP/3 required adding additional information to the signals, leading
+to more complicated processing. Problems found with this approach could not be
+resolved and definition of a HTTP/3 priority signalling feature was removed
+before publication.
 
 Considering the problems with the deployment of RFC 7540 stream priority, and
 the difficulties in adapting it to HTTP/3, continuing to base prioritization on
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -78,14 +78,14 @@ important for an HTTP server to prioritize the HTTP responses, or the chunks of
 those HTTP responses, that it sends.
 
 RFC 7540 {{?RFC7540}} stream priority allowed a client to send a series of
-priority signals that communicate to the server a "priority tree"; this
+priority signals that communicate to the server a "priority tree"; the structure of this tree
 represents the client's preferred relative ordering and weighted distribution of
 the bandwidth among HTTP responses. Servers could use these priority signals as
 input into prioritization decision making.
 
 The design and implementation of RFC 7540 stream priority was observed to have
 shortcomings, explained in {{motivation}}.
-({{!HTTP2=I-D.ietf-httpbis-http2bis}}) has deprecated the use of these stream
+({{!HTTP2=I-D.ietf-httpbis-http2bis}}) has consequently deprecated the use of these stream
 priority signals.
 
 This document describes an extensible scheme for prioritizing HTTP responses
@@ -148,7 +148,7 @@ determine how such images should be prioritized against other responses.
 
 RFC 7540 allows intermediaries to coalesce multiple client trees into a
 single tree that is used for a single upstream HTTP/2 connection. However, most
-intermediaries do not support this. RFC 7540 does not define a method that can
+intermediaries do not support this. Additionally, RFC 7540 does not define a method that can
 be used by a server to express the priority of a response. Without such a
 method, intermediaries cannot coordinate client-driven and server-driven
 priorities.
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -77,19 +77,23 @@ To provide meaningful presentation of a document at the earliest moment, it is
 important for an HTTP server to prioritize the HTTP responses, or the chunks of
 those HTTP responses, that it sends.
 
-RFC 7540 {{?RFC7540}} defined such a prioritization scheme. A client sends a
-series of PRIORITY frames to communicate to the server a "priority tree"; this
-represents the client's preferred ordering and weighted distribution of the
-bandwidth among the HTTP responses. However, the design and implementation of
-this scheme has been observed to have shortcomings, explained in {{motivation}}.
-({{!HTTP2=I-D.ietf-httpbis-http2bis}}) deprecated the scheme defined in
-RFC 7540.
+RFC 7540 {{?RFC7540}} stream priority allowed a client to send a series of
+priority signals that communicate to the server a "priority tree"; this
+represents the client's preferred relative ordering and weighted distribution of
+the bandwidth among HTTP responses. Servers could use these priority signals as
+input into prioritization decision making.
 
-This document defines the Priority HTTP header field that can be used by both
-client and server to specify the precedence of HTTP responses in a standardized,
-extensible, protocol-version-independent, end-to-end format. Along with the
-protocol-version-specific frame for reprioritization, this prioritization scheme
-acts as a substitute for the original prioritization scheme of RFC 7540.
+The design and implementation of RFC 7540 stream priority was observed to have
+shortcomings, explained in {{motivation}}.
+({{!HTTP2=I-D.ietf-httpbis-http2bis}}) has deprecated the use of these stream
+priority signals.
+
+This document describes an extensible scheme for prioritizing HTTP responses
+that uses absolute values. It defines the Priority HTTP header field that can be
+used by both client and server to specify the precedence of HTTP responses in a
+standardized, extensible, protocol-version-independent, end-to-end format. It
+also defines version-specific frame for reprioritization. This prioritization
+scheme and its signals can act as a substitute for RFC 7540 stream priority.
 
 ## Notational Conventions
 
@@ -113,42 +117,38 @@ The term control stream is used to describe the HTTP/2 stream with identifier
 # Motivation for Replacing RFC 7540 Priorities {#motivation}
 
 An important feature of any implementation of a protocol that provides
-multiplexing is the ability to prioritize the sending of information. This was
-an important realization in the design of HTTP/2 leading to RFC 7540.
+multiplexing is the ability to prioritize the sending of information.
 Prioritization is a difficult problem, so it will always be suboptimal,
 particularly if one endpoint operates in ignorance of the needs of its peer.
+Priority signalling allows endpoints to communicate their own view of priority
+needs, which can be combined with other factors that are considered during the
+peer's prioritization decision making.
 
-RFC 7540 {{?RFC7540}} introduced a complex prioritization scheme that uses a
-combination of stream dependencies and weights to describe an unbalanced tree.
-This scheme suffered from poor deployment and interoperability, it was
-deprecated in HTTP/2 ({{HTTP2}}). However, in order to maintain wire
-compatibility, HTTP/2 priority signals are still mandatory to handle. These come
-in three forms. First, a HEADERS frame {{Section 6.2 of HTTP2}} with the
-PRIORITY flag set is an explicit signal that includes an Exclusive flag, Stream
-Dependency field, and Weight field. Second, a HEADERS frame with no PRIORITY
-flag is an implicit signal to use the default priority. Third, the PRIORITY
-frame {{Section 6.3 of HTTP2}}, which is always explicit since it always
-contains an Exclusive flag, Stream Dependency field, and Weight field.
+RFC 7540 stream priority ({{Section 5.3 of ?RFC7540}}) is a complex system where
+clients signal stream dependencies and weights to describe an unbalanced tree.
+It suffered from poor deployment and interoperability and was deprecated in a
+revision of HTTP/2 ({{HTTP2}}). However, in order to maintain wire
+compatibility, HTTP/2 priority signals are still mandatory to handle (see
+{{Section 5.3.2 of HTTP2}}).
 
-Client build an RFC 7540 tree using signals but experience has shown the rich
-flexibility is rarely exercised. Instead they tend to choose a single model
-optimized for a web use case and experiment within the model constraints, or do
-nothing at all. Furthermore, many clients build their prioritization tree in a
-unique way, which makes it difficult for servers to understand their intent and
-act or intervene accordingly.
+Clients can build RFC 7540 trees with rich flexibility but experience has shown
+this is rarely exercised. Instead they tend to choose a single model optimized
+for a single use case and experiment within the model constraints, or do nothing
+at all. Furthermore, many clients build their prioritization tree in a unique
+way, which makes it difficult for servers to understand their intent and act or
+intervene accordingly.
 
-Many RFC 7540 server implementations do not include support for the priority
-scheme. Some instead favor custom server-driven schemes based on heuristics or
-other hints, such as resource content type or request generation
-order. For example, a server, with knowledge of the document structure, might
-want to prioritize the delivery of images that are critical to user experience
-above other images, but below the CSS files. Since client trees vary, it is
-impossible for the server to determine how such images should be prioritized
-against other responses.
+Many RFC 7540 server implementations do not act on HTTP/2 priority signals. Some
+instead favor custom server-driven schemes based on heuristics or other hints,
+such as resource content type or request generation order. For example, a
+server, with knowledge of the document structure, might want to prioritize the
+delivery of images that are critical to user experience above other images, but
+below the CSS files. Since client trees vary, it is impossible for the server to
+determine how such images should be prioritized against other responses.
 
-The RFC 7540 scheme allows intermediaries to coalesce multiple client trees into a
+RFC 7540 allows intermediaries to coalesce multiple client trees into a
 single tree that is used for a single upstream HTTP/2 connection. However, most
-intermediaries do not support this. The scheme does not define a method that can
+intermediaries do not support this. RFC 7540 does not define a method that can
 be used by a server to express the priority of a response. Without such a
 method, intermediaries cannot coordinate client-driven and server-driven
 priorities.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-priority.md
```diff
@@ -108,35 +110,34 @@ The term control stream is used to describe the HTTP/2 stream with identifier
...
-clients build their prioritization tree in a unique way, which makes it
-difficult for servers to understand their intent and act or intervene
-accordingly.
-
-Many HTTP/2 server implementations do not include support for the priority
+an important realization in the design of HTTP/2 leading to RFC 7540.
+Prioritization is a difficult problem, so it will always be suboptimal,
+particularly if one endpoint operates in ignorance of the needs of its peer.
+
+RFC 7540 {{?RFC7540}} introduced a complex prioritization scheme that uses a
```
martinthomson, line 121:
```suggestion
RFC 7540 {{?RFC7540}} introduced a complex prioritization signaling scheme that uses a
```

draft-ietf-httpbis-priority.md
```diff
@@ -108,35 +110,34 @@ The term control stream is used to describe the HTTP/2 stream with identifier
...
+combination of stream dependencies and weights to describe an unbalanced tree.
+This scheme suffered from poor deployment and interoperability, it was
+deprecated in HTTP/2 ({{HTTP2}}). However, in order to maintain wire
```
martinthomson, line 124:
```suggestion
deprecated in a revision of HTTP/2 ({{HTTP2}}). However, in order to maintain wire
```

draft-ietf-httpbis-priority.md
```diff
@@ -108,35 +110,34 @@ The term control stream is used to describe the HTTP/2 stream with identifier
...
+compatibility, HTTP/2 priority signals are still mandatory to handle. These come
+in three forms. First, a HEADERS frame {{Section 6.2 of HTTP2}} with the
+PRIORITY flag set is an explicit signal that includes an Exclusive flag, Stream
+Dependency field, and Weight field. Second, a HEADERS frame with no PRIORITY
```
martinthomson, line 128: You can probably skip the detailed description of fields and frames.

draft-ietf-httpbis-priority.md
```diff
@@ -108,35 +110,34 @@ The term control stream is used to describe the HTTP/2 stream with identifier
...
+flag is an implicit signal to use the default priority. Third, the PRIORITY
+frame {{Section 6.3 of HTTP2}}, which is always explicit since it always
+contains an Exclusive flag, Stream Dependency field, and Weight field.
+
+Client build an RFC 7540 tree using signals but experience has shown the rich
```
martinthomson, line 133:
```suggestion
Clients build an RFC 7540 tree using signals but experience has shown the rich
```

draft-ietf-httpbis-priority.md
```diff
@@ -108,35 +110,34 @@ The term control stream is used to describe the HTTP/2 stream with identifier
...
+flexibility is rarely exercised. Instead they tend to choose a single model
+optimized for a web use case and experiment within the model constraints, or do
```
martinthomson, line 135:
```suggestion
optimized for a single use case and experiment within the model constraints, or do
```

draft-ietf-httpbis-priority.md
```diff
@@ -108,35 +110,34 @@ The term control stream is used to describe the HTTP/2 stream with identifier
...
-that clients tend to choose a single model optimized for a web use case and
-experiment within the model constraints, or do nothing at all. Furthermore, many
-clients build their prioritization tree in a unique way, which makes it
-difficult for servers to understand their intent and act or intervene
-accordingly.
-
-Many HTTP/2 server implementations do not include support for the priority
+an important realization in the design of HTTP/2 leading to RFC 7540.
+Prioritization is a difficult problem, so it will always be suboptimal,
+particularly if one endpoint operates in ignorance of the needs of its peer.
```
martinthomson, line 119: You sail very close to conflating prioritization with priority signals here.  This paragraph could be better if it was crisper on this distinction.  That prioritization is important in a multiplexed protocol is key.  That can be done based on contextual information, but signaling means that an endpoint, particularly an HTTP server, does not operate in ignorance of circumstances.

draft-ietf-httpbis-priority.md
LPardue: Glad you picked up on this. There's a stylistic difference between the two docs and I purposefully tried to stick to this docs style. That's to say that here we've tried to refer to the abstract notion of weights and dependencies as a "scheme", and the signals as just bits on the wire. Signals are input to a concrete instance of the scheme, which itself is context or input for making resuours usage decisions (  along with other stuff outside the scheme). For my money, the signalling wasn't the hard part, the other stuff was.

You'll probably have a good retort to this style :)

draft-ietf-httpbis-priority.md
LPardue: @mnot asked for an explicit definition of what we meant by HTTP/2 signals  a little while ago. I think it adds value.

draft-ietf-httpbis-priority.md
MikeBishop: I'll pile in and say that an actual implementation might have a very simple or very complex internal representation of how priority is allocated.  The weights and dependencies modify a conceptual graph (hopefully consistent between the two endpoints), and that graph is condensed into whatever the implementation actually uses internally.

I tend to think that the weights and dependencies are signaling, because they don't actually specify internals of how an implementation must represent priority; they're inputs.  However, an implementation *could* use them directly as its internal representation, in which case they would also be the priority scheme itself.

LPardue: for clarity: it sounds like you're siding with MT, is that right @MikeBishop ?

MikeBishop: Slightly.  However, I think the reality is that we're not disagreeing that there is a split, but rather which side of the split the shared state sits.

I think the shared state is part of the signal, because the server need not directly implement priorities by strictly following that system.  (As Cloudflare's own past work demonstrated, what's received from the client is only part of what the server might reasonably consider in its scheme.)  It sounds like you consider the shared state to be a scheme, and the output of that scheme is one input to an implementation.

Both are self-consistent, so long as they're used consistently in the document.

LPardue: thanks.

There's a few interesting things on reflection of this document as a whole

* It says it defines a priority scheme
* about half of the uses of "scheme" describe design elements
* about half the uses of "scheme" describe what RFC 7540 said and some foibles
* RFC 7540 doesn't use the term "priority scheme" at all
  * neither does it use "priority signal" or really explicitly use signal in the context of priorities.
  * Instead,  it just says "stream prioritization" and talks very specifically in those terms
  
I'm starting to think this is what is causing a bit of friction, especially as we try to loop in http2bis. Now is a good time to shore up things.

Since http2bis is leaving stream weights and deps behind to focus on mandatory frame formats (signals), I think a path forward could be to rephrase things as "RFC 7540 priorities" and "HTTP/2 priority signals". But continue to use "priority scheme" as a term of art for this thing we are specifying here.

draft-ietf-httpbis-priority.md
LPardue: I've actually done this, preferring to delegate to https://www.ietf.org/archive/id/draft-ietf-httpbis-http2bis-03.html#section-5.3.2

LPardue:
I've pushed an update that implements the changes I outlined in https://github.com/httpwg/http-extensions/pull/1601#discussion_r688693965, PTAL.

@kazuho any thoughts on this?

draft-ietf-httpbis-priority.md
rmarx: To me, a "scheme" has always been more specific (i.e., Round-Robin vs FIFO vs what firefox H2 does) and I've always referred to H2's setup as a "prioritization **system**" in my writings. 

Mapping that to the above, I tend to side more with Martin and Mike that "signal" is a better denominator here, but that's not a hill I'm willing to die on.

draft-ietf-httpbis-priority.md
```diff
@@ -77,17 +77,23 @@ To provide meaningful presentation of a document at the earliest moment, it is
...
+The design and implementation of RFC 7540 stream priority was observed to have
+shortcomings, explained in {{motivation}}.
+({{!HTTP2=I-D.ietf-httpbis-http2bis}}) has deprecated the use of these stream
+priority signals.
+
+This document describes an extensible scheme for prioritizing HTTP responses
+that uses absolute values. It defines the Priority HTTP header field that can be
+used by both client and server to specify the precedence of HTTP responses in a
+standardized, extensible, protocol-version-independent, end-to-end format. It
+also defines version-specific frame for reprioritization. This prioritization
```
rmarx, line 95:
```suggestion
also defines a version-specific frame for reprioritization. This prioritization
```
unless it defines multiple, then it should be "frames", of course

rmarx: Not sure this should be addressed here, but above you say this is protocol-version-independent, while the reprioritization frame is not for H1.

draft-ietf-httpbis-priority.md
rmarx: I'm not sure I see the problem here; to me, the text doesn't really fully cover what you're saying (doesn't explicitly state that you don't need signals to prioritize, but it _is_ implied imo), but it also doesn't really conflate things.

draft-ietf-httpbis-priority.md
```diff
@@ -108,101 +114,100 @@ The term control stream is used to describe the HTTP/2 stream with identifier
...
-impossible for the server to determine how such images should be prioritized
-against other responses.
-
-The HTTP/2 scheme allows intermediaries to coalesce multiple client trees into a
+multiplexing is the ability to prioritize the sending of information.
+Prioritization is a difficult problem, so it will always be suboptimal,
+particularly if one endpoint operates in ignorance of the needs of its peer.
+Priority signalling allows endpoints to communicate their own view of priority
+needs, which can be combined with other factors that are considered during the
+peer's prioritization decision making.
```
rmarx, line 125:
```suggestion
sender's prioritization decision making.
```
In general, I wonder about the usage of "client" and "server" in the introduction and "endpoint" and "peers" here. I agree the latter is more correct in multi-hop setups, but hardly makes it easier to digest. For clarity, switching to client/server here might be better.

draft-ietf-httpbis-priority.md
```diff
@@ -108,101 +114,100 @@ The term control stream is used to describe the HTTP/2 stream with identifier
...
+
+RFC 7540 stream priority ({{Section 5.3 of ?RFC7540}}) is a complex system where
+clients signal stream dependencies and weights to describe an unbalanced tree.
+It suffered from poor deployment and interoperability and was deprecated in a
```
rmarx, line 129: I would remove this and the next sentence from this part. It seems to break the flow of the text and is already stated in the introduction and repeated later when talking about deprecation and handling of both signals etc.

draft-ietf-httpbis-priority.md
```diff
@@ -108,101 +114,100 @@ The term control stream is used to describe the HTTP/2 stream with identifier
...
+at all. Furthermore, many clients build their prioritization tree in a unique
+way, which makes it difficult for servers to understand their intent and act or
+intervene accordingly.
+
+Many RFC 7540 server implementations do not act on HTTP/2 priority signals. Some
+instead favor custom server-driven schemes based on heuristics or other hints,
+such as resource content type or request generation order. For example, a
+server, with knowledge of the document structure, might want to prioritize the
+delivery of images that are critical to user experience above other images, but
+below the CSS files. Since client trees vary, it is impossible for the server to
```
rmarx, line 146: I would re-structure this to be an example of the last sentence of the previous paragraph:

- Client trees are heterogeneous
- Makes it difficult to insert/reprioritize
- Consequently, some servers ignore client signals in favor of other heuristics to make these use cases easier

Now, these two somewhat come across as separate concepts, while to me, they are inherently linked, one the consequence of the other. 

(on a related note: I'm not sure the newly proposed system in this document really fully solves this issue, but that's another discussion :))

draft-ietf-httpbis-priority.md
```diff
@@ -108,101 +114,100 @@ The term control stream is used to describe the HTTP/2 stream with identifier
...
+determine how such images should be prioritized against other responses.
+
+RFC 7540 allows intermediaries to coalesce multiple client trees into a
 single tree that is used for a single upstream HTTP/2 connection. However, most
-intermediaries do not support this. The scheme does not define a method that can
+intermediaries do not support this. RFC 7540 does not define a method that can
```
rmarx, line 151:
```suggestion
intermediaries do not support this. Additionally, RFC 7540 does not define a method that can
```

draft-ietf-httpbis-priority.md
```diff
@@ -108,101 +114,100 @@ The term control stream is used to describe the HTTP/2 stream with identifier
...
-such as the scheme defined in this specification. The
-SETTINGS_DEPRECATE_HTTP2_PRIORITIES setting described below enables endpoints to
-understand their peer's intention. The value of the parameter MUST
-be 0 or 1. Any value other than 0 or 1 MUST be treated as a connection error
-(see {{Section 5.4.1 of HTTP2}}) of type PROTOCOL_ERROR.
+resource exhaustion vectors affecting multiple RFC 7540 implementations. One
+attack, [CVE-2019-9513] aka "Resource Loop", is based on using priority signals
+to manipulate the server's stored prioritization state.
+
+HTTP/2 priority signals depend on in-order delivery, leading to challenges in
```
rmarx, line 162: imo, this is vague... in-order delivery of what? the signals? requests? responses? 

```suggestion
HTTP/2 priority signals depend on their in-order delivery to deterministically construct the intended tree at the receiver, leading to challenges in
```

draft-ietf-httpbis-priority.md
```diff
@@ -108,101 +114,100 @@ The term control stream is used to describe the HTTP/2 stream with identifier
...
+setups seen in practice, at least for the web use case.
+
+## Disabling RFC 7540 Priorities {#disabling}
+
+The problems and insights set out above provided the motivation for deprecating
+RFC 7540 stream priority (see {{Section 5.3 of RFC7540}}).
+
+The SETTINGS_DEPRECATE_RFC7540_PRIORITIES setting is defined by this document in
+order to allow endpoints to explicitly opt out of using HTTP/2 priority signals
+(see {{Section 5.3.2 of HTTP2}}). Endpoints are expected to use an alternative,
```
rmarx, line 181: Here, it's somewhat unclear who sends this setting and who acts upon it. You use "endpoint" for both client and server interchangeably (which is correct, but somewhat confusing). 

As in: does the client use the setting to indicate it won't use the signals? or just the server to indicate it won't? Later text first seems to indicate it's only the server sending it, but at the end, it's mentioned the client can also send... imo, this should be clarified up-front: who can send it and what are the semantics from both sides (e.g., if client send it, does it promise not to send RFC 7540 signals?)

draft-ietf-httpbis-priority.md
```diff
@@ -108,101 +114,100 @@ The term control stream is used to describe the HTTP/2 stream with identifier
...
-behind the server that the client is directly connected to.
-
-The SETTINGS frame precedes any priority signal sent from a client in HTTP/2, so
-a server can determine if it should respect the HTTP/2 scheme before building
-state. A server that receives SETTINGS_DEPRECATE_HTTP2_PRIORITIES with value of
-1 MUST ignore HTTP/2 priority signals.
-
-Where both endpoints disable HTTP/2 priorities, the client is expected to send
-this scheme's priority signal. Handling of omitted signals is described in
+send both the HTTP/2 priority signals and the signals of this prioritization
```
rmarx, line 194: Should this be HTTP/2-bis or RFC 7540? IIUC, any specific details of these signals is removed from -bis, so it should be RFC 7540 to be able to use them in this way?

rmarx:
```suggestion
send both the HTTP/2 priority signals and the signals of the employed alternative scheme
```

Above, you indicate any scheme can be used, with this being just one option. 

This does make guidance about PRIO_UPDATE frames/header field more difficult later, so maybe some other textual adjustment is needed earlier:

e.g., for the remainder of this text ,we assume the use of the scheme defined in this document as the alternate

Or, move the following text from below up and adjust a bit 
> Where both endpoints disable RFC 7540 stream priority, the client is expected to
send this scheme's priority signal

draft-ietf-httpbis-priority.md
```diff
@@ -108,101 +114,100 @@ The term control stream is used to describe the HTTP/2 stream with identifier
...
+scheme (see {{header-field}} and {{h2-update-frame}}). When the client receives
+the first SETTINGS frame that contains the SETTINGS_DEPRECATE_RFC7540_PRIORITIES
+parameter with value of 1, it SHOULD stop sending the HTTP/2 priority signals.
+If the value was 0 or if the settings parameter was absent, it SHOULD stop
+sending PRIORITY_UPDATE frames ({{h2-update-frame}}), but MAY continue sending
+the Priority header field ({{header-field}}), as it is an end-to-end signal that
+might be useful to nodes behind the server that the client is directly connected
+to.
+
+The SETTINGS frame precedes any HTTP/2 priority signal sent from a client, so a
```
rmarx, line 204:
```suggestion
The SETTINGS frame takes precedence over any HTTP/2 priority signal sent from a client, so a
```
It doesn't necessarily precede it in the pure sense of the word if we're talking about server-sent SETTINGS frame, right?

draft-ietf-httpbis-priority.md
```diff
@@ -108,101 +114,100 @@ The term control stream is used to describe the HTTP/2 stream with identifier
...
+server can determine if it needs to allocate any resource to signal handling
```
rmarx, line 205:
```suggestion
server can determine if it needs to allocate any resource to handle signals
```

draft-ietf-httpbis-priority.md
```diff
@@ -108,101 +114,100 @@ The term control stream is used to describe the HTTP/2 stream with identifier
...
+before they arrive. A server that receives SETTINGS_DEPRECATE_RFC7540_PRIORITIES
+with value of 1 MUST ignore HTTP/2 priority signals.
+
+Where both endpoints disable RFC 7540 stream priority, the client is expected to
+send this scheme's priority signal. Handling of omitted signals is described in
```
rmarx, line 210: "this" is somewhat ambiguous here.

```suggestion
send the priority signals defined in this document. Handling of omitted signals is described in
```

draft-ietf-httpbis-priority.md
```diff
@@ -108,101 +114,100 @@ The term control stream is used to describe the HTTP/2 stream with identifier
...
+particularly if one endpoint operates in ignorance of the needs of its peer.
+Priority signalling allows endpoints to communicate their own view of priority
+needs, which can be combined with other factors that are considered during the
+peer's prioritization decision making.
+
+RFC 7540 stream priority ({{Section 5.3 of ?RFC7540}}) is a complex system where
+clients signal stream dependencies and weights to describe an unbalanced tree.
+It suffered from poor deployment and interoperability and was deprecated in a
+revision of HTTP/2 ({{HTTP2}}). However, in order to maintain wire
+compatibility, HTTP/2 priority signals are still mandatory to handle (see
```
rmarx, line 131: should this ref be RFC 7540 priority signals, instead of H2-bis prio signals? IIUC the signals are no longer defined in H2-bis, which is taken as the "canonical H2" for this doc, so this should be RFC 7540 for consistency IIUC?

draft-ietf-httpbis-priority.md
```diff
@@ -424,24 +429,32 @@ initial priority of a response, or to reprioritize a response or push stream. It
 carries the stream ID of the response and the priority in ASCII text, using the
```
rmarx, line 429: IIUC, this does not use HPACK then? Maybe worth it to point that out explicitly.

rmarx (commented):
Most of these are simple language fixes, but some ask for clarification/disambiguation that I feel should probably be fixed in some way.

I also didn't just limit to purely the H2-bis vs RFC 7540 changes, as I wasn't sure if I should open separate issues for some other points (e.g., the structure/buildup of the motivation section). If needed, I can of course still do that.

draft-ietf-httpbis-priority.md
MikeBishop: This second comment I don't believe is an issue.  "[This document] defines the Priority HTTP header field ... in a ... protocol-version-independent, end-to-end format. [This document] **also** defines version-specific frames for reprioritization."  So, a version-independent header field as well as version-specific frames.

In fact, it sounds like what we should clarify is the antecedent of "it" in the second sentence.

rmarx: Seems like I stupidly missed the "version-specific" part... I agree that resolves my comment.

LPardue: > unless it defines multiple, then it should be "frames", of course

yes I think its more accurate to use the plural form here, since we define H2 and H3 PRIORITY_UPDATE frames

> In fact, it sounds like what we should clarify is the antecedent of "it" in the second sentence.

I'll try and accomodate that too in a commit

draft-ietf-httpbis-priority.md
LPardue: You're right, we can substitute in client and server here for simplicity

draft-ietf-httpbis-priority.md
LPardue: This question is tangential to the editorial change at hand. Personally I don't see a problem with the text as it is. But if you feel strongly a new issue can be opened.

draft-ietf-httpbis-priority.md
LPardue: In HTTP/2, SETTINGS has to come before other things because of the preface, so the current text is accurate IMO.

draft-ietf-httpbis-priority.md
LPardue: +1 to Mike's thumbs down. There's resources that might get allocated just to support server-side dependency trees even if a client never sends explicit priority signals.

draft-ietf-httpbis-priority.md
LPardue: I made https://github.com/httpwg/http-extensions/issues/1679 because the question is tangential to this PR.

draft-ietf-httpbis-priority.md
LPardue: A resolution to https://github.com/httpwg/http-extensions/issues/1679 would also touch this

draft-ietf-httpbis-priority.md
LPardue: Only certain frames can drive the HPACK state by HTTP/2 definition. I think mentioning HPACK here would just be confusing.

draft-ietf-httpbis-priority.md
LPardue: Actually, I changed my mind. This paragraph follows the first in speaking to the abstract notion of prioritization and priority signalling. The only to replace that is to change "endpoint" and "peer" to "client or server", which just gets clunky.

draft-ietf-httpbis-priority.md
LPardue: Can you explain more how it breaks the flow? The first two paragraphs in the section introduce the abstract notion of prioritization and priority signals, this paragraph then speaks specifically to RFC 7540 and it's soft deprecation in RFC7540bis. This document needs to provide sufficient context to explain the later hard deprecation, which is different.

draft-ietf-httpbis-priority.md
LPardue: I disagree, they are still defined via the section that is cited. RFC 7540bis can't remove HEADERS priority bits or the PRIORITY frame. *How* an RFC 7540bis implementation chooses to treat them is going to be up to it.

draft-ietf-httpbis-priority.md
LPardue: The substantive comment here relates to old text that has been there for a long time, it's not suitable for fixing here. Please submit a new issue and/or PR if you feel it important to address.

draft-ietf-httpbis-priority.md
LPardue: fixed in ca3c21e59a6d80ca66e1afcb0ccb10c650b53d04

LPardue:
In the interest of making progress I am merging this editorial change.

draft-ietf-httpbis-priority.md
ianswett: I would have gone for MAY here.

LPardue: Can we please continue the discussion on the issue?

PS: the requirement didn't change in this PR, so the discussion of this point is revisiting a design decision, so let's make sure its surfaced to folks.

PR #1602, opened 2021-08-13
Release Signatures 06

jricher:
Minor editorial cleanups, including:

- added client cert example
- updated multiple signature example
- prepare for next revision


swansoffie:
> Minor editorial cleanups, including:
> 
> * added client cert example
> * updated multiple signature example
> * prepare for next revision

PR #1616, opened 2021-08-25
Editorial suggestions from Benjamin Kaduk's IESG review (bcpb6bis)

kaduk:


draft-ietf-httpbis-bcp56bis.md
```diff
@@ -363,9 +363,9 @@ Queries can be performed with GET, often using the query component of the URL; t
...
 
-Applications should not change their state or have other side effects that might be significant to the client, since implementations can and do retry HTTP GET requests that fail, and some GET requests protected by TLS Early Data might be vulnerable to replay attacks (see {{?RFC8470}}). Note that this does not include logging and similar functions; see {{HTTP, Section 9.2.1}}.
+Application processing of GET requests should not change application state or have other side effects that might be significant to the client, since implementations can and do retry HTTP GET requests that fail, and some GET requests protected by TLS Early Data might be vulnerable to replay attacks (see {{?RFC8470}}). Note that this does not include logging and similar functions; see {{HTTP, Section 9.2.1}}.
```
mnot, line 368:
```suggestion
Processing of GET requests should not change application state or have other side effects that might be significant to the client, since implementations can and do retry HTTP GET requests that fail, and some GET requests protected by TLS Early Data might be vulnerable to replay attacks (see {{?RFC8470}}). Note that this does not include logging and similar functions; see {{HTTP, Section 9.2.1}}.
```

draft-ietf-httpbis-bcp56bis.md
```diff
@@ -541,7 +541,7 @@ When used, it is important to carefully specify the scoping and use of cookies;
 
 Clients often need to send multiple requests to perform a task.
 
-In HTTP/1 {{HTTP11}}, parallel requests are most often supported by opening multiple connections. Application performance can be impacted when too many simultaneous connections are used, because connections' congestion control will not be coordinated. Furthermore, it can be difficult for applications to decide when and to select which connection to use to issue a given request, further impacting performance.
+In HTTP/1 {{HTTP11}}, parallel requests are most often supported by opening multiple connections. Application performance can be impacted when too many simultaneous connections are used, because connections' congestion control will not be coordinated. Furthermore, it can be difficult for applications to decide when to issue and to select which connection to use to issue a given request, further impacting performance.
```
mnot, line 544:
```suggestion
In HTTP/1 {{HTTP11}}, parallel requests are most often supported by opening multiple connections. Application performance can be impacted when too many simultaneous connections are used, because connections' congestion control will not be coordinated. Furthermore, it can be difficult for applications to decide when to issue and which connection to use for a given request, further impacting performance.
```

mnot (changes requested):

draft-ietf-httpbis-bcp56bis.md
kaduk: A more elegant solution, thanks.

PR #1618, opened 2021-08-25
Sprinkle some "field" (proxy-status)

kaduk:


draft-ietf-httpbis-proxy-status.md
```diff
@@ -405,13 +405,13 @@ Note that additional information about the error can be recorded in the details
 * Only generated by intermediaries: false
 * Reference: \[this document\]
 
-### HTTP Response Header Too Large
+### HTTP Response Header Field Too Large
 
 * Name: http_response_header_size
 * Description: The intermediary received a response to the request containing an individual header line that was considered too large.
```
kaduk, line 411: (I left this "header line" alone.)

mnot: Technically, it should be 'HTTP response header field line too large' -- it's the field line, not the field as an entirety.

PR #1623, opened 2021-08-26
AltSvc-bis:  Largely unchanged -00

MikeBishop:
@martinthomson, since you're the other editor here, can you take a look?

Resulting file is at https://mikebishop.github.io/http-extensions/altsvc_adopt/draft-ietf-httpbis-rfc7838bis.txt, but RFCdiff doesn't seem to like the URL.


MikeBishop:
Doesn't have to go in the mega-repo; I've worked on it in my own repo and just copied it across.  Just as easy to drop it in a new one, particularly since this is a PR adding a separate file.

mnot:
@martinthomson your reaction gifs only encourage continuing use of the megarepo.

Issue #1620, opened 2021-08-26
Ben Kaduk's COMMENTs on proxy-status


mnot:
- [x] Section 2.1.3

>  The "next-protocol" parameter's value indicates the ALPN protocol
  identifier [RFC7301] used by the intermediary to connect to the next
  hop.  This is only applicable when that connection was actually
  established.
>
>  The value MUST be either an sf-token or sf-binary, representing a TLS
  Application-Layer Protocol Negotiation (ALPN) Protocol ID (see

I do know about "h2c" and Upgrade, but I think there are cases where
neither TLS nor Upgrade is used.  Is there some way to indicate, e.g.,
that HTTP/1.1 without TLS was used?  I do not believe that setting
"http/1.1" here has those semantics in the way that "h2c" does, and it
seems that there may some value in indicating whether a secure transport
was used for the next hop in a similar way to how setting "h2" or "h3"
here would do so.

**Mark**: The simplest thing to do here would be to register `h1c`. Thoughts?

---

- [x] Section 2.1.3

>  [...] If the protocol
  identifier is able to be expressed as an sf-token using ASCII
  encoding, that form MUST be used.

There seems to be some ossification risk of the sf-binary form is never
used.  I would hope that the use of structured fields mitigates that
risk, since sf-binary support is needed in the generic field decoder,
but I don't know that I would specifically rely on it.  The MUST here
prohibits greasing to be sure; maybe the benefit of consistency
outweighs the risk, though.

**Mark**: I think the analysis is about right, and no change is necessary.

---

- [x] Section 2.3.13

TLS protocol errors are typically accompanied by a TLS alert message
being generated; would the 'alert-id' and 'alert-message' extra
parameters be applicable here (in addition to for "tls_alert_received"
down in §2.3.15)?

**Mark**: I think the intent here is that when an alert is received, `tls_alert_received` is more appropriate; this is a more general fallback, and we should clarify that. @PiotrSikora?

---

- [x] Section 2.3.15

>     -  alert-message: an sf-token containing the applicable
>        description string from the TLS Alerts registry.  See
>        [RFC8446].

I don't think there's a formal restriction on what characters are
allowed in description strings in the TLS Alerts registry, so that,
e.g., one could in theory have an alert description containing
characters like ";" that are not allowed in an sf-token.  It's so
unlikely to happen in practice that I'm not actually concerned, but I
will ask if we want to formally provide some behavior to allow for such
cases.

[...]

PR #1628, opened 2021-09-01
Add next-tls

mnot:
For #1620.


draft-ietf-httpbis-proxy-status.md
```diff
@@ -173,6 +174,19 @@ For example:
 Proxy-Status: "proxy.example.org"; next-protocol=h2
 ~~~
 
+## next-tls
+
+The `next-encryption` parameter's value indicates the version (as a float, not as encoded in TLS records) of TLS {{TLS}} used by the intermediary to connect to the next hop.
```
MikeBishop, line 179:
```suggestion
The `next-tls` parameter's value indicates the version (as a float, not as encoded in TLS records) of TLS {{TLS}} used by the intermediary to connect to the next hop.
```
I presume this is a remnant of name ideation.

MikeBishop (commented):
This feels very much underspecified.  What does the absence of this field mean -- TLS wasn't used on the next hop, or the proxy opted not to provide that information?  Assuming the latter, how does the proxy indicate that the next hop didn't use encryption, or used encryption other than TLS?

PR #1633, opened 2021-09-01
If you don't use abbrev, don't set it to the empty string

martinthomson:


LPardue (approved):

mnot:
See https://github.com/martinthomson/i-d-template/pull/308

martinthomson:
Probably still worth taking this.  Setting abbrev to the empty string might cause some yaml parsers to produce values of None or null, but it might result in documents containing no short name (not that that is as important now that documents don't get paginated, but still).

reschke:
The abbreviated title should be used in HTML too (for CSS paged media footer info).

PR #1638, opened 2021-09-02
Add next-secure

mnot:
For #1620.


mnot:
ping @kaduk

draft-ietf-httpbis-proxy-status.md
```diff
@@ -173,6 +174,23 @@ For example:
...
+
+The `next-secure` parameter's value indicates whether the connection used by the intermediary to
+obtain this response was secured; typically, using TLS {{TLS}}.
+
+Its value MUST be a sf-boolean.
+
+For example:
+
+~~~ http-message
+Proxy-Status: "proxy.example.org"; next-protocol=http11;
```
MikeBishop, line 187:
```suggestion
Proxy-Status: "proxy.example.org"; next-protocol=http/1.1;
```
`http11` isn't a registered ALPN token.

MikeBishop (commented):
This probably also warrants some updated language in `next-protocol` to indicate that if `next-secure` is false, the protocol stack does not include TLS even if the ALPN token refers to a TLS-based protocol.

Also, it's probably worth stating that (I presume) the recipient cannot infer anything about the security of the next hop if this parameter is not present.  How that interacts with an ALPN token that suggests the presence of TLS gets... interesting.

MikeBishop:
Passing on another corner-case from someone else's comment:  What if the proxy is configured not to validate the upstream certificate?  Is that secure or not secure?

mnot-fastly:
See most recent commits; they should cover that.

MikeBishop:
There's only one commit on this (admittedly, closed) PR; do you need to push, perhaps?

PR #1640, opened 2021-09-02
Tweak param language

### LINKED ISSUE #1639: parameter consistency
I noticed that parameters are generally defined in terms of what happened to generate this response, whether or not the next hop was actually contacted.

The exception is `next-protocol`, which says:

> This is only applicable when that connection was actually established.

I think it's potentially valuable for a response to indicate what protocol it was fetched over even if it comes from cache.

The remedy would be to remove that statement and align this with the other parameters by talking about 'this response'. Probably also making that more prominent in `received-status`.

Thoughts, @PiotrSikora?

mnot:
Fixes #1639


PiotrSikora (approved):
Thanks!

PR #1641, opened 2021-09-02
Update references (fixes #1631)

### LINKED ISSUE #1631: expect-ct: update references
This spec is finally getting close to RFC publication, but since it was finished several referenced documents have been updated. We should review the updates here (leaving this to AUTH48 might lead to breakage in "deep" links).

reschke:
- updates refs into HTTP specs
- updates refs into 6962bis (as per https://www.ietf.org/rfcdiff?url1=draft-ietf-trans-rfc6962-bis-42.txt&url2=draft-ietf-trans-rfc6962-bis-30.txt) - this needs review


reschke:
That said, updating 7230/4 will cause the expect-ct to depend on the new core specs, which will cause another delay.

PR #1643, opened 2021-09-03
Expect-CT editorial

mnot:
Just a few small things...


reschke (changes requested):
The change to include HTTP is not sufficient, we need to update the refs to 7230 and 7234 as well (and more). See separate PR.

PR #1650, opened 2021-09-08
tweaking initial header fiel description

dret:
it seems not quite necessary to say that it is a new field...


mnot:
Expect-CT is published; closing.

Issue #1533, opened 2021-06-08
Registration policy for digest algorithm values


mnot:
This specification obsoletes 3230 and updates the IANA registry to point to it, but doesn't specify what the registration policy actually is.

ioggstream:
Related to #1394

ioggstream:
See #1567

ioggstream:
@LPardue do we need `Values to be added to this registry require IETF Review (see Section 4.8 of {{!RFC8126}}).` ?

LPardue:
I'm sure I made this comment somewhere but I've lost it. I don't think we should mandate RFC required, that a very high bar for algorithms (and their users) that most likely developed the algoithm elsewhere already. Instead [Specification Required](https://datatracker.ietf.org/doc/html/rfc8126#section-4.6) seem like a better option here

ioggstream:
@LPardue imho

* registering new algos should be done via an RFC,
* this doesn't imply that those algorithms should be defined in an RFC.

While it is fair to delegate the registration to an external specification,
I think the approach of an RFC (eg  https://datatracker.ietf.org/doc/rfc5843/ )
is a lot more clean and improves the overall interoperability.

It is important to note that  any type of RFC is sufficient (currently
   Standards Track, BCP, Informational, Experimental, or Historic).

In any case, I don't think this should be a game-stopper, so if there's
general consensus on "Specification Required", that's ok.

LPardue:
I don't think running an RFC process can provide any meaningful value to the registration requirements.

I mean, that's going to add like a year of overhead to something as simple as "use algorithm $foo with token `foo`, the output is encoded using base64".

seanturner:
The key to Specification Required is the Designated Experts. If they do they job you basically get what you need. If there's any question from the DEs they can always kick it back out for further review.

ioggstream:
@LPardue Ok then. Since it seems there's general agreement on Specification Required, it's fine to me. Move on :)

PR #1652, opened 2021-09-09
Fix: #1533. Add registration policy for digest algorithms.

ioggstream:
## This PR

- add registration policy cc: @seanturner and @mnot

## It's done

- according to https://www.rfc-editor.org/rfc/rfc8126.html#section-4.7

## Note

Using RFC policy avoids the issues we had to face to refresh the current table.

See #1533

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1267,7 +1267,7 @@ IANA is asked to update the "Reference" for this registry
 to refer this document.
 
 New registrations will be permitted
-through the Specification Required policy {{!BCP26}}.
+through the RFC Required policy {{!BCP26}}.
 
 
 ## Obsolete "contentMD5" token in Digest Algorithm {#iana-contentMD5}
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1273,8 +1273,8 @@ to refer this document
 and to inizialize the registry with the tokens 
 defined in {{algorithms}}.
 
-New registrations will be permitted
-through the RFC Required policy (see Section 4.7 of {{!RFC8126}}).
+This registry uses the Specification
+Required policy (Section 4.6 of {{!RFC8126}}).
 
 
 ## Obsolete "contentMD5" token in Digest Algorithm {#iana-contentMD5}
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1270,7 +1270,7 @@ Values](https://www.iana.org/assignments/http-dig-alg/) registry.
 IANA is asked to update the "Reference" for this registry
 to refer this document
 and to inizialize the registry with the tokens 
-defined in {{digest-algorithms}}.
+defined in {{algorithms}}.
 
 New registrations will be permitted
 through the RFC Required policy (see Section 4.7 of {{!RFC8126}}).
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -417,6 +418,9 @@ All digest-algorithm token values are case-insensitive
 but lower case is preferred;
 digest-algorithm token values MUST be compared in a case-insensitive fashion.
 
+Every digest-algorithm defines its computation procedure and,
```
ioggstream, line 421: Mark's feedback on case-sensitivity of checksum.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1263,6 +1267,15 @@ This memo sets this specification to be the establishing document for the [HTTP
 Algorithm
 Values](https://www.iana.org/assignments/http-dig-alg/) registry.
 
+IANA is asked to update the "Reference" for this registry
+to refer this document
+and to inizialize the registry with the tokens 
+defined in {{digest-algorithms}}.
```
ioggstream, line 1273:
```suggestion
defined in {{algorithms}}.
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1263,6 +1267,15 @@ This memo sets this specification to be the establishing document for the [HTTP
 Algorithm
 Values](https://www.iana.org/assignments/http-dig-alg/) registry.
 
+IANA is asked to update the "Reference" for this registry
+to refer this document
+and to inizialize the registry with the tokens 
+defined in {{algorithms}}.
+
+New registrations will be permitted
+through the RFC Required policy (see Section 4.7 of {{!RFC8126}}).
```
LPardue, line 1276: as per my comment on the issue, I think specification required is fine. So the suggestion here is

```suggestion
Registrations or this registry use the Specification
Required policy (Section 4.6 of {{!RFC8126}}).
The designated expert or experts verify that a
specification exists and is readily accessible.
```

LPardue (changes requested):

draft-ietf-httpbis-digest-headers.md
seanturner: +1 to Specification Required. You could also shorten it to "This registry use the Specification
Required policy (Section 4.6 of {{!RFC8126}}).

ioggstream: In case of "Specification Required", how to ensure interoperability with the overall HTTP work if the Specification is published outside IETF (eg. W3C, OASIS, ...) ? 

Further notes are here https://github.com/httpwg/http-extensions/issues/1533#issuecomment-916462729

martinthomson: You don't.  It might be nice in theory to insist on people only deploying things that have been carefully designed and vetted, but that doesn't happen.  If you make the requirements on registrations very difficult to achieve, what you end up with is people not registering things.  They will still use the codepoints and deploy the code, so what you end up with is no reliable record of what codepoints are safe to use.  Even "Specification Required" can be too much to avoid that happening, but we should at least encourage people to write down what they are doing.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1263,6 +1268,15 @@ This memo sets this specification to be the establishing document for the [HTTP
...
```
ioggstream, line 1277:
```suggestion
This registry uses the Specification
Required policy (Section 4.6 of {{!RFC8126}})
```

ioggstream:
@LPardue now it's Specification Required :)

LPardue (approved):

PR #1664, opened 2021-09-10
Fixes #1627 - Remove outdated comments related to Secure attribute

### LINKED ISSUE #1627: RFC6265bis: The Secure Attribute section may be incorrectly warning that a secure cookie can be overwritten by an insecure origin
[Section-4.1.2.5](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-08#section-4.1.2.5) states "An active network attacker can overwrite Secure cookies from an insecure channel, disrupting their integrity".

This warning appears to be out of date since [Leave Secure Cookies Alone](https://datatracker.ietf.org/doc/html/draft-west-leave-secure-cookies-alone-05) updated the spec to disallow insecure origins (schemes) from overwriting Secure cookies set by secure origins.

If there is in fact some other mechanism in which an insecure origin can overwrite a Secure cookie set by a secure origin then section 8.6 should be updated to list it (ignoring any origins which the UA might consider trustworthy, such as localhost).

miketaylr:
PTAL @sbingler

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -660,11 +660,6 @@ attribute, the user agent will include the cookie in an HTTP request only if
 the request is transmitted over a secure channel (typically HTTP over Transport
 Layer Security (TLS) {{RFC2818}}).
 
-Although seemingly useful for protecting cookies from active network attackers,
-the Secure attribute protects only the cookie's confidentiality. An active
-network attacker can overwrite Secure cookies from an insecure channel,
-disrupting their integrity (see {{weak-integrity}} for more details).
-
 #### The HttpOnly Attribute {#attribute-httponly}
 
 The HttpOnly attribute limits the scope of the cookie to HTTP requests. In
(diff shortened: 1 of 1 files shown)


sbingler (approved):
LGTM

miketaylr:
Mind taking a look and merging @mikewest?

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -660,11 +660,6 @@ attribute, the user agent will include the cookie in an HTTP request only if
 the request is transmitted over a secure channel (typically HTTP over Transport
 Layer Security (TLS) {{RFC2818}}).
 
-Although seemingly useful for protecting cookies from active network attackers,
-the Secure attribute protects only the cookie's confidentiality. An active
-network attacker can overwrite Secure cookies from an insecure channel,
-disrupting their integrity (see {{weak-integrity}} for more details).
```
mikewest, line 666: Perhaps we could also add a note about `__Secure-` to the "Weak Integrity" section as a partial mitigation? I'll land this and send you a PR to that effect.

mikewest (approved):

PR #1665, opened 2021-09-10
Fixes #1508 - Add additional example to third-party cookie policy.

### LINKED ISSUE #1508: RFC 6265bis should allow user agents to block a subset of cookies during third-party requests
Currently, the relevant wording is:

Section 5.3:
> When a user agent receives a Set-Cookie header field in an HTTP
   response, the user agent MAY ignore the Set-Cookie header field in
   its entirety.  For example, the user agent might wish to block
   responses to "third-party" requests from setting cookies (see
   Section 7.1).

Section 5.5:
> A user agent MAY omit the Cookie header in its entirety.  For
   example, the user agent might wish to block sending cookies during
   "third-party" requests from setting cookies (see Section 7.1).

These indicate that a user agent can block setting/sending **all** cookies during third-party requests. This bakes in an assumption that the user agent considers all cookies to be "third-party" during such a request. Chrome is planning on breaking this assumption in the future (notably for https://github.com/cfredric/sameparty/issues/4): some cookies (those with the [SameParty](https://github.com/cfredric/sameparty) attribute) may be considered first-party, so will not be blocked in those cases.

Therefore, it would be nice if the spec indicated that user agents may block a _subset_ of cookies during third-party requests. Section 7.1 could also explicitly mention that user agents might wish to block a subset of cookies during third-party requests, as an alternative to refusing to send the Cookie header at all.

miketaylr:


miketaylr:
PTAL @sbingler

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1831,7 +1831,9 @@ might block third-party cookies entirely by refusing to send Cookie header field
 process Set-Cookie header fields during third-party requests. They might take a less
 draconian approach by partitioning cookies based on the first-party context,
 sending one set of cookies to a given third party in one first-party context,
-and another to the same third party in another.
+and another to the same third party in another. Or they might even send or set
```
sbingler, line 1834: nit: There are a couple usages of "third party" in this paragraph, can you change them to "third-party"?

sbingler, line 1834: > even send or set

Can you change to "even allow"

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1831,7 +1831,9 @@ might block third-party cookies entirely by refusing to send Cookie header field
...
+some third-party cookies, but block others depending on user-agent cookie policy
```
sbingler, line 1835: > some third-party cookies, but block others

I think we can drop the comma
"some third-party cookies but block others"

sbingler (changes requested):

draft-ietf-httpbis-rfc6265bis.md
cfredric: I think "third-party" is correct when used as an adjective, but "third party" is correct when used as a noun (which the usages you're referring to seem to be).

See https://www.merriam-webster.com/dictionary/third-party, https://www.depo.com/wp-content/uploads/2015/02/hyphenate.pdf for example.

sbingler: Ah that makes sense, thanks Chris.

Consider this issue closed.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1831,7 +1831,9 @@ might block third-party cookies entirely by refusing to send Cookie header field
...
+or user controls.
```
cfredric, line 1836: > depending on user-agent cookie policy or user controls

Do we want to include "cookie attributes" as something that could be considered?

draft-ietf-httpbis-rfc6265bis.md
miketaylr: dog

miketaylr: Done.

draft-ietf-httpbis-rfc6265bis.md
miketaylr: I don't think it's strictly necessary - if a UA can define a cookie policy, attributes can be part of that policy (even if not explicitly stated in the rfc).

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1831,7 +1831,9 @@ might block third-party cookies entirely by refusing to send Cookie header field
 process Set-Cookie header fields during third-party requests. They might take a less
 draconian approach by partitioning cookies based on the first-party context,
 sending one set of cookies to a given third party in one first-party context,
-and another to the same third party in another.
+and another to the same third party in another. Or they might even allow or set
```
sbingler, line 1834: > Or they might even allow or set

Why not just "Or they might even allow"?

miketaylr: Sigh, because I'm multi-tasking. That was unintentional. 😭

sbingler (approved):
LGTM

miketaylr:
Mind taking a look and merging @mikewest?

miketaylr:
(I'm assuming this build nonsense has nothing to do with me):

```
/root/.cache/xml2rfc/reference.RFC.8297.xml: fetching from https://www.rfc-editor.org/refs/bibxml/reference.RFC.8297.xml
Error: Unable to parse the XML document: /dev/stdin
 /dev/stdin: Line 1195: ID RFC8820 already defined
```

mnot:
@miketaylr it appears that a cached reference (in GitHub artefacts) became corrupted somehow; I deleted it and re-ran, so it's fine now. Indeed nothing to do with you :)

Pinging @martinthomson FYI.

martinthomson:
Thanks for letting me know.  Nothing much I can do about that; if it happens again, then it might be something that kramdown-rfc2629 can improve (it is the one that generates and reads these cache entries).

PR #1666, opened 2021-09-13
Structured Fields first

### LINKED ISSUE #1606: Parsing as Structured Fields: MAY or SHOULD?
Right now, parsing of targeted Cache-Control is specified like this:

> Targeted fields MAY be parsed as a Dictionary Structured Field [RFC8941], and implementations are encouraged to use a parser for that format in the interests of robustness, interoperability and security. [...] However, implementers MAY reuse a Cache-Control parser for simplicity.

The idea here is that while we'd like the interop benefits of SF, we recognise that some -- perhaps many -- implementers will have a strong motivation to reuse their Cache-Control parsers.

This seems like it might be the worst of both worlds, in that we now have two possible implementation strategies. Given that there are several [SF implementations](https://github.com/httpwg/wiki/wiki/Structured-Fields) now,  I wonder whether should reconsider this. E.g.,

1) Strengthen the first MAY to a SHOULD, making the field sf-first, or even
2) Require SF, don't allow reuse of CC parsers

mnot:
This strengthens the language around Structured Fields so that it's clear they're preferred, and targeted fields are now primarily defined as structured fields.

It still leaves using a Cache-Control parser as an option. If we want to go further, we could remove the paragraph starting with 'However,...' and the two bullet points below that.

Fixes #1606.

### DIFF
--- a/draft-ietf-httpbis-targeted-cache-control.md
+++ b/draft-ietf-httpbis-targeted-cache-control.md
@@ -148,11 +148,12 @@ Because cache directives are not defined in terms of structured data types, it i
 
 For example, the max-age directive ({{Section 5.2.2.1 of HTTP-CACHING}}) has an integer value; no-store ({{Section 5.2.2.5 of HTTP-CACHING}}) always has a boolean true value, and no-cache ({{Section 5.2.2.4 of HTTP-CACHING}}) has a value that can either be boolean true or a string containing a comma-delimited list of field names.
 
-Implementations SHOULD NOT use or generate values that violate specified constraints on the directive's value (e.g. coerce a max-age with a decimal value into an integer). Likewise, implementations SHOULD ignore parameters on directives, unless otherwise specified.
+Implementations SHOULD NOT generate or consume values that violate these inferred constraints on the directive's value (e.g. coerce a max-age with a decimal value into an integer). Likewise, implementations SHOULD ignore parameters on directives, unless otherwise specified.
 
-However, implementers MAY reuse an existing parser for the Cache-Control field value ({{Section 5.2 of HTTP-CACHING}}) for simplicity. If they do so, they SHOULD implement the following constraints, to aid in a smooth transition to a full Structured Field parser and prevent interoperability issues:
+Sending implementations MUST generate valid Structured Fields. Receiving implementations SHOULD use a Structured Fields parser, but MAY reuse an existing parser for the Cache-Control field value ({{Section 5.2 of HTTP-CACHING}}). In doing so, they SHOULD implement the following constraints, to aid in a smooth transition to a full Structured Field parser and prevent interoperability issues:
 
-* If a directive is repeated in the field value (e.g., "max-age=30, max-age=60"), the last value 'wins' (60, in this case)
+* Directive names are all lowercase (e.g., "MAX-AGE=60" is considered an error).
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-targeted-cache-control.md
```diff
@@ -136,18 +140,23 @@ Furthermore, they SHOULD implement other cache directives (including extension c
...
 
-However, implementers MAY reuse a Cache-Control parser for simplicity. If they do so, they SHOULD observe the following points, to aid in a smooth transition to a full Structured Field parser and prevent interoperability issues:
+Sending implementations MUST generate valid Structured Fields. Receiving implementations SHOULD use a Structured Fields parser, but MAY reuse an existing parser for the Cache-Control field value ({{Section 5.2 of HTTP-CACHING}}). In doing so, they SHOULD implement the following constraints, to aid in a smooth transition to a full Structured Field parser and prevent interoperability issues:
 
-* If a directive is repeated in the field value (e.g., "max-age=30, max-age=60"), the last value 'wins' (60, in this case)
+* Directive names are all lowercase (e.g., "MAX-AGE=60" is considered an error).
+* If a directive is repeated in the field value (e.g., "max-age=30, max-age=60"), the last value 'wins' (60, in this case).
 * Members of the directives can have parameters (e.g., "max-age=30;a=b;c=d"), which should be ignored unless specified.
```
martinthomson, line 157: so much "should"

s/should be/are/

draft-ietf-httpbis-targeted-cache-control.md
```diff
@@ -136,18 +140,23 @@ Furthermore, they SHOULD implement other cache directives (including extension c
...
+For example, the max-age directive ({{Section 5.2.2.1 of HTTP-CACHING}}) has an integer value; no-store ({{Section 5.2.2.5 of HTTP-CACHING}}) always has a boolean true value, and no-cache ({{Section 5.2.2.4 of HTTP-CACHING}}) has a value that can either be boolean true or a string containing a comma-delimited list of field names.
 
-When an implementation parses a targeted field as a Structured Field, each cache directive will be assigned a value. For example, max-age has an integer value; no-store’s value is boolean true, and no-cache’s value can either be boolean true or a list of field names. Implementations SHOULD NOT accept other values (e.g. coerce a max-age with a decimal value into an integer). Likewise, implementations SHOULD ignore parameters on directives, unless otherwise specified.
+Implementations SHOULD NOT generate or consume values that violate these inferred constraints on the directive's value (e.g. coerce a max-age with a decimal value into an integer). Likewise, implementations SHOULD ignore parameters on directives, unless otherwise specified.
```
martinthomson, line 151: more "SHOULD" where "MUST" would do (re: the latter; the former is obviously a consequence of permitting the use of a CC parser)

martinthomson (commented):
This seems like the right balance: this is a profile of SF that a CC parser will accept, assuming that a CC parser won't choke on parameterization of "no-cache" or similar.

draft-ietf-httpbis-targeted-cache-control.md
mnot: see 6c4216e1

ioggstream:
Reverted in https://github.com/httpwg/http-extensions/commit/c7dfa098b54f7692a068f1bba4edc3f74ea25a24  for the record

PR #1672, opened 2021-09-14
RFC6265bis: [Editorial] Note `__Secure-` as mitigation.

mikewest:
The "Weak Integrity" section correctly notes that non-secure cookies can be set by
network attackers, and will be sent to secure origins. This commit notes that the
`__Secure-` prefix is a mitigation against that attack.


mikewest:
@miketaylr WDYT?

miketaylr (approved):
LGTM, thanks!

PR #1678, opened 2021-09-17
RFC 6265bis: Add domain attribute '.' prefix/suffix handling to storage model

### LINKED ISSUE #1583: RFC 6265bis: Domain attribute '.' prefix handling only applies to Set-Cookie header parsing
RFC6265bis mentions that a preceding '.' character in a domain attribute should be removed when parsing Set-Cookie headers but doesn't mention this in the more general Storage Model section.

From chatting with @chlily1 about this, it'd be desirable to have this behavior listed in that section as well. Since there are other ways cookies can come in to the cookie store (Ex: Cookie Store API), having these constraints be enforced when adding cookies from any source to the cookie store would better ensure cookie format consistency.

arichiv:
The general storage model should be consistent with the Set-Cookie header parsing model.

The current behavior can be tested as follows:

1. `test1=set1; Domain=github.com` (Should Accept)
2. `test2=set2; Domain=.github.com` (Should Accept)
3. `test3=set3; Domain=github.com.` (Should Reject)

Per-browser results were:

- Chrome 94.0.4606.50 - Consistent
- Firefox: 93.0b6 - Consistent
- Safari: 15.0 (17612.1.28.5) - Didn't reject 3

Fixes #1583

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1443,7 +1443,10 @@ user agent MUST process the cookie as follows:
     1.  Let the domain-attribute be the attribute-value of the last
         attribute in the cookie-attribute-list with both an
         attribute-name of "Domain" and an attribute-value whose
-        length is no more than 1024 octets.
+        length is no more than 1024 octets. (Note that a leading %x2E
+        ("."), if present, is ignored even though that character is not
+        permitted, but a trailing %x2E ("."), if present, will cause
+        the user agent to ignore the attribute.)
 
     Otherwise:
 


mikewest:
Do we have web platform tests for this behavior? The addition sounds fine to me, and is purely editorial given the algorithm, but it'll ideally be locked in with tests as well.

sbingler (approved):
Change LGTM

sbingler:
> Do we have web platform tests for this behavior? The addition sounds fine to me, and is purely editorial given the algorithm, but it'll ideally be locked in with tests as well.

It looks like we do https://wpt.fyi/results/cookies/domain
Unless you have something else in mind?

miketaylr:
We also have https://github.com/web-platform-tests/wpt/blob/master/cookies/attributes/resources/domain-child.sub.html

arichiv:
Looks like full coverage to me

mikewest (approved):
Thanks for checking on the tests!

PR #1685, opened 2021-09-22
Use titles for example messages

MikeBishop:
Rather than using Request: and Response: indications in the text, I think it looks cleaner if these are done using figure titles.  There are relatively few cross-references, but for the one that is present, this also permits anchors and references.  Please do check that I'm representing these accurately, however.

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -785,11 +785,7 @@ Content-Type: application/json
 
 {"hello": "world"}
 ~~~
-
-Here is a gzip-compressed JSON object
-using a content coding.
-
-Request:
+{: title="Request containing a JSON object without any content coding"}
 
 ~~~ http-message
 PUT /entries/1234 HTTP/1.1
@@ -799,6 +795,7 @@ Content-Encoding: gzip
 
 H4sIAItWyFwC/6tWSlSyUlAypANQqgUAREcqfG0AAAA=
 ~~~
+{: title="Request containing a gzip-encoded JSON object"}
 
 Now the same content conveys a malformed JSON object.
 
@@ -811,10 +808,9 @@ Content-Type: application/json
 
 H4sIAItWyFwC/6tWSlSyUlAypANQqgUAREcqfG0AAAA=
 ~~~
+{: title="Request containing malformed JSON"}
 
 A Range-Request alters the content, conveying a partial representation.
-
-Request:
 
 ~~~ http-message
 GET /entries/1234 HTTP/1.1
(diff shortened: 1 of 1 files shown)


ioggstream (approved):
LGTM

draft-ietf-httpbis-digest-headers.md
```diff
@@ -785,11 +785,7 @@ Content-Type: application/json
 
 {"hello": "world"}
 ~~~
-
-Here is a gzip-compressed JSON object
-using a content coding.
-
-Request:
+{: title="Request containing a JSON object without any content coding"}
```
LPardue, line 788: Did you miss the `Request:` instance above on L779?

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1220,27 +1196,13 @@ Note that a `204 No Content` response without content but with the same
...
-~~~ http-message
-PATCH /books/123 HTTP/1.1
-Host: foo.example
-Content-Type: application/merge-patch+json
-Accept: application/json
-Accept-Encoding: identity
-Digest: sha-256=bWopGGNiZtbVgHsG+I4knzfEJpmmmQHf7RHDXA3o1hQ=
-
-{"title": "New Title"}
-~~~
```
LPardue, line 1240: Was this deletion intentional?

LPardue: Oh I see, you ref to the previous example - clever.

LPardue (changes requested):
Thanks Mike, this looks neat (literally). There's a couple of nits still though, so I'll let you push an update and then we can merge

ioggstream:
@MikeBishop I like using figure titles, thank you! I think that probably still retaining Request/Response is more readable since on my html rendering, the request's figure title is closer to the response. 

Probably add some spacing between req. and resp. could work too.

WDYT?

draft-ietf-httpbis-digest-headers.md
MikeBishop: Yes, it was identical to the previous request, so I didn't see a need to repeat it.

MikeBishop:
Hm, I wish there were a way to put the title above the figure, but still part of the figure itself.  @martinthomson, any ~~wizardry~~ suggestions?

martinthomson:
What you want doesn't seem to be possible at the CSS layer.  Figures are generated and rendered as simple block-level elements in HTML, with the caption coming after.  There is only so much you can do to reorder things (tables are different, there you have [caption-side](https://developer.mozilla.org/en-US/docs/Web/CSS/caption-side)).

Some changes to xml2rfc might enable switching things around in stylesheets (you could render figures as tables pretty easily).  But that would still not be down to individual documents.

So changes to the RFC 7991 format is where you are at.  I'm going to assume that this is something you don't want to undertake for this.

ioggstream:
I propose to merge this while we look for a solution and eventually re-add Request / Response alongside with captions

LPardue (approved):

PR #1699, opened 2021-09-28
[WIP] Remove unneeded MESSAGING references.

ioggstream:
## This PR

- replaces unneeded references from MESSAGING with the newer counterparts in I-D semantics

## TODO 

- [ ] verify if terms still occur
- [ ] add references for each term?

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -147,12 +147,16 @@ Additionally, all changes to components not covered by the signature are conside
 
 {::boilerplate bcp14}
 
-The terms "HTTP message", "HTTP request", "HTTP response",
-`absolute-form`, `absolute-path`, "effective request URI",
-"gateway", "header field", "intermediary", `request-target`,
-"sender", and "recipient" are used as defined in {{MESSAGING}}.
+The rule `absolute-path`
+is defined in Section 4.1 of {{SEMANTICS}}.
 
-The term "method" is to be interpreted as defined in Section 4 of {{SEMANTICS}}.
+The terms
+"HTTP message", "HTTP request", "HTTP response,
+"sender", "recipient", "absolute form",
+"intermediary", "gateway"
+"field", "header field",
+"target URI", "recomposted target URI",
+and "method" are to be interpreted as defined in {{SEMANTICS}}.
 
 For brevity, the term "signature" on its own is used in this document to refer to both digital signatures and keyed MACs.  Similarly, the verb "sign" refers to the generation of either a digital signature or keyed MAC over a given input string.  The qualified term "digital signature" refers specifically to the output of an asymmetric cryptographic signing operation.
 


draft-ietf-httpbis-message-signatures.md
```diff
@@ -147,12 +147,16 @@ Additionally, all changes to components not covered by the signature are conside
...
-The term "method" is to be interpreted as defined in Section 4 of {{SEMANTICS}}.
+The rule `absolute-path`
+is defined in Section 4.1 of {{SEMANTICS}}.
+
+The terms
+"HTTP message", "HTTP request", "HTTP response,
+"sender", "recipient", "absolute form",
+"intermediary", "gateway"
+"field", "header field",
+"target URI", "recomposted target URI",
```
LPardue, line 158: typo, composted means something a little different 😆 

But also, you appear to be switching from `effective request URI` to `recomposed target URI` here but there are still instances of effective request URI in the text. It's not clear to me if the term we are importing should be simply `target URI`

ioggstream: Still WIP ;) I need to check if all those terms are still present in the current text.

Would it be helpful to reference the various sections of SEMANTICS for each term?

LPardue: I think being specific for the `absolute-path` rule is good but for the rest I'm happy with just pointing at the document. However if you did want to point out a section for each, a different way to present that is

> The following items from HTTP are used: 
> * the rule `absolute path` ({{Section 4.1 of SEMANTICS}}).
> * `message`  ({{Section xyz of SEMANTICS}}).
> * etc etc

jricher: @LThat kind of link doesn't work with the build

LPardue: That is absolutely the style we should be using. If it's broken in your env then you need to update your environmental

reschke: @jricher - you may have to updated your version of kramdown-rfc2629.

jricher: Good to know, I've just updated to 1.5.6 on my laptop so hopefully that's all it was. I assumed it was some subtle difference between kramdown-rfc2629 and mmark, which I'd seen creep in before.

ioggstream: @jricher use a bash function

```
i-d-template(){
docker run -ti -v $PWD:/code --rm martinthomson/i-d-template --entrypoint /bin/sh martinthomson/i-d-template
}
```

to update image, use docker and clean up lib/:

```
docker pull martinthomson/i-d-template --entrypoint /bin/sh martinthomson/i-d-template
rm lib/ -fr
```

ioggstream:
Closing as stale. In case I'll reopen.

PR #1702, opened 2021-09-29
Don't shout about SETTINGs

martinthomson:


LPardue:
To be fair, RFC 8441 says `Using a SETTINGS parameter`. Didn't you say you liked consistency? ;)

martinthomson:
RFC 8441 is wrong.  Consistency with wrongness should be avoided.

RyanTheOptimist (approved):

Issue #1332, opened 2020-11-25
rfc6265bis empty domain attribute


essen:
The behavior for when the domain attribute is empty is confusing when there are multiple domain attributes.

The [spec](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-06#section-5.3.3) says:

```
   If the attribute-name case-insensitively matches the string "Domain",
   the user agent MUST process the cookie-av as follows.

   1.  If the attribute-value is empty, the behavior is undefined.
       However, the user agent SHOULD ignore the cookie-av entirely.
```

This means that when there's a single empty domain attribute, and the implementation follows this SHOULD, then this attribute is removed. No problem there. But when there's two attributes including an empty domain attribute, then only the empty attribute is removed. This means that the three following set-cookie headers are equivalent:

```
Set-Cookie: foo=bar; domain=foo.example.org; domain=
Set-Cookie: foo=bar; domain=; domain=foo.example.org
Set-Cookie: foo=bar; domain=foo.example.org
```

When the domain is `foo.example.org` all 3 cookies end up stored. When the domain is `home.example.org` all 3 are dropped.

The first case is confusing because it is in stark contrast with the path attribute where an empty path equals `default-path` and the final path wins, so an empty final path has the effect of "resetting" the path. I would also suggest the first case is not intuitive.

An implementation that does not follow the SHOULD would have undefined behavior. This is the only place in the spec where this is the case I believe. It would be good to get rid of it.

I would suggest first changing the SHOULD into a MUST so as to avoid differing behaviors. I would also suggest either having behavior similar to path, or invalidating the cookie entirely.

(This is a followup to #1136 and the relevant http-state test was as follow:)

```
Set-Cookie: foo=bar; domain=foo.example.org; domain=
Location: http://home.example.org:8888/cookie-parser-result?optional-domain0042
```

miketaylr:
This test now lives at https://wpt.fyi/results/cookies/attributes/domain/domain.sub.html?label=master&label=experimental&aligned 

https://github.com/web-platform-tests/wpt/blob/489679e4fccedcb37ae4d3e028b5d1a56adbc769/cookies/attributes/resources/domain-child.sub.html#L311-L316
```
        {
          cookie: `test=42; domain=${www1Host}; domain=`,
          expected: "",
          name: "No cookie returned for domain mismatch (with domain mismatch as first domain attribute and second a bare attribute)",
[...]

PR #1709, opened 2021-10-04
Issue #1332 - Specify empty domain attribute handling.

miketaylr:
PTAL @sbingler.

(Not 100% sure I should set this to the canonicalized request-host, or the empty string, will think about it more tomorrow).

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1250,21 +1250,24 @@ user agent MUST process the cookie-av as follows.
 If the attribute-name case-insensitively matches the string "Domain", the user
 agent MUST process the cookie-av as follows.
 
-1.  If the attribute-value is empty, the behavior is undefined. However, the
-    user agent SHOULD ignore the cookie-av entirely.
+1.  If the attribute-value is empty:
 
-2.  If the first character of the attribute-value string is %x2E ("."):
-
-    1.  Let cookie-domain be the attribute-value without the leading %x2E
-        (".") character.
+    1. Let cookie-domain be the canonicalized request-host.
 
     Otherwise:
 
-    1. Let cookie-domain be the entire attribute-value.
+    2.  If the first character of the attribute-value string is %x2E ("."):
 
-3.  Convert the cookie-domain to lower case.
+        1.  Let cookie-domain be the attribute-value without the leading %x2E
+        (".") character.
 
-4.  Append an attribute to the cookie-attribute-list with an attribute-name
+        Otherwise:
+
+        2. Let cookie-domain be the entire attribute-value.
+
+2.  Convert the cookie-domain to lower case.
+
+3.  Append an attribute to the cookie-attribute-list with an attribute-name
     of Domain and an attribute-value of cookie-domain.
 
 ### The Path Attribute


miketaylr:
Note: this change intends to document what Safari and Firefox currently do, and I'll update WPTs to match once we land it.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1250,21 +1250,24 @@ user agent MUST process the cookie-av as follows.
...
-2.  If the first character of the attribute-value string is %x2E ("."):
+    1. Let cookie-domain be the canonicalized request-host.
 
-    1.  Let cookie-domain be the attribute-value without the leading %x2E
+    Otherwise:
+
+    2.  If the first character of the attribute-value string is %x2E ("."):
+
+        1.  Let cookie-domain be the attribute-value without the leading %x2E
         (".") character.
```
chlily1, line 1262: What if the entire attribute-value is `.`?

Maybe this step should come first: strip off the leading `.` first before mapping `""` to request-host. (See suggestion in other comment.)

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1250,21 +1250,24 @@ user agent MUST process the cookie-av as follows.
...
-    user agent SHOULD ignore the cookie-av entirely.
+1.  If the attribute-value is empty:
 
-2.  If the first character of the attribute-value string is %x2E ("."):
+    1. Let cookie-domain be the canonicalized request-host.
 
-    1.  Let cookie-domain be the attribute-value without the leading %x2E
+    Otherwise:
+
+    2.  If the first character of the attribute-value string is %x2E ("."):
```
chlily1, line 1259: Hmm, I get what this is trying to say, but the indentation makes it somewhat confusing. Can we unnest this somehow?

e.g.

step 0: let cookie-domain be the attribute-value.

step 1: if cookie-domain starts with `.`, let cookie-domain be cookie-domain without its leading character.

step 2: if cookie-domain is empty, let cookie-domain be the canonicalized request host.

sbingler: Agreed, we should flatten this.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1250,21 +1250,24 @@ user agent MUST process the cookie-av as follows.
 If the attribute-name case-insensitively matches the string "Domain", the user
 agent MUST process the cookie-av as follows.
 
-1.  If the attribute-value is empty, the behavior is undefined. However, the
-    user agent SHOULD ignore the cookie-av entirely.
+1.  If the attribute-value is empty:
 
-2.  If the first character of the attribute-value string is %x2E ("."):
+    1. Let cookie-domain be the canonicalized request-host.
```
sbingler, line 1255: If what we want to do is match Firefox's (and presumably Safari's) behavior as documented [here](https://github.com/httpwg/http-extensions/issues/1332#issuecomment-933967006) we shouldn't be setting the cookie-domain, that'd make a cookie that isn't host-only. We should keep it empty.

See [Step 8 of Section 5.5](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-08#section-5.5).

> If the domain-attribute is non-empty:
...
**Otherwise**:
        1.  Set the cookie's host-only-flag to true.

sbingler, line 1255: If what we want to do is match Firefox's (and presumably Safari's) behavior as documented [here](https://github.com/httpwg/http-extensions/issues/1332#issuecomment-933967006) we shouldn't be setting the cookie-domain, that'd make a cookie that isn't host-only. We should keep it empty.

See [Step 8 of Section 5.5](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-08#section-5.5).

> If the domain-attribute is non-empty:
...
**Otherwise**:
        1.  Set the cookie's host-only-flag to true.

sbingler (commented):
Dropped a couple thoughts in. To be clear, we're attempting to match Firefox's behavior [here](https://github.com/httpwg/http-extensions/issues/1332#issuecomment-933967006)?

chlily1 (approved):
Thanks!

PR #1720, opened 2021-10-06
The server doesn't override priority

martinthomson:
The server only informs the client of what it did.  This also informs intermediaries, who might
factor that into how they operate.

I tweaked the next sentence here for grammar, though I think that sentence needs a little work too.

### DIFF
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -376,8 +376,8 @@ where to send registration requests.
 
 The Priority HTTP header field can appear in requests and responses. A client
 uses it to specify the priority of the response. A server uses it to inform
-the client that the priority was overwritten. An intermediary can use the
-Priority information from client requests and server responses to correct or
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-priority.md
```diff
@@ -376,8 +376,8 @@ where to send registration requests.
 
 The Priority HTTP header field can appear in requests and responses. A client
 uses it to specify the priority of the response. A server uses it to inform
-the client that the priority was overwritten. An intermediary can use the
-Priority information from client requests and server responses to correct or
+the client that the priority that was applied. An intermediary can use
```
MikeBishop, line 379:
```suggestion
the client and intemediaries of the priority that was applied. An intermediary can use
```
IIRC, the header in the response is primarily for the benefit of intermediaries -- a client doesn't really care what priority was actually applied, but an intermediary can consider the server's suggestion.

PR #1721, opened 2021-10-06
Merge reprioritization section with frame section

martinthomson:
It's one paragraph that references the next paragraph.  It doesn't need to stand alone.


MikeBishop (approved):
I was worried about losing the anchor, but it appears you've deleted the only reference to it.

PR #1725, opened 2021-10-07
Relax requirements of NO_RFC7540 setting receiver

### LINKED ISSUE #1714: Changes to SETTINGS_NO_RFC7540_PRIORITIES
I couldn't see this being discussed in an issue so far.

> Detection of a change by a receiver MUST be treated as a connection error of type PROTOCOL_ERROR.

This seems unnecessary.  While there is no value to changing the value, it creates a special rule for handling of the setting.  In particular, endpoints that don't distinguish between SETTINGS-in-preface and other SETTINGS frames might have trouble generating the special handling necessary to police this rule.

While I can see the value to an endpoint in terms of being able to defer creating the whole priority tree mess for new connections until they receive the preface, and maybe avoid creating that stuff entirely.  But as long as priority signals are advisory, you get most of the value from a strong recommendation to use the preface, noting that making a change later might be too late.  No special exceptions would be needed.

I'm sure that the editors will disagree with me here.  That's OK, but it's worth noting why this special handling is included in the specification.  It's not immediately obvious.

LPardue:
Fixes #1714 by taking @MikeBishop's suggestion of MAY and bolts it to an "if you can detect" caveat

### DIFF
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -206,8 +206,8 @@ HTTP2}}) of type PROTOCOL_ERROR.
 
 Endpoints MUST send this SETTINGS parameter as part of the first SETTINGS frame.
 A sender MUST NOT change the SETTINGS_NO_RFC7540_PRIORITIES parameter value
-after the first SETTINGS frame. Detection of a change by a receiver MUST be
-treated as a connection error of type PROTOCOL_ERROR.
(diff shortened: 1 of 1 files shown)


LPardue:
Since this is a change of normative requirement, lets solicit for some more WG input before landing

wtarreau:
I noticed the same while reading the document and thought that MAY would be more suitable. Or in more precisely, "MUST ignore any change and MAY treat it as a connection error".

PR #1732, opened 2021-10-08
Standardize maximum Expires/Max-Age

### LINKED ISSUE #1600: RFC6265bis: Standardize maximum Max-Age cookie attribute value/handling
RFC6265bis specifies a 1024 octet limit on all cookie attribute values, but given that the Max-Age attribute value is the number of seconds until the cookie expires, a much lower upper bound must be used in practice.  From some experimentation, these limits and the handling of Max-Age attributes that exceed them vary by browser (see below), and in some cases the differences could be used as a browser fingerprinting vector.  It might be worth having the RFC define appropriate maximum values for the Max-Age field in addition to specifying how values that exceed the max should be handled.

Results of Max-Age attribute experimentation:

In Chrome, if the Max-Age value is sufficiently high (Ex: `9999999999999`), the cookie will get set and DevTools shows it as being a `Session` cookie (in other words, it appears that the attribute is effectively ignored).  If the Max-Age value is something smaller like `7999999999999`, the cookie gets set and the Max-Age value appears to get used directly (DevTools shows it as having a huge date like `+255531-07-10T10:45:20.000Z`). 

In Firefox, if the Max-Age value is above roughly `9223372036854775807-Math.floor(Date.now() / 1000)`, the whole cookie will fail to be set. Otherwise, the cookie gets set and the Max-Age attribute value appears to get used directly (even to the point where large Max-Age values result in an expiration date with such a large year component that `Invalid Date` is shown under the `Expires / Max-Age` column in the Firefox Developer Tools UI).

In Safari, any Max-Age value above `604800` (one week in seconds) appears to be treated as having a value of `604800`, but very large values for Max-Age (even 1000 digits) still result in the cookie being set.

arichiv:
The current spec is ambiguous on (1) what the maximum `Expires`/`Max-Age` attribute values can be and (2) whether the two must be consistent. This resolves both by requiring:

1. The maximum attribute value to be 400 days in the future or less
2. The maximum set for Max-Age and Expires be consistent.

Why 400 days? The goal was to get close to 13 months so that functions one might perform annually (e.g., selecting insurance benefits for the next year) would work even as specific dates varied slightly. Since `Expires` deals in a specific date while `Max-Age` deals in delta-seconds, it seems ideal to select a value with an unambiguous meaning. If the cap were set at 13 months (instead of 400 days) in the future, the maximum TTL varies by the month it was set in and whether or not it's a leap year.

Who is compliant with this modification? Currently, only Safari sets a cap lower than 400 days (one week).

closes #1600

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1247,7 +1247,7 @@ user agent MUST process the cookie-av as follows.
 2.  If the remainder of attribute-value contains a non-DIGIT character, ignore
     the cookie-av.
 
-3.  Let delta-seconds be the attribute-value converted to an integer
+3.  Let delta-seconds be the attribute-value converted to an integer.
  
 4.  Let cookie-age-limit be the maximum age of the cookie (which must be 400 days
     or less, see {{attribute-max-age}}).


chlily1:
Sure, I don't feel strongly about it. If there's to be a hard upper bound on a limit, 400 days seems reasonable.

Separately, maybe there should be suggestion for a lower bound (a minimum capability that servers can expect but should not rely on, a la https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis#section-6.1).

Anyway, this lgtm if @sbingler and @miketaylr approve.

sbingler:
For my own sake I want to clarify I'm how I'm interpreting this proposed change and if it aligns with everyone else.
(I'm not suggesting we use this phrasing in the spec, it's only for illustrative purposes.)


After this change is made:
- UAs **MUST** clamp the upper range of `max-age` to some value `upper_max_age_limit`.
- `upper_max_age_limit` **SHOULD** be equal to 400 days
- `upper_max_age_limit` **MUST** be greater than or equal to 7 days 
  - (The number here is open to discussion but Safari already implements a 7 day limit for cookies set by JS.)

Sites are then free to set a cookie with a `max-age` anywhere from `[0, upper_max_age_limit]` (`0` being shorthand for the UA's  "earliest representable date and time"), and if they exceed the upper range the value will be revised down to `upper_max_age_limit`.

Does this sound agreeable? Does it align with everyone else's thoughts?

miketaylr:
+1 to @mikewest's comment, lower-bound requirements don't seem very useful to add to the spec.

chlily1:
> It can't be a guarantee, because people can (and do) clear cookies earlier. If it isn't a guarantee, it's not clear what value it would provide.

There's some phrasing in https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis#section-6.1 which suggests (SHOULD) some minimum capabilities that can be expected but that aren't guaranteed. I think it's still valuable to provide a ballpark guideline or an expected order of magnitude of the value, even if it's not guaranteed. (And servers are explicitly advised to degrade gracefully rather than rely on them.)

arichiv:
I'm down with leaving the lower limit to future work/discussion. I'm also open to changing the MUST <= 400 days to be a SHOULD if we revisit in a few months to see whether Chrome & Firefox followed suit. If all three major engines followed, then the MUST seems easier to defend that our current juncture.

PR #1737, opened 2021-10-11
Caching subsection

### LINKED ISSUE #1552: Caching considerations
We should figure out/document:

1) **Establishing the cache key** - a cache needs to know how to compute a cache key for the request. The obvious (and very inefficient) way to do this is to use the entire request body as part of the key.

2) **Request canonicalisation** - we should allow caches with knowledge of canonicalisation algorithms for a given format to use them. 

3) **Indicating canonical form in requests** - if a client (UA or intermediary) canonicalises, it would be nice if they could assert that in the request, so a downstream cache can assume it's there and just look for a byte-for-byte match, rather than re-canonicalising. This would allow a client to manufacture cache misses, but that's already possible in URLs...

4) **Indicating cache key in requests** - E.g., a hash of the request body. Needs security analysis, though.

5) **Indicating cache key in responses** - If the server can instruct caches as to what future requests would match it, that would be very advantageous. Necessarily requires knowledge of the request format, but we could define common ones for url encode, XML, JSON.

mnot:
Fixes #1552

### DIFF
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -413,6 +413,7 @@ Dubois, Camille, camille.dubois@example.net
 </section>
 <section title="Since draft-ietf-httpbis-safe-method-w-body-01" anchor="changes.since.01">
 <ul>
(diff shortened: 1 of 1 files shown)


mnot:
Content-Location doesn't have any effect on caching. It would be odd to add caching implications to it.

PR #1746, opened 2021-10-13
Content-Location with QUERY

### LINKED ISSUE #1745: Describe role of Content-Location and Location fields
@reschke mentioned in #1552 that there should be a way for the server to indicate a GET-able equivalent to the query if one exists; this has the benefit of being more easily cacheable by intermediaries which don't understand QUERY.

Based on SEMANTICS, I think the right mechanism for this would be Content-Location.  A server can optionally indicate the GET-able location of the query to which it is responding.

MikeBishop:
Fixes #1745.

### DIFF
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -249,10 +249,10 @@ q=foo&amp;limit=10&amp;sort=-published
       </t>
 
       <t>
-        If a successful response includes a Content-Location header field, the
-        response content also reflects the current state of the indicated
-        resource and MAY also be used to satisfy future requests for that
-        resource.
+        If a successful response includes a Content-Location header field with
+        the same origin as the requested resource, the response content also
+        reflects the current state of the indicated resource and MAY also be
+        used to satisfy future requests for that resource.
       </t>
 
     </section>
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -300,6 +300,7 @@ select surname, givenname, email limit 10
 <artwork type="http-message">
 HTTP/1.1 200 OK
 Content-Type: text/csv
+Content-Location: http://example.org/contacts/query123
 
 surname, givenname, email
 Smith, John, john.smith@example.org
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -212,13 +212,13 @@ q=foo&amp;limit=10&amp;sort=-published
 
     <t>
       A server &MAY; create or locate a resource that identifies the query
-      operation for future use. If the server does so, the server claims that a
-      client can send a GET request to the indicated URI to repeat the query
-      operation just performed without resending the query parameters. The URI
-      of the resource can be included in the Location header field of the
-      response. This resource might be temporary; if a future request fails, the
-      client &SHOULD; retry using the original QUERY resource and the previously
-      submitted parameters again.
+      operation for future use. If the server does so, the URI of the resource
+      can be included in the Location header field of the response. This
+      represents a claim that a client can send a GET request to the indicated
+      URI to repeat the query operation just performed without resending the
+      query parameters. This resource might be temporary; if a future request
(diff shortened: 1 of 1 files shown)


mnot:
`Content-Location` doesn't indicate whether the resource on the other end will return *these* results in perpetuity, or whether it'll return the results to the query (which may change over time). 

So it might be better to be explicit -- if the server wants to create a resource for these results, use `201 Created` with `Location` that explicitly points to a resource that represents this query, and which can change over time. The server can choose whether to include the query results on the 201, of course.

asbjornu:
Could perhaps a `Link` header be used to disambiguate the two use-cases?

```http
Link: </updated-results>; rel="latest-version", </these-results>; rel="memento"
```

The chosen link relations are probably not appropriate, I just picked some that exists to provide a complete example.

MikeBishop:
Per discussion in Brisbane side-meeting:

- Need to add text indicating what it means if the C-L target errors
- Add text clarifying that the resource can be cached despite the different URI (this is cautioned against in SEMANTICS)

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -200,6 +200,14 @@ q=foo&amp;limit=10&amp;sort=-published
       using an HTTP GET request. Various non-normative examples of successful
       QUERY responses are illustrated in <xref target="examples" />.
     </t>
+
```
reschke, line 203:
```suggestion
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -200,6 +200,14 @@ q=foo&amp;limit=10&amp;sort=-published
       using an HTTP GET request. Various non-normative examples of successful
       QUERY responses are illustrated in <xref target="examples" />.
     </t>
+    
+    <t>
+      A successful response to a QUERY request &MAY; include a Content-Location
```
reschke, line 205:
```suggestion
      A successful response to a QUERY request can include a Content-Location
```

MikeBishop (commented):
Review feedback from @reschke.

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -201,6 +201,17 @@ q=foo&amp;limit=10&amp;sort=-published
       QUERY responses are illustrated in <xref target="examples" />.
     </t>
 
+    <t>
+      A successful response to a QUERY request &MAY; include a Content-Location
+      header field (see <xref target="HTTP" section="8.7"/>) containing an
+      identifier for a specific resource corresponding to the results of the
+      operation. A client can send a GET request for the indicated URI to repeat
```
martinthomson, line 208:
```suggestion
      operation. A client can send a GET request for the indicated URI to repeat
```

So is this "A client can" or "This represents a claim from the server that a client can"

martinthomson (commented):
I see nothing here about cacheability or origins.  I know that I left early, but I thought that we had concluded that the server claims about the Content-Location resource would be treated differently based on whether the indicated resource was on the same origin (or not).

MikeBishop:
@martinthomson, I've added language restricting it to same-origin; is there a stronger statement you were expecting to see here?

reschke (approved):

reschke:
@mnot , @royfielding , @martinthomson - should we go ahead with this?

royfielding:
No.

See https://github.com/httpwg/http-extensions/issues/1745#issuecomment-2130499002

reschke:
IETF 120 feedback: define both (the resource that identfies the result of *this* query), and a resource that can be used with GET to repeat the QUERY itself.

reschke:
This looks good (right, @royfielding ?)

Looking at the PR, I was going to ask whether Location and Content-Location can be trailers, but found that this is consistent with RFC 9110.

Will have a look at the full text later to see how in integrates, and whether we may want to have more examples.

reschke:
OK, I think we should add an example to section 4.

@MikeBishop - can you move this PR to the WG repo, so I can add to it?

MikeBishop:
@reschke, GitHub doesn't let me change the source branch of a PR, only the target. I've pushed the same branch to the upstream repo and opened #2851 with the same commit history. You should have full access to that one.

PR #1759, opened 2021-10-19
Fix #1399 - Add a grammar corresponding to the parsing algorithm

### LINKED ISSUE #1399: RFC 6265bis: Set-Cookie parsing algorithm should enforce more of the syntax requirements
There is an ABNF syntax given for the Set-Cookie header [in section 4.4.1](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-07#section-4.1.1), however many aspects of this are not enforced in the actual parsing algorithm [in section 5.3](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-07#section-5.3).

The syntax given in 4.4.1 is only a "SHOULD", not a hard requirement for the server:
```
Servers SHOULD NOT send Set-Cookie headers that fail to conform to
   the following grammar:
```

There is an explicit caveat about this in section 5.3:
```
   NOTE: The algorithm below is more permissive than the grammar in
   Section 4.1.  For example, the algorithm strips leading and trailing
   whitespace from the cookie name and value (but maintains internal
   whitespace), whereas the grammar in Section 4.1 forbids whitespace in
   these positions.  User agents use this algorithm so as to
   interoperate with servers that do not follow the recommendations in
   Section 4.
```

The algorithm in 5.3 should enforce more of the requirements given in the recommended syntax, at the very least the requirements on validity of characters/octets (whitespace, control characters, etc.), and specify the intended behavior in these cases. (For context and examples of such discrepancies, see [here](https://github.com/httpwg/http-extensions/issues/1136#issuecomment-772872621) and [here](https://crbug.com/1174647).)

recvfrom:
This PR add a grammar that corresponds to the algorithm parsing section.  

A grammar currently exists in section 4.4.1 but it's much stricter than what user agents actually enforce, and there have been several instances where that grammar was assumed to correspond with the cookie parsing algorithm defined in section 5.  By having a separate grammar in section 5, hopefully we can reduce confusion while also preserving the high-level view that a grammar provides.

I'm not sure whether it's worth keeping around the existing grammar in section 4.4.1, but in this initial work I have not attempted to remove it.

CC @sbingler and @miketaylr

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1219,6 +1219,39 @@ said to "receive a cookie" from the request-uri with name cookie-name,
 value cookie-value, and attributes cookie-attribute-list. (See {{storage-model}}
 for additional requirements triggered by receiving a cookie.)
 
+### Syntax {#ua-abnf-syntax}
+
+Based on the parsing algorithms defined above, the following grammar defines
+the syntax requirements enforced by user agents when parsing set-cookie-strings:
+
+~~~ abnf
+set-cookie            = set-cookie-string
+set-cookie-string     = cookie-pair *( BWS ";" OWS cookie-av)
+cookie-pair           = *1(BWS cookie-name BWS "=") BWS cookie-value BWS
+                          ; the sum of the lengths of cookie-name and
+                          ; cookie-value must be less than or equal to 4096
+                          ; octets.
+
+cookie-name           = *4096(cookie-name-octet)
+cookie-value          = *4096(cookie-value-octet)
+cookie-name-octet     = %x09 / %x20-3A / %x3C / %x3E-7E / %x80-FF
+                          ; octets excluding non-whitespace CTLs,
+                          ; semicolon, and equals.
+cookie-value-octet    = %x09 / %x20-3A / %x3C-7E / %x80-FF
+                          ; octets excluding non-whitespace CTLs and
+                          ; semicolon.
+
+cookie-attr           = 1*cookie-name-octet BWS cookie-attr-eq-value BWS
+cookie-attr-eq-value  = "" / ("=" BWS optional-attr-value)
+    ; cookie-attr defines the base syntax for all cookie attributes. More
+    ; specific grammars for each of the recognized attribute names can be
+    ; found in the sections below.
+
+optional-attr-value   = *1024(cookie-value-octet)
+ignored-attr-eq-value = "" / ("=" BWS ignored-atrr-value)
+ignored-attr-value    = optional-attr-value
+~~~
+
 ### The Expires Attribute
 
 If the attribute-name case-insensitively matches the string "Expires", the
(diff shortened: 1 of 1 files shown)


recvfrom:
I believe I've address all feedback with the most recent commit, but please let me know if I missed anything.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1378,6 +1385,68 @@ with
...
+cookie-name-octet  = %x09 / %x20-3A / %x3C / %x3E-7E / %x80-FF
+                       ; octets excluding non-whitespace CTLs,
+                       ; semicolon, and equals
+cookie-value-octet = %x09 / %x20-3A / %x3C-7E / %x80-FF
+                       ; octets excluding non-whitespace CTLs and
+                       ; semicolon
+
+cookie-av          = expires-av / max-age-av / domain-av /
+                     path-av / secure-av / httponly-av /
+                     samesite-av / extension-av
```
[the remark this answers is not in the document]
recvfrom: Good point - I moved the attribute grammars into separate ABNFs. Thanks for the feedback!

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1236,6 +1276,15 @@ user agent MUST process the cookie-av as follows.
...
     of Expires and an attribute-value of expiry-time.
 
+Based on the parsing algorithm above, the syntax requirements for the Expires
+attribute are represented by the grammar below. Expires attributes not
+conforming to this grammar are ignored.
+
+~~~ abnf
+expires-av = "Expires" BWS "=" BWS cookie-date BWS
+               ; cookie-date is defined in the "Dates" section.
+~~~
```
reschke, line 1286: At the "Expires" is detected, the parsing into name and value has already happened.

Thus, I would just define the syntax for the value, as in:

~~~
expires-v = cookie-date
~~~

Note this is how multi-level grammars are defined in the core HTTP specs as well.

reschke (changes requested):

draft-ietf-httpbis-rfc6265bis.md
recvfrom: Ah, I see - thank you for the example @reschke.

It seems like we'd lose a bit of completeness if we only include the attribute value in the grammar.  For instance, the grammar wouldn't explicitly tie the "Expires" keyword to the Expires attribute, and wouldn't convey whether "ExPiReS" is treated the same way as "Expires".  Having the grammar format used here align with the convention in the core HTTP specs seems desirable as well, though.

Does anyone else have thoughts on this?

reschke: > . For instance, the grammar wouldn't explicitly tie the "Expires" keyword to the Expires attribute

It by definition can't do that anyway, as long as the grammar allows for extensibilty - an ABNF-driven parser would simply switch to the code branch for new attributes, and then the value syntax would remain unconstrained.

You really can't do this in ABNF; it needs to be done in prose.

recvfrom: If our goal is to have an ABNF-driven parser function correctly, it seems like we'd need to go back to the using the previous definition of `cookie-av`. To address the original concern of "any valid cookie-av will match extension-av", maybe we could do something like the following:
```
cookie-av          = expires-av / max-age-av / domain-av /
                     path-av / secure-av / httponly-av /
                     samesite-av / extension-av

extension-av       = 1*cookie-name-octet BWS extension-eq-value BWS
extension-eq-value = "" / ("=" BWS optional-value)
    ; attributes solely matching extension-av should be treated as unrecognized by
    ; cookie parsers strictly conforming to this RFC
```

What do you think?

reschke: I don't think you can solve this by adding prose (comments) to the ABNF.

The ABNF should be something that can translate to how a parser processes the input. Parsers in general parse first generic structures, then look at specific micro syntaxes of values.

That's what we're doing in the base spec, and that's what I'll keep suggesting to do.

Issue #1687, opened 2021-09-22
"Want-Digest" and Bad Request


MikeBishop:
The Digest mechanism seems to be oriented such that it's always optional -- senders can disregard receiver preferences; receivers can disregard digests in unsupported formats, or disregard them entirely.

As such, the example in https://httpwg.org/http-extensions/draft-ietf-httpbis-digest-headers.html#appendix-C.3 is a bit confusing.  It shows a server responding with a 400 because the client would prefer to receive a digest in a format the server does not support.  However, I don't see any text in this document stating that servers SHOULD/MUST/SHOULD-NOT, etc. reject requests on the basis of requested digests.

Am I just missing it, or does this need more explanatory text?

LPardue:
I agree this is an overloaded example. I think we need to explain more and/or split out the cases.

An alternative interpretation of this example is that the server required a digestvwith the request. There's some ambiguity. I kind of dislike want-digest but we're sort of stuff with it.

LPardue:
Further, to address Mike's point. There's lots of things that could happen and a lack of normative language about what to do. The danger with examples is they become defacto normative unintentionally.

ioggstream:
@MikeBishop there are two different examples, to clarify that both ignoring Digest and requiring it are legitimate. 

1. 
https://github.com/httpwg/http-extensions/blob/main/draft-ietf-httpbis-digest-headers.md#server-selects-algorithm-unsupported-by-client shows that the receiver can ignore want-digest

2. https://github.com/httpwg/http-extensions/blob/main/draft-ietf-httpbis-digest-headers.md#server-does-not-support-client-algorithm-and-returns-an-error shows that the receiver is free to enforce a given policy

I think it's worth improving the explanation though.

PR #1761, opened 2021-10-19
Fix: #1687. Improve  description of want-digest and bad request.

ioggstream:
## This PR

- improves the example description wrt #1687 
- fixes the response, since `Want-Digest` cannot be used to advertise the list of supported algorithms (like Accept) but only the list of the acceptable ones.


LPardue:
I editorialized the words a bit. I'm happy with this so I'll squash and merge.

Issue #1774, opened 2021-10-28
RFC6265bis: Consider minimum Max-Age/Expiration cookie attribute value/handling


arichiv:
We came to an answer to the maximum (400 days) in #1732, but punted on defining a minimum (if any).

One suggestion with plurality support was: "SUOULD be 7 days or more"

miketaylr:
I think this should be out of scope for the 6265bis work, and perhaps considered as an improvement to a future revision to the (new) cookies RFC, given multi-implementor support. 

Right now there's already [signal that we shouldn't pursue this](https://github.com/httpwg/http-extensions/pull/1775#pullrequestreview-792457090).

PR #1775, opened 2021-10-28
Minimum Max-Age/Expiration cookie attribute

arichiv:
We came to an answer to the maximum (400 days) in #1732, but punted on defining a minimum (if any).

One suggestion with plurality support was: "SUOULD be 7 days or more"

#1774


miketaylr:
> As I said on the other issue. I don't think that this is necessary, nor is it useful. Shorter durations will happen. And shorter durations are still potentially useful.

I'm lukewarm on the idea, but think we shouldn't consider this right now as in scope for the 6265bis work. Maybe better to discuss in the issue: https://github.com/httpwg/http-extensions/issues/1774#issuecomment-956333921

PR #1776, opened 2021-10-29
Security and Privacy Considerations for Signatures

jricher:
Adds security and privacy considerations and applies various cleanups including those suggested by early security review

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -354,7 +354,7 @@ The signature parameters component value is the serialization of the signature p
 
 * `created`: Creation time as an `sf-integer` UNIX timestamp value. Sub-second precision is not supported. Inclusion of this parameter is RECOMMENDED.
 * `expires`: Expiration time as an `sf-integer` UNIX timestamp value. Sub-second precision is not supported.
-* `nonce`: A random unique value generated for this signature.
+* `nonce`: A random unique value generated for this signature as an `sf-string` value.
 * `alg`: The HTTP message signature algorithm from the HTTP Message Signature Algorithm Registry, as an `sf-string` value.
 * `keyid`: The identifier for the key material as an `sf-string` value.
 
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -383,7 +383,7 @@ NOTE: '\' line wrapping per RFC 8792
   created=1618884475;expires=1618884775
 ~~~
 
-Note that an HTTP message could contain multiple signatures, but only the signature parameters used for the current signature are included in the entry.
+Note that an HTTP message could contain [multiple signatures](#signature-multiple), but only the signature parameters used for the current signature are included in the entry.
 
 ### Method {#content-request-method}
 
@@ -1041,9 +1041,9 @@ The signer MAY include the `Signature` field as a trailer to facilitate signing 
 
 Multiple `Signature` fields MAY be included in a single HTTP message. The signature labels MUST be unique across all field values.
 
-## Multiple Signatures
+## Multiple Signatures {#signature-multiple}
 
-Multiple distinct signatures MAY be included in a single message. Since `Signature-Input` and `Signature` are both defined as Dictionary Structured fields, they can be used to include multiple signatures within the same HTTP message by using distinct signature labels. For example, a signer may include multiple signatures signing the same message components with different keys or algorithms to support verifiers with different capabilities, or a reverse proxy may include information about the client in fields when forwarding the request to a service host, including a signature over the client's original signature values.
+Multiple distinct signatures MAY be included in a single message. Each distinct signature MUST have a unique label. Since `Signature-Input` and `Signature` are both defined as Dictionary Structured fields, they can be used to include multiple signatures within the same HTTP message by using distinct signature labels. These multiple signatures could be added all by the same signer or could come from several different signers. For example, a signer may include multiple signatures signing the same message components with different keys or algorithms to support verifiers with different capabilities, or a reverse proxy may include information about the client in fields when forwarding the request to a service host, including a signature over the client's original signature values.
 
 The following is a non-normative example of header fields a reverse proxy sets in addition to the examples in the previous sections.
 
@@ -1401,6 +1401,12 @@ To provide confidentiality at the transport level, TLS or its equivalent can be 
 ## Oracles {#privacy-oracle}
 
 It is important to balance the need for providing useful feedback to developers on error conditions without providing additional information to an attacker. For example, a naive but helpful server implementation might try to indicate the required key identifier needed for requesting a resource. If someone knows who controls that key, a correlation can be made between the resource's existence and the party identified by the key. Access to such information could be used by an attacker as a means to target the legitimate owner of the resource for further attacks.
+
+## Required Content {#privacy-required}
+
+A core design tenet of this specification is that all message components covered by the signature need to be available to the verifier in order to recreate the signature input string and verify the signature. As a consequence, if an application of this specification requires that a particular field be signed, the verifier will need access to the value of that field. 
+
+For example, in some complex systems with intermediary processors this could cause the surprising behavior of an intermediary not being able to remove privacy-sensitive information from a message before forwarding it on for processing, for fear of breaking the signature. A possible mitigation for this specific situation would be for the intermediary to verify the signature itself, then modifying the message to remove the privacy-sensitive information. The intermediary can ad its own signature at this point to signal to the next destination that the incoming signature was validated, as is shown in the example in {{signature-multiple}}.
 
 --- back
 
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -1296,7 +1296,22 @@ The table below contains the initial contents of the HTTP Signature Specialty Co
 
 # Security Considerations {#security}
 
-## Use TLS {#security-tls}
+In order for an HTTP message to be considered covered by a signature, all of the following conditions have to be true:
+
+- a signature is expected or allowed on the message
+- the signature exists on the message
+- the signature is verified against the identified key material and algorithm
+- the key material and algorithm are appropriate for the context of the message
+- the signature is within expected time boundaries
+- the signature covers the expected content, including any critical components
+
+## Signature Verification Skipping {#security-ignore}
+
+HTTP Message Signatures only provide security if the signature is verified by the verifier. Since the message to which the signature is attached remains a valid HTTP message without the signature fields, it is possible for a verifier to ignore the output of the verification function and still process the message. Common reasons for this could be relaxed requirements in a development environment or a temporary suspension of enforcing verification during debugging an overall system. Such temporary suspensions are difficult to detect under positive-example testing since a good signature will always trigger a valid response whether or not it has been checked.
+
+To detect this, verifiers should be tested using both valid and invalid signatures, ensuring that the invalid signature fails as expected.
+
+## Use of TLS {#security-tls}
 
 The use of HTTP Message Signatures does not negate the need for TLS or its equivalent to protect information in transit. Message signatures provide message integrity over the covered message components but do not provide any confidentiality for the communication between parties.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-message-signatures.md
```diff
@@ -1283,22 +1291,122 @@ The table below contains the initial contents of the HTTP Signature Specialty Co
...
+
+If a verifier wants to trigger a new signature from a signer, it can send the `Accept-Signature` header field with a new `nonce` parameter. An attacker that is simply replaying a signature would not be able to generate a new signature with the chosen nonce value.
+
+## Insufficient Coverage {#security-coverage}
+
+Any portions of the message not covered by the signature are susceptible to modification by an attacker without affecting the signature. An attacker can take advantage of this by introducing a header field or other message component that will change the processing of the message but will not be covered by the signature. Such an altered message would still pass signature verification, but when the verifier processes the message as a whole, the unsigned content injected by the attacker would subvert the trust conveyed by the valid signature.
+
+To combat this, an application should require as much of the message as possible to be signed, within the limits of the application and deployment. The verifier should only trust message components that have been signed.
```
jyasskin, line 1317: Would it make sense to recommend that signature verifiers be structured to return a whole message instead of a boolean result, so that they can strip out the unverified components?

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1283,22 +1291,122 @@ The table below contains the initial contents of the HTTP Signature Specialty Co
...
+
+If a verifier wants to trigger a new signature from a signer, it can send the `Accept-Signature` header field with a new `nonce` parameter. An attacker that is simply replaying a signature would not be able to generate a new signature with the chosen nonce value.
+
+## Insufficient Coverage {#security-coverage}
+
+Any portions of the message not covered by the signature are susceptible to modification by an attacker without affecting the signature. An attacker can take advantage of this by introducing a header field or other message component that will change the processing of the message but will not be covered by the signature. Such an altered message would still pass signature verification, but when the verifier processes the message as a whole, the unsigned content injected by the attacker would subvert the trust conveyed by the valid signature.
```
jyasskin, line 1315: I think "the verifier processes the message as a whole" conflates two components of the system: the piece that verifies the signature, and the piece that then processes the message. Could you introduce separate terms for those?

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1283,22 +1291,122 @@ The table below contains the initial contents of the HTTP Signature Specialty Co
...
+
+To combat this, an application should require as much of the message as possible to be signed, within the limits of the application and deployment. The verifier should only trust message components that have been signed.
+
+## Cryptography and Signature Collision {#security-collision}
+
+The HTTP Message Signatures specification does not define any of its own cryptographic primitives, and instead relies on other specifications to define such elements. If the signature algorithm or key used to process the signature input string is vulnerable to any attacks, the resulting signature will also be susceptible to these same attacks.
+
+A common attack against signature systems is to force a signature collision, where the same signature value an be applied to multiple different inputs. Since this specification relies on reconstruction of the input string based on an HTTP message, and the list of components signed is fixed in the signature, it is difficult but not impossible for an attacker to effect such a collision. An attacker would need to manipulate the HTTP message and its covered message components in order to make the collision effective.
```
jyasskin, line 1323:
```suggestion
A common attack against signature systems is to force a signature collision, where the same signature value can be applied to multiple different inputs. Since this specification relies on reconstruction of the input string based on an HTTP message, and the list of components signed is fixed in the signature, it is difficult but not impossible for an attacker to effect such a collision. An attacker would need to manipulate the HTTP message and its covered message components in order to make the collision effective.
```

jyasskin: And I'm not sure that "be applied to" is the right term here ... "where the same signature value can verify multiple different inputs"?

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1283,22 +1291,122 @@ The table below contains the initial contents of the HTTP Signature Specialty Co
...
+
+To counter this, only vetted keys and signature algorithms should be used to sign HTTP messages. The HTTP Message Signatures Algorithm Registry is one source of potential trusted algorithms.
+
+## Key Theft {#security-keys}
+
+A foundational assumption of signature-based cryptographic systems is that the signing key is not compromised by an attacker. If the keys used to sign the message are exfiltrated or stolen, the attacker will be able to generate their own signatures using those keys. As a consequence, signers have to protect any signing key material from exfiltration, capture, and use by an attacker.
+
+## Modification of Required Message Parameters {#security-modify}
+
+An attacker could effectively deny a service by modifying an otherwise benign signature parameter or signed message component. While rejecting a modified message is the desired behavior, consistently failing signatures could lead to the verifier signature checking in order to make systems work again.
```
jyasskin, line 1333: I don't understand "could lead to the verifier signature checking".

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1283,22 +1291,122 @@ The table below contains the initial contents of the HTTP Signature Specialty Co
...
+
+To counteract this, implementations should use fully compliant and trusted parsers for all structured field processing, both on the signer and verifier side.
+
+## Choosing Message Components {#signature-components}
+
+Applications of HTTP Message Signatures need to decide which message components will be covered by the signature. Depending on the application, some components could be expected to be changed by intermediaries prior to the signature's verification. If these components are covered, such changes would, by design, break the signature. In such circumstances, it is often tempting for the verifier to simply turn off signature verification entirely in order to allow these otherwise good messages to pass. 
+
+However, the HTTP Message Signature standard allows for flexibility in determining which components are signed precisely so that a given application can choose the appropriate portions of the message that need to be signed, avoiding problematic components. For example, a web application framework that relies on rewriting query parameters might avoid use of the `@query` content in favor of sub-indexing the using `@query-params` instead.
```
jyasskin, line 1381:
```suggestion
However, the HTTP Message Signature standard allows for flexibility in determining which components are signed precisely so that a given application can choose the appropriate portions of the message that need to be signed, avoiding problematic components. For example, a web application framework that relies on rewriting query parameters might avoid use of the `@query` content in favor of sub-indexing it using `@query-params` instead.
```
I think?

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1283,22 +1291,122 @@ The table below contains the initial contents of the HTTP Signature Specialty Co
...
+
+An attacker could effectively deny a service by modifying an otherwise benign signature parameter or signed message component. While rejecting a modified message is the desired behavior, consistently failing signatures could lead to the verifier signature checking in order to make systems work again.
+
+If such failures are common within an application, the signer and verifier should compare their generated signature input strings with each other to determine which part of the message is being modified.
+
+## Mismatch of Signature Parameters from Message {#security-mismatch}
+
+The verifier needs to make sure that the signed message components match those in the message itself. This specification encourages this by requiring the verifier to derive these values from the message, but lazy cacheing or conveyance of the signature input string to a processing system could lead to downstream verifiers accepting a message that does not match the presented signature.
+
+## Symmetric Cryptography {#security-symmetric}
```
jyasskin, line 1341: In systems that allow both symmetric and asymmetric signatures in the same format, there have also been attacks where the attacker replaces, say, RSA with HMAC and uses the public key to generate a "valid" signature: https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/#RSA-or-HMAC-.

https://httpwg.org/http-extensions/draft-ietf-httpbis-message-signatures.html#name-application-of-http-message mentions this kind of consideration, but I think it should also be in the security considerations.

draft-ietf-httpbis-message-signatures.md
jricher: I think that works, we'll play with the language here.

draft-ietf-httpbis-message-signatures.md
jricher: By "verifier" we mean "the party that verifies the message with its signature" and not "the verification function". The function returns a boolean with the message as input, so the caller of that function would have access to the whole message anyway. I think having more precise terms isn't a bad idea, as long as we're using them consistently.

draft-ietf-httpbis-message-signatures.md
jricher: Should have been:

```suggestion
However, the HTTP Message Signature standard allows for flexibility in determining which components are signed precisely so that a given application can choose the appropriate portions of the message that need to be signed, avoiding problematic components. For example, a web application framework that relies on rewriting query parameters might avoid use of the `@query` content in favor of sub-indexing the query value using `@query-params` instead.
```

draft-ietf-httpbis-message-signatures.md
jricher: That's what this section was getting at, but I think it's worth calling out this specific example as well. Thanks!

draft-ietf-httpbis-message-signatures.md
jricher: See other comment: the verifier calls the verification function and so the boolean response makes sense here, but more precise term usage is a good call.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1283,22 +1291,161 @@ The table below contains the initial contents of the HTTP Signature Specialty Co
...
+
+For example, if a buggy parser of the `@signature-input` value does not enforce proper closing of quotes around string values within the list of component identifiers, an attacker could take advantage of this and inject additional content into the signature input string through manipulating the `Signature-Input` field value on a message.
+
+To counteract this, implementations should use fully compliant and trusted parsers for all structured field processing, both on the signer and verifier side.
+
+## Choosing Message Components {#signature-components}
+
+Applications of HTTP Message Signatures need to decide which message components will be covered by the signature. Depending on the application, some components could be expected to be changed by intermediaries prior to the signature's verification. If these components are covered, such changes would, by design, break the signature. In such circumstances, it is often tempting for the verifier to simply turn off signature verification entirely in order to allow these otherwise good messages to pass.
```
martinthomson, line 1412: This is the second time that you have mentioned "just turn it off".  I get that you want to acknowledge this tension that is created by security mechanisms that make things not work, but I think that you should be a little more careful in phrasing these to avoid creating the wrong impression.

martinthomson (commented):
This is very long.  I'm afraid I ran out of time and didn't read it all.  Maybe I'll get back to it later.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1283,22 +1291,161 @@ The table below contains the initial contents of the HTTP Signature Specialty Co
...
+
+To detect this, verifiers should be tested using both valid and invalid signatures, ensuring that the invalid signature fails as expected.
+
+## Use of TLS {#security-tls}
+
+The use of HTTP Message Signatures does not negate the need for TLS or its equivalent to protect information in transit. Message signatures provide message integrity over the covered message components but do not provide any confidentiality for the communication between parties.
+
+TLS also protects the signature data itself from being captured by an attacker, which is an important step in preventing [signature replay](#security-replay).
```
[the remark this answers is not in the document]
jricher: The assumption on the use of TLS is that TLS is not itself compromised. Can this be stated more clearly? This section does still need to point to BCP195.

draft-ietf-httpbis-message-signatures.md
jricher: Ah thanks -- the other section was added after we wrote this paragraph, we can remove that text here. And I agree, it does need to be phrased carefully. However, I have seen too many security systems not acknowledge the fact that when security gets in the way of functionality, functionality naturally wins. I think that's particularly true for something like this, as described in the other section.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1283,22 +1291,161 @@ The table below contains the initial contents of the HTTP Signature Specialty Co
...
+
+## Cryptography and Signature Collision {#security-collision}
+
+The HTTP Message Signatures specification does not define any of its own cryptographic primitives, and instead relies on other specifications to define such elements. If the signature algorithm or key used to process the signature input string is vulnerable to any attacks, the resulting signature will also be susceptible to these same attacks.
+
+A common attack against signature systems is to force a signature collision, where the same signature value successfully verifies against multiple different inputs. Since this specification relies on reconstruction of the input string based on an HTTP message, and the list of components signed is fixed in the signature, it is difficult but not impossible for an attacker to effect such a collision. An attacker would need to manipulate the HTTP message and its covered message components in order to make the collision effective.
```
richanna, line 1338: These security considerations make me want to reframe signature verification as "verify a signature against your requirements" rather than just "verify a signature against its input". E.g., verification steps are:
1. Identify the set of message components required/expected/allowed to be signed, keys and algs permitted, etc.
2. Verify that the signature input matches requirements.
3. Verify that the corresponding signature matches signature input and message. (i.e., existing verification instructions)

This highlights the importance of identifying requirements and testing against those first and foremost - a "valid" signature that doesn't cover what's needed is irrelevant - and emphasizes this as a critical part of the verifier's job. As it is, this is tacked on at the end of the verification process. I want to encourage library interfaces that look like `verify(message, required_components)` not `verify(message)`.

jricher: I agree with that in principle, and I get where it's coming from. We've got some precedent in the world of JOSE implementations where you send not only the JWS but also the required header and (sometimes) body claims for a JWT. My only concern is that there is a separate definitive step of doing the signature verification calculation, and we don't want that to get lost in the overall discussion. How can we frame it so that we mostly talk about the larger piece but still can call out the smaller piece when necessary? Namely when we define how the algorithms are applied.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1283,22 +1292,167 @@ The table below contains the initial contents of the HTTP Signature Specialty Co
...
 
-There are a number of security considerations to take into account when implementing or utilizing this specification.  A thorough security analysis of this protocol, including its strengths and weaknesses, can be found in {{WP-HTTP-Sig-Audit}}.
+For example, in some complex systems with intermediary processors this could cause the surprising behavior of an intermediary not being able to remove privacy-sensitive information from a message before forwarding it on for processing, for fear of breaking the signature. A possible mitigation for this specific situation would be for the intermediary to verify the signature itself, then modifying the message to remove the privacy-sensitive information. The intermediary can ad its own signature at this point to signal to the next destination that the incoming signature was validated, as is shown in the example in {{signature-multiple}}.
```
ryan-castner, line 1449: typo: `The intermediary can ad its own signature` => `The intermediary can add its own signature`

PR #1777, opened 2021-11-01
add parentheses around URI

reschke:


draft-ietf-httpbis-targeted-cache-control.md
```diff
@@ -185,7 +185,7 @@ This specification does not place any specific requirements on implementations t
 
 ## Defining Targeted Fields
 
-A targeted field for a particular class of cache can be defined by requesting registration in the Hypertext Transfer Protocol (HTTP) Field Name Registry <https://www.iana.org/assignments/http-fields/>, listing this specification as the specification document. The Comments field of the registration should clearly define the class of caches that the targeted field applies to.
+A targeted field for a particular class of cache can be defined by requesting registration in the Hypertext Transfer Protocol (HTTP) Field Name Registry (<https://www.iana.org/assignments/http-fields/>), listing this specification as the specification document. The Comments field of the registration should clearly define the class of caches that the targeted field applies to.
```
martinthomson, line 188: My experience with inline links is that xml2rfc makes a real hash of this format.  Try this:

```suggestion
A targeted field for a particular class of cache can be defined by requesting registration in the Hypertext Transfer Protocol (HTTP) Field Name Registry ([](https://www.iana.org/assignments/http-fields/)), listing this specification as the specification document. The Comments field of the registration should clearly define the class of caches that the targeted field applies to.
```

Issue #855, opened 2019-07-10
Shall we use sh-binary serialization for `id-sha` digest-algoritms?


ioggstream:
## Hypothesis
Considering that:

1- every digest-algorithm define his own serialization method (eg. you don't need to use just b64)

2- Structured headers allow to [serialize byte-sequences](https://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html#binary) like the following:

```
sh-binary = "*" *(base64) "*"
base64    = ALPHA / DIGIT / "+" / "/" / "="
``` 

3- the newly defined digest-algorithms `id-sha-256` and `id-sha-512` do not have a 
    predefined encoding format, so that we can chose which one to use.

We could:

- use for `id-sha-256|512` the `sh-binary` serialization

## Notes

Pro:
- compatibility with structured headers

Cons:
- syntax will be different from the (previous) sha-xxx

martinthomson:
I think that it would be best to consider digests as octet sequences.  You can encode all sorts of stuff into one of those.

LPardue:
I'm not sure that answers @ioggstream original question as intended. If we rephrase the question to be:

Spec says "The encoded digest output uses the encoding format defined for the specific digest-algorithm.". Presently, SHA based algorithms encode using base64 and Unixsum uses ASCII decimal-digit string. `id-sha-256` and `id-sha-512` are new digest-algorithms added in this update but presently have no defined encoding format; should they also be base64 encoded or something else?

@martinthomson are you suggesting that the encoding format be octects?

martinthomson:
Yes, in that every digest algorithm produce a single output that is a sequence of octets and that the new SH be used to encode that.  I know that this doesn't fit with existing code, but you could carve out exceptions for legacy identifiers (and then discourage their use).

dlongley:
It would be great to consider [Multibase](https://tools.ietf.org/html/draft-multiformats-multibase-01) (see also: https://github.com/multiformats/multibase) to solve this problem for current and future algorithms.

LPardue:
My inner pragmatist leans toward simply using base64 encoded for id-sha-256|512. This is partly driven by the possibility of endpoints generating digests `id-sha-256` and `sha-256` with associated `content-encoding: identity`. Having different encoding formats for effectively the same thing seems weird.

ioggstream:
@dlongley I will propose a PR to multibase so that it will add an algorithm to the iana digest-algorithm table 

[...]

PR #1779, opened 2021-11-09
Fix: #855. Remove id-sha-*.

ioggstream:
## This PR

- strip out id-* algorithms from this spec

## This is done

- id- are complex to address in this document
- they will be better addressed in https://github.com/ioggstream/draft-polli-id-digest-algorithms

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1003,11 +1003,7 @@ Digest: sha-256=4REjxQ4yrqUVicfSKYNO/cF9zNj5ANbzgDZt3/h3Qxo=
 
 ## Client and Server Provide Full Representation Data
 
-The response contains two digest values:
-
-- one with no content coding applied, which in this case accidentally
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -1028,7 +1001,7 @@ Digest: sha-256=4REjxQ4yrqUVicfSKYNO/cF9zNj5ANbzgDZt3/h3Qxo=
 ~~~
 {: title="Empty response with Digest"}
 
-## Client and Server Provide Full Representation Data, Client Uses id-sha-256.
```
LPardue, line 1031: Is the comment about the second digest value at https://github.com/httpwg/http-extensions/pull/1779/files#diff-de34152d297a9eb29b85b24962949ba2e309478a5c32e1d94fd4c3ee27b4584dL1037 still correct with this change?

LPardue (commented):
LGTM modulo the comment I made

ioggstream:
Fix typo. Thx @LPardue

PR #1785, opened 2021-11-15
Be clearer about handling of h2 priority

martinthomson:
Not a complete solution, but hopefully helpful.  More to come for those who are less cautious than
I.


LPardue (approved):
Thank you, this looks clear and accurate to me.

PR #1787, opened 2021-11-16
Relocate appendix text to introduction

LPardue:
This editorial change combines most references to how this document relates to RFC 3230. By dropping the appendix, it makes the flow of the document more smooth, which is reflected in the ToC.


ioggstream (approved):
LGTM

PR #1791, opened 2021-11-17
Add binary messages

martinthomson:
I'm pleased to see that adding the file is all that is necessary.

I didn't retain the history of the file; this is a simple copy.  It's a bit of work to figure out
how to do that.

### DIFF
--- a/draft-ietf-httpbis-binary-message.md
+++ b/draft-ietf-httpbis-binary-message.md
@@ -13,6 +13,7 @@ venue:
   mail: ietf-http-wg@w3.org
   arch: https://lists.w3.org/Archives/Public/ietf-http-wg/
   repo: https://github.com/httpwg/http-extensions/labels/binary-messages
+github-issue-label: binary-messages
 
 stand_alone: yes
 pi: [toc, sortrefs, symrefs, docmapping]
--- a/draft-ietf-httpbis-binary-message.md
+++ b/draft-ietf-httpbis-binary-message.md
@@ -1,0 +1,634 @@
+---
+title: "Binary Representation of HTTP Messages"
+abbrev: Binary HTTP Messages
+docname: draft-ietf-httpbis-binary-message-latest
+category: std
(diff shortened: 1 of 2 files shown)


draft-ietf-httpbis-binary-message.md
```diff
@@ -0,0 +1,634 @@
+---
+title: "Binary Representation of HTTP Messages"
+abbrev: Binary HTTP Messages
+docname: draft-ietf-httpbis-binary-message-latest
+category: std
+ipr: trust200902
+area: ART
+workgroup: HTTPBIS
```
mnot, line 8: Just 'HTTP', not the short name

draft-ietf-httpbis-binary-message.md
```diff
@@ -0,0 +1,634 @@
...
+  QUIC: RFC9000
+
+informative:
+
+
+--- abstract
+
+This document defines a binary format for representing HTTP messages.
+
+
```
mnot, line 38: Add the Note to Readers?

draft-ietf-httpbis-binary-message.md
martinthomson: I thought that we had to use the short name here.  I guess that this is display, not metadata.

draft-ietf-httpbis-binary-message.md
```diff
@@ -0,0 +1,641 @@
...
+area: ART
+workgroup: HTTP
+venue:
+  group: HTTP
+  type: Working Group
+  home: https://httpwg.org/
+  mail: ietf-http-wg@w3.org
+  arch: https://lists.w3.org/Archives/Public/ietf-http-wg/
+  repo: https://github.com/httpwg/http-extensions/labels/binary-messages
+
```
mnot, line 16: Needs `github-issue-label` too. (we might want to put that under  `venue` eventually, but that can be done separately.

PR #1795, opened 2021-11-22
HTTP/3 websocket should be standards track

LPardue:
I assume the current category is just an oversight


martinthomson (approved):

RyanTheOptimist (approved):
Oh, yes good point! Thanks!

PR #1798, opened 2021-11-29
Fix digest in method-rsa-pss-sha512 definition

clehner:
It looks like SHA-512 was intended here, as that is what the description and name use.
Maybe someone had in mind to add an additional rsa-pss-sha256 method.


jricher:
Good catch, thank you!

PR #1799, opened 2021-11-29
clarify priority parameter registration

LPardue:
During IANA review of the document it was noted that the instructions for populating the new HTTP Priority Parameters registry with the urgency and incremental parameters was not that clear. So this change lists both `u` and `i` to make it clear what we want.


kazuho (approved):
LGTM. Thank you for the quick work.

PR #1800, opened 2021-11-29
cite experiments of simpler priority schemes

LPardue:
In the Genart review of this document, it was requested to provided citations of such experiments. These two are the best examples I can think of, which came up during the motivation to form a priorities design team and take on the work.

@rmarx does this reference to your paper look ok?

### DIFF
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -42,6 +42,31 @@ informative:
     date: 2019-03-01
     target: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-9513
 
+  MARX:
+    target: https://www.scitepress.org/Link.aspx?doi=10.5220/0008191701300143
+    title: "Of the Utmost Importance: Resource Prioritization in HTTP/3 over QUIC"
+    date: 2019-09
+    author:
+    -
+      name: Robin Marx
+    -
+      name: Tom De Decker
+    -
+      name: Peter Quax
+    -
+      name: Wim Lamotte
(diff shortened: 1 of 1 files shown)


kazuho (approved):

rmarx:
@LPardue looks good to me! Also happy this is getting included; I'd considered asking for it myself, but didn't want to seem vain. 

The link should probably be https://www.doi.org/10.5220/0008191701300143 instead, which auto-redirects to what you have now but is more canonical. 

That said though, the official hosting of the paper is pay-to-access, which I'm not a particularly big fan of. 
Alternatives could be https://h3.edm.uhasselt.be/ or https://www.researchgate.net/publication/335977773_Of_the_Utmost_Importance_Resource_Prioritization_in_HTTP3_over_QUIC,
though those are slightly less guaranteed to keep working for many years down the line...

Your choice :)

draft-ietf-httpbis-priority.md
```diff
@@ -42,6 +42,31 @@ informative:
     date: 2019-03-01
     target: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-9513
 
+  MARX:
+    target: https://www.scitepress.org/Link.aspx?doi=10.5220/0008191701300143
```
LPardue, line 46:
```suggestion
    target: https://www.doi.org/10.5220/0008191701300143
```

LPardue:
I'll stick with the www.doi.org link for know, folks that dislike paywalls can find alternative access as long as they know what they are trying to view.

Issue #1802, opened 2021-12-03
TSVART review - editorial comments


LPardue:
Comments from Bob Briscoe's review:

General (all sections).

- [x]  E#0a) Are precedence and priority interchangeable with urgency?

It's called a priority field, and the parameter is called urgency, but
sometimes the term priority or precedence is used to describe urgency. Please
go through the draft using priority or urgency consistently, and remove
precedence unless it's there for good reason.

§1. Introduction

- [ ] E#1a) The "For example" in para 2 is actually the general point not the
example, whereas the example is in the previous para.

- [x] E#1b) The last para "   The prioritization scheme and priority signals defined
herein can act as a substitute for RFC 7540 stream priority." would fit better
3 paras earlier, just after "HTTP/2 [HTTP2] has consequently deprecated the use
of these stream priority signals."

- [ ] E#1c) It would help to state what the interaction model is for the priority
field. I believe it's not as simple as just "e2e" (see technical comment
earlier about §5). Although I've suggested that normative text about this ought
to be in the §5, it needs to be stated early on (probably in the intro, not
just the abstract), because the reader needs it to understand §§2 & 4.

§1.1.  Notational Conventions

- [ ] E#1d) Was HTTP/2 priority only from C-S?

   The term HTTP/2 priority signal is used to describe the priority
   information sent from clients to servers in HTTP/2 frames;

Neither RFC7540 nor the http2bis draft says that an HTTP/2 priority signal
cannot be sent by a server. It may be that this was the unstated intention but,
if it wasn't, the above sentence is incorrect.

§2.  Motivation for Replacing RFC 7540 Priorities

- [x] E#2a) Not just absence

CURRENT:

   compatibility (see Section 5.3.2 of [HTTP2]), which means that they
   might still be used in the absence of alternative signaling, such as
   the scheme this document describes.

PROPOSED:

   compatibility (see Section 5.3.2 of [HTTP2]), because they
   might still be used by other nodes.

REASONING:
7540 priority fields are not only used in the absence of alternative
signalling. They are used by the client in parallel to extensible priorities
before it receives the server's SETTINGS frame.

2.1.1. Advice when Using Extensible Priorities as the Alternative

- [ ] E#2b)
CURRENT:

   might be useful to nodes behind the server that the client is
   directly connected to.

PROPOSED:

   might be useful to a server behind the directly connected node.

REASONING:
Removes ambiguity - I initially read this as "nodes (behind the server) that"
rather than "...the server that".

[...]

PR #1804, opened 2021-12-03
Integrate RFC 7540 replacement with deprecation notage

LPardue:
addresses E#1b of #1802


kazuho (approved):
This is better. Thank you for the PR.

PR #1805, opened 2021-12-03
H2 priority elements might always be used

LPardue:
Addresses E#2a of #1802 

Bob correctly highlights that absence of a replacement might not be the only condition under which the prior signals might be used. So lets be more accurate.


kazuho (approved):
Thank you for the PR. Looks good.

PR #1810, opened 2021-12-03
trim text to focus on why e2e header

LPardue:
Addresses E#14b, E#14c, E#14d of #1802

This cuts some text to reduce the waffle factor ;)﻿


kazuho (approved):
This is nice condensation.

PR #1813, opened 2021-12-05
Tweak the server scheduling text to clarify what is covered

LPardue:
Addresses E#10a of #1802

Bob comments that the "no guidance is provided" s﻿ounds like it is in conflict with the rest of the section that includes some normative recommendations. The intent of this text was to highlight there is no (or at least very little) about how servers are supposed to combine priority parameters with the million other things that they have to worry about. 

This change shuffles the sentences around to try and make the original intent clearer.


kazuho (approved):
This is much better than the original. Thank you for the PR.

PR #1823, opened 2021-12-06
Recommend 501 for unknown protocols

### LINKED ISSUE #1773: Server advertise support for extended CONNECT but does not support websocket as a :protocol
I think this is an issue. and we need graceful way to fall back to fallback. 

I do not believe that HTTP_1_1_REQUIRED and 421 response code are the right solution.

I think it would be good to make this part of WebSockets over HTTP3 spec.

edit: remove double negation -> I think that   HTTP_1_1_REQUIRED and 421 response code are not the answer.

RyanTheOptimist:
Fixes #1773

### DIFF
--- a/draft-ietf-httpbis-h3-websockets.md
+++ b/draft-ietf-httpbis-h3-websockets.md
@@ -75,7 +75,8 @@ SETTINGS_ENABLE_CONNECT_PROTOCOL value is 0x08 (decimal 8), as in HTTP/2.
 If a server which advertises support for Extended CONNECT but receives an
 Extended CONNECT request with a ":protocol" value that is unknown or is
 not supported, the server SHOULD respond to the request with a 501 (Not
-Implemented) status code ({{Section 15.6.2 of HTTP}}).
+Implemented) status code ({{Section 15.6.2 of HTTP}}). A server MAY
+provide more information via a Problem Details resoponse {{!RFC7807}}.
 
 # Security Considerations
 
--- a/draft-ietf-httpbis-h3-websockets.md
+++ b/draft-ietf-httpbis-h3-websockets.md
@@ -71,6 +71,10 @@ in HTTP/2 as defined {{!RFC8441}}. {{Appendix A.3 of HTTP3}} requires that
 HTTP/3 settings be registered separately for HTTP/3. The
 SETTINGS_ENABLE_CONNECT_PROTOCOL value is 0x08 (decimal 8), as in HTTP/2.
 
+If a server receives an Extended CONNECT request with a ":protocol"
+value that is unknown or is not supported, the server SHOULD respond
+to the request with a 501 status code, see {{Section 6.6.2 of RFC7231}}.
+
 # Security Considerations
 
 This document introduces no new security considerations beyond those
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-h3-websockets.md
```diff
@@ -71,6 +71,10 @@ in HTTP/2 as defined {{!RFC8441}}. {{Appendix A.3 of HTTP3}} requires that
 HTTP/3 settings be registered separately for HTTP/3. The
 SETTINGS_ENABLE_CONNECT_PROTOCOL value is 0x08 (decimal 8), as in HTTP/2.
 
+If a server receives an Extended CONNECT request with a ":protocol"
+value that is unknown or is not supported, the server SHOULD respond
+to the request with a 501 status code, see {{Section 6.6.2 of RFC7231}}.
```
LPardue, line 76: The build failure is this reference. FWIW you probably want to point to https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#status.501 and you could do so by adding `HTTP` to your normative refs and then using that here.

draft-ietf-httpbis-h3-websockets.md
```diff
@@ -71,6 +71,10 @@ in HTTP/2 as defined {{!RFC8441}}. {{Appendix A.3 of HTTP3}} requires that
...
+
```
LPardue, line 77: @mnot suggested supplying a header with this error and that sounds good. But I'm not sure which one he intended. Proxy-Status would not seem right though

draft-ietf-httpbis-h3-websockets.md
LPardue: Stil lthink you're better off target RFC-to-be replacement for RFC7231 :D

RyanTheOptimist: Thanks! Sorry, I totally missed your comment when I was fixing the build error. Hopefully I got it right this time! :)

draft-ietf-httpbis-h3-websockets.md
RyanTheOptimist: Agreed. Hopefully @mnot can suggest an appropriate header here.

draft-ietf-httpbis-h3-websockets.md
LPardue: heh 409 conflict ;)

draft-ietf-httpbis-h3-websockets.md
RyanTheOptimist: Body sounds like a fine place for the explanation to me.

Re: Alt-Svc, that's a fair point. It makes me wonder if we should consider (in a different draft) adding an httpOnly= tag to Alt-Svc to restrict the advertisement to HTTP traffic, and not other protocols (or some similar mechanism) to avoid the potential problem @martinthomson describes?

martinthomson: I don't think that is backwardly compatible.  Nor would a flag that says "just send this request (or requests to this resource) to the alternative".

mnot: Body is fine; perhaps mention RFC7807 as an option.

mnot (approved):

ddragana (approved):

draft-ietf-httpbis-h3-websockets.md
RyanTheOptimist: Thanks for the suggestion. Mentioned RFC 7807

PR #1832, opened 2021-12-10
Padding

martinthomson:
This adds the ability to pad messages to the binary HTTP format.

The design is pretty simple: add zero-valued bytes to the end.


tfpauly (approved):

LPardue:
What if someone adds a non-zero byte in the padding block?

chris-wood (approved):

Issue #1834, opened 2021-12-11
Define or cite "incremental digest-algorithm"


LPardue:
In @jricher's WGLC review, he notes several instances of the "incremental digest-algorithm" and comments

> no additional information or explanation as to what that would be or how it would work

We should define or cite examples of these. Failing that, we should consider removing the text, since it's a bit hand-wavy.

ioggstream:
@LPardue agree removing the text.

PR #1848, opened 2021-12-11
Fix: #1834. Remove incremental algorithms.

ioggstream:
## This PR

Fix: #1834. Remove incremental algorithms.

@LPardue tweak the text as you prefer and merge :)

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -219,6 +219,10 @@ The term "checksum" describes the output of the application of an algorithm
 to a sequence of bytes,
 whereas digest is only used in relation to the value of the fields.
 
+When it is convenient to do so,
+the sender and the receiver can dynamically compute the digest value
+while streaming the content.
+
 # The Digest Field {#digest}
 
 The `Digest` field contains a comma-separated list of one or more representation
@@ -281,10 +285,6 @@ that the recipient will ignore it.
 In this case,
 `Digest` MAY be merged into the header section; see {{Section 6.5.1 of SEMANTICS}}.
 
-When an incremental digest-algorithm
-is used, the sender and the receiver can dynamically compute a checksum
-while streaming the content.
-
 A non-comprehensive set of examples showing the impacts of
 representation metadata, payload transformations and HTTP methods on `Digest` is
 provided in {{examples-unsolicited}} and {{examples-solicited}}.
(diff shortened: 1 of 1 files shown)


LPardue:
The new text looks ok but since it only appears in the Representation Digest section, it gives the impression that it doesn't apply to content digest.

ioggstream:
@LPardue PTAL

draft-ietf-httpbis-digest-headers.md
```diff
@@ -419,6 +411,10 @@ Insecure digest algorithms MAY be used to preserve integrity against accidental
 NOT be used in a potentially adversarial setting; for example, when signing the digest of content for
 authenticity.
 
+When it is convenient to do so,
+the sender and the receiver can dynamically compute the checksum value
+while streaming the content.
+
```
LPardue, line 417: As an idea, what do you think about moving these 3 paragraphs to above the declaration of the registry? I could propose a follow up PR to this one if its easier

ioggstream: @LPardue Great, thanks!

ioggstream: I did it... PTAL

draft-ietf-httpbis-digest-headers.md
```diff
@@ -375,10 +367,14 @@ Want-Content-Digest: sha-512;q=0.3, sha-256;q=1, unixsum;q=0
 
 Digest-algorithm values are used to indicate a specific digest computation.
 
-~~~
+~~~ abnf
    digest-algorithm = token
 ~~~
 
+When it is convenient to do so,
```
jricher, line 374: I'd like to see some indication of what would make it "convenient" here -- this is as unclear as "incremental" was before.

ioggstream: Probably it's better to mention incremental algorithms then :P

```suggestion
When using algorithms based on incremental computation,
(see {{?RFC1624}}) 
```

ioggstream: alternatively just pass the ball to implementers.

```suggestion
When implementers deem convenient to do so,
```

ioggstream: @LPardue if none of the above solutions work, I'd just remove these 3 lines since checksum implementers already know that :)

LPardue: lets just delete it

Issue #1841, opened 2021-12-11
Explain the reasons why algorithm agility could be bad


LPardue:
From @jricher's WGLC review:

> §8.6: This draft's algorithm agility effectively allows an attacker to choose and substitute the algorithm applied to the message, which could allow the attacker to choose a weaker algorithm and apply a collision hash to a request. Signing the digest headers, with things like message signatures, mitigates this by fixing the values in the digest header. Also, a recipient should have a limited set of algorithms that it should accept for a given application, and not just take whatever the message creator (or attacker) sends it.

ioggstream:
@jricher is it ok to integrate the security consideration on signatures? Do you think we should describe stuff in other sections too?

ioggstream:
re-reading, this is explained in "replacing rfc 3230". I'll add a link in the related sec-

PR #1851, opened 2021-12-11
Fix: #1841. Reference Algorithm substitution.

ioggstream:
Fix: #1841.

-  Reference Algorithm substitution.
- details on algorithm agility

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -635,8 +635,12 @@ or by sending multiple representation-data-digest values from which the receiver
 Endpoints are advised that sending multiple values consumes resources,
 which may be wasted if the receiver ignores them (see {{digest}}).
 
-Mechanisms such as TLS and digital signatures mitigate substitution attacks
-(see Section 1 of {{?RFC6211}}) on digest-algorithms.
+Differently from [RFC3230], Algorithm Agility cannot be achieved
+parameterizing digest algorithms for security reasons
+(see {{replacing-rfc-3230}}).
+Mechanisms such as TLS and digital signatures can
+mitigate substitution attacks on digest-algorithms
+(see Section 1 of {{?RFC6211}}).
 
 ## Duplicate digest-algorithm in field value
 
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -635,12 +635,12 @@ or by sending multiple representation-data-digest values from which the receiver
 Endpoints are advised that sending multiple values consumes resources,
 which may be wasted if the receiver ignores them (see {{digest}}).
 
-Differently from [RFC3230], Algorithm Agility cannot be achieved
-parameterizing digest algorithms for security reasons
-(see {{replacing-rfc-3230}}).
-Mechanisms such as TLS and digital signatures can
-mitigate substitution attacks on digest-algorithms
-(see Section 1 of {{?RFC6211}}).
+While algorithm agility allows the migration to stronger algorithms
+it does not prevent the use of weaker algorithms.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -635,6 +635,13 @@ or by sending multiple representation-data-digest values from which the receiver
 Endpoints are advised that sending multiple values consumes resources,
 which may be wasted if the receiver ignores them (see {{digest}}).
```
LPardue, line 636: The point I believe @jricher made in #1841 is that it's just as easy to downgrade to weak algorithms by stripping off better ones. Since digest has no in built mechanisms to prevent downgrade, if people care then they should protect the digest fields. That could be done, for example, by using TLS or digital signatures. It could also be done by endpoints restricting the algorithms that they support - potentially rejecting processing of messages that do not contain a digest contained in their supported set,

The algorithm substitution attack mentioned in the next paragraph and described in RFC6211 seems to be something a bit different. It seems to predicate on the verification depending on the length of the hash, not the so much algorithm itself. Since the algorithm identifier is always included in digest, I don't know how much this one applies. Perhaps if there is a digest algorithm that encodes its parameters in the representation-data-digest, as we suggest in Section 1.3. But again the only way to protect from manipulation of field values is to use something like TLS or sigs.

So how about rephrasing the paragraph addition to something like:

While algorithm agility allows the migration to stronger algorithms it does not prevent the use of weaker algorithms. Digest and Content-Digest do not provide any mitigiations for downgrade or substitution (see Section 1 of {{?RFC6211}} attacks of the digest-algorithm. To protect against such attacks, endpoints could restrict their set of supported algorithms to stronger ones and protect the value of Digest and Content-Digest by using TLS and/or digital signatures.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -635,6 +635,13 @@ or by sending multiple representation-data-digest values from which the receiver
...
 
+Differently from [RFC3230], Algorithm Agility cannot be achieved
+parameterizing digest algorithms for security reasons
+(see {{replacing-rfc-3230}}).
+Mechanisms such as TLS and digital signatures can
+mitigate substitution attacks on digest-algorithms
+(see Section 1 of {{?RFC6211}}).
```
ioggstream, line 643:
```suggestion
While algorithm agility allows the migration to stronger algorithms
it does not prevent the use of weaker algorithms.
Digest fields do not provide any mitigiations for downgrade or substitution 
(see Section 1 of {{?RFC6211}} attacks of the digest-algorithm.
To protect against such attacks, endpoints could restrict their set of supported algorithms
to stronger ones and protect the fields value by using TLS and/or digital signatures.
```

draft-ietf-httpbis-digest-headers.md
ioggstream: @LPardue @jricher PTAL

draft-ietf-httpbis-digest-headers.md
```diff
@@ -635,6 +635,13 @@ or by sending multiple representation-data-digest values from which the receiver
 Endpoints are advised that sending multiple values consumes resources,
 which may be wasted if the receiver ignores them (see {{digest}}).
 
+While algorithm agility allows the migration to stronger algorithms
+it does not prevent the use of weaker algorithms.
+Digest fields do not provide any mitigiations for downgrade or substitution 
+(see Section 1 of {{?RFC6211}} attacks of the digest-algorithm.
```
LPardue, line 641:
```suggestion
attacks (see Section 1 of {{?RFC6211}}) of the digest-algorithm.
```

I made a typo in my orginal proposal, sorry

LPardue (approved):
LGTM modulot the one nit

Issue #1837, opened 2021-12-11
Discuss hash collisions in security considerations


LPardue:
There's a few mentions of collisions peppered through the code. We should probably have a paragraph (at least) on the topic in the Security Considerations.

ioggstream:
Ok, though this is mentioned in algorithm definitions (sha, md5).

LPardue:
Since @jricher raised this issue, I'd like his opinion too. 

As Roberto says, we note the collision attacks in the algorithm definitions. What more would a specific section in the Security considerations have to say?

jricher:
Choosing a secure algorithm is a security consideration, and they're only as secure as the underlying algorithm. It seems silly but it should be pointed out explicitly as a security consideration here that the digest is only as good as the algorithm used to create it. The registry is a great way to make that decision easier but the underlying consideration is that some digests are easier to force collisions -- for example, by adding semantically irrelevant padding to the content.

ioggstream:
@jricher pls see/integrate the linked PR

jricher:
@ioggstream I did, it doesn't go far enough. That's just a consideration for the registry, but like I said above it needs to be spelled out more as considerations for implementors, not just the experts running the registry.

ioggstream:
@jricher do you suggest to add the following?

1- the digest is only as good as the algorithm used to create it.
2- some digests are easier to force collisions -- for example, by adding semantically irrelevant padding to the content.


1. while it's tautological, I'm ok with adding this
2. that some digests are easier to force collisions is implied by the fact of having insecured SHA-1 and MD5 
3. I'm ok wrt mentioning "adding semantically irrelevant padding to the content" together with the altering content using `content-enconding` in the "signature" section.

WDYT?

jricher:
@ioggstream All of that would be good to add, but it needs to be in a security consideration paragraph about choosing algorithms, not in the instructions about registering algorithms. No implementor will read the IANA paragraphs that the PR currently targets.

PR #1853, opened 2021-12-13
Fix: #1837. Mention collisions.

ioggstream:
## This PR

Fix: #1837. Mention collisions.

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -554,7 +554,7 @@ to choose hashing algorithms from the IANA Hash Algorithms for HTTP Digest Field
 
 An endpoint might have a preference for algorithms,
 such as preferring "standard" algorithms over "insecure" ones
-(e.g. vulnerable to collisions, as in {{NO-MD5}} and {{NO-SHA}}).
+(e.g. vulnerable to collisions, as in {{NO-MD5}} and {{NO-SHA1}}).
 Transition from weak algorithms is supported
 by negotiation of hashing algorithm using `Want-Representation-Digest` or `Want-Content-Digest` (see {{want-fields}})
 or by sending multiple representation-data-digest values from which the receiver chooses.
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -556,7 +556,7 @@ The "standard" algorithms listed in this document are suitable for many purposes
 including adversarial situations where hash functions might need
 to provide resistance to collision, first-preimage and second-preimage attacks.
 Algorithms listed as "insecure" either provide none of these properties,
-or are known to be weak (see {{NO-MD5}} and {{NO-SHA1}}).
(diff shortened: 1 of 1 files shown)


jricher (changes requested):
As mentioned in the linked issue, this PR is not complete: there needs to be a discussion of collisions and, more broadly, algorithm selection in the security considerations, since this draft leaves the choice of which algorithms to accept entirely up to the implementor. I would also expect to see a discussion of the danger of a receiver of Digest headers blindly accepting "all" algorithms defined in the draft or registry, even if they aren't deprecated. For example, a system in which all "good" clients use SHA512 but still accepts SHA256 is only really protected by SHA256 since that's the lightest weight item an attacker can make use of.

ioggstream:
@martinthomson WDYT?

martinthomson:
If you are talking about the standard assumptions about what a cryptographic hash function provides, you need to talk about what properties you are getting (or, conversely, what the not-secure options don't provide).  So not just collisions, but first- and second-preimage resistance.

Justin also points out that you need to consider the target security level in making any recommendation.  As in, it might be that a 256-bit hash suffices in some cases, but not others.  I think that is secondary to what I think we need to be saying here, but it's probably worth including something.

> The "standard" algorithms listed in this document are suitable for many purposes, including adversarial situations where hash functions might need to provide resistance to collision, first-preimage and second-preimage attacks.  Algorithms listed as "insecure" either provide none of these properties, or are known to be weak (see {{NO-MD5}} and {{NO-SHA1}}).
>
> For adversarial situations, which of the "standard" algorithms are acceptable will depend on the level of protection the circumstances demand. As there is no negotiation, endpoints that depend on a (digest|hash) for security will be vulnerable to attacks on the weakest algorithm they are willing to accept.

ioggstream:
Thanks @martinthomson! @jricher can you PTAL?

LPardue (approved):
LGTM, thank you all!

draft-ietf-httpbis-digest-headers.md
```diff
@@ -627,8 +627,17 @@ To help endpoints distinguish weaker algorithms from stronger ones,
 this document adds to the IANA Digest Algorithm Values registry
 a new "Status" field containing the most recent appraisal of the digest-algorithm.
 
-An endpoint might have a preference for algorithms,
-such as preferring "standard" algorithms over "insecure" ones.
+The "standard" algorithms listed in this document are suitable for many purposes,
+including adversarial situations where hash functions might need
+to provide resistance to collision, first-preimage and second-preimage attacks.
+Algorithms listed as "insecure" either provide none of these properties,
+or are known to be weak (see {{NO-MD5}} and {{NO-SHA1}}).
```
LPardue, line 634:
```suggestion
or are known to be weak (see {{NO-MD5}} and {{NO-SHA}}).
```

ref typo causing build failure

LPardue:
apologies, the ref is `NO-SHA` in main. I changed that to avoid table width alignment problems. if you rebase, it should work. If you need to rename `NO-SHA` to `NO-SHA1` that is fine too.

Issue #1844, opened 2021-12-11
Review use of checksum term


LPardue:
From @jricher's WGLC review:

> Throughout document, "checksum" and "digest" are used interchangeably, is that intentional? It would probably be better to pick one, and I suggest "digest".

Personally, I think the use of both terms is ok. Mainly because digest is overloaded in the document. But it wouldn't hurt us to review the document for usages and perhaps add a note in notational conventions about the use of terms.

ioggstream:
`checksum` is a generic checksum computation,
`digest` should be the actual digest algoritm computation. We need to clarify this and probably introduce new terminology.

PR #1854, opened 2021-12-13
Fix: #1844. Consistent use of checksum.

ioggstream:
## This PR

Fix: #1844. Consistent use of "checksum".

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -411,7 +411,7 @@ Registrations MUST include the following fields:
      - e.g. according to the type and status of the primary document
      in which the algorithm is defined;
      "insecure" when the algorithm is insecure;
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -407,7 +411,7 @@ Registrations MUST include the following fields:
      - e.g. according to the type and status of the primary document
      in which the algorithm is defined;
      "insecure" when the algorithm is insecure;
-     "reserved" when Digest algorithm references a reserved token value
+     "reserved" when the digest algorithm references a reserved token value
```
LPardue, line 414:
```suggestion
     "reserved" when the algorithm references a reserved token value
```

this would make it consistent with the line above

ioggstream: rebased and merged. Thx

PR #1855, opened 2021-12-13
Corruption of data is not always only accidental

### LINKED ISSUE #1838: Corruption might not be by accident
From @jricher's WGLC review:

> §8.1: It's not just accidental corruption, but also malicious changes. The important part here is that it only protects the message body and not the rest.

Corruption could also be non-accidental. Let's remove the accidental qualifier or add words about explicit manipulation (for good or bad purposes).

LPardue:
Fixes #1838 

Spotted another mention of accidental while I was at it, so changed that too.


ioggstream (approved):
LGTM

PR #1856, opened 2021-12-13
Simplify IANA algorithm descriptions

### LINKED ISSUE #1836: Reduce cruft in the IANA algorith value descriptions
From @jricher's WGLC review

> Some of the existing algorithm descriptions should be toned down. Specifically, CRC32c claiming "better hamming distance" and talking about its uses is completely out of place in this registry.

This seems like cruft. The important things for the table are clear pointers to the algorithm definition and the encoding of values created by the algorithm.

LPardue:
Fixes #1836

There's not really any reason to describe anything other than the algorithm used and the format of encoding.


ioggstream (approved):
LGTM

PR #1862, opened 2021-12-17
Signature Cleanup

### LINKED ISSUE #1537: [SIGNATURES] Define a Registry Process
The IANA registries process needs to be defined in the specification. In addition, a few questions on how the IANA registries are maintained need to be contemplated before text can be finalized:

* Can an algorithm be registered from an I-D? Is a final RFC required for registration?
* Are aliases allowed? Can "ecdsa-2013" refer to "ecdsa-p256-sha256"?
* What is the appeals process (when there is disagreement with the Experts)?
* What is the process when there is non-response from the Experts?

### LINKED ISSUE #1538: [SIGNATURES] Are algorithm aliases allowed in the registry?
Will the algorithm registry allow the registration of aliases for algorithms? For example:

* recommended-signature-2015 (aliased to eddsa-2015... or just this value -- no aliases)
* recommended-nist-2013 (aliased to ecdsa-2013... or just this value -- no aliases)

### LINKED ISSUE #1607: Use of `Date` for creation time - potential conflict with some caching deployments
Background: I've added and have been field-testing HTTP response signatures in my business's API, per the earlier [Cavage HTTP draft spec](https://tools.ietf.org/id/draft-cavage-http-signatures-12.html). I hope that this is OK to post my issue here — trying to save other teams time in resolving this issue, granted they encounter it. I attempted to deliver a similar message to the ietf-http-wg mailing list, but since I'm not a W3 member, I don't think that's allowed, or at least it doesn't seem like it is.

With that said —

As you all may know, response signatures often utilize the standard `Date` response header to help prevent replay attacks, verify message integrity, among other things. In fact, the Cavage spec even defines the `Date` header as the default test for the response `Signature` header.

I wanted to give a PSA that Cloudflare rewrites the `Date` response header at the edge. Meaning, if a particularly unlucky request leaves a server at `14:01:26.999` with a `Date` header `Thu, 12 Aug 2021 14:01:26 GMT` and takes ~1ms to process at the edge by Cloudflare, the `Date` header from Cloudflare to the client ends up being `Thu, 12 Aug 2021 14:01:27 GMT` (note the change to the time's second). As of this issue, there is no way to disable this behavior.

Obviously, this is problematic when it comes to a cryptographic signature of an HTTP response, since the signature is invalidated once the `Date` header is modified. This ends up causing sporadic, hard-to-debug response signature verification failures, because the server is sending the correct dataset, but the client is not receiving the correct dataset. It's my understanding that the `Date` header is "the date and time at which the message _originated._"

I understand that the original draft spec by Cavage had the default test utilize the `Date` header, so this renders Cloudflare unusable for even the default test if signature reliability is a requirement (and it should be). I'm not sure if [the newer version of the spec](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-message-signatures) also defaults to using `Date`, but regardless — if a signature contains the `Date` header and is proxied by Cloudflare, signatures will occasionally be invalidated.

A workaround is to use `X-Date` or similar instead of `Date`, but I do not find that to be a satisfactory workaround. And that doesn't take into account a default test that utilizes the standard `Date` header, which could be implemented more widely granted the draft spec is standardized in the future.

The other solution is to, of course, disable any proxies that behave in this way.

### LINKED ISSUE #1811: signatures: format of references into SEMANTICS
RFC Editor recommends either:

  Section xxx of [DOC]

or

  [DOC], Section xxx.

(Also note that there's a kramdown format that actually generates the proper deep link).

jricher:
Prepare for next release:

- update introduction, especially expirations
- add signature targets and update IANA
- clarify sign/verify algorithms
- editorial fixes

Closes #1537
Closes #1538
Closes #1607
Closes #1811


draft-ietf-httpbis-message-signatures.md
```diff
@@ -1872,17 +1894,16 @@ The internal service can validate the proxy's signature and therefore be able to
...
 Mark Adamcin,
 Mark Allen,
 Paul Annesley,
-Karl Böhlmark,
-Stéphane Bortzmeyer,
+{{{Karl Böhlmark}}},
+{{{Stéphane Bortzmeyer}}},
 Sarven Capadisli,
 Liam Dennehy,
-ductm54,
```
jricher, line 1885: Removed because I don't know who this is and we don't have a full name.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1872,17 +1894,16 @@ The internal service can validate the proxy's signature and therefore be able to
...
+This specification was initially based on the draft-cavage-http-signatures internet draft.  The editors would like to thank the authors of that draft, Mark Cavage and Manu Sporny, for their work on that draft and their continuing contributions. The specification also includes contributions from the draft-oauth-signed-http-request internet draft and other similar efforts.
 
 The editors would also like to thank the following individuals for feedback, insight, and implementation of this draft and its predecessors (in alphabetical order):
 Mark Adamcin,
 Mark Allen,
 Paul Annesley,
-Karl Böhlmark,
-Stéphane Bortzmeyer,
+{{{Karl Böhlmark}}},
+{{{Stéphane Bortzmeyer}}},
```
jricher, line 1904: Syntax required for non-ascii characters in names.

PR #1866, opened 2021-12-21
Make it clearer why and how clients send PRIORITY_UPDATE

LPardue:


kazuho (approved):
Nice improvements. Thank you for the PR.

PR #1872, opened 2021-12-24
Cite only Marx paper

LPardue:
I was noted that the citation to Pat Meenan's blog post lacked scientific experimental data and a decent permalink. Since we already acknowledge Pat's original scheme proposal in the acknowledgements, lets just remove this one. 

To avoid anyone picking up on a lack of multiple citations to match the multiple experiments statement, lets drop that qualifier too.


kazuho (approved):
Makes sense. Thank you for the PR.

Issue #1596, opened 2021-08-09
Editorial: good (client cert) header field description in section 2


b---c:
based on some twitter feedback https://twitter.com/dret/status/1422583903894786051

PR #1874, opened 2021-12-27
Client certificate chain

### LINKED ISSUE #1521: Certificate chain data
@b---c wrote:
> TODO: It has been suggested that more information about the certificate chain
might be needed/wanted by the backend application (to independently evaluate the
cert chain, for example, although that seems like it would be terribly
inefficient) and that any intermediates as well as the root should also be
somehow conveyed, which is an area for further discussion should this draft
progress. One potential approach suggested by a few folks is to allow some
configurability in what is sent along with maybe a prefix token to indicate
what's being sent - something like `Client-Cert: FULL \<cert> \<intermediate>
\<anchor>` or `Client-Cert: EE \<cert>` as the strawman. Or a perhaps a
parameter or other construct of {{?RFC8941}} to indicate what's being sent. It's
also been suggested that the end-entity certificate by itself might sometimes be
too big (esp. e.g., with some post-quantum signature schemes). Hard to account
for it both being too much data and not enough data at the same time. But
potentially opening up configuration options to send only specific attribute(s)
from the client certificate is a possibility for that. In the author's humble
opinion the end-entity certificate by itself strikes a good balance for the vast
majority of needs and avoids optionality. But, again, this is an area for
further discussion should this draft progress.

MikeBishop:
Fixes #1521, based on #1873.

### DIFF
--- a/draft-ietf-httpbis-client-cert-field.md
+++ b/draft-ietf-httpbis-client-cert-field.md
@@ -119,37 +119,29 @@ Finished messages.
 
 # HTTP Header Field and Processing Rules
 
-## Encoding
-
-The field-values of the HTTP header field defined herein utilize the following
-encoded form.
-
-A certificate is represented in text as an `EncodedCertificate`, which is the
-base64-encoded (Section 4 of {{!RFC4648}}) DER {{!ITU.X690.1994}} PKIX certificate. The
-encoded value MUST NOT include any line breaks, whitespace, or other additional
-characters. ABNF {{?RFC5234}} syntax for `EncodedCertificate` is shown in the
-figure below.
-
-~~~
- EncodedCertificate = 1*( DIGIT / ALPHA / "+" / "/" ) 0*2"="
-
- DIGIT = <Defined in Section B.1 of [RFC5234]>  ; A-Z / a-z
- ALPHA = <Defined in Section B.1 of [RFC5234]>  ; 0-9
-~~~
-
 ## Client-Cert HTTP Header Field {#header}
 
-In the context of a TLS terminating reverse proxy (TTRP) deployment, the TTRP
+In the context of a TLS terminating reverse proxy (TTRP) deployment, the proxy
 makes the TLS client certificate available to the backend application with the
-following header field.
+Client-Cert HTTP header field. This field contains the end-entity certificate
+used by the client in the TLS handshake.
 
-Client-Cert:
-: The end-entity client certificate as an `EncodedCertificate` value.
+Client-Cert is an Item Structured Header {{!RFC8941}}.  Its value MUST be a
+Byte Sequence ({{Section 3.3.5 of RFC8941}}).  Its ABNF is:
 
-The `Client-Cert` header field defined herein is only for use in HTTP requests
-and MUST NOT be used in HTTP responses.  It is a single HTTP header field value
-as defined in Section 3.2 of {{?RFC7230}}, which MUST NOT have a list of values or
-occur multiple times in a request.
+~~~
+ Client-Cert = sf-binary
+~~~
+
+The binary sequence is the DER {{!ITU.X690.1994}} PKIX certificate. The encoded
+value MUST NOT include any line breaks, whitespace, or other additional
+characters. A binary sequence which cannot be successfully parsed as a
+certificate MUST be ignored.
+
+The `Client-Cert` header field is only for use in HTTP requests and MUST NOT be
+used in HTTP responses.  It is a single HTTP header field value as defined in
+Section 3.2 of {{?RFC7230}}, which MUST NOT have a list of values or occur
+multiple times in a request.
 
 ## Processing Rules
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-client-cert-field.md
```diff
@@ -121,35 +121,73 @@ Finished messages.
...
+
+The first certificate in the list SHOULD directly certify the end-entity
+certificate provided in the `Client-Cert` header; each following certificate
+SHOULD directly certify the one immediately preceding it.  Because certificate
+validation requires that trust anchors be distributed independently, a
+certificate that specifies a trust anchor MAY be omitted from the chain,
+provided that the server is known to possess any omitted certificates.
+
+However, for maximum compatibility, servers SHOULD be prepared to handle
+potentially extraneous certificates and arbitrary orderings.
```
jricher, line 190: Arbitrary ordering worries me here -- I think we can have a well-defined _order_ without necessarily having it be a complete (as in, no-gaps) certificate chain with specific endpoints. Maybe something like:

> The certificates in the `Cient-Cert-Chain` header field MUST be increasingly general, with the later certificates signing earlier certificates in the list.

... Or something along those lines. I think the warning about extraneous (and missing) certificates is reasonable, but I think the order should be predictable.

b---c: Omitting the trust anchor, for the reasons mentioned, seems like a very reasonable thing. But I don't really see reason or value in allowing for gaps elsewhere, extraneous unrelated certificates, or unpredictable ordering. I plan to take editing pass over this section and try to tighten it up as such.

b---c:
This addresses #1521 and #1551

b---c:
also would fix #1596

draft-ietf-httpbis-client-cert-field.md
MikeBishop: I took the text here almost verbatim from TLS 1.3.  It recommends the proper ordering, but does not require it.  It certainly could be that they have constraints we don't -- legacy code being one, obviously -- but I'd want to have a concrete reason to diverge from what TLS uses in transferring the same thing.

draft-ietf-httpbis-client-cert-field.md
```diff
@@ -231,29 +244,30 @@ employ the `Client-Cert` or `Client-Cert-Chain` header field in requests.
...
+connection. Use of the header fields outside that intended use
 case, however, may undermine the protections afforded by TLS client certificate
-authentication. Therefore steps MUST be taken to prevent unintended use, both in
+authentication. Therefore, steps MUST be taken to prevent unintended use, both in
 sending the header field and in relying on its value.
 
-Producing and consuming the `Client-Cert` header field SHOULD be a configurable
-option, respectively, in a TTRP and backend server (or individual application in
+Producing and consuming the `Client-Cert` and `Client-Cert-Chain` header
+fields SHOULD be a configurable
```
MikeBishop, line 260:
```suggestion
fields SHOULD be configurable
```

draft-ietf-httpbis-client-cert-field.md
```diff
@@ -179,15 +194,13 @@ NOT be used in HTTP responses.  It MAY have a list of values or occur multiple
 times in a request.  For header compression purposes, it might be advantageous
 to split lists into multiple instances.
 
-The first certificate in the list SHOULD directly certify the end-entity
+The first certificate in the list directly certifies the end-entity
 certificate provided in the `Client-Cert` header; each following certificate
-SHOULD directly certify the one immediately preceding it.  Because certificate
+directly certifies the one immediately preceding it.  Because certificate
```
MikeBishop, line 199: Replied in the now-obsolete thread, but the SHOULDs come from RFC 8446's version of these requirements.  If we're making this mandatory (implicit MUST), this needs to specify what happens when it's violated.

SHOULD means exactly that -- it's better to do it this way, and things might go wrong if you don't.

MikeBishop (commented):
@b---c, thanks for building on this.

draft-ietf-httpbis-client-cert-field.md
bc-pi: I am blissfully ignorant of the constraints and objectives that went into that bit of TLS 1.3 but, while the same thing is being transferred, the context of use is quite different. In TLS it's part of the elaborate authentication ceremony during the establishment of a connection.  With the Cient-Cert-Chain header, it's conveying the information post-authentication to another entity on every request on the established connection. My reasoning is more: why wouldn't we have the TTRP convey that information in a defined and predictable way?

draft-ietf-httpbis-client-cert-field.md
bc-pi: My hope/goal was to place constraints on the sender here that lead to a predictable output. Do it this way. Full stop. 

I'm not sure what happens when it's violated with must or should. I see what you are saying but the implicit MUSTs are meat to mean - do it this way, otherwise it's wrong.

draft-ietf-httpbis-client-cert-field.md
jricher: I agree with @b---c here, if this spec can be tighter in its generation requirements, I think it should be. Reaching out to the TLS community is not a bad idea though.

martinthomson: This is the result of the way in which certification paths are constructed in practice, with co-signing and other things causing the ordering to be non-obvious.

Clearly, the order in which certificates appear can be linear, at least in the sense that the certificates can form a tree rooted on the EE certificate.  The problem is that it often needs to be a tree and not a chain because of the way in which different trust anchors accumulate in different endpoints.

This is far less likely to be a problem with client certificates as servers tend to carry the compatibility burdens, but I can't imagine that client certificates are completely immune to those sorts of compatibility problems either.

More fundamentally though: if the list of certificates that come out of TLS is not validated by the intermediary that inserts the header field, why do you think we can specify validation rules?  That is, if the certification path is constructed and validated by the back end system, the header field might be inserted with no knowledge of its contents. The only rule that makes any sense here is that the list of certificates match what the client included in its TLS handshake (or post-handshake authentication if you want to go there).

b---c: My expectation of what Cient-Cert-Chain header is supposed to carry, which I think the text (largely from Mike) in the PR is in line with and is consistent with the thrust of the document in general, it that it's the chain/path that was used by the intermediary to validate the client's EE certificate. The intermediary that inserts the header field does the validation during the TLS handshake and the cert path that was constructed during the handshake is the chain that is conveyed to the back end.

martinthomson: I don't think that you can assume that the intermediary will be performing path construction and validation, nor can you assume that the back-end server trusts it to do that.  While the backend might need to trust that the intermediary validates the in-TLS signature relative to the EE certificate, this design allows the backend to perform all the path construction independently.

If the back-end has different trust anchors then the intermediary, it will want to have all the information presented in the TLS handshake.

b---c: Honestly, I'm at a bit of a loss as to how to respond (and have spent way more time than I'd like trying to do so). That is exactly the assumption I want to make and what I view as an objective of the document to facilitate - where the proxy does the validation work and only exposes the information to back end. I view separating out the validation responsibilities like that as being rife with potential for subtle unexpected problems.  But maybe I'm in the weeds on this. I was rather skeptical of the purported need to expose more than the client's EE cert in the first place. Yet a number of folks expressed a need/want for it. And here we are. 

I also tried to look at docs of a few proxies and TLS implementations to get an idea of existing practice and/or what's even generally possible. And, to me anyway, everything is pretty ambiguous about what is exposed - what the client provided or what the server constructed. I suppose that if one reads the text of this PR without the bias I came with, it's consistent with that. 

All of this rambling is to say that l relent and I'll put that sentence back as well as the SHOULDs in the paragraph prior.

b---c: b251935ece5105c3b98aa078610e0fdeed9581bd puts that sentence back in the text as well as the SHOULDs into the paragraph prior.

Issue #1372, opened 2021-01-10
Third-party cookies


mnot:
In [the current draft's privacy considerations](https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#name-third-party-cookies):

> Given this risk to user privacy, some user agents restrict how third-party cookies behave, and those restrictions vary widly. For instance, user agents might block third-party cookies entirely by refusing to send Cookie headers or process Set-Cookie headers during third-party requests. They might take a less draconian approach by partitioning cookies based on the first-party context, sending one set of cookies to a given third party in one first-party context, and another to the same third party in another.
>
> This document grants user agents wide latitude to experiment with third-party cookie policies that balance the privacy and compatibility needs of their users. However, this document does not endorse any particular third-party cookie policy.

Is this still true?

chlily1:
Which part of it are you asking about, specifically?

The part about "grants user agents wide latitude" is certainly true. It allows UAs to [ignore Set-Cookie headers in their entirety](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-07#section-5.3) and [omit the Cookie header entirely](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-07#section-5.5). I suppose it does not technically allow for selectively withholding certain cookies from the Cookie header. (Maybe that should be allowed?)

The part about not endorsing any third-party cookie policy is probably also fine. It gives examples of a third-party cookie policy that drops Set-Cookie and Cookie headers entirely on third-party requests, but I don't think this constitutes an endorsement.

chlily1:
@mnot Does that answer your question? Please reopen if not, or if you think there are any changes needed.

mnot:
I was wondering if, given how much has changed in the intervening years, we could go a little further and say something about how third-party cookies cannot be relied upon to work by sites because they are abused, and current techniques for mitigating that abuse are not standardised.

Happy to write up a proposal if that would help.

sbingler:
I don't feel strongly here but if you think some updated wording would help I'm happy to review a proposal.

mnot:
See PR.

annevk:
Given that the PR got merged, should this be closed?

There's agreement emerging to block cross-site cookies, but perhaps that's best covered in a new issue.

jwrosewell:
[...]

PR #1878, opened 2022-01-05
Noodle on third-party cookies

mnot:
For #1372.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1820,7 +1820,7 @@ IDNA2003 {{RFC3490}}.
 
 # Privacy Considerations
 
-The primary privacy risk associated with cookies is their ability to correlate user activity. This
+Cookies' primary privacy risk is their ability to correlate user activity. This
 can happen on a single site, but is most problematic when activity is tracked across different,
 seemingly unconnected Web sites to build a user profile.
 
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1859,7 +1859,7 @@ these mitigations are necessary to improve Web privacy, they are not sufficient 
 ## Third-Party Cookies {#third-party-cookies}
 
 A "third-party" or cross-site cookie is a cookie that is associated with embedded content (such as
-scripts, images, stylesheets, frames) that is served from a different server than the one that
+scripts, images, stylesheets, frames) that is obtained from a different server than the one that
 hosts the primary resource (usually, the Web page that the user is viewing). Third-party cookies
 are often used to correlate users' activity on different sites.
 
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1822,7 +1822,14 @@ IDNA2003 {{RFC3490}}.
 
 The primary privacy risk associated with Cookies is their ability to correlate user activity. This can happen on a single site, but is most problematic when activity is tracked across different, seemingly unconnected Web sites to build a profile of the user.
 
-Over time, this capability (warned against explicitly in {{?RFC2109}} and all of its successors) has become widely used to correlate user activity across sites, known also as tracking. Reasons are varied and include protection against fraud and other forms of undesirable traffic, assembling information on users, targeting advertisements at specific users or at users with specified attributes, measuring how often ads are shown to users, and to recognize when an ad resulted in a change in user behavior.
+Over time, this capability (warned against explicitly in {{?RFC2109}} and all of its successors) has become widely used for varied reasons including:
+
+* authenticating users across sites,
+* assembling information on users,
+* protecting against fraud and other forms of undesirable traffic,
+* targeting advertisements at specific users or at users with specified attributes,
+* measuring how often ads are shown to users, and
+* recognizing when an ad resulted in a change in user behavior.
 
 While in some instances these purposes may not be privacy-invasive, the potential abuses of tracking imply a need for significant limitations on cookies. However, any such limitations need to be imposed with care, because cookies are widely used for a variety of purposes that are critical to the functioning of a Web site (and the Web).
 
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1860,7 +1860,7 @@ It is RECOMMENDED that user agents implement third-party cookie policies that ba
 privacy and compatibility needs of their users. However, this document does not yet endorse any
 particular third-party cookie policy.
 
-As a result, resources SHOULD NOT rely upon third-party cookies functioning reliably.
+As a result, resources cannot rely upon third-party cookies functioning reliably.
 
 
 ## Cookie Policy {#cookie-policy}
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1820,7 +1820,7 @@ IDNA2003 {{RFC3490}}.
 
 # Privacy Considerations
 
-The primary privacy risk associated with Cookies is their ability to correlate user activity. This can happen on a single site, but is most problematic when activity is tracked across different, seemingly unconnected Web sites to build a profile of the user.
+The primary privacy risk associated with cookies is their ability to correlate user activity. This can happen on a single site, but is most problematic when activity is tracked across different, seemingly unconnected Web sites to build a profile of the user.
 
 Over time, this capability (warned against explicitly in {{?RFC2109}} and all of its successors) has become widely used for varied reasons including:
 
@@ -1831,11 +1831,13 @@ Over time, this capability (warned against explicitly in {{?RFC2109}} and all of
 * measuring how often ads are shown to users, and
 * recognizing when an ad resulted in a change in user behavior.
 
-While in some instances these purposes may not be privacy-invasive, the potential abuses of tracking imply a need for significant limitations on cookies. However, any such limitations need to be imposed with care, because cookies are widely used for a variety of purposes that are critical to the functioning of a Web site (and the Web).
+While not every use of cookies is necessarily problematic for privacy, their potential for abuse has become a widespread concern in industry and society. In response to these concerns, user agents have actively constrained cookie functionality in various ways (as allowed and encouraged by previous specifications), while avoiding disruption to features they judge desirable for the health of the Web.
 
-This document describes limited mitigations against the privacy risks associated with cookies below. These mitigations are the most widely implemented and compatible at the time of writing.  It is expected that implementations will impose stricter and more well-defined limitations on cookies.  Future versions of this document might codify those limitations based upon experience with those limitations. If functions that currently rely on cookies can be supported by separate, targeted mechanisms, stricter limitations might be more feasible.
+It is too early to declare consensus on how the privacy impact of cookies can be successfully mitigated in detail; ongoing developments in user agents are best characterised as experiments that can provide input into that eventual consensus.
 
-Cookies are not the only mechanism that can be used to track users across sites, so while these mitigations are necessary to improve Web privacy, they are not sufficient on their own.
+Instead, this document describes limited, general mitigations against the privacy risks associated with cookies that enjoy wide deployment at the time of writing.  It is expected that implementations will continue to experiment and impose stricter, more well-defined limitations on cookies over time.  Future versions of this document might codify those mechanisms based upon deployment experience. If functions that currently rely on cookies can be supported by separate, targeted mechanisms, they might be documented in separate specifications and stricter limitations on cookies might become feasible.
+
+Note that cookies are not the only mechanism that can be used to track users across sites, so while these mitigations are necessary to improve Web privacy, they are not sufficient on their own.
 
 
 ## Third-Party Cookies {#third-party-cookies}
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1858,13 +1858,10 @@ these mitigations are necessary to improve Web privacy, they are not sufficient 
 
 ## Third-Party Cookies {#third-party-cookies}
 
-In a Web browsing context, a "third-party" or cross-site cookie is a cookie that is associated with
-content that is served from a different server than the one that hosts the primary resource (that
-is, the HTML). HTML documents can embed content (such as scripts, images, stylesheets, frames) that
-is served from other domains. These third-party resources can use cookies to track the user even if
-they never visit that server directly. For example, if a user visits a site that contains content
-from a third party and then later visits another site that contains content from the same third
-party, the third party can track the user between the two sites.
+A "third-party" or cross-site cookie is a cookie that is associated with embedded content (such as
+scripts, images, stylesheets, frames) that is served from a different server than the one that
+hosts the primary resource (usually, the Web page that the user is viewing). Third-party cookies
+are often used to correlate users' activity on different sites.
 
 Because of their inherent privacy issues, most user agents now limit third-party cookies in a
 variety of ways. Some completely block third-party cookies by refusing to process third-party
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1833,13 +1833,13 @@ Cookies are not the only mechanism that can be used to track users across sites,
 
 ## Third-Party Cookies {#third-party-cookies}
 
-A cookie that is sent to a site other than the one it originated from is known as a "third-party"
-cookie. When rendering a document, a user agent can request resources from sites other than that
-where that document was obtained from (such as advertising networks). Those third-party resources
-can use cookies to track the user even if they never visit that server directly. For example, if a
-user visits a site that contains content from a third party and then later visits another site that
-contains content from the same third party, the third party can track the user between the two
-sites.
+In a Web browsing context, a "third-party" or cross-site cookie is a cookie that is associated with
+content that is served from a different server than the one that hosts the primary resource (that
+is, the HTML). HTML documents can embed content (such as scripts, images, stylesheets, frames) that
+is served from other domains. These third-party resources can use cookies to track the user even if
+they never visit that server directly. For example, if a user visits a site that contains content
+from a third party and then later visits another site that contains content from the same third
+party, the third party can track the user between the two sites.
 
 Because of their inherent privacy issues, most user agents now limit third-party cookies in a
 variety of ways. Some completely block third-party cookies by refusing to process third-party
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1822,7 +1822,7 @@ IDNA2003 {{RFC3490}}.
 
 The primary privacy risk associated with cookies is their ability to correlate user activity. This
 can happen on a single site, but is most problematic when activity is tracked across different,
-seemingly unconnected Web sites to build a profile of the user.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1820,44 +1820,41 @@ IDNA2003 {{RFC3490}}.
...
 
-Cookies are often criticized for letting servers track users. For example, a
-number of "web analytics" companies use cookies to recognize when a user returns
-to a web site or visits another web site. Although cookies are not the only
-mechanism servers can use to track users across HTTP requests, cookies
-facilitate tracking because they are persistent across user agent sessions and
-can be shared between hosts.
+The primary privacy risk associated with Cookies is their ability to correlate user activity. This can happen on a single site, but is most problematic when activity is tracked across different, seemingly unconnected Web sites to build a profile of the user.
+
+Over time, this capability (warned against explicitly in {{?RFC2109}} and all of its successors) has become widely used to authenticate users across sites, to combat fraud and other forms of undesirable traffic, to target advertisements at specific users or at users with specified attributes, to measure how often ads are shown to users, and to recognise when an ad resulted in a change in user behavior.
```
martinthomson, line 1825:
```suggestion
Over time, this capability (warned against explicitly in {{?RFC2109}} and all of its successors) has become widely used to correlate user activity across sites, known also as tracking. Reasons are varied and include protection against fraud and other forms of undesirable traffic, assembling information on users, targeting advertisements at specific users or at users with specified attributes, measuring how often ads are shown to users, and to recognize when an ad resulted in a change in user behavior.
```

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1820,44 +1820,41 @@ IDNA2003 {{RFC3490}}.
...
+
+While in some instances these purposes may not be privacy-invasive, the potential abuses of tracking imply a need for significant limitations on cookies. However, any such limitations need to be imposed with care, because cookies are widely used for a variety of purposes that are critical to the functioning of a Web site (and the Web).
```
martinthomson, line 1827: I think that you are making the wrong distinction.  The purposes are always tracking, but you are implying a value judgment they are objectionable from a privacy standpoint (though this is of course subjective).

```suggestion
Though some of these purposes might not be objectionable from a privacy perspective, the potential for this tracking capability of cookies to be abused is being carefully reviewed. Some client implementations are deploying constraints on cookies with an aim to limit the potential for abuse.  These limitations need to be carefully balanced against the disruption that might result, especially where Web sites depend on preexisting features of cookies for critical functions.
```

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1820,44 +1820,41 @@ IDNA2003 {{RFC3490}}.
...
+
+As a result, this document recommends limited mitigations against the privacy risks associated with cookies below. It is expected that future updates will impose stricter and more well-defined limitations on cookies, based upon experiments now being undertaken in user agents. Note that some functions currently relying on cookies might eventually be supported by separate, targeted mechanisms.
```
martinthomson, line 1829:
```suggestion
This document describes limited mitigations against the privacy risks associated with cookies below. These mitigations are the most widely implemented and compatible at the time of writing.  It is expected that implementations will impose stricter and more well-defined limitations on cookies.  Future versions of this document might codify those limitations based upon experience with those limitations. If functions that currently rely on cookies can be supported by separate, targeted mechanisms, stricter limitations might be more feasible.
```

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1820,44 +1820,41 @@ IDNA2003 {{RFC3490}}.
...
-policy.
-
-Third-party cookie blocking policies are often ineffective at achieving their
-privacy goals if servers attempt to work around their restrictions to track
-users. In particular, two collaborating servers can often track users without
-using cookies at all by injecting identifying information into dynamic URLs.
-
-## Cookie policy {#cookie-policy}
+A cookie that is sent to a site other than the one it originated from is known as a "third-party"
+cookie. When rendering a document, a user agent can request resources from sites other than that
```
martinthomson, line 1837: This first sentence is not right, though the remainder of the definition is accurate.  Cookies are always sent back to the entity that set them, but a third-party cookie is a cookie set by a server that is other than the primary entity with which a user interacts.

Maybe: 

> In a Web browsing context, a "third-party" or cross-site cookie is a cookie that is associated with content that is served from a different server than the one that hosts the primary resource (that is, the HTML).  HTML documents can embed content (such as scripts, images, stylesheets, frames) that is served from other domains.  These "third-party" resources [...]

By the way, [Tess recommends "cross-site"](https://tess.oconnor.cx/2020/10/parties).

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1820,44 +1820,41 @@ IDNA2003 {{RFC3490}}.
...
+where that document was obtained from (such as advertising networks). Those third-party resources
+can use cookies to track the user even if they never visit that server directly. For example, if a
```
martinthomson, line 1839: The key here is that third-party resources will get the same cookies no matter how they are embedded.  That's the key piece, and this doesn't quite get to that.  If the third-party server sees the same cookie in a variety of contexts, they can link those contexts to the same person (and thereby track them).

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1820,44 +1820,41 @@ IDNA2003 {{RFC3490}}.
...
+variety of ways. Some completely block third-party cookies by refusing to process third-party
+Set-Cookie header fields and refusing to send third-party Cookie header fields. Some partition
+cookies based upon the first-party context, so that different cookies are sent depending on the
+site being browsed. Some block cookies based upon user agent cookie policy and/or user controls.
+
+It is RECOMMENDED that user agents implement third-party cookie policies that balance the
+privacy and compatibility needs of their users. However, this document does not yet endorse any
+particular third-party cookie policy.
+
+As a result, resources SHOULD NOT rely upon third-party cookies functioning reliably.
```
martinthomson, line 1854: Isn't this purely factual?  That is, s/SHOULD NOT/cannot/ ?

A normative statement implies that this is the way we want it to be.  I don't think that would be right.

mnot: People assign somewhat magical qualities to 2119 language; I'm invoking that here.

draft-ietf-httpbis-rfc6265bis.md
mnot: Thanks, I knew terminology was going to be an issue...

draft-ietf-httpbis-rfc6265bis.md
mnot: committing, but it needs some tweaks and you're losing federated auth...

draft-ietf-httpbis-rfc6265bis.md
martinthomson: I'm saying that I don't think that invoking the mystical 2119 invocation is right.  You are saying that they are unreliable, which is a fact.  But by invoking 2119 you are saying that people shouldn't use third-party cookies, which I don't agree with.  Though perhaps the fine folks at Apple would disagree, I would argue that partitioned cookies are perfectly fine (partitioned storage more generally).  Useful even.

That's an ongoing debate that you would have this draft take a position on.

draft-ietf-httpbis-rfc6265bis.md
mnot: Went in a different direction - see revision.

draft-ietf-httpbis-rfc6265bis.md
mnot: Fair enough.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1820,44 +1820,50 @@ IDNA2003 {{RFC3490}}.
...
-using cookies at all by injecting identifying information into dynamic URLs.
-
-## Cookie policy {#cookie-policy}
+In a Web browsing context, a "third-party" or cross-site cookie is a cookie that is associated with
+content that is served from a different server than the one that hosts the primary resource (that
+is, the HTML). HTML documents can embed content (such as scripts, images, stylesheets, frames) that
+is served from other domains. These third-party resources can use cookies to track the user even if
+they never visit that server directly. For example, if a user visits a site that contains content
+from a third party and then later visits another site that contains content from the same third
+party, the third party can track the user between the two sites.
```
martinthomson, line 1851:
```suggestion
party, that third party might be able to use cookies to track the user between the two sites.
```

draft-ietf-httpbis-rfc6265bis.md
mnot: I see how that's central to Mozilla's approach, but I don't think this level of detail / specificity is necessary here.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1820,44 +1820,50 @@ IDNA2003 {{RFC3490}}.
...
+
+Because of their inherent privacy issues, most user agents now limit third-party cookies in a
+variety of ways. Some completely block third-party cookies by refusing to process third-party
+Set-Cookie header fields and refusing to send third-party Cookie header fields. Some partition
+cookies based upon the first-party context, so that different cookies are sent depending on the
+site being browsed. Some block cookies based upon user agent cookie policy and/or user controls.
+
+It is RECOMMENDED that user agents implement third-party cookie policies that balance the
+privacy and compatibility needs of their users. However, this document does not yet endorse any
+particular third-party cookie policy.
```
martinthomson, line 1861: This is a bit of a non-statement really.  I'm all for magical thinking, but wouldn't this be better removed?  You already explain that things are unreliable.

If you want to recommend something, how about:

```suggestion
As the situation is fluid, this document does not endorse any particular policy for third-party cookies.  However, user agents SHOULD adopt a policy for third-party cookies that is as restrictive as compatibility constraints permit, thereby providing the best possible privacy.  Any such policy might become stricter as compatibility constraints change over time.

Consequently, resources cannot rely on a specific treatment for cookies in third-party contexts.
```

draft-ietf-httpbis-rfc6265bis.md
mnot: what value does 'might be able to' add?

draft-ietf-httpbis-rfc6265bis.md
mnot: I see it more as air cover than magical thinking. See edits.

draft-ietf-httpbis-rfc6265bis.md
martinthomson: ...because "can" implies an absolute possibility (I can use legs to climb stairs), in this case whether tracking is possible depends on what policies are in place with respect to cross site cookies.

mnot: This para is already super-wordy. Will try to refactor.

martinthomson (approved):

chlily1 (approved):

jwrosewell:
I'm concerned about the general direction of this PR and the document in general. A reliable method of sharing information between data controllers is needed to enable smaller organisations to band together to offer comparable services to larger organisations that are able to operate their services under a single domain or origin or have sufficient brand presence to nudge users to login across multiple services.

This PR emphasis unspecified concerns in the industry and appears to endorse a direction set by internet gatekeepers without evidence of those concerns and without providing any specifics concerning the equivalent guidance for user agents to support lawful use cases.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1820,44 +1820,62 @@ IDNA2003 {{RFC3490}}.
...
+* recognizing when an ad resulted in a change in user behavior.
+
+While not every use of cookies is necessarily problematic for privacy, their potential for abuse
+has become a widespread concern in industry and society. In response to these concerns, user agents
+have actively constrained cookie functionality in various ways (as allowed and encouraged by
+previous specifications), while avoiding disruption to features they judge desirable for the health
+of the Web.
+
+It is too early to declare consensus on how the privacy impact of cookies can be successfully
+mitigated in detail; ongoing developments in user agents are best characterised as experiments that
```
mnot, line 1844: Slight clarification based on private feedback;

```suggestion
It is too early to declare consensus on which specific mechanism(s) should be used to mitigate the privacy impact of cookies; ongoing developments in user agents are best characterised as experiments that
```

mnot:
@jwrosewell concern isn't actionable. What specific statements are incorrect?

jwrosewell:
The technical standards environment has been conditioned over a long period to tolerate and advance discrimination against suppliers and smaller operators. The text needs to be balanced to include the position of competition and privacy regulators if it is to properly inform the reader concerning the state of data sharing including third party cookies. As it stands the text perpetuates selective prosecution by internet gatekeepers.

mnot:
James, this isn't a competition regulator or a courtroom -- it's recording the consensus of the technical community, which has had a long history of balancing privacy concerns against technical capabilities. The proposed text is largely a statement of fact and history, not creating policy. The one exception I see - recommending that UAs adopt restricted third-party cookie policies -- is something that the various iterations of the Cookie spec have done since the very beginning.

Again, if you believe there's an inaccuracy, please point it out.

sbingler (approved):
This looks good, thanks for your effort!

jwrosewell:
*Completeness*
 
> The proposed text is largely a statement of fact and history, not creating policy.

In which case it should reflect all the facts and all the history rather than being selective.

*Experiment*

The latest PR contains the following text.

> …ongoing developments in user agents are best characterised as *experiments* that can provide input into that eventual consensus.

Checking the meaning of the word experiment with Websters provides the following definition.

-	test or trial
-	a tentative procedure or policy
-	an operation or procedure carried out under controlled conditions in order to discover an unknown effect or law, to test or establish a hypothesis, or to illustrate a known law

All web browser vendors have adopted concrete and widely communicated policies on the subject and many have widely deployed implementations. There is nothing experimental.

*Established practice*

The IETF considers factors beyond the text of past specifications. RFC 8890 is one such example. The sharing of state between different origins or domains via HTTP and other methods has enabled a great deal of innovation and benefits to end users including free and easy access to services. End users have many service providers to choose from. We have moved on from the days of closed services such as Compuserve, Prodigy and AOL. The guidance in future specifications must show how benefits to end users are preserved and enhanced by mitigating the potential harms.

Lukasa (approved):
This looks reasonable to me: it seems like a measured reflection of the current consensus and doesn't overstretch to try to declare a specific mitigation at this time. Good stuff.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1820,44 +1820,61 @@ IDNA2003 {{RFC3490}}.
...
+* measuring how often ads are shown to users, and
+* recognizing when an ad resulted in a change in user behavior.
+
+While not every use of cookies is necessarily problematic for privacy, their potential for abuse
+has become a widespread concern in industry and society. In response to these concerns, user agents
+have actively constrained cookie functionality in various ways (as allowed and encouraged by
+previous specifications), while avoiding disruption to features they judge desirable for the health
+of the Web.
+
+It is too early to declare consensus on which specific mechanism(s) should be used to mitigate the privacy impact of cookies; ongoing developments in user agents are best characterised as experiments that
```
johnwilander, line 1843: I think some browsers are well beyond what would be characterized as experiments.

johnwilander (approved):
Thanks for this PR. Switching from my co-editor hat to my Apple WebKit hat, I just wanted to express support for this.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1820,44 +1820,61 @@ IDNA2003 {{RFC3490}}.
...
+
+* authenticating users across sites,
+* assembling information on users,
+* protecting against fraud and other forms of undesirable traffic,
+* targeting advertisements at specific users or at users with specified attributes,
+* measuring how often ads are shown to users, and
+* recognizing when an ad resulted in a change in user behavior.
+
+While not every use of cookies is necessarily problematic for privacy, their potential for abuse
+has become a widespread concern in industry and society. In response to these concerns, user agents
```
mnot, line 1838:
```suggestion
has become a widespread concern in the Internet community and broader society. In response to these concerns, user agents
```

I think this is more appropriate -- the IETF isn't just industry

jwrosewell:
The text should further modified to incorporate the positions of law makers.

Extract from [An EU Strategy on Standardisation](https://ec.europa.eu/docsroom/documents/48598/attachments/2/translations/en/renditions/native).

> "A particular critical situation relates to internet standardisation to promote a free, open, accessible, inclusive and secure global internet. In recent years, international standardisation on internet protocols has become increasingly politicised, at the risk of limiting the evolution of the global open internet and hampering the digitisation process across the world."

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1820,44 +1820,61 @@ IDNA2003 {{RFC3490}}.
 
 # Privacy Considerations
 
-Cookies are often criticized for letting servers track users. For example, a
-number of "web analytics" companies use cookies to recognize when a user returns
-to a web site or visits another web site. Although cookies are not the only
-mechanism servers can use to track users across HTTP requests, cookies
-facilitate tracking because they are persistent across user agent sessions and
-can be shared between hosts.
+The primary privacy risk associated with cookies is their ability to correlate user activity. This
```
mnot, line 1823:
```suggestion
Cookies' primary privacy risk is their ability to correlate user activity. This
```

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1820,44 +1820,61 @@ IDNA2003 {{RFC3490}}.
...
+can happen on a single site, but is most problematic when activity is tracked across different,
+seemingly unconnected Web sites to build a profile of the user.
```
mnot, line 1825:
```suggestion
seemingly unconnected Web sites to build a user profile.
```

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1820,44 +1820,61 @@ IDNA2003 {{RFC3490}}.
...
+* measuring how often ads are shown to users, and
+* recognizing when an ad resulted in a change in user behavior.
+
+While not every use of cookies is necessarily problematic for privacy, their potential for abuse
+has become a widespread concern in the Internet community and broader society. In response to these concerns, user agents
+have actively constrained cookie functionality in various ways (as allowed and encouraged by
+previous specifications), while avoiding disruption to features they judge desirable for the health
+of the Web.
+
+It is too early to declare consensus on which specific mechanism(s) should be used to mitigate the privacy impact of cookies; ongoing developments in user agents are best characterised as experiments that
```
mnot, line 1843:
```suggestion
It is too early to declare consensus on which specific mechanism(s) should be used to mitigate cookies' privacy impact; user agents' ongoing changes in how they are handled are best characterised as experiments that
```

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1820,44 +1820,61 @@ IDNA2003 {{RFC3490}}.
...
-However, this document does not endorse any particular third-party cookie
-policy.
-
-Third-party cookie blocking policies are often ineffective at achieving their
-privacy goals if servers attempt to work around their restrictions to track
-users. In particular, two collaborating servers can often track users without
-using cookies at all by injecting identifying information into dynamic URLs.
-
-## Cookie policy {#cookie-policy}
+A "third-party" or cross-site cookie is a cookie that is associated with embedded content (such as
```
mnot, line 1860:
```suggestion
A "third-party" or cross-site cookie is one that is associated with embedded content (such as
```

mikewest (approved):

PR #1883, opened 2022-01-10
Add a section on differences with protocols

martinthomson:
These aren't THAT surprising, but it is worth being explicit.

### DIFF
--- a/draft-ietf-httpbis-binary-message.md
+++ b/draft-ietf-httpbis-binary-message.md
@@ -558,7 +558,7 @@ versions.  In particular, this format does not allow for:
 * framing of responses that depends on the corresponding request (such as HEAD)
   or the value of the status code (such as 204 or 304)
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-binary-message.md
```diff
@@ -61,10 +62,10 @@ messages that can be conveyed outside of an HTTP protocol. This enables the
 transformation of entire messages, including the application of authenticated
 encryption.
 
-This format is informed by the framing structure of HTTP/2 ({{H2}})
-and HTTP/3 ({{?H3=I-D.ietf-quic-http}}). In comparison, this format simpler by
-virtue of not including either header compression ({{?HPACK=RFC7541}},
-{{?QPACK=I-D.ietf-quic-qpack}}) or a generic framing layer.
+This format is informed by the framing structure of HTTP/2 ({{H2}}) and HTTP/3
+({{H3}}). In comparison, this format simpler by virtue of not including either
```
LPardue, line 66: Old error but there's a missing word before simpler

PR #1890, opened 2022-01-17
Editorial suggestions for targeted-cc from Benjamin Kaduk's IESG review

kaduk:


mnot (changes requested):
LGTM except for one thing.

PR #1893, opened 2022-01-20
Digest SF proposal 1: three headers

LPardue:
This is a rough proposal to enable support for Structured Fields for Digest headers. It is intended to illustrate the high level changes to structure. Minor errors or stale examples would be fixed up should this proposal advance.

In this proposal:

1. we keep the Digest header with it's old syntax. We provide the prose to clarify that the old term Instance is analagous to Representation today. 
2. we define a new Representation-Digest header that functions the same as Digest but is an SF Dictionary. Ths is RECOMMENDED over Digest.
3. we define a new Content-Digest header that is an SF Dictionary
4. Update the algorithm table to accommodate the hodge-podge of algorithms and their tokens or key

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -122,8 +122,9 @@ This document obsoletes [RFC3230].
 
 This document is structured as follows:
 
-- {{digest}} describes concepts related to representation digests. It also
-  defines the Digest request and response header and trailer field,
+- {{representation-digest}} describes concepts related to representation
+  digests,
+- {{digest}} defines the Digest request and response header and trailer field,
 - {{content-digest}} defines the Content-Digest request and response header and trailer field,
 - {{want-fields}} defines the Want-Digest and Want-Content-Digest request and response header and
   trailer field,
@@ -148,12 +149,9 @@ To support use-cases where a simple checksum of the content bytes is required,
 this document introduces the `Content-Digest` request and response header and trailer field; see
 {{content-digest}}.
 
-`Digest` and `Content-Digest` contain checksums that are calculated using one of
-the digest-algorithms listed in the HTTP Digest Algorithm Values Registry (see
-{{algorithms}}), which are encoded in the format associated with each encoding.
-Algorithm agility is supported. The `Want-Digest` and `Want-Content-Digest`
-fields allows endpoints to express interest in `Digest` and `Content-Digest`
-respectively, and preference of algorithms in either.
+`Digest` and `Content-Digest` support algorithm agility. The `Want-Digest` and
+`Want-Content-Digest` fields allows endpoints to express interest in `Digest` and
+`Content-Digest` respectively, and preference of algorithms in either.
 
 Digest field calculations are tied to the `Content-Encoding`
 and `Content-Type` header fields. Therefore, a given resource may have multiple
@@ -219,28 +217,48 @@ The term "checksum" describes the output of the application of an algorithm
 to a sequence of bytes,
 whereas digest is only used in relation to the value of the fields.
 
+# Representation Digest {#representation-digest}
+
+The representation digest is an integrity mechanism for HTTP resources
+which uses a checksum  that is calculated independently of the content
+(see {{Section 6.4 of SEMANTICS}}).
+It uses the representation data (see {{Section 8.1 of SEMANTICS}}),
+that can be fully or partially contained in the content, or not contained at all.
+
+This takes into account the effect of the HTTP semantics on the messages;
+for example, the content can be affected by Range Requests or methods such as HEAD,
+while the way the content is transferred "on the wire" is dependent on other
+transformations (e.g. transfer codings for HTTP/1.1 - see {{Section 6.1 of
+HTTP11}}). To help illustrate how such things affect `Digest`,
+several examples are provided in {{resource-representation}}.
+
+A representation digest consists of
+the value of a checksum computed on the entire selected `representation data`
+(see {{Section 8.1 of SEMANTICS}}) of a resource identified according to {{Section 6.4.2 of SEMANTICS}}
+together with an indication of the algorithm used:
+
+~~~ abnf
+   representation-data-digest = digest-algorithm "="
+                                <encoded checksum output>
+~~~
+
+When a message has no representation data
+it is still possible to assert that no representation data was sent
+computing the representation digest on an empty string
+(see {{usage-in-signatures}}).
+
+The checksum is computed using one of the digest-algorithms listed in
+the HTTP Digest Algorithm Values Registry (see {{algorithms}})
+and then encoded in the associated format.
+
 # The Digest Field {#digest}
 
-The `Digest` field contains a comma-separated list of one or more representation
-digest values. It can be used in both requests and responses.
-
-A representation digest is an integrity mechanism for HTTP resources that uses a
-checksum  calculated independently of the content (see {{Section 6.4 of
-SEMANTICS}}).
-A representation digest consists of the value of a checksum computed on
-the entire selected `representation data` (see {{Section 8.1 of SEMANTICS}})
-of a resource identified according to {{Section 6.4.2 of SEMANTICS}},
-together with an indication of the algorithm used.
-Note that representation data can be fully or partially contained in the content,
-or not contained at all.
-When a message has no representation data it is possible to
-assert that no representation data was sent by using the representation digest
-on an empty string (see {{usage-in-signatures}}).
+The `Digest` field contains a comma-separated list of one or more representation digest values as
+defined in {{representation-digest}}. It can be used in both requests and
+responses.
 
 ~~~ abnf
    Digest = 1#representation-data-digest
-   representation-data-digest = digest-algorithm "="
-                                <encoded checksum output>
 ~~~
 
 For example:
@@ -249,13 +267,6 @@ For example:
 Digest: sha-512=WZDPaVn/7XgHaAy8pmojAkGWoRx2UFChF41A2svX+TaPm
                 AbwAgBWnrIiYllu7BNNyealdVLvRwE\nmTHWXvJwew==
 ~~~
-
-Basing `Digest` on representation data takes into account the effect of HTTP
-semantics on messages; for example, the content can be affected by Range
-Requests or methods such as HEAD, while the way the content is transferred "on
-the wire" is dependent on other transformations (e.g. transfer codings for
-HTTP/1.1 - see {{Section 6.1 of HTTP11}}). To help illustrate how such things
-affect `Digest`, several examples are provided in {{resource-representation}}.
 
 A `Digest` field MAY contain multiple representation-data-digest values.
 For example, a server may provide representation-data-digest values using different algorithms,
@@ -487,7 +498,7 @@ does not describe the target resource,
 the representation digest MUST be computed on the
 representation-data.
 This is the only possible choice because representation digest requires complete
-representation metadata (see {{digest}}).
+representation metadata (see {{representation-digest}}).
 
 In responses,
 
@@ -1169,7 +1180,7 @@ Digest: sha-256=KPqhVXAT25LLitV1w0O167unHmVQusu+fpxm65zAsvk=
 An origin server sends `Digest` as trailer field, so it can calculate digest-value
 while streaming content and thus mitigate resource consumption.
 The `Digest` field-value is the same as in {{example-full-representation}} because `Digest` is designed to
-be independent from the use of one or more transfer codings (see {{digest}}).
+be independent from the use of one or more transfer codings (see {{representation-digest}}).
 
 ~~~ http-message
 GET /items/123 HTTP/1.1
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -84,10 +84,13 @@ informative:
 --- abstract
 
 This document defines HTTP fields that support integrity checksums.
-The Digest field can be used for the integrity of HTTP representations.
-The Content-Digest field can be used for the integrity of HTTP message content.
-Want-Digest and Want-Content-Digest can be used to indicate a sender's desire to
-receive integrity fields respectively.
+The Digest field can be used for the integrity of HTTP representations using a
+legacy syntax. The new Representation-Digest field can be used for the integrity
+of HTTP representations using Structured Fields. The new Content-Digest field
+can be used for the integrity of HTTP message content using Structured Fields.
+A sender's desire to receive integrity fields can be indicated using the legacy
+Want-Digest field or new Want-Representation-Digest and Want-Content-Digest
+fields.
 
 This document obsoletes RFC 3230.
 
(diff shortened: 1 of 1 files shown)


jricher:
As stated on the list, I support this approach.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -122,44 +125,62 @@ This document obsoletes [RFC3230].
...
-Digest field calculations are tied to the `Content-Encoding`
+The HTTP fields defined in this document can be used for HTTP integrity. Senders
+choose a hashing algorithm and calculate a checksum from an input related to the
+HTTP message, the algorithm identifier and checksum are transmitted in an HTTP
+field. Receivers can validate the checksum for integrity purposes. Hashing
+algorithms are registered in the HTTP Digest Algorithm Values Registry (see
+{{algorithms}}).
+
+Selecting the data on which checksums are calculated depends on the use case of
+HTTP messages. This document provides different headers for HTTP representation
```
ioggstream, line 150:
```suggestion
HTTP messages. This document provides different fields for HTTP representation
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -122,44 +125,62 @@ This document obsoletes [RFC3230].
...
+header and trailer field is defined to support checksums of content
+({{Section 3.2 of SEMANTICS}}).
+
+`Digest`, `Representation-Digest` and `Content-Digest` support hashing algorithm
+agility. The `Want-Digest`, `Want-Representation-Digest` and
+`Want-Content-Digest` fields are used respectively to express interest in their
+counterpart along with a preference list of algorithms.
+
+`Digest`, `Representation-Digest` and `Content-Digest` are collectively termed
+integrity fields.
```
ioggstream, line 177: +1 :clap:

draft-ietf-httpbis-digest-headers.md
```diff
@@ -196,9 +217,16 @@ behavior. A digest-algorithm can still be parameterized by defining its own way
 to encode parameters into the representation-data-digest, in such a way as to
 mitigate security risks related to its computation.
 
-The algorithm table has been updated to reflect the current state of the art,
-(see {{algorithms}}).
+While this document updates `Digest` and `Want-Digest`, their syntax remains
+unmodified. The custom syntax is now treated as legacy. It is RECOMMENDED that
```
ioggstream, line 221:
```suggestion
unmodified. The original syntax is now treated as legacy. It is RECOMMENDED that
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -219,28 +250,42 @@ The term "checksum" describes the output of the application of an algorithm
...
+several examples are provided in {{resource-representation}}.
+
+When a message has no representation data it is still possible to assert that no
+representation data was sent computing the representation digest on an empty
+string (see {{usage-in-signatures}}).
+
+The `Digest` an `Representation-Digest` HTTP fields carry a set of
+representation digests; the fields differ only in their wire syntax. Each entry
+in the set conveys the hashing algorithm used to compute the associated computed
+checksum.
```
ioggstream, line 275:
```suggestion
in the set conveys the hashing algorithm used to compute the associated
checksum.
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -219,28 +250,42 @@ The term "checksum" describes the output of the application of an algorithm
 to a sequence of bytes,
 whereas digest is only used in relation to the value of the fields.
 
-# The Digest Field {#digest}
+# Representation Digest {#representation-digest}
```
ioggstream, line 253: This section is back again :DDD

draft-ietf-httpbis-digest-headers.md
```diff
@@ -281,70 +319,112 @@ that the recipient will ignore it.
...
+The `Representation-Digest` HTTP field can be used in requests and responses. It
+is a Structured Fields Dictionary (see {{Section 3.2 of
+STRUCTURED-FIELDS}}):
+
+~~~ abnf
+Representation-Digest   = sf-dictionary
+~~~
+
+`Representation-Digest` member-keys convey the hashing algorithm (see
+{{algorithms}}) used to compute the checksum. Member-values are the the output
```
ioggstream, line 333:
```suggestion
{{algorithms}}) used to compute the checksum. Member-values are the output
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -250,13 +295,6 @@ Digest: sha-512=WZDPaVn/7XgHaAy8pmojAkGWoRx2UFChF41A2svX+TaPm
...
 ~~~
 
-Basing `Digest` on representation data takes into account the effect of HTTP
-semantics on messages; for example, the content can be affected by Range
-Requests or methods such as HEAD, while the way the content is transferred "on
-the wire" is dependent on other transformations (e.g. transfer codings for
-HTTP/1.1 - see {{Section 6.1 of HTTP11}}). To help illustrate how such things
-affect `Digest`, several examples are provided in {{resource-representation}}.
-
 A `Digest` field MAY contain multiple representation-data-digest values.
```
ioggstream, line 298: This `MAY` is redundant, given the abnf rule...

draft-ietf-httpbis-digest-headers.md
```diff
@@ -281,70 +319,112 @@ that the recipient will ignore it.
...
 ~~~ http-message
-Content-Digest: sha-512=WZDPaVn/7XgHaAy8pmojAkGWoRx2UFChF41A2svX+TaPm
-                        AbwAgBWnrIiYllu7BNNyealdVLvRwE\nmTHWXvJwew==
+Content-Digest: sha-512=:WZDPaVn/7XgHaAy8pmojAkGWoRx2UFChF41A2svX+TaPm
+                        AbwAgBWnrIiYllu7BNNyealdVLvRwE\nmTHWXvJwew==:
 ~~~
 
-A `Content-Digest` field MAY contain multiple content-digest values,
-similarly to `Digest` (see {{digest}})
+Since the `Content-Digest` field is a Dictionary, it can contain multiple
```
ioggstream, line 389: @LPardue since the original document, this paragraph is repeated multiple times. Considered that here https://github.com/httpwg/http-extensions/pull/1893/files#r793058641 the `MAY` is redundant, do you think we can just state this sentence once and for all somewhere?

LPardue: I think we can deduplicate text. With headers it was bad enough with 3 it gets worse.

But I'm holding off investing time on deep editorial work until there is consensus to adopt or reject either proposal

draft-ietf-httpbis-digest-headers.md
```diff
@@ -358,152 +438,213 @@ Examples:
 ~~~ http-message
 Want-Digest: sha-256
 Want-Digest: sha-512;q=0.3, sha-256;q=1, unixsum;q=0
```
ioggstream, line 440: As an implementation consideration, both sf-list and sf-dictionary can probably parse want-digest.

```In [2]: d=http_sfv.Dictionary()

In [3]: d.parse(b'sha-512;q=0.3, sha-256;q=1, unixsum;q=0')

In [4]: d.to_json()
Out[4]: 
{'sha-512': [True, {'q': Decimal('0.3')}],
 'sha-256': [True, {'q': 1}],
 'unixsum': [True, {'q': 0}]}
```
or

```
In [5]: d=http_sfv.List()

In [6]: d.parse(b'sha-512;q=0.3, sha-256;q=1, unixsum;q=0')

In [7]: d.to_json()
Out[7]: 
[[{'__type': 'token', 'value': 'sha-512'}, {'q': Decimal('0.3')}],
 [{'__type': 'token', 'value': 'sha-256'}, {'q': 1}],
 [{'__type': 'token', 'value': 'unixsum'}, {'q': 0}]]
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -358,152 +438,213 @@ Examples:
...
+   Want-Representation-Digest = sf-list
+   Want-Content-Digest = sf-list
 ~~~
 
-All digest-algorithm token values are case-insensitive
-but lower case is preferred;
-digest-algorithm token values MUST be compared in a case-insensitive fashion.
+List items convey hashing algorithm preferences (see {{algorithms}}). Items can
+be parameterized with the `q` parameter, which can be used to express an
+ascending relative weighted preference between 0 and 1. `q` MUST be sf-decimal.
```
ioggstream, line 463: Referencing the qvalue rule mentioned in Notational conventions.

```suggestion
List items convey hashing algorithm preferences (see {{algorithms}}).
Items can be parameterized with the `q` parameter, which can be used to express an
ascending relative weighted preference between 0 and 1.
`q` MUST be sf-decimal compatible to the "qvalue" rule.
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -358,152 +438,213 @@ Examples:
...
+NOT be used in a potentially adversarial setting; for example, when signing integrity fields' values for
 authenticity.
 
-The registry is initialized with the tokens listed below.
+The registry is initialized with the algorithms listed below.
 
   {: vspace="0"}
-  sha-512
-  : * Digest Algorithm: sha-512
+  SHA-512
```
ioggstream, line 522: why uppercase?

draft-ietf-httpbis-digest-headers.md
LPardue: digest-algorithm is a `token`, while list could be a list of `sf-token` and sf-dictionary uses `key`

```
  token          = 1*tchar

     tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
                    / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
                    / DIGIT / ALPHA
                    ; any VCHAR, except delimiters
```
```
sf-token = ( ALPHA / "*" ) *( tchar / ":" / "/" )
```

```
key           = ( lcalpha / "*" )
                   *( lcalpha / DIGIT / "_" / "-" / "." / "*" )
   lcalpha       = %x61-7A ; a-z
```

so there's the possibility that someone has an algorithm identifier that won't actually parse correctly to sf-token or key

draft-ietf-httpbis-digest-headers.md
```diff
@@ -358,152 +438,213 @@ Examples:
...
+  : * Algorithm: SHA-512
     * Description: The SHA-512 algorithm [RFC6234].
-      The output of this algorithm is encoded using base64 [RFC4648].
+    * Digest-algorithm token: sha-512
+    * Digest-algorithm output format: Base64 [RFC4648]
+    * Digest Dictionary key: sha-512
+    * Digest List item: sha-512
```
ioggstream, line 528: Not sure we should repeat the same token multiple time, maybe the wg will help in fixing that

draft-ietf-httpbis-digest-headers.md
```diff
@@ -358,152 +438,213 @@ Examples:
...
 representation-data.
 This is the only possible choice because representation digest requires complete
-representation metadata (see {{digest}}).
+representation metadata (see {{representation-digest}}).
 
 In responses,
 
 - if the representation describes the status of the request,
-  `Digest` MUST be computed on the enclosed representation
+  `Digest` or `Representation-Digest` MUST be computed on the enclosed representation
```
ioggstream, line 636: wdyt of referencing representation-data digest which applies to both fields?

draft-ietf-httpbis-digest-headers.md
LPardue: My idea here is to modify the table to be rows of Algorithms, not rows of tokens. The encoding rules for legacy digest and SF has the potential for incompatibility. It's feasible that somebody has defined a `token` that cannot be encoded as a `key` - so we have multiple columns in the table that spells this out explicitly.

I'm open to other options here but we do need to consider a robust solution given our constraints. This is robust if a bit ugly.

To also answer you question below, T

draft-ietf-httpbis-digest-headers.md
LPardue: agree

draft-ietf-httpbis-digest-headers.md
LPardue: I know we merged it as part of WGLC feedback but I think it helps to restore it when Represenation concept applies to `Digest` and `Representation-Digest` equally.

draft-ietf-httpbis-digest-headers.md
LPardue: yep we should do that

draft-ietf-httpbis-digest-headers.md
LPardue: I think this is an area for some WG feedback. 

My approach here is to define a new parameter `q`. It's based on qvalue but it's compatibility is arguable, because the definition of compatibility is not well defined. If there are additional constraints for the parameter, we should state these explicitly. But we should also think about whether the expectations for acting on this `q` are as clear as qvalue.

If we assume that future SF-based fieIds might want a qvalue-like parameter. Perhaps we should consider making this document the canonical definition of `q` parameter. (it could be a standalone document too but there's more bootstrapping to that).

draft-ietf-httpbis-digest-headers.md
LPardue: I like it!

draft-ietf-httpbis-digest-headers.md
ioggstream: Just tracking down stuff, here

draft-ietf-httpbis-digest-headers.md
```diff
@@ -281,70 +319,112 @@ that the recipient will ignore it.
...
-                        AbwAgBWnrIiYllu7BNNyealdVLvRwE\nmTHWXvJwew==
+Content-Digest: sha-256=:4REjxQ4yrqUVicfSKYNO/cF9zNj5ANbzgDZt3/h3Qxo=:,
+                sha-512=:WZDPaVn/7XgHaAy8pmojAkGWoRx2UFChF41A2svX+TaPm
+                        AbwAgBWnrIiYllu7BNNyealdVLvRwE\nmTHWXvJwew==:
 ~~~
 
-A recipient MAY ignore any or all of the content-digests in a Content-Digest
-field. This allows the recipient to choose which digest-algorithm(s) to use for
-validation instead of verifying every received content-digest.
+A recipient MAY ignore any or all of members of the Content-Digest field. This allows
```
ioggstream, line 401:
```suggestion
A recipient MAY ignore any or all of members of `Content-Digest`. This allows
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -281,70 +319,112 @@ that the recipient will ignore it.
...
-the `Digest` field.
+## The Want-Digest field {#want-digest}
 
-`Want-Content-Digest` indicates the sender's desire to receive a content digest
-on messages associated with the request URI and representation metadata, using
-the `Content-Digest` field.
+The `Want-Digest` field indicates the sender's desire to receive a
+representation digest on messages associated with the request URI and
+representation metadata, using the `Digest` field. It uses a legacy list
+encoding.
```
ioggstream, line 424:
```suggestion
syntax.
```

draft-ietf-httpbis-digest-headers.md
ioggstream: > this document the canonical definition of q parameter

@reschke WDYT about SF and qvalue?

draft-ietf-httpbis-digest-headers.md
ioggstream: > It's feasible that somebody has defined a token that cannot be encoded as a key

if such tokens do not exist yet, I will just restrict the space of possible values for all fields.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -863,7 +980,7 @@ Host: foo.example
 ~~~ http-message
 HTTP/1.1 200 OK
 Content-Type: application/json
-Digest: sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
+Representation-Digest: sha-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
```
ioggstream, line 983: Here and beyond, `sha-256=:X48....PE=:`

draft-ietf-httpbis-digest-headers.md
LPardue: Since there is no requirement to register algorithms before usage, I'm not sure how enforceable changing things is. 

`token` has a definition and trying to change that charset now in relation to digest seems like a backwards compatibility breaking change.

ioggstream:
we opted for #1919

PR #1899, opened 2022-01-26
Add Ed25519, update algorithm text

### LINKED ISSUE #1509: Add Ed25519 support: algorithm definition and examples
It's been *a very long time* since I added the RSA signature test vectors to the appendix of the HTTP Signatures spec. We should replace them with more modern examples. My suggestion would be to use ed25519 examples, or secp256r1 (P-256) examples. The former would need a [new signature algorithm specified](https://github.com/httpwg/http-extensions/issues/1510), which would be a good thing.

jricher:
Adds text to clarify algorithms in several places:

- encode ECDSA signature as raw bytes, mark ECDSA and RSAPSS as non-deterministic
- note non-determinism of ECDSA and RSAPSS in examples, add security consideration of nondeterministic algorithms.
- Add ed25519, closes #1509

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -1000,7 +1000,9 @@ The mask generation function is `MGF1` as specified in {{RFC8017}} with a hash f
 The salt length (`sLen`) is 64 bytes.
 The hash function (`Hash`) SHA-512 {{RFC6234}} is applied to the signature input string to create the digest content to which the verification function is applied.
 The verifier extracts the HTTP message signature to be verified (`S`) as described in {{verify}}.
-The results of the verification function are compared to the http message signature to determine if the signature presented is valid.
+The results of the verification function indicate if the signature presented is valid.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-message-signatures.md
```diff
@@ -1013,16 +1018,32 @@ Use of this algorithm can be indicated at runtime using the `hmac-sha256` value
 To sign using this algorithm, the signer applies the `ECDSA` algorithm {{FIPS186-4}} using curve P-256 with the signer's private signing key and
 the signature input string ({{create-sig-input}}).
 The hash SHA-256 {{RFC6234}} is applied to the signature input string to create
-the digest content to which the digital signature is applied.
-The resulting signed content byte array is the HTTP message signature output used in {{sign}}.
+the digest content to which the digital signature is applied, (`M`).
+The signature algorithm returns two integer values, `r` and `s`. These are both encoded in big-endian unsigned integers, zero-padded to 32-octets each. These encoded values are concatenated into a single 64-octet array consisting of the encoded value of `r` followed by the encoded value of `s`. The resulting concatenation of `(r, s)` is byte array of the HTTP message signature output used in {{sign}}.
```
yaronf, line 1022: You might want to add an informative ref to JWA, https://www.rfc-editor.org/rfc/rfc7518#section-3.4

PR #1900, opened 2022-01-26
Changed "specialty" to "derived" components

### LINKED ISSUE #1888: Security Consideration: Lazy header parsers
It's been reported that the HTTP header parsers on some platforms are particularly lazy in what they allow for header content, including allowing header field names to start with the `@` symbol. If an implementation of HTTP Signatures uses one of these platforms naively, that implementation could end up in a situation where an attacker could try to override one of the derived content values, like `@method` by sending a fake header in the request to override the actual value derivation.

Advice in both processing algorithm and security considerations will be to check the first character and only look for header field values if it's not the reserved `@` character.

### LINKED ISSUE #1889: Rename "Specialty" to "Derived"
All specialty content identifiers point to information that is derived from the HTTP message itself, or from the context of the message being processed. Suggest we change terminology from "specialty content" to "derived content" and associated terms.

jricher:
Closes #1889
﻿Closes #1888


draft-ietf-httpbis-message-signatures.md
```diff
@@ -300,11 +300,11 @@ The following example shows canonicalized values for different component identif
...
 ~~~
 
-## Specialty Components {#specialty-content}
+## Derived Components {#derived-components}
 
-Message components not found in an HTTP field can be included in the signature input by defining a component identifier and the canonicalization method for its component value.
+In addition to HTTP fields, there are a number of different components that can be derived from the control data, processing context, or other aspects of the HTTP message being signed. Such derived components can be included in the signature input by defining a component identifier and the derivation method for its component value.
 
-To differentiate specialty component identifiers from HTTP fields, specialty component identifiers MUST start with the "at" `@` character. This specification defines the following specialty component identifiers:
+To differentiate for identifiers derived components from identifiers for HTTP fields, derived component identifiers MUST start with the "at" `@` character. This specification defines the following derived component identifiers:
```
LPardue, line 307: I find this sentence quite hard to grok. Are you trying to say something like

Derived component identifiers MUST start with the "at '@' character. This avoids derived component identifiers being mistake for HTTP fields.

LPardue (approved):
this looks good modulo the one comment I made

draft-ietf-httpbis-message-signatures.md
jricher: Yes, something like that. I'll reword it, thanks.

PR #1902, opened 2022-01-27
[6265bis] Specify that no decoding of Set-Cookie line should occur

### LINKED ISSUE #1517: RFC 6265bis should specify that no decoding of Set-Cookie line should occur
UAs should not decode any URL-encoded (or otherwise encoded) characters while storing a cookie, e.g. a cookie such as `__%53ecure-foo=bar` should not be interpreted as a `__Secure-` cookie.

I landed some WPTs for this in https://github.com/web-platform-tests/wpt/pull/28066. Chrome, Edge, Firefox, and Safari [seem to](https://wpt.fyi/results/cookies/attributes/secure-non-secure.html?label=experimental&label=master&aligned) do the right thing.

I think this could just be a clarifying note in the parsing algorithm section.

johannhof:
Fixes #1517

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1131,6 +1131,11 @@ explicitly. Horizontal tab (%x09) is excluded from the CTL characters that
 lead to set-cookie-string rejection, as it is considered whitespace, which is
 handled separately.
 
+NOTE: The set-cookie-string may contain octet sequences that appear
+percent-encoded as per {{Section 2.1 of RFC3986}}. However, a user agent
+MUST NOT decode these sequences and instead parse the individual octets
+as specified in this algorithm.
+
 A user agent MUST use an algorithm equivalent to the following algorithm to
 parse a set-cookie-string:
 


draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1131,6 +1131,11 @@ explicitly. Horizontal tab (%x09) is excluded from the CTL characters that
 lead to set-cookie-string rejection, as it is considered whitespace, which is
 handled separately.
 
+NOTE: The set-cookie-string may contain octet sequences that appear
+percent-encoded as per {{Section 2.1 of RFC3986}}. However, a user agent
```
chlily1, line 1135: I think the section number goes outside of the {{}}.

chlily1, line 1135: This should be added to the informative references at the top.

DiamondG3m:
```

```~~

draft-ietf-httpbis-rfc6265bis.md
johannhof: With `{{Section 2.1 of RFC3986}}` it creates an additional deeplink into that section:
image

Which I think is nice, so I'll keep it like that if you don't mind :)

johannhof: I think it's already there, but maybe I'm missing something.

chlily1: TIL, that's great. :)

chlily1: Nope, I skimmed too quickly. My bad!

chlily1:
Thanks! This Lgtm but I'll let Steven take a look too

sbingler:
This looks to me as well, thanks!

PR #1903, opened 2022-01-28
Update Examples, Publish -08

jricher:
- update canonicalization rules, remove weird headers from examples
- changed dictionaries and messages.
- updated crypto primative functions
- regenerate examples

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -249,7 +249,7 @@ Following are non-normative examples of canonicalized values for header fields, 
 
 ~~~ http-message
 Host: www.example.com
-Date: Tue, 07 Jun 2014 20:51:35 GMT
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-message-signatures.md
```diff
@@ -1900,7 +2046,7 @@ Without a signature, the internal service would need to trust that the incoming
 NOTE: '\' line wrapping per RFC 8792
 
 "@path": /foo
-"@query": Param=value&pet=Dog
+"@query": param=Value&Pet=dog
```
bblfish, line 2049: should there not have been a question mark there?

jricher: @bblfish Yes, that is fixed in #1908, please review the new examples there if you can, and we can merge them.

PR #1904, opened 2022-01-28
Editorial nits in h3-websockets

kaduk:
pseudo-header is hyphenated per [HTTP3], and expand out a
"headers" straggler.


RyanTheOptimist (approved):
Thanks, this is great!

PR #1908, opened 2022-01-31
Update signature examples for query

### LINKED ISSUE #1901: Signature 07: Signing String date wrong in B.2.3
The Example from [Appendix B.2.3](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-07.html#appendix-B.2.3) uses the following request defined in  B.2

```scala
POST /foo?param=value&pet=dog HTTP/1.1
Host: example.com
Date: Tue, 20 Apr 2021 02:07:55 GMT
Digest: SHA-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
Content-Length: 18

{"hello": "world"}
```
The signing string is meant to be

```text
"date": Tue, 20 Apr 2021 02:07:56 GMT
"@method": POST
"@path": /foo
"@query": ?param=value&pet=dog
"@authority": example.com
"digest": SHA-256=X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=
"content-length": 18
"@signature-params": ("date" "@method" "@path" "@query" \
  "@authority" "content-type" "digest" "content-length")\
  ;created=1618884475;keyid="test-key-rsa-pss"
```
But that does not work because in the request we have
```http
Date: Tue, 20 Apr 2021 02:07:55 GMT
```
and the date in the signing string is one minute later
```text
"date": Tue, 20 Apr 2021 02:07:56 GMT
```

jricher:
Update query parameter generation in examples, closes #1901
﻿

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -837,7 +837,7 @@ If covered components reference a component identifier that cannot be resolved t
  * The component identifier identifies a field that is not present in the message or whose value is malformed.
  * The component identifier indicates that a structured field serialization is used (via the `sf` parameter), but the field in question is known to not be a structured field or the type of structured field is not known to the implementation.
  * The component identifier is a dictionary member identifier that references a field that is not present in the message, is not a Dictionary Structured Field, or whose value is malformed.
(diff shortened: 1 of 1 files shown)


yaronf:
I verified that B.2.3 is fixed, but did not implement the "ttrp" example.

PR #1958, opened 2022-02-09
Varints not minimal varints

### LINKED ISSUE #1943: Binary messages: is variable length integer encoding minimal
The binary messages draft uses QUIC variable-length integers. Those allow encoding numbers multiple ways (for example 42 can be encoded minimally as 0x2A but also non-minimally as 0x402A). The draft should clarify whether using non-minimal encoding is allowed or not.

martinthomson:
For the avoidance of doubt.

This makes encoding easier if you don't care about wasting a few bytes.

Closes #1943.


DavidSchinazi (approved):
Thanks!

PR #1959, opened 2022-02-09
Fix field length issues

### LINKED ISSUE #1944: Binary messages: why is Known-Length Field length at least 2?
The draft currently shows:
```
Known-Length Field Section {
  Length (i) = 2..,
  Field Line (..) ...,
}
```

This Length field isn't well defined. Is it the length of the Known-Length Field Section in bytes or in lines? If in bytes, does it include or exclude the length of the Length field itself? Either way, I don't understand why Length is at least 2. Either it's possible for there to not be any field lines, in which case maybe it should be zero. And if there needs to be at least one field line, then (we should say so, and) the minimum number should be 3, because a field line is two lengths and at least one byte of field name.

### LINKED ISSUE #1945: Binary message: are empty field values allowed?
The Known-Length Messages section states that `Both name and value are sequences of bytes that cannot be zero length.` whereas the Header and Trailer Field Lines section allows empty field values. Which is it?

martinthomson:
Closes #1945 by using a citation rather than a redundant and wrong statement.
Closes #1944 by removing the minimum length (zero is totally fine).

### DIFF
--- a/draft-ietf-httpbis-binary-message.md
+++ b/draft-ietf-httpbis-binary-message.md
@@ -125,7 +125,7 @@ Message with Known-Length {
 }
 
 Known-Length Field Section {
-  Length (i) = 2..,
+  Length (i),
   Field Line (..) ...,
 }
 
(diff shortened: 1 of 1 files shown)


DavidSchinazi (approved):
Thanks!

PR #1960, opened 2022-02-09
Why truncation

### LINKED ISSUE #1946: Binary message: what is truncation?
The draft mentions that messages can be truncated in some places, but not in others. I think it would help to explain why one would want to truncated a message.

martinthomson:
Basically, it saves the effort of writing stuff that you know to be empty.  It's
not a critical feature, but it turns out to be a simple way to deal with not
sending trailers or a body.  In the extreme, you can also drop headers to
produce the simplest and shortest HTTP message ever: 0x0140c8.

CLoses #1946.

### DIFF
--- a/draft-ietf-httpbis-binary-message.md
+++ b/draft-ietf-httpbis-binary-message.md
@@ -154,10 +154,7 @@ length-prefixed value. Both name and value are sequences of bytes that cannot
 be zero length.
 
 The format allows for the message to be truncated before any of the length
-prefixes that precede the field sections or content. Truncation can be used to
-reduce the size of messages that have no data in trailing field sections or
-content.  A message that is truncated at any other point is invalid; see
-{{invalid}}.
+prefixes that precede the field sections or content; see {{padding}}.
 
 The variable-length integer encoding means that there is a limit of 2^62-1
 bytes for each field section and the message content.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-binary-message.md
```diff
@@ -154,8 +154,9 @@ length-prefixed value. Both name and value are sequences of bytes that cannot
 be zero length.
 
 The format allows for the message to be truncated before any of the length
-prefixes that precede the field sections or content. This reduces the overall
-message size. A message that is truncated at any other point is invalid; see
+prefixes that precede the field sections or content. Truncation can be used to
+reduce the size of messages that have no content for subsequent field sections
```
DavidSchinazi, line 158: Reusing the word "content" twice to mean different things in one sentence isn't ideal.

DavidSchinazi (commented):
This new text helps me understand what you mean, but it would be nice to define this more formally. Something along the lines of: "if the last section of the message is empty, it MAY be omitted by the encoder in place of adding a length field equal to zero. Decoders MUST treat missing truncated fields as equivalent to having been sent with the length field sent to zero."

martinthomson:
Hey David, I've made some larger edits here to factor the truncation text better.  I've included your normative language.

DavidSchinazi (approved):
Thanks

PR #1961, opened 2022-02-09
Information status is response only

### LINKED ISSUE #1947: Binary messages: do known-length requests carry Known-Length Informational Response?
I suspect that requests do not carry the Known-Length Informational Response field (otherwise how do you know how many of them there are?) but that should be spelled out explicitly.

I'd suggest creating a different diagram for known-length request vs known-length responses to make this easier to reason about, and same for indeterminate length

martinthomson:
The diagram resulted in a few double-takes that might be helped by a little bit of a note pointing
this out.

Closes #1947.

### DIFF
--- a/draft-ietf-httpbis-binary-message.md
+++ b/draft-ietf-httpbis-binary-message.md
@@ -148,8 +148,8 @@ a trailer section with a length prefix.
 
 Response messages that contain informational status codes result in a different
 structure; see {{informational}}.  Note that while the Known-Length
-Informational Response field is shown in {{format-known-length}}, it is only
-permitted in response messages.
+Informational Response field is shown in {{format-known-length}}, it can only
+appear in response messages.
 
 Fields in the header and trailer sections consist of a length-prefixed name and
 length-prefixed value. Both name and value are sequences of bytes that cannot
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-binary-message.md
```diff
@@ -147,7 +147,9 @@ a header section with a length prefix, binary content with a length prefix, and
 a trailer section with a length prefix.
 
 Response messages that contain informational status codes result in a different
-structure; see {{informational}}.
+structure; see {{informational}}.  Note that while the Known-Length
+Informational Response field is shown in {{format-known-length}}, it is only
+permitted in response messages.
```
DavidSchinazi, line 152: "permitted" isn't the right word - this is about a field that just cannot be present, not about an optional field that is not allowed to be present?

martinthomson: s/it is only permitted in/it cannot appear in/ perhaps?

DavidSchinazi: Or "it only appears". I still think a separate diagram would make this easier to read personally though

PR #1964, opened 2022-02-11
Signature Document Cleanups

### LINKED ISSUE #1867: Signatures: Move keys to PKCS#8 PEM Format
1. Neither Java nor JS subtleCrypto come out of the box with PKCS1 PEM support.  (for java [see here](https://stackoverflow.com/posts/6560187/revisions)).
2. It is not immediately obvious to work out that the PEM formats of the private key in in PKCS1 format (the spec could point that out).
3. When building an implementation it helps to have the test suite clearly map to the data in the spec (e.g. [here](https://github.com/bblfish/bobcats/blob/ab9c7fa1759f3fc58abdbf10e8c5431451a045a0/core/shared/src/test/scala/bobcats/SigningHttpMessages.scala#L256)) so as to make error reporting clearer. That is even though using the following openssl command to convert the keys work, it makes it difficult to see if an error was introduced at some point.
```zsh
$ openssl pkcs8 -topk8 -inform PEM -in spec.private.pem -out private.pem -nocrypt
```
4. To support PKCS1 the developer has to depend on further less well tested libraries and add additional code (as I [did here](https://github.com/bblfish/bobcats/blob/main/core/jvm/src/test/scala/bobcats/util/BouncyJavaPEMUtils.scala)) leading possibly to new bugs. Furthermore, I will need to go through the same process for JavaScript now, doubling the work.

Since PKCS8 is as expressive as PKCS1 ([according to this article](https://www.starkandwayne.com/blog/credhub-keysmust-be-pem-encoded-pkcs1-keys/) and see [this nice StackExchange answer](https://stackoverflow.com/a/48960291/396079)), it would make life easier for devs if the keys were all in PKCS8 format. Unless there really is something that PKCS1 offers that PKCS8 does not?

I will try doing this using the Web Crypto API next. But there we find that [importKey](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/importKey) supports the PKCS#8 format.

### LINKED ISSUE #1907: Signatures: "alg" for JWS signatures
This sentence is correct:

> There is no use of the explicit alg signature parameter when using JOSE signing algorithms, as they can be signaled using JSON Web Keys or other mechanisms.

But it's true for most reasonable uses of `alg`, and people will still be using this parameter. So why not add:

> If the `alg` parameter is used anyway, its value MUST be the algorithm's name (Sec. 3.1 of RFC 7518), such as "ES256".

### LINKED ISSUE #1921: Signatures: Host header
There are quite a few occurrences of the `Host` header in the draft.

From an implementation point of view, Go "swallows" the `Host` header and it is not part of the standard Request structure. There's a good reason for the Go library to do it, and in fact the draft recommends to use @authority instead. Can we remove `Host` from examples, e.g. Sec. 2.1, so we don't train people to do the wrong thing?

### LINKED ISSUE #1951: Signatures: @query clarification
Also: If there is no query string should the corresponding signing text be
`@query: ?` or `@query: `
ie. should the query value be empty or just consist of one `?` 

Should we return  `@query: ` with this request
```
POST /path HTTP/1.1
Host: www.example.com
```
and return  `@query: ?` on this request
```
GET /path? HTTP/1.1
Host: www.example.com
```

This is what I opted for in my implementation. See
* [http4s @query selector](https://github.com/bblfish/httpSig/blob/e1dd912b00ed960891e99e1f02b6b4d5bd5a9e02/http4s/shared/src/main/scala/run/cosy/http4s/headers/HeaderSelector.scala#L217) shows that http4s makes the distinction between blank queries and empty queries
* [akka @query selector](https://github.com/bblfish/httpSig/blob/e1dd912b00ed960891e99e1f02b6b4d5bd5a9e02/akka/src/main/scala/run/cosy/akka/http/headers/AkkaMessageSelectors.scala#L188) is simpler
* [two extra tests to distinguish those cases](https://github.com/bblfish/httpSig/blob/e1dd912b00ed960891e99e1f02b6b4d5bd5a9e02/ietfSig/shared/src/test/scala/run/cosy/http/auth/HttpMessageSigningSuite.scala#L349)

I think this is a corner case that is worth explicating very clearly, perhaps even with a test.

jricher:
Includes several document cleanups for signatures, to publish as -09

Closes #1921
closes #1907
closes #1867
closes #1951

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -1624,16 +1624,28 @@ It is recommended that implementers first detect and validate the `Signature-Inp
 
 # Examples
 
+The following non-normative examples are provided as a means of testing implementations of HTTP Message Signatures. The signed messages given can be used to create the signature input strings with the stated parameters, creating signatures using the stated algorithms and keys.
+
+The private keys given can be used to generate signatures, though since several of the demonstrated algorithms are nondeterministic, the results of a signature are expected to be different from the exact bytes of the examples. The public keys given can be used to validate all signed examples.
+
 ## Example Keys {#example-keys}
 
 This section provides cryptographic keys that are referenced in example signatures throughout this document.  These keys MUST NOT be used for any purpose other than testing.
 
 The key identifiers for each key are used throughout the examples in this specification. It is assumed for these examples that the signer and verifier can unambiguously dereference all key identifiers used here, and that the keys and algorithms used are appropriate for the context in which the signature is presented.
 
+The components for each private key in PEM format can be displayed by executing the following OpenSSL command:
+
+~~~
+openssl pkey -text
+~~~
+
+This command was tested with all the example keys on OpenSSL version 1.1.1m. Note that some systems cannot produce or use these keys directly, and may require additional processing.
+
 ### Example Key RSA test {#example-key-rsa-test}
 
 The following key is a 2048-bit RSA public and private key pair, referred to in this document
-as `test-key-rsa`:
+as `test-key-rsa`. This key is encoded in PEM Format, with no encryption.
 
 ~~~
 -----BEGIN RSA PUBLIC KEY-----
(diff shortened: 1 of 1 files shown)


richsalz (approved):
Looks good!

Issue #1939, opened 2022-02-07
Parser for Domain attributes


annevk:
If a resource on `127.0.0.1` sets a cookie with a `Domain` attribute set to `127.1`, does that work? (Spoiler: it works if it was passed through the [host parser](https://url.spec.whatwg.org/#concept-host-parser), which seems like a prerequisite to determine whether it's an IP address or not as needed by Domain Matching.)

(I raised this in #1707 but it seems that's exclusively used for the non-ASCII bytes discussion.)

sbingler:
Good question, what should happen?

[Section 5.5](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-09#section-5.5) Step 9, sub-step 1 states
```
 9.   If the domain-attribute is non-empty:
        1.  If the canonicalized request-host does not domain-match the
            domain-attribute:
               1.  Ignore the cookie entirely and abort these steps.
```
It looks like "domain-match" is doing the heavy lifting in that alogithm.

[Section 5.1.3](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-09#section-5.1.3) Domain Matching
has the following algorithm
```
A string domain-matches a given domain string if at least one of the
   following conditions hold:

   *  The domain string and the string are identical.  (Note that both
      the domain string and the string will have been canonicalized to
      lower case at this point.)

   *  All of the following conditions hold:

      -  The domain string is a suffix of the string.

      -  The last character of the string that is not included in the
         domain string is a %x2E (".") character.

      -  The string is a host name (i.e., not an IP address).
```
Because this is an IP address, only the top condition is applicable meaning that since a cookie with a `Domain=127.1` wouldn't be identical to the string (the request-host) the implementor should ignore the cookie entirely.

That's straightforward enough, assuming we agree on my interpretation here.

> does that work?

Also a good and (unfortunately) different question.

Well for Chrome 98 it seems we do set the cookie, oops.

This feels like it's an implementation bug, as the spec's instructions seem reasonable, but do you have a different take Anne?

sbingler:
Are you suggesting that 
`-  The string is a host name (i.e., not an IP address).`
should be modified to say
`-  The domain string is a host name (i.e., not an IP address).`?

annevk:
Maybe that's good enough. Ideally the spec would just reuse the URL Standard data types, but that's probably not attainable.

sbingler:
[...]

PR #1969, opened 2022-02-16
[6265bis] Require ASCII for domain attributes (fixes #1707)

### LINKED ISSUE #1707: IDN names for domain attributes
We should specify what to do if the domain attribute in a cookie is:

1. punycode (A-labels, xn--etc)
2. unicode (U-labels)

I think that the right behaviour here is to only accept punycode, but we should ensure that this is what people do.

(This is a small piece of #1073 that might just be tractable.)

johannhof:
See #1707 for discussion and rationale.

I wasn't entirely sure where to put this new restriction, but it does seem like the UA should parse the domain attribute list and only reject the cookie if the final domain attribute contains invalid characters (as opposed to, say, reject if any domain attribute contains invalid characters). So I went with that.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -530,7 +530,7 @@ data in a cookie-value SHOULD encode that data, for example, using Base64
 {{RFC4648}}.
 
 The domain-value is a subdomain as defined by {{RFC1034}}, Section 3.5, and
-as enhanced by {{RFC1123}}, Section 2.1. The domain-value MUST be a string of
+as enhanced by {{RFC1123}}, Section 2.1. Thus, domain-value is a string of
 {{USASCII}} characters, such as one obtained by applying the "ToASCII" operation
 defined in {{Section 4 of RFC3490}}.
 
@@ -1473,7 +1473,7 @@ user agent MUST process the cookie as follows:
     1.  Let the domain-attribute be the empty string.
 
 8.  If the domain-attribute contains a character that is not in the range of {{USASCII}}
-    characters (%x00 - %x7F), abort these steps and ignore the cookie entirely.
+    characters, abort these steps and ignore the cookie entirely.
 
 9.  If the user agent is configured to reject "public suffixes" and the
     domain-attribute is a public suffix:
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1475,11 +1475,6 @@ user agent MUST process the cookie as follows:
 8.  If the domain-attribute contains a character that is not in the range of {{USASCII}}
     characters, abort these steps and ignore the cookie entirely.
 
-    NOTE: The user agent MAY also reject any cookies that contain characters not
-    in the LDH (letter, digit, hyphen) range of %x41-5A / %x61-7A / %x30-39 / %x2D,
-    which is the character range of the canonicalized request-host the
-    domain-attribute is compared against.
-
 9.  If the user agent is configured to reject "public suffixes" and the
     domain-attribute is a public suffix:
 
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -530,7 +530,9 @@ data in a cookie-value SHOULD encode that data, for example, using Base64
 {{RFC4648}}.
 
 The domain-value is a subdomain as defined by {{RFC1034}}, Section 3.5, and
-as enhanced by {{RFC1123}}, Section 2.1.
+as enhanced by {{RFC1123}}, Section 2.1. The domain-value MUST be a string of
+{{USASCII}} characters, such as one obtained by applying the "ToASCII" operation
+defined in {{Section 4 of RFC3490}}.
 
 Per the grammar above, the cookie-value MAY be wrapped in DQUOTE characters.
 Note that in this case, the initial and trailing DQUOTE characters are not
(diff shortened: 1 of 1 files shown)


johannhof:
@sbingler @martinthomson @annevk ^ happy to get feedback on this :)

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -530,7 +530,9 @@ data in a cookie-value SHOULD encode that data, for example, using Base64
 {{RFC4648}}.
 
 The domain-value is a subdomain as defined by {{RFC1034}}, Section 3.5, and
-as enhanced by {{RFC1123}}, Section 2.1.
+as enhanced by {{RFC1123}}, Section 2.1. The domain-value MUST be a string of
+{{USASCII}} characters, such as one obtained by applying the "ToASCII" operation
+defined in {{Section 4 of RFC3490}}.
```
annevk, line 535: I think this is redundant with how subdomain is defined in the referenced RFCs so I'd not include this. (If you really want to include it I'd make it a statement of fact. ("Thus, the domain-value is a string of {{USASCII}} characters, such as one ...")

annevk (commented):
Looks good!

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1470,7 +1472,10 @@ user agent MUST process the cookie as follows:
 
     1.  Let the domain-attribute be the empty string.
 
-8.  If the user agent is configured to reject "public suffixes" and the
+8.  If the domain-attribute contains a character that is not in the range of {{USASCII}}
+    characters (%x00 - %x7F), abort these steps and ignore the cookie entirely.
```
sbingler, line 1476: It feels odd to include characters we specifically avoid elsewhere as valid (such as %x00). I guess the earlier parts of the spec handle those cases where necessary so this should be ok, it's strikes me as add.

Just noting this for posterity. No change required.

sbingler (approved):
Looks good, aside from Anne's suggestion.

Thanks!

draft-ietf-httpbis-rfc6265bis.md
martinthomson: I think that we can use a narrower range for acceptance here that will ensure that rejected characters (%x00) don't get included.  A domain name can be just limited to LDH (letter, digit, hyphen) + '.'.

It seems we're very deliberately avoiding mention of IP addresses, so we can probably not specify tighter validation rules (along the lines of what RFC 1034 specifies), but somewhat tighter is probably wise.

martinthomson (commented):
I think that removing %x00 from the allowed set of characters is worthwhile, if only because it means that the logic and the explanatory text (with Anne's tweak) line up better.

draft-ietf-httpbis-rfc6265bis.md
johannhof: I agree it's a bit weird to explicitly list those. I'm not sure to what extent it's better to just remove the explicit mention of the character range vs. having a narrower definition. The UA will perform an equality check to a canonicalized URL host right after this step, so there's not a lot of worth to UAs in checking for LDH explicitly. Browsers have very optimized isASCII functions that would have to be sidestepped for this.

(Even this step is mostly added to make it clear to UAs that they should not perform canonicalization on the domain attribute, strictly speaking it wouldn't be necessary since the spec doesn't mandate canonical/A-label domain attributes)

If we do go for referencing LDH it might be useful to declare it in the terminology, I filed #1979 in any case.

draft-ietf-httpbis-rfc6265bis.md
johannhof: I think that's a good suggestion! I might be reading the referenced RFCs wrong but for me they operate more under an assumption of ASCII than strictly requiring it? In any case, I do think it's worthwhile calling this out in the server section so I'll go with what you suggested.

draft-ietf-httpbis-rfc6265bis.md
annevk: I like this simpler check for that reason too. We don't actually want to reject `..` or some such at this point.

(IP addresses we should tackle as part of #1939.)

johannhof:
Thanks all for taking a look! I made the minimal necessary changes first but would still want to wait for @martinthomson to respond to my comment on his LDH suggestion. I don't have strong feelings on that either way.

draft-ietf-httpbis-rfc6265bis.md
martinthomson: I'm OK with that, but this is a good point to point out that the domain canonicalization will ensure that invalid domain names will be rejected.  The oddness (and inconsistency) is pretty apparent and you've had two people tripped up by it already.

johannhof: Added a note to help clear the confusion, lmk what you think :)

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1470,7 +1472,15 @@ user agent MUST process the cookie as follows:
...
     1.  Let the domain-attribute be the empty string.
 
-8.  If the user agent is configured to reject "public suffixes" and the
+8.  If the domain-attribute contains a character that is not in the range of {{USASCII}}
+    characters, abort these steps and ignore the cookie entirely.
+
+    NOTE: The user agent MAY also reject any cookies that contain characters not
+    in the LDH (letter, digit, hyphen) range of %x41-5A / %x61-7A / %x30-39 / %x2D,
+    which is the character range of the canonicalized request-host the
+    domain-attribute is compared against.
```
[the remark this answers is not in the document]
johannhof: Alright, let's punt this on the IP address discussion then.

johannhof:
@chlily1 @johnwilander  @mikewest  I think we've reached some level of agreement on this PR, would any of you mind giving this a quick look and merging if you don't have any concerns?

Issue #1968, opened 2022-02-16
Migration guide


ioggstream:
## I expect

support for migrating from Digest to the new fields

PR #1971, opened 2022-02-16
Fix: #1968. Migrating from RFC3230.

ioggstream:
## This PR

-  supports current implementers of [RFC3230] in the migration to the new integrity fields
- should allow them to replace any reference to [RFC3230] with this document

Fix: #1968

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -471,7 +471,7 @@ enclosed representation refers to the resource identified by its value and
 `Representation-Digest` is computed accordingly. An example is given in {{post-not-request-uri}}.
 
 
-# Security Considerations
+# Security Considerations {#security}
 
 ## HTTP Messages Are Not Protected In Full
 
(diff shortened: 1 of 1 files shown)


LPardue:
This seems like too much RFC3230 detail to me. We don't need to duplicate ABNF for instance. IMO all we need to say is something like

> HTTP digests are computed by applying a hashing algorithm to input data. RFC 3230 defined the input data as an "instance", a term it also defined. The concept of instance has since been superseded by the HTTP semantic term "representation". It is understood that some implementations of RFC 3230 mistook "instance" to mean HTTP content. Using content for the HTTP Digest field is an error that leads to interoperability problems between peers that implement RFC 3230.

> For the uncertainty of doubt, RFC 3230 was only ever intended to use what HTTP now defines as  "selected representation data". The semantic concept of digest and representation are explained alongside the definition of the Representation-Digest field {{representation-digest}}. While the syntax of Digest and Representation-Digest fields are different, the consideration this document gives to Representation-Digest apply equally to Digest because they operate on the same input data. See {{state-changing-requests}}, {{security}} and {{usage-in-signatures}}.

> Implementations of RFC 3230 could never communicate the digest of HTTP message content in the Digest field. Content-Digest now provides that capability.

ioggstream:
> This seems like too much RFC3230 detail to me

The goal is providing enough information to current implementers to avoid referencing RFC3230 in their documents.
If providing those information in an appendix would do the job (I'm engaging with them), are you ok with it?

LPardue:
Duplicating parts of any spec is generally a bad smell.

If they want to use an obsolete spec, then they have to refer to the obsolete spec because that's where all the normative statements are. They can additionally profile or describe their use of RFC3230 by pointing to text in this doc that explains all the problems with the old one.

mnot:
+1 to @LPardue's points. Also, it's important to remember that this is a specification, not a user guide -- while some amount of prose to help folks out is good, we don't want to over-do it. Extrinsic materials (e.g., blog entries, books, articles, papers) can and will inform use of the feature, and are more able to both talk to the audience in an appropriate tone, and be refined over time.

ioggstream:
@mnot do you think this could be addressed separately in a short, informational RFC targeted at the current community of implementers? If this is the case, I'll follow that road.

ioggstream:
@LPardue Used your text, thanks!

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1255,6 +1255,29 @@ Content-Type: application/problem+json
...
+mistook "instance" to mean HTTP content. 
+Using content for the Digest field is an error 
+that leads to interoperability problems between peers that implement RFC 3230.
+
+For the uncertainty of doubt, RFC 3230 was only ever intended
+to use what HTTP now defines as "selected representation data".
+The semantic concept of digest and representation are explained
+alongside the definition of Representation-Digest {{representation-digest}}.
+
+While the syntax of Digest and Representation-Digest are different,
```
LPardue, line 1273:
```suggestion
While the syntax of Digest and Repr-Digest are different,
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1255,6 +1255,29 @@ Content-Type: application/problem+json
...
+the considerations and examples this document gives to Representation-Digest
```
LPardue, line 1274:
```suggestion
the considerations and examples this document gives to Repr-Digest
```

LPardue (commented):
nit: need to update header name

PR #1972, opened 2022-02-16
Fix: #1954. No normative text in security considerations.

### LINKED ISSUE #1954: No normative statements in Security Considerations
## I expect

- no normative statements in security considerations

ioggstream:
## This PR

- no normative text in security considerations.
Fix #1954

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -473,7 +473,7 @@ enclosed representation refers to the resource identified by its value and
 
 # Security Considerations
 
-## HTTP Messages Are Not Protected In Full
+## HTTP Messages Are Not Protected In Full {#sec-limitations}
 
 This document specifies a data integrity mechanism that protects HTTP
 `representation data` or content, but not HTTP header and trailer fields, from
@@ -512,11 +512,9 @@ digest validation failure at the recipient, preventing the application from
 accessing the representation. Such an attack consumes the resources of both
 endpoints. See also {{digest-and-content-location}}.
 
-Digest fields SHOULD always be used over a connection that provides integrity at
-the transport layer that protects HTTP fields.
-
-A `Representation-Digest` field using NOT RECOMMENDED hashing algorithms SHOULD NOT be used in
-signatures.
+As explained in {{sec-limitations}}, integrity fields rely
+on transport layer integrity protection mechanisms,
+while {{algorithms}} provides further guidance on algorithm's choice.
 
 Using signatures to protect the checksum of an empty representation
 allows receiving endpoints to detect if an eventual payload has been stripped or added.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -512,11 +512,9 @@ digest validation failure at the recipient, preventing the application from
...
 endpoints. See also {{digest-and-content-location}}.
 
-Digest fields SHOULD always be used over a connection that provides integrity at
-the transport layer that protects HTTP fields.
-
-A `Representation-Digest` field using NOT RECOMMENDED hashing algorithms SHOULD NOT be used in
-signatures.
+As explained in {{sec-limitations}}, integrity fields rely
+on transport layer integrity protection mechanisms,
+while {{algorithms}} provides further guidance on algorithm's choice.
```
LPardue, line 517: Thanks.

Seeing it like this, is there even a need to say anything here? The referenced sections apply to the usage of digest and don't appear to add any further insight or considerations when used with Signature. 

@jricher WDYT about just deleting the two sentences and not adding anything more?

jricher: I actually like the cross-reference here, I would keep it this way.

jricher (approved):

draft-ietf-httpbis-digest-headers.md
LPardue: Interesting. If someone is using Digests with signature, they should really apply all the recommondations and apply all the considerations of the document. There seems to me no need to call out any particular section as unique to Signatures, unless there is something to say about why. Is that the case here?

The rest of this section _does_ achieve the goal of signature-specific considerations. It's just these that stick out to me.

ioggstream: @LPardue  {{algorithms}} actually references a signature-related normative text, so I think it should stay. The TLS part in 2022 is probably redundant for a careful reader.

Can we agree on removing the transport layer text and leave the {{algorithms}} reference? cc: @jricher

jricher: I still think they can both stay, but at this stage I'll leave it up to your discretion if you think it is clearer without them.

LPardue: So reading the text in {{algorithms}} the relevant part seems to be about saying signatures is a potential adversarial setting. 

But we already have the Algorithm Agility security consideration section, which now contains nice text about considerations for choosing secure or insecure algorithms.

So instead I would say

"Signatures are likely to be deemed an adversarial setting when applying integrity digests; see {{algorithm agility}}."

And still drop the part about transport integrity.

ioggstream: deal!

LPardue (approved):

Issue #1941, opened 2022-02-07
Simplify the text explaining fields


LPardue:
```suggestion
This document defines the `Representation-Digest` field ({{representation-digest}}) that contains a value
```

_Originally posted by @martinthomson in https://github.com/httpwg/http-extensions/pull/1919#discussion_r799059536_

PR #1973, opened 2022-02-16
Fix: #1941. Editorial alignments

ioggstream:
## This PR

- same boilerplate text for C-Digest and R-Digest
- avoid ABNF in favor of Types
- representation-data -> representation data
- representatio-data-digest -> digests (clarified in Notational Convetions);
- integrity fields -> Integrity fields (@LPardue let me know if you meant it lowercase, it's ok to me in any way);
- digest algorithm -> hashing algorithm,

Note: it'd be cool if we could write something like

```
~~~ abnf
R-Digest = sf-dictionary
member-value = sf-binary
```

but it's ok anyway

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -150,7 +150,7 @@ data and HTTP content.
 
 This document defines the `Representation-Digest` request and response header
 and trailer field ({{representation-digest}}) that contains a digest value
-computed by applying a hashing algorithm to *selected representation data*
+computed by applying a hashing algorithm to "selected representation data"
 ({{Section 3.2 of SEMANTICS}}). Basing `Representation-Digest` on the selected
 representation makes it straightforward to apply it to use-cases where the
 transferred data requires some sort of manipulation to be considered a
@@ -188,7 +188,7 @@ This specification does not define means for authentication, authorization or pr
 [RFC3230] defined the `Digest` and `Want-Digest` HTTP fields for HTTP integrity.
 It also coined the term "instance" and "instance manipulation" in order to
 explain concepts that are now more universally defined, and implemented, as HTTP
-semantics such as *selected representation data* ({{Section 3.2 of SEMANTICS}}).
+semantics such as "selected representation data" ({{Section 3.2 of SEMANTICS}}).
 
 Experience has shown that implementations of [RFC3230] have interpreted the
 meaning of "instance" inconsistently, leading to interoperability issues. The
@@ -236,7 +236,7 @@ Integrity preference fields: collective term for `Want-Representation-Digest` an
 
 The `Representation-Digest` HTTP field can be used in requests and responses to
 communicate digests that are calculated using a hashing algorithm applied to
-the entire *selected representation data* (see {{Section 8.1 of SEMANTICS}}).
+the entire "selected representation data" (see {{Section 8.1 of SEMANTICS}}).
 
 Representations take into account the effect of the HTTP semantics on
 messages. For example, the content can be affected by Range Requests or methods
@@ -245,8 +245,8 @@ dependent on other transformations (e.g. transfer codings for HTTP/1.1 - see
 {{Section 6.1 of HTTP11}}). To help illustrate HTTP representation concepts,
 several examples are provided in {{resource-representation}}.
 
-When a message has no representation data it is still possible to assert that no
-representation data was sent by computing the digest on an empty
+When a message has no "representation data" it is still possible to assert that no
+"representation data" was sent by computing the digest on an empty
 string (see {{usage-in-signatures}}).
 
 `Representation-Digest` is a Structured Fields `Dictionary` (see {{Section 3.2 of
@@ -410,7 +410,6 @@ the `Content-Digest` field.
 `Want-Representation-Digest` and `Want-Content-Digest` are Structured Fields
 List (see {{Section 3.2 of STRUCTURED-FIELDS}}) where:
 
-* members cannot have parameters;
 * keys convey the hashing algorithm (see {{algorithms}});
 * values MUST be of type `Integer` ({{Section 3.3.1 of STRUCTURED-FIELDS}})
   in the range 0 to 10 inclusive.
@@ -486,7 +485,7 @@ The entries in {{iana-hash-algorithm-table}} are registered by this document.
 ## HTTP Messages Are Not Protected In Full {#sec-limitations}
 
 This document specifies a data integrity mechanism that protects HTTP
-*representation data* or content, but not HTTP header and trailer fields, from
+"representation data" or content, but not HTTP header and trailer fields, from
 certain kinds of corruption.
 
 Integrity fields are not intended to be a general protection against malicious tampering with
@@ -495,10 +494,10 @@ as transport-layer security or digital signatures.
 
 ## End-to-End Integrity
 
-Integrity fields can help detect  *representation data* or content modification due to implementation errors,
+Integrity fields can help detect  "representation data" or content modification due to implementation errors,
 undesired "transforming proxies" (see {{Section 7.7 of SEMANTICS}})
 or other actions as the data passes across multiple hops or system boundaries.
-Even a simple mechanism for end-to-end  *representation data* integrity is valuable
+Even a simple mechanism for end-to-end  "representation data" integrity is valuable
 because a user agent can validate that resource retrieval succeeded before handing off to a
 HTML parser, video player etc. for parsing.
 
@@ -1104,7 +1103,7 @@ Note that a `204 No Content` response without content but with the same
 
 ## Error responses
 
-In error responses, the representation data does not necessarily refer to the
+In error responses, the "representation data" does not necessarily refer to the
 target resource. Instead, it refers to the representation of the error.
 
 In the following example, a client sends the same request from {{fig-patch}} to
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -212,9 +212,10 @@ have names that more clearly articulate the intended usages.
 This document uses the Augmented BNF defined in [RFC5234] and updated by
 [RFC7405].
 
-The terms Dictionary, List, Byte Sequence, sf-dictionary,
-sf-integer and sf-binary are imported from
-{{!STRUCTURED-FIELDS=RFC8941}}.
+This document uses the Boolean, Byte Sequence, 
+Dictionary, Integer and List types from
+{{!STRUCTURED-FIELDS=RFC8941}} along with
+the sf-dictionary and sf-list ABNF rules.
 
 The definitions "representation", "selected representation", "representation
 data", "representation metadata", "user agent" and "content" in this document are to be
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -408,7 +408,7 @@ on messages associated with the request URI and representation metadata, using
 the `Content-Digest` field.
 
 `Want-Representation-Digest` and `Want-Content-Digest` are Structured Fields
-List (see {{Section 3.2 of STRUCTURED-FIELDS}}) where:
+Dictionary (see {{Section 3.2 of STRUCTURED-FIELDS}}) where:
 
 * keys convey the hashing algorithm (see {{algorithms}});
 * values MUST be of type `Integer` ({{Section 3.3.1 of STRUCTURED-FIELDS}})
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -287,7 +287,7 @@ Representation-Digest: \
 
 A recipient MAY ignore any or all `dict-member`s.
 This allows the recipient to choose which hashing algorithm(s) to use for
-validation instead of verifying every received `dict-member`.
+validation instead of verifying every `dict-member`.
 
 A sender MAY send a `dict-member` without knowing whether the
 recipient supports a given hashing algorithm, or even knowing that the recipient
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -236,7 +236,7 @@ Integrity preference fields: collective term for `Want-Representation-Digest` an
 
 The `Representation-Digest` HTTP field can be used in requests and responses to
 communicate digests that are calculated using a hashing algorithm applied to
-the entire selected  *representation data* (see {{Section 8.1 of SEMANTICS}}).
+the entire *selected representation data* (see {{Section 8.1 of SEMANTICS}}).
 
 Representations take into account the effect of the HTTP semantics on
 messages. For example, the content can be affected by Range Requests or methods
@@ -251,9 +251,9 @@ string (see {{usage-in-signatures}}).
 
 `Representation-Digest` is a Structured Fields `Dictionary` (see {{Section 3.2 of
 STRUCTURED-FIELDS}}) where:
+* members cannot have parameters;
 * keys convey the hashing algorithm (see {{algorithms}})
-  used to compute the digest
-  and have no parameters;
+  used to compute the digest;
 * values MUST be of type `Byte Sequence`,
   which contain the output of the digest calculation.
 
@@ -346,9 +346,9 @@ communicate digests that are calculated using a hashing algorithm applied to
 the actual message content (see {{Section 6.4 of SEMANTICS}}). It is a
 Structured Fields Dictionary (see {{Section 3.2 of STRUCTURED-FIELDS}})
 where:
+* members cannot have parameters;
 * keys convey the hashing algorithm (see {{algorithms}})
-  used to compute the digest
-  and have no parameters;
+  used to compute the digest;
 * values MUST be `Byte Sequences`
   containing the output of the digest calculation.
 
@@ -410,9 +410,8 @@ the `Content-Digest` field.
 `Want-Representation-Digest` and `Want-Content-Digest` are Structured Fields
 List (see {{Section 3.2 of STRUCTURED-FIELDS}}) where:
 
-* members convey hashing algorithm preferences (see {{algorithms}});
-* keys convey the hashing algorithm (see {{algorithms}})
-  and have no parameters;
+* members cannot have parameters;
+* keys convey the hashing algorithm (see {{algorithms}});
 * values MUST be of type `Integer` ({{Section 3.3.1 of STRUCTURED-FIELDS}})
   in the range 0 to 10 inclusive.
   1 is the least preferred, 10 is the most preferred,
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -267,8 +270,8 @@ Representation-Digest: \
   iYllu7BNNyealdVLvRwEmTHWXvJwew==:
 ~~~
 
-Since `Representation-Digest` is a Dictionary, it can contain multiple
-members. This could be used, for example, to attach multiple digests
+The `sf-dictionary` syntax, can
```
ioggstream, line 273: @LPardue this text is now identical in both C-Digest and R-Digest. I guess if there's a way to avoid repetitions

draft-ietf-httpbis-digest-headers.md
```diff
@@ -282,11 +285,11 @@ Representation-Digest: \
   iYllu7BNNyealdVLvRwEmTHWXvJwew==:
 ~~~
 
-A recipient MAY ignore any or all of members of `Representation-Digest`.
+A recipient MAY ignore any or all `dict-member`s.
```
ioggstream, line 288: @LPardue we could replace `dict-member`s with just "members" or "digests" (which is already used here https://github.com/httpwg/http-extensions/pull/1973/files#diff-de34152d297a9eb29b85b24962949ba2e309478a5c32e1d94fd4c3ee27b4584dL271 )

draft-ietf-httpbis-digest-headers.md
```diff
@@ -370,10 +374,10 @@ List (see {{Section 3.2 of STRUCTURED-FIELDS}}):
 ~~~
 
 Dictionary members convey hashing algorithm preferences (see {{algorithms}}).
-Member-keys convey the hashing algorithm (see {{algorithms}}), member-values convey
-an ascending relative weighted preference between 0 and 10 inclusive. 1 is the
+Member-keys convey the hashing algorithm (see {{algorithms}}),
+member-values convey an ascending relative weighted preference
+and their syntax is `sf-integer` in the range 0 to 10 inclusive. 1 is the
```
ioggstream, line 379: Removed duplicated text.

LPardue:
Thanks for these. 

I'm still unsure of the best way to reference structured fields things. I've seen all kinds of permutations across drafts, so I sent an email to the list to see if anyone has an opinion we might be able to align to. I agree that being able to express the type constraints in ABNF would be nice but I don't think that's standard and might be a step too far.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -253,9 +254,11 @@ STRUCTURED-FIELDS}}):
...
 ~~~
 
-`Representation-Digest` member-keys convey the hashing algorithm (see
-{{algorithms}}) used to compute the digest. Member-values are the output
-of the digest calculation. Member-values MUST be of type sf-binary.
+where member-keys convey the hashing algorithm (see
+{{algorithms}}) used to compute the digest
+and have no parameters.
+Member-values are the output of the digest calculation
+and their syntax is `sf-binary`.
```
LPardue, line 261: How about we just move this to above the ABNF. So it reads

> `Representation-Digest` is a Structured Fields Dictionary (see {{Section 3.2 of
STRUCTURED-FIELDS}}). Member-keys convey ....
> 
> ~~~ abnf

LPardue, line 261: Separately I'm not a big fan of saying "the syntax is sf-xyz" because it puts a lot of attention on syntax, which doesn't matter, and it doesn't put a normative requirement on the type, which is very important. 

How about "Member-values are the output of the digest calculation, the type MUST be Byte Sequence."?

This comment applies to other instances too.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -282,11 +285,11 @@ Representation-Digest: \
   iYllu7BNNyealdVLvRwEmTHWXvJwew==:
 ~~~
 
-A recipient MAY ignore any or all of members of `Representation-Digest`.
+A recipient MAY ignore any or all `dict-member`s.
 This allows the recipient to choose which hashing algorithm(s) to use for
-validation instead of verifying every received representation-data-digest.
+validation instead of verifying every received `dict-member`.
```
LPardue, line 290: Good catch of representation-data-digest

``` suggestion
validation instead of verifying every `dict-member`.
```

LPardue, line 290: Good catch of representation-data-digest

``` suggestion
validation instead of verifying every `dict-member`.
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -521,7 +524,7 @@ signatures.
 Using signatures to protect the checksum of an empty representation
 allows receiving endpoints to detect if an eventual payload has been stripped or added.
 
-Any mangling of digest fields, including de-duplication of representation-data-digest values
+Any mangling of digest fields, including de-duplication of `dict-member`s
 or combining different field values (see {{Section 5.2 of SEMANTICS}})
 might affect signature validation.
```
LPardue, line 529: Dictionaries already describe how to treat duplicate entries. So we probably need a separate issue to fix up this paragraph.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -556,7 +559,7 @@ An endpoint might have a preference for algorithms,
 such as preferring "standard" algorithms over "insecure" ones.
 Transition from weak algorithms is supported
 by negotiation of hashing algorithm using `Want-Representation-Digest` or `Want-Content-Digest` (see {{want-fields}})
-or by sending multiple representation-data-digest values from which the receiver chooses.
+or by sending multiple digest calculations from which the receiver chooses.
```
LPardue, line 562: or simply
```suggestion
or by sending multiple digests from which the receiver chooses.
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -630,7 +633,7 @@ Specification document(s):  {{want-fields}} of this document
 
 --- back
 
-# Resource Representation and Representation-Data {#resource-representation}
+# Resource Representation and Representation Data {#resource-representation}
```
LPardue, line 636: This is a weird title now that we removed the term `Representation-Data`. Maybe there's a better alternative?

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1382,7 +1386,7 @@ _RFC Editor: Please remove this section before publication._
 ## Since draft-ietf-httpbis-digest-headers-01
 {:numbered="false"}
 
-* Digest of error responses is computed on the error representation-data #1004
+* Digest of error responses is computed on the error representation data #1004
```
LPardue, line 1389: I don't think we need to worry about updating terms in the changelog. That can quickly become an impossible task.

draft-ietf-httpbis-digest-headers.md
ioggstream: I  thought that syntax and type were tighly coupled, but if it's not the case (eg. a type can have different syntaxes)
it's ok.

As a follow up to your email, I think that a formal way to express relations between items - that is defining a "schema" for structured fields - would be great, but I suspect that it's not trivial. Moreover this could have sense only if fields are serialized differently from rfc8941.

draft-ietf-httpbis-digest-headers.md
ioggstream: Once a field is signed, even a compliant mangling invalidates the signature. I see this line as a simple advice to the users.

draft-ietf-httpbis-digest-headers.md
ioggstream: We removed `representation-data-digest`. Representation data is the concept defined in SEMANTICS.

draft-ietf-httpbis-digest-headers.md
ioggstream: it was just a s//g ;)

draft-ietf-httpbis-digest-headers.md
ioggstream: "digest" is actually mentioned in Notational Conventions in relation to the values that we have in the fields... let's use it!

draft-ietf-httpbis-digest-headers.md
```diff
@@ -150,7 +150,7 @@ data and HTTP content.
 
 This document defines the `Representation-Digest` request and response header
 and trailer field ({{representation-digest}}) that contains a digest value
-computed by applying a hashing algorithm to `selected representation data`
+computed by applying a hashing algorithm to *selected representation data*
```
LPardue, line 153: I think we'll end up revisiting how best to draw attention to these terms. Emboldening doesn't seem like something the RFC editor would do but we can cross that bridge later on.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -211,19 +212,20 @@ have names that more clearly articulate the intended usages.
 This document uses the Augmented BNF defined in [RFC5234] and updated by
 [RFC7405].
 
-The terms Dictionary, List, sf-dictionary, sf-integer and sf-binary are imported from
+The terms Dictionary, List, Byte Sequence, sf-dictionary,
+sf-integer and sf-binary are imported from
 {{!STRUCTURED-FIELDS=RFC8941}}.
```
LPardue, line 217: Borrowing MT's suggestion from https://github.com/httpwg/http-extensions/pull/1977/files#diff-4dc65e5537b9f2d8c77d359e6b7a032db79f294b488d2ae2b0554aa96f9b9258R133

```suggestion
This document uses the Boolean, Byte Sequence, 
Dictionary, Integer and List types from
{{!STRUCTURED-FIELDS=RFC8941}} along with
the sf-dictionary and sf-list ABNF rules.
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -233,7 +235,7 @@ Integrity preference fields: collective term for `Want-Representation-Digest` an
 
 The `Representation-Digest` HTTP field can be used in requests and responses to
 communicate digests that are calculated using a hashing algorithm applied to
-the entire selected `representation data` (see {{Section 8.1 of SEMANTICS}}).
+the entire selected  *representation data* (see {{Section 8.1 of SEMANTICS}}).
```
LPardue, line 238: above the attention is drawn to the term 

selected representation data

I can't recall if there's a reason this one is different by excluding selected.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -243,19 +245,21 @@ dependent on other transformations (e.g. transfer codings for HTTP/1.1 - see
 several examples are provided in {{resource-representation}}.
 
 When a message has no representation data it is still possible to assert that no
-representation data was sent computing the representation digest on an empty
+representation data was sent computing the digest on an empty
```
LPardue, line 248: old typo can be fixed too

```suggestion
representation data was sent by computing the digest on an empty
```

draft-ietf-httpbis-digest-headers.md
```diff
@@ -243,19 +245,21 @@ dependent on other transformations (e.g. transfer codings for HTTP/1.1 - see
...
 string (see {{usage-in-signatures}}).
 
-`Representation-Digest` is a Structured Fields Dictionary (see {{Section 3.2 of
-STRUCTURED-FIELDS}}):
+`Representation-Digest` is a Structured Fields `Dictionary` (see {{Section 3.2 of
+STRUCTURED-FIELDS}}) where:
+* keys convey the hashing algorithm (see {{algorithms}})
+  used to compute the digest
+  and have no parameters;
```
LPardue, line 255: I don't follow this line, keys can't have parameters by design. If you mean that there are no parameters defined for Dictionary members, that probably needs to be done as a third bullet.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -340,16 +343,18 @@ An example is given in {{post-not-request-uri}}.
...
 communicate digests that are calculated using a hashing algorithm applied to
 the actual message content (see {{Section 6.4 of SEMANTICS}}). It is a
-Structured Fields Dictionary (see {{Section 3.2 of STRUCTURED-FIELDS}}):
+Structured Fields Dictionary (see {{Section 3.2 of STRUCTURED-FIELDS}})
+where:
+* keys convey the hashing algorithm (see {{algorithms}})
+  used to compute the digest
+  and have no parameters;
+* values MUST be `Byte Sequences`
+  containing the output of the digest calculation.
```
LPardue, line 352: Same points here, as for Representation digest

draft-ietf-httpbis-digest-headers.md
```diff
@@ -403,18 +407,22 @@ on messages associated with the request URI and representation metadata, using
 the `Content-Digest` field.
 
 `Want-Representation-Digest` and `Want-Content-Digest` are Structured Fields
-List (see {{Section 3.2 of STRUCTURED-FIELDS}}):
+List (see {{Section 3.2 of STRUCTURED-FIELDS}}) where:
+
+* members convey hashing algorithm preferences (see {{algorithms}});
```
LPardue, line 412: we already state that above, so this seems duplicative
```suggestion
```

draft-ietf-httpbis-digest-headers.md
ioggstream: Sure, I already filed an issue on kramdown for that ;)

draft-ietf-httpbis-digest-headers.md
ioggstream: I'll check the spec and fix

draft-ietf-httpbis-digest-headers.md
ioggstream: https://github.com/cabo/kramdown-rfc2629/issues <-

draft-ietf-httpbis-digest-headers.md
```diff
@@ -243,19 +246,21 @@ dependent on other transformations (e.g. transfer codings for HTTP/1.1 - see
...
 When a message has no representation data it is still possible to assert that no
-representation data was sent computing the representation digest on an empty
+representation data was sent by computing the digest on an empty
 string (see {{usage-in-signatures}}).
 
-`Representation-Digest` is a Structured Fields Dictionary (see {{Section 3.2 of
-STRUCTURED-FIELDS}}):
+`Representation-Digest` is a Structured Fields `Dictionary` (see {{Section 3.2 of
+STRUCTURED-FIELDS}}) where:
+* members cannot have parameters;
```
LPardue, line 254: This is something I sidestepped during the initial change to SF. The SF grammar allows parameters, so member-values *can* have them. The SF says

```
   Both Items and Inner Lists allow parameters as an extensibility
   mechanism; this means that values can later be extended to
   accommodate more information, if need be.  To preserve forward
   compatibility, field specifications are discouraged from defining the
   presence of an unrecognized parameter as an error condition.
```

When updating Digest we decided parameters could be a security problem. If we want prohibit the use of SF parameters, then we need some form of normative text to state that and probably justify why. I'd suggest doing that in a separate paragraph.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -403,18 +408,21 @@ on messages associated with the request URI and representation metadata, using
 the `Content-Digest` field.
 
 `Want-Representation-Digest` and `Want-Content-Digest` are Structured Fields
-List (see {{Section 3.2 of STRUCTURED-FIELDS}}):
+List (see {{Section 3.2 of STRUCTURED-FIELDS}}) where:
+
+* members cannot have parameters;
```
LPardue, line 413: This case is a bit different, what is good justification to prohibit parameters on want-* fields?

draft-ietf-httpbis-digest-headers.md
```diff
@@ -479,20 +486,20 @@ The entries in {{iana-hash-algorithm-table}} are registered by this document.
...
 ## End-to-End Integrity
 
-Representation-Digest and Content-Digest can help detect `representation data` or content modification due to implementation errors,
+Integrity fields can help detect  *representation data* or content modification due to implementation errors,
 undesired "transforming proxies" (see {{Section 7.7 of SEMANTICS}})
 or other actions as the data passes across multiple hops or system boundaries.
-Even a simple mechanism for end-to-end `representation data` integrity is valuable
-because user-agent can validate that resource retrieval succeeded before handing off to a
+Even a simple mechanism for end-to-end  *representation data* integrity is valuable
+because a *user agent* can validate that resource retrieval succeeded before handing off to a
```
LPardue, line 502: The grammer fix is good but there doesn't seem to be any need to treat user agent as special by adding '*'

draft-ietf-httpbis-digest-headers.md
LPardue: Furthermore, if we move the parameter discussion elsewhere, then we can get rid of the list format and use a prose style similar to what's already there.

draft-ietf-httpbis-digest-headers.md
ioggstream: Ok.

draft-ietf-httpbis-digest-headers.md
ioggstream: @LPardue I added a paragraph in the security consideration.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -346,7 +347,8 @@ communicate digests that are calculated using a hashing algorithm applied to
 the actual message content (see {{Section 6.4 of SEMANTICS}}). It is a
 Structured Fields Dictionary (see {{Section 3.2 of STRUCTURED-FIELDS}})
 where:
-* members cannot have parameters;
+
+* members MUST NOT have parameters (see {{sec-agility}});
```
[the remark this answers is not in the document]
LPardue: Furthermore, prohibiting parameters suffers can lead to the same fate of validation DoS.

So the text that discusses normative behaviour of parameters has to be more nuanced than a single line.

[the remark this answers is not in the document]
LPardue: Understood but it's likely that such an editorial comment would come up during our later review anyway. I'll take a go and implementing my suggestion.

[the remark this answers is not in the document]
ioggstream: > I'll take a go and implementing my suggestion. 

:rocket:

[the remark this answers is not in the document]
ioggstream: Parameters will be managed in another PR

ioggstream:
@LPardue I deferred parameters to another PR. After you refactor the list as prose, feel free to merge :)

draft-ietf-httpbis-digest-headers.md
```diff
@@ -403,18 +408,20 @@ on messages associated with the request URI and representation metadata, using
 the `Content-Digest` field.
 
 `Want-Representation-Digest` and `Want-Content-Digest` are Structured Fields
-List (see {{Section 3.2 of STRUCTURED-FIELDS}}):
+List (see {{Section 3.2 of STRUCTURED-FIELDS}}) where:
```
LPardue, line 411:
```suggestion
Dictionary (see {{Section 3.2 of STRUCTURED-FIELDS}}) where:
```

LPardue (approved):
Thanks @ioggstream, theres a lot of good fixes in here and the effort is appreciated.

PR #1976, opened 2022-02-16
Add support for absent header fields

### LINKED ISSUE #1864: Signing Empty / Absent Headers
Currently, an empty header can be included, but an absent header can't be guaranteed. 

Proposal: add a boolean parameter for header field identifiers to mark a specific field as "not present" in the target message. Its value is an empty string. (Or should it be some special flag?)

If the verifier finds any header value with that name in the target message, throw an error.

jricher:
Closes #1864
﻿

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -341,6 +341,30 @@ The following example shows canonicalized values for different component identif
 
 Note that the value for `key="c"` has been re-serialized.
 
+### Absent HTTP Fields
+
+If the signer wants to signal that an HTTP field is not present in the signed message at the time the signature is generated, the signer can include that field in the signature with the parameter `absent` to indicate that the field MUST NOT be present in the message.
+
+The component value is the empty string.
+
+For example, for the following message:
+
+~~~ http-message
+HTTP/1.1 201 No Content
+Date: Wed, 16 Feb 2022 04:20:23 GMT
+~~~
+
+The signer wishes to indicate that a `Content-Encoding` header is not included in the response being signed, which can be shown as presented using the signature input string format discussed in {{create-sig-input}}:
+
+~~~
+NOTE: '\' line wrapping per RFC 8792
+
+"content-encoding";absent: \
+
+~~~
+
+Note: these are shown here using the line wrapping algorithm in {{RFC8792}} due to limitations in the document format that strips trailing spaces from diagrams.
+
 ## Derived Components {#derived-components}
 
 In addition to HTTP fields, there are a number of different components that can be derived from the control data, processing context, or other aspects of the HTTP message being signed. Such derived components can be included in the signature input by defining a component identifier and the derivation method for its component value.
@@ -850,6 +874,7 @@ If covered components reference a component identifier that cannot be resolved t
  * The component identifier indicates that a structured field serialization is used (via the `sf` parameter), but the field in question is known to not be a structured field or the type of structured field is not known to the implementation.
  * The component identifier is a dictionary member identifier that references a field that is not present in the message, is not a Dictionary Structured Field, or whose value is malformed.
  * The component identifier is a dictionary member identifier or a named query parameter identifier that references a member that is not present in the component value, or whose value is malformed. E.g., the identifier is `"example-dict";key="c"` and the value of the `Example-Dict` header field is `a=1, b=2`, which does not have the `c` value.
+ * The component identifier is a field with the parameter value `absent` but the field exists in the source message.
 
 In the following non-normative example, the HTTP message being signed is the following request:
 
@@ -2184,6 +2209,7 @@ Jeffrey Yasskin.
   - -09
      * Explained key formats better.
      * Removed "host" and "date" from most examples.
+     * Added "absent" header parameter.
 
   - -08
      * Editorial fixes.
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -289,7 +289,7 @@ NOTE: '\' line wrapping per RFC 8792
 
 Note: these are shown here using the line wrapping algorithm in {{RFC8792}} due to limitations in the document format that strips trailing spaces from diagrams.
 
-Any HTTP field content identifiers MAY have the following parameters in specific circumstances.
+Any HTTP field component identifiers MAY have the following parameters in specific circumstances.
 
 sf
 : A boolean flag indicating that the field value is to be canonicalized using strict encoding
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -289,6 +289,18 @@ NOTE: '\' line wrapping per RFC 8792
 
 Note: these are shown here using the line wrapping algorithm in {{RFC8792}} due to limitations in the document format that strips trailing spaces from diagrams.
 
+Any HTTP field content identifiers MAY have the following parameters in specific circumstances.
+
+sf
+: A boolean flag indicating that the field value is to be canonicalized using strict encoding
+of the structured field value. {{http-header-structured}}
+
+key
+: A string parameter used to select a single member value from a dictionary structured field. {{http-header-dictionary}}
+
+absent
+: A boolean flag indicating that the field does not occur in the target message. {{http-header-absent}}
+
 ### Canonicalized Structured HTTP Fields {#http-header-structured}
 
 If value of the the HTTP field in question is a structured field ({{!RFC8941}}), the component identifier MAY include the `sf` parameter to indicate it is a known structured field. If this
@@ -341,7 +353,7 @@ The following example shows canonicalized values for different component identif
 
 Note that the value for `key="c"` has been re-serialized.
 
-### Absent HTTP Fields
+### Absent HTTP Fields {#http-header-absent}
 
 If the signer wants to signal that an HTTP field is not present in the signed message at the time the signature is generated, the signer can include that field in the signature with the parameter `absent` to indicate that the field MUST NOT be present in the message.
 


draft-ietf-httpbis-message-signatures.md
```diff
@@ -341,6 +341,30 @@ The following example shows canonicalized values for different component identif
...
+HTTP/1.1 201 No Content
+Date: Wed, 16 Feb 2022 04:20:23 GMT
+~~~
+
+The signer wishes to indicate that a `Content-Encoding` header is not included in the response being signed, which can be shown as presented using the signature input string format discussed in {{create-sig-input}}:
+
+~~~
+NOTE: '\' line wrapping per RFC 8792
+
+"content-encoding";absent: \
```
martinthomson, line 362: The use of '\' line wrapping confused me here.  It reads line you aren't terminating the line, but I think that the blank line that follows implies that the line is terminated.  The note clarifies that the terminator is to capture the trailing space, but that seems unnecessary (see above).

martinthomson (changes requested):
I'm not sure that this description is clear enough.  The way this works is that the field is listed in the signature input list:

```
Signature-Input: sig=("absent-field");created=1645056898;alg="rsa-pss-sha512"
```

But if the header field is not present in the message, the canonical string will include a line that includes ";absent" between the quoted field name and the colon.

Also, I realize that this is a bikeshed thing, but why not a simpler encoding?  Even just dropping the colon and space would work (it's still an injective transform).  Or with something in place of the colon+space (";absent" is fine, but "-" would work just as well).

draft-ietf-httpbis-message-signatures.md
jricher: We did something similar with the "empty header" examples for the same reasons. The tools strip off the trailing space from examples, which messed up people trying to build the examples correctly. Happy to hear other solutions, but this does give the right output when the line unwrapping is applied.

jricher:
No, the signature input would be:

```
Signature-Input: sig=("absent-field";absent);created=1645056898;alg="rsa-pss-sha512"
```

And then the line would be added to the signature input as:

```
"absent-field";absent: 
```

Just like any other content identifier. The value is an empty string, which is what shows up after the colon and space.

yaronf:
My usual question: why does `absent` apply to regular headers but not to dictionary members, or to query params?

martinthomson:
Ok, this is probably just me, but I have something of a visceral reaction to:

```
Signature-Input: sig=("absent-field";absent);created=1645056898;alg="rsa-pss-sha512"
```

Why does this need to be specified there?  Either the field is present or absent.  The signature will guarantee that.  It doesn't need to be distinguished any place other than in the input to the signature.  That might address Yaron's concern as well, I think.

jricher:
It's to ensure a field is not injected by an intermediary.

jricher:
@yaronf this PR is about **prohibited** fields, not optional ones. If they're tagged "absent" then they MUST NOT show up in the resulting message. Honestly "optional" feels like it's opening up a door for security bugs and implementation quirks. And the signer is going to know whether the field is there or not, and what its value is.  If that's changed by the time it hits the verifier, then the signature fails. Any decision on the optionality of a component needs to happen before the covered components are chosen, and the state of that field is fixed by the time the signature creation process is started.

Also note, an empty field value is allowed and has a defined value syntax, see https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-08.html#section-2.1-10 for the details.

jricher:
I also want to note about the alternate encoding suggestions: remember that the HTTP signature input is never parsed by any party. The signature input parameters are not reconstructed from the input string -- and it should be considered a write-only buffer by the generator. As such, as long as a given line is unique for a given message and can be created deterministically, it doesn't need to use any different constructions for a given line. I think this greatly simplifies the code. Right now, ever single line is of the form:

```
sf-string params ":" SP component-value
```

Where "component-value" is defined by the component and its parameters.

Tagging a header field as "absent" here already creates a different line than when it's present but without a value:

```
"header": 
"header";absent: 
```

So a signer or verifier will never create the wrong line, as the flagging is explicit.

yaronf:
> @yaronf it's an indication that if the field shows up, the verifier can assume that something unexpected happened. It's increased security. Think of it as an extra layer of enforcement on the guidelines of "don't trust a value that isn't signed" that are already listed in the security considerations.

I think we have a disconnect on the goals of the verifier. To me, a typical verifier is not out to detect malicious injections. All they want is to provide a simple policy (a list of assertions, and in the simplest case a list of required fields) and then leave it to the signature verification layer to ensure the message is trustworthy, before it can be sent to the rest of the application.

Seen this way, if all I want is to enforce the assertion "field X is absent", the verifier library can do it just fine without any signature syntax. Just check that the thing doesn't exist!

draft-ietf-httpbis-message-signatures.md
```diff
@@ -289,6 +289,18 @@ NOTE: '\' line wrapping per RFC 8792
 
 Note: these are shown here using the line wrapping algorithm in {{RFC8792}} due to limitations in the document format that strips trailing spaces from diagrams.
 
+Any HTTP field content identifiers MAY have the following parameters in specific circumstances.
```
yaronf, line 292: First, content identifiers -> component identifiers.

I think this paragraph is over-generalizing. This would only make sense if Structured Fields were a recursive structure.

Agreed that the `absent` modifier can apply to any identifier (my misgivings about it aside).

I'm less sure about `sf`: can it be applied meaningfully to query parameters?

And `key` is very specific to dictionaries, so it can only apply to full headers.

The last one is `req` that was discussed elsewhere, that one could apply to all component identifiers.

jricher:
@yaronf I agree that a verifier can check for absence of a field, and will need to do so if the application requires it, but without the signature syntax it won't know **that the intent of the signer is that the field is absent**. I'm also confused by your separation of "verifier' and "signature verification layer" in the comment above; I would have assumed they are the same, can you elaborate on what the distinction you're making is?

draft-ietf-httpbis-message-signatures.md
jricher: This paragraph is very specifically talking about HTTP field component identifiers, not all component identifiers. Query parameters are a derived component, not a field.

yaronf:
Sorry for the invented terminology. I thought of the "verifier" as the main business logic, and it calls a "signature verification layer" (a.k.a. my library...) to validate the incoming message, including verification of the signature and and additional policy enforcement.

Honestly I don't see the value in validating the **intent** of the signer. I believe the discussion now is a use case discussion. You think this is a real-life use case, I don't.

draft-ietf-httpbis-message-signatures.md
yaronf: Fair enough. But the paragraph still leaves open the question whether you can have both `key` and `absent` in the same field component identifier, i.e. if you can signal the intent to **not** include a particular dictionary member.

martinthomson:
@jricher, I am trying to understand why you think that Signature-Input needs to be built this way.

It's not a specification, where you need to know, based on the value of the string, whether the field is absent or not.  The signer signs a message, which either has the field present or absent.  It knows.  Same for the verifier.

@yaronf's last few comments shows that he understands this perspective.

jricher:
@martinthomson Because it's a detached signature on an object that can be changed in the middle between the signer and verifier, I believe this is a useful signal. Signing an present header says "This field is here and I know its value". Signing an absent header says "this field is not there". If I sign a present header and it's not there anymore when the verifier gets it, that's an error. If I sign an absent header and it's there when the verifier gets it, that's an error. Yes of course they both know that a field is there or not when they sign or verify the message -- the important thing here is whether that field was still there or not at the other end.

martinthomson:
I'm not following your logic.  Wouldn't this invite mistakes where a verifier that receives this detached signature sees that something should be absent, so (incorrectly) omits it from the signature, even when it is present?

jricher:
@martinthomson That is not what the signature input creation algorithm dictates. Yes, someone could make that mistake in implementing this -- but then the signature would fail anyway because the identifier would not be present in the signature input string at the verifier. And if it's present and the verifier ignores the `absent` parameter and adds the header with a value, then it will also fail because the value will not be blank, as is prescribed by the definition of the `absent` parameter.

So yes, someone could make mistakes when implementing support for this, but those mistakes will at least fail closed and result in rejected signatures. I would argue that this is the correct state because the signatures are not being implemented correctly and therefore should be rejected.

martinthomson:
I'm suggesting that upon receipt of this:

```
Signature-Input("not-present";absent);blah=blah
Not-Present: ?1
...
```
A verifier might choose to pretend that `Not-Present` is not present because the `Signature-Input` told it to do that.

On the other hand, if it were just 
```
Signature-Input("not-present");blah=blah
Not-Present: ?1
...
```

Then they have no additional information that might guide them toward a bad choice.

LPardue:
I don't follow this example. In my naive understanding the signer is saying "for this message to be valid, there must not be a header field with this name". If an implementation pretends there wasn't a header when there was, then that's broken; just like if synthesised a header that should have been present but wasn't.

martinthomson:
Yes, but we're talking about potential for misuse.  You might reasonably say "don't do that", but we know that this is not what happens.  The latter example gives no clues as to what is valid, so it isn't open to that sort of abuse.

LPardue:
The latter example seems to be, don't support a signal for absent header. Pr in otherworld, don't add this feature?

martinthomson:
No, I'm saying that you can sign over absent headers - that only requires that there is an unambiguous encoding difference between absent and empy - but you don't need to signal that they are absent in `Signature-Input`.

LPardue:
Oh I think I see what you're saying now. That's not how I understood it earlier.

To replay, a signer would see Signature-Input including `Not-Present` field and be provided an HTTP message that didn't include `Not-Present`. The signature output would encode that somehow as "absent". The validator would receive a signature and, if `Not-Present` was now on the input message, the signature comparison would fail.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -289,6 +289,18 @@ NOTE: '\' line wrapping per RFC 8792
 
 Note: these are shown here using the line wrapping algorithm in {{RFC8792}} due to limitations in the document format that strips trailing spaces from diagrams.
 
+Any HTTP field component identifiers MAY have the following parameters in specific circumstances.
+
+sf
```
mnot, line 294: I _think_ you need to allow the top-level type of the field to be specified here (Dictionary, List, or Item) -- that information is necessary input.

draft-ietf-httpbis-message-signatures.md
mnot: I take it the whitespace is required by the canonical format you've chosen? If so, what about removing it?

martinthomson:
yes, given that s/signature output/signature input/.  it's confusing because there is the `Signature-Input` field and the "signature input" as the canonical bytes that are fed to the signature function.  I'm suggesting that you need the signal on the latter, but having it on the former is a hazard.

yaronf:
> yes, given that s/signature output/signature input/. it's confusing because there is the `Signature-Input` field and the "signature input" as the canonical bytes that are fed to the signature function. I'm suggesting that you need the signal on the latter, but having it on the former is a hazard.

"Signature input" or "signature input string" appears 87 times in the current draft... I would have replaced them by "signed bytes" to eliminate this confusion. An even better though incompatible solution would be to change `Signature-Input` to `Signed-Components`.

Either of them is a lot of work, and this doesn't seem critical to me.

LPardue:
To agree with Yaron's point, one is an abstract definition of the input and the other is a concrete input based on the HTTP message. Making sure people understand that unambiguously right seems important and not _that_ much editorial work if there are more clear terms.

draft-ietf-httpbis-message-signatures.md
jricher: It was my understanding that this information would be known based on the field itself, by both sides. So putting this on `Representation-Siganture` you know that it's always a Dictionary. I wouldn't want to have a way to try and signal that within the signature parameters.

jricher:
I would be fine with renaming "signature input string" to something else to make it more clear. I am not a fan of any of the suggestions made here so far, though, so I've submitted a new issue: #1986

jricher:
@martinthomson I understand what you're proposing and disagree that it makes things better. You lose the intentionality of saying "this is what I meant to sign" and instead fall into a space where things fail because of some other internal, hidden thing. I've debugged enough detached signature systems in my lifetime and have been bitten by implied state like this many many times. I also disagree with you that flagging a field as absent is going to make developers leave it out of the signature string. As we've discussed it's a broken implementation if they do it, and it's not what the text says to do. And if they did, then that would still fail the signature validation process.

@mnot One of the considerations that I had going into this was to keep the signature base string generation algorithm simple and deterministic. Each line right now is always the same exact format:

- content identifier (with parameters)
- colon
- space
- content value

If we have a different encoding for empty and absent fields or values, then we have to call out those branches in the generation algorithm. I think this is doable if we can declare a branch like "if the value is not `null` or the empty string, add a colon and a space then the value, otherwise next step", and I can submit a separate PR that looks at that separately from the `absent` flag.

yaronf:
> If we have a different encoding for empty and absent fields or values, then we have to call out those branches in the generation algorithm. I think this is doable if we can declare a branch like "if the value is not `null` or the empty string, add a colon and a space then the value, otherwise next step", and I can submit a separate PR that looks at that separately from the `absent` flag.

As an implementor I value the consistency where we always have `:<sp>`, even at the cost of line-final spaces in some cases. Let's not forget that this string is never visible on the wire.

mnot:
@jricher I meant _all_ fields -- the space doesn't add anything, and in this case causes an issue.

jricher:
@mnot the space has the particular feature of not being allowed as a starting character in header field values, which are the primary target of signature inputs. Having a character to act as a delimiter that is not allowed to start the input is a well-trodden path for avoiding a class of bugs in canonicalization where a confusing input can look like it's been delimited already. We have the benefit of not ever needing to parse the value that's created, which is really nice, but it's still good in my opinion to have something that's more unambiguous for the generation of the string in the first place.

mnot:
OK, I'll ask - why doesn't the colon serve that purpose?

jricher:
The colon is allowed as the first character in a field value, the space isn't. If anything, it's the colon that's redundant here, but I believe part of the original design (imported from Cavage) was to mimic how header fields are serialized. Ultimately I don't see a good reason to change that -- avoiding awkward formatting limits of the tools for a couple examples isn't a good reason to change the entire syntax.

mnot:
If the colon is required in the serialisation, it's unambiguous, and can't be confused with the first character of the field value. What am I missing here?

martinthomson:
I'm with @mnot here.  "`"example": \r\n`" (present but empty, with a space) and "`"example":\r\n`" (present but empty without a space) and "`"example"\r\n`" (absent) and "`"example": :\r\n`" (present with a value of a single colon) are all distinct.

jricher:
With the current serialization format plus this PR, a (syntactically valid) value of `Example: :` you get:

```
"example": :\n   (present)
"example": \n    (empty)
"example";absent: \n (absent)
```

If we remove the space and adopt the no-value absent line we get:

```
"example"::\n   (present)
"example":\n    (empty)
"example"\n (absent)
```

While still distinct, I'd argue it's easier for someone developing and debugging something like this to look at the first one and think it's probably an error, and overcorrect it to look like the second value, especially if one side has already made the mistake and you're trying to just make things work. This is about making it clear to an implementor what to do in every case at every step, especially when things go wrong -- which they will.

So let's talk about parameter names, which as a token can include the colon character unescaped. So somebody adds the parameter "a=b:" to a field and you end up with the even more confusing:

```
"example";a=b:::\n   (present)
"example";a=b::\n    (empty)
"example";a=b:\n (absent)
```

And then things get fun, because with that parameter added, you now have a case where the presence of the parameter `;a=b:` and a field value of `:` now serializes to the same as a parameter value of `;a=b` and a field value of `::`, both of them are:

```
"example";a=b:::
```

Using the space, which can neither end a parameter name or value nor start the field value, prevents that. This is a small collision, but it's exactly the kind of collision that's unlikely to happen on its own, but it's the kind of thing that can act as a foothold for other kinds of injection and manipulation.

We can avoid this entire class of problems by having a clear and unambiguous serialization for each distinct case. Using the space and colon we end up with two different things:

```
"example";a=b: ::\n
"example";a=b:: :\n
```

So as I said above, it's the colon that's redundant, and the space is needed to stave off this kind of collision. Without that extra colon but with a space we still have unambiguous encodings:

```
"example";a=b ::\n
"example";a=b: :\n
```

But, the colon isn't really raising any issues on its own here, and getting rid of it and the space doesn't get rid of the ambiguity of creating:

```
"example";a=b::\n
```

While we don't have to worry about ambiguous parsing, we do need to be concerned about ambiguous generation. Anything you can give an attacker to force a change that they control into the input of the signature that's different from what's intended is a bad design. This is all just for the signature base string -- hopefully the rest of the application will be checking for things like valid parameter names and values and valid field values, but if they aren't and you can get the signature to pass with something other than what was used on the signer's end, then you've undermined the whole system.

And finally, I still do not see the value of changing the core serialization mechanic at this point without really good reason to do so. The proposed solutions make the input string prettier, but do not make it easier or less ambiguous to create. Remember, this string is never seen except in debugging situations (and the draft examples), and it is never sent over the wire. Deterministic generation is absolutely important here.

Now: This whole conversation does raise the question: does an empty field (present with no value) have a semantic difference from an absent field? My reading of SEMANTICS points to "yes", or at least "it could depending on the field definition". If that isn't the case, and they are equivalent, they should be serialized the same way. But that's separate from the underlying question of how to encode all the values in the string.

martinthomson:
I thought that we'd already resolved the present vs. empty thing.  There is a semantic difference.

As for the parameters thing, why?  Field values might have parameters, but why would the name need a value?  I know that the grammar of Inner List supports the notion, but why would you allow that here?

I'm happy to concede that we simply disagree about signaling absence in the `Signature-Input`, but isn't this the essence of that disagreement?  That is, why seek to make things more complicated: ban the use of parameters in this instance (or at least say that there are currently none and refuse to define what to do with them for the moment) and avoid the complications.

richanna:
> And then things get fun, because with that parameter added, you now have a case where the presence of the parameter ;a=b: and a field value of : now serializes to the same as a parameter value of ;a=b and a field value of ::

The value of the `@signature-input` component is different for each of those, so the signatures will still be different. So technically that's still safe, but it makes me nervous. I'd definitely prefer it if there were no serialization collisions between semantically distinct components.

> This whole conversation does raise the question: does an empty field (present with no value) have a semantic difference from an absent field?

It can, depending on the field definition. For example, see [`Accept-Encoding`](https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.4):

```
A request without an Accept-Encoding header field implies that the
user agent has no preferences regarding content-codings.
...
An Accept-Encoding header field with a combined field-value that is
empty implies that the user agent does not want any content-coding in
response.
```
(hat-tip: Willy Tarreau on the mailing list for pointing out this example)

---

I want to step back and consider use cases and failure modes involving absent header fields. With the published draft, signature input generation fails at [step 2.4](https://httpwg.org/http-extensions/draft-ietf-httpbis-message-signatures.html#section-2.3-4.2.2.4.2.2) if the component list includes an absent header field. There is no way for the the signer to explicitly signal that they did not include a particular header field. If the header field is present in the message, the verifier has no way to determine if it was included by the signer or injected along the way. But this is true of all unsigned components. Message signatures provide the verifier with information about the components that were included in the signature input. They do not provide information on components not included in the signature input.

This PR changes that by allowing the message signature to assert something about a component that is not covered – specifically, `example-header;absent` asserts that that any `example-header` present in the message did not come from the signer. Is this useful information to the verifier? Since `example-header` is absent from the signature input, the verifier already knows that the signature provides no information about its value. But is there a meaningful difference between "this signature doesn't tell me anything about this value" and "this signature tells me this value didn't come from the signer"? Under what circumstances would that difference trigger a change in a verifier's behavior?

Given the degree to which proxies can modify messages, including this information might cause more harm than good. Consider an overzealous client that includes `Accept-Encoding;absent` in their signature input. Suppose requests pass through a proxy that adds `Accept-Encoding: compress, gzip`, as a bandwidth optimization. Now the service receives a request with an invalid signature. The service could decide that since `Accept-Encoding` isn't in its list of meaningful, must-be-signed components, that it will ignore the value for the purposes of signature verification and build the input string as if it were absent. That might even be safe, but it's awfully complicated and seems like the kind of code that shows up as the root cause of a CVE a year later.

Here's another example: a client developer heard about bad guys spoofing `X-Forwarded-For`, so they decide to include `X-Forwarded-For;absent` so no one can replay their signatures in requests with `X-Forwarded-For` header fields. Take that, bad guys. ...Except the service's load balancer adds `X-Forwarded-For`, and even adds a signature over it. So now just like the earlier example the service host receives a request with an invalid signature.

The fundamental issue here is that the signer is presuming that a header field being injected by an intermediary is necessarily an error state, and in so doing they are making an unfounded assertion about content they have not seen and know nothing about. It is neither the signer's responsibility nor place to dictate to the verifier how it should treat unsigned components.

jricher:
The overzealous signer is a problem whether it's signing present fields it shouldn't or signing absent fields. The verifier hack to insert an "ignore" on an absent field is a compelling argument against allowing this signal, though.

I still don't think I fully agree, but I can see I'm probably in the rough on this. I'm going to leave the PR open for now and hopefully anyone who wants to see this feature (or something like it) added will pipe up, and we can have a sidebar conversation at IETF113.

And thanks for the semantic conversation -- I missed that message on the list (the thread got binned to the spam folder), but it aligns with what I thought the case was.

jricher:
After review of the previous discussion, I still think this is something we **should not add**. 

I am strongly against the proposed alternative of having a signer include a covered component that does not exist by using the same syntax as a covered component that does exist,  and having that process create a valid signature.

PR #1977, opened 2022-02-17
priority: use SF terms, not ABNF

### LINKED ISSUE #1975: Use SF terms, not ABNF
Priority currently uses sf- prefixed ABNF to describe datatypes in prose. As discussed, this should use the terminology from the draft instead (e.g., "Integer, Boolean").

LPardue:
Closes #1975.

We're not in AUTH48 yet. I suggest we hold merging the change until that time.

### DIFF
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -285,8 +285,8 @@ Replacing or adding a Priority header field overrides any signal from a client
 and can affect prioritization for all subsequent recipients.
 
 For both the Priority header field and the PRIORITY_UPDATE frame, the set of
-priority parameters is encoded as a Structured Fields Dictionary (see
-{{Section 3.2 of STRUCTURED-FIELDS}}).
+priority parameters is encoded as a Dictionary (see {{Section 3.2 of
+STRUCTURED-FIELDS}}).
 
 This document defines the urgency (`u`) and incremental (`i`) priority
 parameters. When receiving an HTTP request that does not carry these priority
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -137,8 +137,8 @@ servers might act upon signals.
 
 {::boilerplate bcp14-tagged}
 
-The terms "Dictionary", "sf-boolean", "sf-dictionary", and "sf-integer" are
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-priority.md
```diff
@@ -130,8 +130,8 @@ consideration and guidance about how servers might act upon signals.
 
 {::boilerplate bcp14-tagged}
 
-The terms Dictionary, sf-boolean, sf-dictionary, and sf-integer are imported from
-{{!STRUCTURED-FIELDS=RFC8941}}.
+The terms Boolean, Dictionary, and Integer, along with the sf-dictionary rule,
```
martinthomson, line 133: I don't think that you want to say that RFC 8941 defines the term "Boolean".  Maybe instead "This document uses the Boolean, Dictionary, and Integer types from {{!SF}} along with the `sf-dictionary` ABNF rule."

LPardue: Thanks. "Boolean" is so generic, it was part of the reason why I gravitated to `sf-boolean` in prose in the first place.

mnot: In prose, you can try "Boolean Structured Type" or similar.

LPardue: The pattern I've seen so far is people saying "Structured Fields Dictionary" or "Structured Fields List". So "Structured Fields Boolean" would be consistent with that but its a bit wordy. Perhaps settling on a common acronym of `SF` in front of these would work.

LPardue: Applying this to "SF Member" would also be help in specs that need to talk generically about the things in List or Dictionary without wanting to have to make up a term of their own.

draft-ietf-httpbis-priority.md
```diff
@@ -328,7 +329,8 @@ responses that have impact on user interaction.
 
 ## Incremental
 
-The incremental parameter (`i`) takes an sf-boolean as the value that indicates
+The incremental parameter (`i`) takes a Boolean (see
+{{Section 3.3.6 of STRUCTURED-FIELDS}}) as the value that indicates
```
ioggstream, line 333:
```suggestion
The incremental parameter (`i`) is a Boolean (see
{{Section 3.3.6 of STRUCTURED-FIELDS}}) that indicates
```

PR #1978, opened 2022-02-17
priority: update author name

LPardue:
Note: this is a test of the GitHub CI as much as anything.

kramdown-rfc2629 v 1.5.26 added the ability for non-ascii names in markdown that don't generate XML that xml2rfc would reject.

If you try this locally and see issues, try to update your tools at https://github.com/martinthomson/i-d-template/blob/main/doc/SETUP.md#update

### DIFF
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -25,7 +25,7 @@ github-issue-label: priorities
 
 author:
   -
-    name:
+    fullname:
       :: 奥一穂
       ascii: Kazuho Oku
     org: Fastly
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -25,8 +25,9 @@ github-issue-label: priorities
 
 author:
   -
-    ins: K. Oku
-    name: Kazuho Oku
+    name:
+      :: 奥一穂
+      ascii: Kazuho Oku
     org: Fastly
     email: kazuhooku@gmail.com
   -


LPardue:
CI error

```
>draft-ietf-httpbis-priority.xml
/usr/lib/ruby/gems/3.0.0/gems/kramdown-rfc2629-1.5.25/lib/kramdown-rfc/refxml.rb:53:in `authorps_from_hash': undefined method `split' for {":"=>"奥一穂", "ascii"=>"Kazuho Oku"}:Hash (NoMethodError)
```

How do we update it?

draft-ietf-httpbis-priority.md
```diff
@@ -24,8 +24,9 @@ github-issue-label: priorities
 
 author:
   -
-    ins: K. Oku
-    name: Kazuho Oku
+    name:
```
martinthomson, line 27: Try:

```suggestion
    fullname:
```

To suppress the splitting stuff that kramdown does.

LPardue:
after your suggestion is applied

```
 xml2rfc -q -s 'Setting consensus="true" for IETF STD document' --rfc-base-url https://www.rfc-editor.org/rfc/ --id-base-url https://datatracker.ietf.org/doc/html/ --cache=/root/.cache/xml2rfc draft-ietf-httpbis-priority.xml -o draft-ietf-httpbis-priority.txt --text --no-pagination
/github/workspace/draft-ietf-httpbis-priority.xml(15): Warning: Expected a valid submissionType (stream) setting, one of IETF, IAB, IRTF, independent, but found None.  Will use 'IETF'
/github/workspace/draft-ietf-httpbis-priority.xml(20): Error: Expected an asciiFullname= attribute to match the unicode fullname value
Not creating output file due to errors (see above)
``` 

still need an updated kramdown :D

martinthomson:
I was able to get something sensible with:

```diff
 author:
   -
-    ins: K. Oku
-    name: Kazuho Oku
+    fullname: 奥一穂
+    asciiFullname: Kazuho Oku
```

But I needed to update kramdown-rfc2629.  I will poke the CI image build.

LPardue:
any of these variants same to work exactly the same for me locally as long at kramdown is up to date

martinthomson:
It has been 14 days since the CI image was updated (we're probably less than 24h from the next scheduled update).  It's just finished build a fresh copy, so try again.

LPardue:
preview: https://httpwg.org/http-extensions/lucas/non-ascii-author/draft-ietf-httpbis-priority.html

kazuho (approved):
@LPardue @martinthomson Thank you so much for all of these. I do not know how to express my thanks, this is just awesome.

martinthomson:
Very happy to do it.  I consider it a basic human right to be able to be identified as you choose.  Work to accommodate is effort well spent.  I only hope that we've made it easier for the next person.

@hildjj should be happy to learn that his efforts are _finally_ paying out.

LPardue:
> Very happy to do it. I consider it a basic human right to be able to be identified as you choose. Work to accommodate that right is effort well spent. I only hope that we've made it easier for the next person.

> @hildjj should be happy to learn that his efforts are finally paying out.

100% agree, people should feel like they can contribute to the IETF and be their real selves. I'm glad we can take a step closer towards making that a reality. Still more to go.

DiamondG3m (approved):

PR #1985, opened 2022-02-19
message-signatures: fix typo in ed25519-test-key name

bpowers:
The sample request (and elsewhere) use `test-key-ed25519` (one `d` not two) -- I think this was just a small type to fix!


jricher (approved):
Good catch, thanks! Since this is just in the descriptions, the examples don't need to be re run.

PR #1993, opened 2022-03-04
Rename to Repr-Digest

LPardue:
Fixes \1966

- Rename Representation-Digest to Repr-Digest
- Remove line wrapping on simple Repr-Digest examples


ioggstream (approved):
If we want, we can reopen this issue at the next interim.

PR #1994, opened 2022-03-06
Release signature-09

### LINKED ISSUE #1192: Transformations to Via header field value break verification
Intermediaries are permitted to strip comments from the `Via` header field value, and consolidate related sequences of entries.  The canonicalization rules do not account for these changes, and thus they cause signature verification to fail if the `Via` header is signed. At the very least, guidance on signing or not signing `Via` headers needs to be included.

### LINKED ISSUE #1193: Case changes to case-insensitive header field values break verification
Some header field values are case-insensitive, in whole or in part. The canonicalization rules do not account for this, thus a case change to a covered header field value causes verification to fail.

jricher:
Includes some new text and several small cleanups.

Closes #1192
Closes #1193
﻿

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -245,6 +245,8 @@ If the combined value is not available for a given header, the following algorit
 
 The resulting string is the canonicalized component value.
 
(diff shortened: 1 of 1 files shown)


reschke:
did you forget to merge it?

jricher:
Sorry, there was ice cream cake and I got distracted

PR #1996, opened 2022-03-09
Pseudo-fields aren't allowed in trailers

### LINKED ISSUE #1930: Pseudo-fields that are defined by protocol extensions MAY be included.
S 3.6 says that extension pseudo header MAY be included but they have to appear before other field lines.

An easy detail to miss  buried in HTTP/2 is the requirement

> Trailers MUST NOT include [pseudo-header fields](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#PseudoHeaderFields) ([Section 8.3](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#PseudoHeaderFields)). An endpoint that receives pseudo-header fields in trailers MUST treat the request or response as [malformed](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#malformed) ([Section 8.1.1](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#malformed)).

does that also apply here?

martinthomson:
Closes #1930.


LPardue (approved):
LGTM thanks

PR #1997, opened 2022-03-09
Acknowledge David S

martinthomson:
@chris-wood WDYT?

### DIFF
--- a/draft-ietf-httpbis-binary-message.md
+++ b/draft-ietf-httpbis-binary-message.md
@@ -688,5 +688,5 @@ information in {{media-type}} for the media type "message/bhttp".
 # Acknowledgments
 {:numbered="false"}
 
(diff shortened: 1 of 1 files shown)


chris-wood (approved):

DavidSchinazi (approved):
👋

DavidSchinazi:
@martinthomson out of curiosity what does this `{{{Person Name}}}` syntax do? The finished product looks the same and I couldn't find documentation for it in kramdown-rfc

martinthomson:
That produces `<xref fullname="Person Name"/>`, which ensures that there is a smaller diff when the RPC gets their hands on the document later.  What is really cool is that you can do `{{{ǝɯɐN uosɹǝԀ}}}{: asciiFullname="Person Name"}` and a real name can be shown (with an ASCII translation).  Or, if you leave off the attribute, you can just put arbitrary unicode in the draft...

DavidSchinazi:
Thanks!

PR #1999, opened 2022-03-14
Use singular `@query-param` instead of `@query-params`

### LINKED ISSUE #1998: [signatures] Singular `@query-param` instead of `@query-params`
If it wasn't your thinking already, I'd like to suggest that the `@query-params` derived component be the singular `@query-param` instead.

When canonicalizing individual query parameters, each component identifier references a single query parameter name/value pair. The use of singular `@query-param` feels more appropriate to me:

```
"@query-param";name="foo": bar
```

In the rare case where there are multiple query parameters with the same name, each one's canonical representation is still referring to a single name/value pair, so the singular still feels more appropriate:

```
"@query-param";name="baz": bar1
"@query-param";name="baz": bar2
```

When used in the list of covered components, we still include the `name` parameter, so each component identifier is *usually* [1] referring to a single query parameter name/value pair:

```
"@query-param";name="foo" "@query-param";name="qux"
```

[1]: However, in that rare case of multiple query parameters having the same name, one instance of `"@query-param";name="baz"` would cover all instances of that repeated query parameter named `baz`. Perhaps the plural `@query-params` does make more sense for this scenario.

But if there must be a singular/plural mismatch somewhere, I'd rather see it in the more rare and exceptional case. So I would propose using the name `@query-param` instead of `@query-params`.

In [http_sig_examples.py](https://github.com/bspk/sig-example-scripts/blob/main/http_sig_examples.py), I see that a recent addition implemented it as the singular `@query-param`. I wonder if drafters already had this change in mind or if it might have been an oversight that organically tended toward what feels more natural.

tyler-ham:
Closes #1998

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -385,8 +385,8 @@ This specification defines the following derived component identifiers:
 @query
 : The query portion of the target URI for a request. ({{content-request-query}})
 
-@query-params
-: The parsed query parameters of the target URI for a request. ({{content-request-query-params}})
+@query-param
+: A parsed query parameter of the target URI for a request. ({{content-request-query-param}})
 
 @status
 : The status code for a response. ({{content-status-code}}).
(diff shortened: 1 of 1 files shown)


jricher (approved):
Looks good to me, will leave for a beat to get feedback from others.

Issue #2023, opened 2022-03-18
[signatures] IANA, references to "token endpoint authorization method"


tyler-ham:
There are several references to "token endpoint authorization method" in section 6. For example,

> Reference to the document(s) that specify the **token endpoint authorization method**, preferably including a URI...

I suspect these were copied from RFC 7662 but not yet updated for the new context.

* Section 6.1.1 (HTTP Signature Algorithms Registry - Registration Template): suggest "algorithm" instead of "token endpoint authorization method"
* Section 6.2.1 (HTTP Signature Metadata Parameters Registry - Registration Template): suggest "parameter" instead of "token endpoint authorization method"
* Section 6.3.1 (HTTP Signature Derived Component Identifier Registry - Registration Template): suggest "derived component" instead of "token endpoint authorization method"

PR #2024, opened 2022-03-18
Fix references to "token endpoint authorization method"

tyler-ham:
For #2023


jricher:
Thanks, copy-paste error from where I pulled the boilerplate from.

jricher (approved):

PR #2049, opened 2022-03-29
message-signatures: fix Content-Digest sha-512 digest on response

bobby-stripe:
I'm not sure what the current digest is, but it doesn't match the 23-byte HTTP body of that message.
We can verify the expected/new digest with:

```
$ echo -n '{"message": "good dog"}' | sha512sum | cut -d ' ' -f 1 | xxd -r -p | base64
mEWXIS7MaLRuGgxOBdODa3xqM1XdEvxoYhvlCFJ41QJgJc4GTsPp29l5oGX69wWdXymyU0rjJuahq4l5aGgfLQ==
```

(also verified in a go implementation of message signatures)

As the digest changed, the Signature for the ECDSA P-256 example also changed, I've included a new signature over the updated digest that verifies based on the key in the spec.

Thanks!

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -2069,8 +2069,8 @@ NOTE: '\' line wrapping per RFC 8792
 
 "@status": 200
 "content-type": application/json
-"content-digest": sha-512=:JlEy2bfUz7WrWIjc1qV6KVLpdr/7L5/L4h7Sxvh6\
-  sNHpDQWDCL+GauFQWcZBvVDhiyOnAQsxzZFYwi0wDH+1pw==:
+"content-digest": sha-512=:mEWXIS7MaLRuGgxOBdODa3xqM1XdEvxoYhvlCFJ4\
+  1QJgJc4GTsPp29l5oGX69wWdXymyU0rjJuahq4l5aGgfLQ==:
 "content-length": 23
 "@signature-params": ("@status" "content-type" "content-digest" \
   "content-length");created=1618884473;keyid="test-key-ecc-p256"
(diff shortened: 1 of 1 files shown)


jricher:
@bobby-stripe Thanks, the signature examples are generated from a script, so it should be patched there first or else any errors might creep back in with future updates. These seem to be the offending lines:

https://github.com/bspk/sig-example-scripts/blob/main/http_sig_examples.py#L185-L191

We should probably be calculating the digest value within the appropriate section of the script here instead of hardcoding it in the source:

https://github.com/bspk/sig-example-scripts/blob/main/http_sig_examples.py#L1063-L1128

Once the generator is patched so that it gives the right values, we can take in this PR.

yaronf:
Both changes validate for me. Thanks!

bobby-stripe:
thanks @jricher !  added a PR there: https://github.com/bspk/sig-example-scripts/pull/2/files

jricher:
I've updated the example generation package to generate the digest values throughout all the examples, and that's at bspk/sig-example-scripts#3 now. The script now generates the same content digest as the PR here, and this is included in the signature base. 

```
"@status": 200
"content-digest": sha-512=:mEWXIS7MaLRuGgxOBdODa3xqM1XdEvxoYhvlCFJ4\
  1QJgJc4GTsPp29l5oGX69wWdXymyU0rjJuahq4l5aGgfLQ==:
"content-length": 23
"@signature-params": ("@status" "content-type" "content-digest" "co\
  ntent-length");created=1618884473;keyid="test-key-ecc-p256"
```

The changes need to be propagated to httpsig.org as well, but we'll track that separately.

Thanks for finding and fixing this issue with the spec's examples!

PR #2053, opened 2022-03-29
State even more clearly intermediary unfairness

### LINKED ISSUE #2011: Section 13
[Section 13](https://www.rfc-editor.org/authors/rfc9218.html#section-13):  "progress made ... can become different" reads oddly.  Does "can become different" mean "shift" or "change"?

Original:

     When
     HTTP requests are forwarded through intermediaries, progress made by
     each connection originating from end clients can become different
     over time, depending on how intermediaries coalesce or split requests
     into backend connections.

LPardue:
Fixes #2011 

This ended up being larger than I wanted, I'm not a huge fan of that. But I think this scope of change is needed if we want to be crystal clear, even if the words I've proposed need to be tweaked.

Based on @MikeBishop's suggestion on the the issue (thanks!)

### DIFF
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -805,12 +805,14 @@ priorities might use the relative priority of streams when choosing probe data.
 # Fairness {#fairness}
 
 Typically, HTTP implementations depend on the underlying transport to maintain
-fairness between connections competing for bandwidth. When HTTP requests are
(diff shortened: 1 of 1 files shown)


kazuho (approved):
Looks good.

I might propose splitting sentence between what happens without priority signals and what happens with. But I think that's a matter of taste, so I would be fine with what is being proposed.

PR #2058, opened 2022-04-04
Intermediary signal processing/production, again

### LINKED ISSUE #2003: trouble following "that overrides that"
[Section 4](https://www.rfc-editor.org/authors/rfc9218.html#section-4):  We had trouble following "that overrides that".  To what does the second "that" refer - the client's signal or something else?  It is also not clear to us what is performing these actions.

Original:

     Sending a
     PRIORITY_UPDATE frame preserves the signal from the client carried by
     the Priority header field, but provides a signal that overrides that
     for the next hop; see [Section 14](https://www.rfc-editor.org/authors/rfc9218.html#section-14)

Possibly (highlighting what is sending the frames):
 An intermediary can send a
 PRIORITY_UPDATE frame to preserve the signal from the client that is 
 carried by the Priority header field; however, the next hop can send 
 a signal that overrides the client's signal; see [Section 14](https://www.rfc-editor.org/authors/rfc9218.html#section-14)

LPardue:
Closes #2003

### DIFF
--- a/draft-ietf-httpbis-priority.md
+++ b/draft-ietf-httpbis-priority.md
@@ -277,11 +277,15 @@ defined in Sections {{<h2-update-frame}} and {{<h3-update-frame}}. Frames
 transmit priority parameters on a single hop only.
 
 Intermediaries can consume and produce priority signals in a PRIORITY_UPDATE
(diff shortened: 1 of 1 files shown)


kazuho (approved):
This is better. Thank you!

PR #2060, opened 2022-04-04
Try to avoid saying payload where it is not needed

### LINKED ISSUE #2047: Payload of incremental responses are used in parts, or chunks,...
The RFC editor has suggested a grammatical change here that is correct - s/Payload/Payloads. 

However, it got me thinking that we probably want to align with HTTP semantics and use "content" instead,  something more like 

> The content of incremental responses is used as parts, or chunks, are received.

LPardue:
Fixes #2047


mnot (commented):
Looks reasonable from a http terminology perspective

PR #2064, opened 2022-04-06
TAG Security & Privacy Questionnaire Answers for *Cookie Expires/Max-Age attribute upper limit*

arichiv:


mnot:
These sorts of things might live in httpwg/admin or httpwg/wiki -- although I could see an argument for just making it an issue on this repo, rather than a file in it.

PR #2085, opened 2022-05-03
Switched to generic request-response binding

### LINKED ISSUE #1905: Signatures: "Related response" feature
I think the "related response" feature, as in:

>If used in a related-response, the `@path` identifier refers to the associated component value of the request that triggered the response message being signed.

is not a good idea, for two reasons:

* It only applies (and can only apply) to specialty/derived components, and cannot apply to normal headers and query params which people might also want to bind between requests and responses.
* It is subsumed by the `@request-response` feature, and request-response is simpler and more secure. You decide once which request fields need to be signed, and the response is then bound to the entire verified request.

So I would suggest to remove this feature, and allow `@path` and similar to only apply to requests.

jricher:
Closes #1905

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -395,21 +395,15 @@ This specification defines the following derived components:
 @status
 : The status code for a response. ({{content-status-code}}).
 
-@request-response
-: A signature from a request message that resulted in this response message. ({{content-request-response}})
-
 Additional derived component names MAY be defined and registered in the HTTP Signatures Derived Component Name Registry. ({{content-registry}})
 
-Derived components can be applied in one or more of three targets:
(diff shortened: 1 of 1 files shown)


yaronf (approved):
LGTM.

richanna (approved):
LGTM. Simple and powerful. I like it.

jricher:
Thanks, I still need to update the examples (and update the tools that generate the examples) before this gets merged.

yaronf:
The new example *almost* validates: the **reqeust**'s signature doesn't verify in my code.

My signature base:
```
"@method": POST
"@authority": example.com
"@path": /foo
"content-digest": sha-512=:WZDPaVn/7XgHaAy8pmojAkGWoRx2UFChF41A2svX+TaPm+AbwAgBWnrIiYllu7BNNyealdVLvRwEmTHWXvJwew==:
"content-length": 18
"@signature-params": ("@method" "@authority" "@path" "content-digest" "content-length" "content-type");created=1618884475;keyid="test-key-rsa-pss"
```

PR #2087, opened 2022-05-03
Remove note to ignore Domain attribute with trailing .

### LINKED ISSUE #1758: RFC6265bis: Handling a trailing dot in the domain attribute
@miketaylr I vaguely remember us having a discussion related to this from a few weeks ago, but I stumbled across it again today when reading the spec and figured I'd open an issue for it.

There are multiple places in the spec that say:

> Note that a leading %x2E ("."), if present, is ignored even though that character is not permitted, but a trailing %x2E ("."), if present, will cause the user agent to ignore the attribute.

I doesn't look like there's a step in the domain attribute parsing algorithm to actually ignore the attribute if a trailing dot is present, though.  My first thought was maybe this is implicitly handled by the domain-match algorithm.  It's unclear to me what the canonicalized version of `www.example.com.` would be, and whether that'd domain-match with `www.example.com`.  Assuming they don't domain-match, the algorithm states that the whole cookie should be ignored instead of just the attribute being ignored.  Assuming they do domain-match, then the attribute wasn't ignored.

Are we missing a step that would cause the attribute to be ignored in this case?

sbingler:
Closes #1758

The spec had a note to handle trailing `.`s in the `Domain`'s value by ignoring the attribute. Since `example.com` and `example.com.` are considered distinct domains a trailing `.` should instead result in a rejected cookie. By removing this note the spec will rely on the domain matching algorithm to (correctly) reject the cookie.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1465,8 +1465,7 @@ user agent MUST process the cookie as follows:
         attribute-name of "Domain" and an attribute-value whose
         length is no more than 1024 octets. (Note that a leading %x2E
         ("."), if present, is ignored even though that character is not
-        permitted, but a trailing %x2E ("."), if present, will cause
-        the user agent to ignore the attribute.)
+        permitted.)
 
     Otherwise:
 


sbingler:
@annevk PTAL

chlily1 (approved):
LGTM. Thanks for the attention to detail!

annevk (approved):
Thanks!

annevk:
Ah, https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis#section-4.1.2.3 also needs updating. I'll file a new issue I guess.

PR #2094, opened 2022-05-08
Fix references in binary messages

martinthomson:
Make HTTP/* informative references; fix their display.
Also, suppress some noise from xml2rfc.


mnot (commented):
LGTM

PR #2097, opened 2022-05-09
Clarify Integrity preference field meaning in responses

### LINKED ISSUE #1846: Review want-digest meaning
From @jricher WGLC review:

> §5: It's unclear what "Want-Digest" and "Want-Content-Digest"  mean in the context of a response. Does that mean the server wants a digest on one or more future requests? (That seems the most reasonable). The relevant text we have for "Accept-Signature" is currently the following (this itself probably needs additional work):

> 	When the Accept-Signature field is used in an HTTP Response message, the field indicates that the server desires the client to sign its next request to the server with the identified parameters, and the target message is the client's next request. The client can choose to also continue signing future requests to the same server in the same way.

LPardue:
Fixes #1846

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -401,7 +401,7 @@ on messages associated with the request URI and representation metadata, using
 the `Repr-Digest` field.
 
 If `Want-Content-Digest` or `Want-Repr-Digest` are used in a response, it
-indicates that the server  would like the client to provide the respective
+indicates that the server would like the client to provide the respective
 Integrity field on future requests.
 
 `Want-Content-Digest` and `Want-Repr-Digest` are of type `Dictionary`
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -392,14 +392,18 @@ preferences using the
...
 on messages associated with the request URI and representation metadata, using
 the `Content-Digest` field.
 
-`Want-Repr-Digest` indicates the sender's desire to receive a representation digest
+`Want-Repr-Digest` indicates that the sender would like to receive a representation digest
 on messages associated with the request URI and representation metadata, using
 the `Repr-Digest` field.
 
+If `Want-Content-Digest` or `Want-Repr-Digest` are used in a response, it
+indicates that the server  would like the client to provide the respective
```
LPardue, line 404:
```suggestion
indicates that the server would like the client to provide the respective
```

ioggstream (approved):
LGTM

PR #2098, opened 2022-05-10
Mention HTTP Message Signatures as an illustration

LPardue:
Editorial change @jricher and I talked about over lunch in IETF 113.

Initially I was going to include an example of a Signature that covered content-digest, but I stepped back from doing that for a few reasons:

* the sigs draft already has a few examples that contain a content-digest
* I don't want to have to explain all the items or concepts that appear in HTTP signatures examples, that belongs in the sigs draft itself and we shouldn't duplicate it
* we probably aren't running the same infrastructure to make sure the examples stay correct

What might work is for the sigs doc to provide some slimmed-down examples that focus on signing `Content-Digest` and `Repr-Digest` and then we could reference them from this draft. But I'm open to other ideas.

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -80,6 +80,7 @@ informative:
   PATCH: RFC5789
   NO-MD5: RFC6151
   NO-SHA: RFC6194
+  SIGNATURES: I-D.ietf-httpbis-message-signatures
 
 --- abstract
 
@@ -180,7 +181,9 @@ different digest values when transferred with HTTP.
 Integrity fields do not provide integrity for
 HTTP messages or fields. However, they can be combined with other mechanisms that
 protect metadata, such as digital signatures, in order to protect
-the phases of an HTTP exchange in whole or in part.
+the phases of an HTTP exchange in whole or in part. For example, HTTP Message
+Signatures {{SIGNATURES}} could be used to sign Integrity fields, thus providing
+coverage for HTTP content or representation data.
 
 This specification does not define means for authentication, authorization or privacy.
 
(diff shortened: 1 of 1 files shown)


jricher:
I think an example of a signed request or response with Content-Digest would be reasonable, and it honestly doesn't have to go into explaining how the signatures work against the digest, just showing the digest in the signed components list would be sufficient.

```
HTTP/1.1 200 OK
Content-Digest: \
  sha-256=:X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=:
Signature-Input: cdsigex=("content-digest" "content-type" "@status")
Signature: cdsigex=:7ajtA8nCss0JpwEjvSzvwShEC1csS8pquZJ0vrAnM\
  2mRw+lEropP5Ea/+5tx9gAOZDzuU0talUe357Yf3VG/2g==:

{"hello": "world"}
```

This doesn't even have parameters like key and alg, but can be introduced with "This example shows covering the content digest field in a response using HTTP Message Signatures [reference here]", and I think that's probably good enough. Yes, it's not directly verifiable from the digest spec, but doing so would require a lot more information than is necessary -- this isn't about showing how signatures work, it's about showing how the signatures cover the digest. I think just showing the Content-Digest in the signature's content is enough to make the point.

ioggstream:
Agree with mentioning signatures without examples.

Signatures i-d is the correct place for examples containing a signed *-digest field.

LPardue:
Thanks for the example Justin. To me, it raises more questions than it answers though. Primarily because it's not clear what the example shows unless we describe to some extent Signature-Input and the process of signature generation. If the argument is to just give an example of how to include a field in the Signature-Input, that seems so trivial I'm not sure the value. 

Instead, my thinking was we could modify https://github.com/httpwg/http-extensions/pull/2098/files#diff-de34152d297a9eb29b85b24962949ba2e309478a5c32e1d94fd4c3ee27b4584dR508 to deep link into the Signature's draft, for example saying something like 

> one possible approach is to include them as part of HTTP Message Signatures, as shown in {{Section X.Y of SIGNATURES}}.

LPardue:
Going to merge this as is, any possible future changes can be one additively without massive change.

PR #2099, opened 2022-05-11
s/a a/a

LPardue:
Typi fix


martinthomson (approved):
I had to read that three times to spot your change.  Something about that text blinded me to that.

PR #2105, opened 2022-05-20
Document Style updates for Signatures

jricher:
This updates the document to comply with the style guides:

- clean up structured field references and remove ABNF
- format field names per style guide
- update semantics and messaging references
- udpate self-references
- update registry tables

This does not (yet) modify the registry initial contents from tables to dictionary lists because I'm not sure how to format those for adding multiple 
entries at once.

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -113,7 +113,7 @@ This document also provides a mechanism for a potential verifier to signal to a 
 
 HTTP permits and sometimes requires intermediaries to transform messages in a variety of ways.  This may result in a recipient receiving a message that is not bitwise equivalent to the message that was originally sent.  In such a case, the recipient will be unable to verify a signature over the raw bytes of the sender's HTTP message, as verifying digital signatures or MACs requires both signer and verifier to have the exact same signature base.  Since the exact raw bytes of the message cannot be relied upon as a reliable source for a signature base, the signer and verifier must independently create the signature base from their respective versions of the message, via a mechanism that is resilient to safe changes that do not alter the meaning of the message.
 
-For a variety of reasons, it is impractical to strictly define what constitutes a safe change versus an unsafe one.  Applications use HTTP in a wide variety of ways, and may disagree on whether a particular piece of information in a message (e.g., the body, or the getDate` header field) is relevant.  Thus a general purpose solution must provide signers with some degree of control over which message components are signed.
+For a variety of reasons, it is impractical to strictly define what constitutes a safe change versus an unsafe one.  Applications use HTTP in a wide variety of ways, and may disagree on whether a particular piece of information in a message (e.g., the body, or the Date header field) is relevant.  Thus a general purpose solution must provide signers with some degree of control over which message components are signed.
 
 HTTP applications may be running in environments that do not provide complete access to or control over HTTP messages (such as a web browser's JavaScript environment), or may be using libraries that abstract away the details of the protocol (such as [the Java HTTPClient library](https://openjdk.java.net/groups/net/httpclient/intro.html)).  These applications need to be able to generate and verify signatures despite incomplete knowledge of the HTTP message.
 
@@ -341,7 +341,7 @@ The resulting string is used as the component value in {{http-header}}.
 
 If a given field is known by the application to be a Dictionary structured field, an individual member in the value of that Dictionary is identified by using the parameter `key` and the Dictionary member key as a String value.
 
-An individual `member_value` of a Dictionary Structured Field is canonicalized by applying the serialization algorithm described in {{Section 4.1.2 of STRUCTURED-FIELDS}} on the `member_value` and its parameters, without the dictionary key. Specifically, the value is serialized as an Item or Inner List (the two possible values of a Dictionary member).
+An individual member value of a Dictionary Structured Field is canonicalized by applying the serialization algorithm described in {{Section 4.1.2 of STRUCTURED-FIELDS}} on the `member_value` and its parameters, without the dictionary key. Specifically, the value is serialized as an Item or Inner List (the two possible values of a Dictionary member).
 
 Each parameterized key for a given field MUST NOT appear more than once in the signature base. Parameterized keys MAY appear in any order in the signature base, regardless of the order they occur in the source Dictionary.
 
@@ -449,7 +449,7 @@ The signature parameters component value is serialized as a parameterized inner 
 6. The output contains the signature parameters component value.
 
 Note that the Inner List serialization from {{Section 4.1.1.1 of STRUCTURED-FIELDS}} is used for the covered component value instead of the List serialization from {{Section 4.1.1 of STRUCTURED-FIELDS}}
-in order to facilitate parallelism with this value's inclusion the Signature-Input field,
+in order to facilitate parallelism with this value's inclusion in the Signature-Input field,
 as discussed in {{signature-input-header}}.
 
 This example shows a canonicalized value for the parameters of a given signature:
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-message-signatures.md
```diff
@@ -1396,22 +1402,21 @@ Signature:  sig1=:LAH8BjcfcOcLojiuOBFWn0P5keD3xAOuJRGziCLuD8r5MW9S0\
...
+While a signer is free to attach a signature to a request or response without prompting, it is often desirable for a potential verifier to signal that it expects a signature from a potential signer using the Accept-Signature field.
 
-When the `Accept-Signature` field is used in an HTTP Response message, the field indicates that the server desires the client to sign its next request to the server with the identified parameters, and the target message is the client's next request. The client can choose to also continue signing future requests to the same server in the same way.
+The message to which the requested signature is applied is known as the "target message". When the Accept-Signature field is sent in an HTTP request message, the field indicates that the client desires the server to sign the response using the identified parameters, and the target message is the response to this request. All responses from resources that support such signature negotiation SHOULD either be uncacheable or contain a Vary header field that lists Accept-Signature, in order to prevent a cache from returning a response with a signature intended for a different request.
```
yaronf, line 1409: I think the SHOULD should be a MUST if this is a security concern.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -423,26 +430,26 @@ The signature parameters component name is `@signature-params`. This message com
...
 4. Determine an order for any signature parameters. Once this order is chosen, it cannot be changed.
-5. Append the parameters to the `inner-list` in the chosen order according to {{Section 4.1.1.2 of RFC8941}},
+5. Append the parameters to the `inner-list` in the chosen order according to {{Section 4.1.1.2 of STRUCTURED-FIELDS}},
     skipping parameters that are not available or not used for this message signature.
 6. The output contains the signature parameters component value.
 
-Note that the `inner-list` serialization is used for the covered component value instead of the `sf-list` serialization
-in order to facilitate this value's inclusion in message fields such as the `Signature-Input` field's dictionary,
+Note that the Inner List serialization from {{Section 4.1.1.1 of STRUCTURED-FIELDS}} is used for the covered component value instead of the List serialization from {{Section 4.1.1 of STRUCTURED-FIELDS}}
+in order to facilitate parallelism with this value's inclusion the Signature-Input field,
```
richanna, line 452:
```suggestion
in order to facilitate parallelism with this value's inclusion in the Signature-Input field,
```

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1537,20 +1546,54 @@ Specification document(s):
...
+|`@query`| The full query of the request URI | Active | Request | {{content-request-query}} of {{&SELF}}|
+|`@query-param`| | A single named query parameter | Active | Request | {{content-request-query-param}} of {{&SELF}}|
+|`@status`| The status code of the response | Active | Response | {{content-status-code}} of {{&SELF}}|
 {: title="Initial contents of the HTTP Signature Derived Component Names Registry." }
 
+## HTTP Signature Component Parameters Registry {#component-param-registry}
+
+This document defines a component identifiers, which can be parameterized in specific circumstances to provide unique modified behavior. IANA is asked to create and maintain a new registry typed "HTTP Signature Component Parameters" to record and maintain the set of non-field component names and the methods to produce their associated component values. Initial values for this registry are given in {{iana-component-param-contents}}.  Future assignments and modifications to existing assignments are to be made through the Expert Review registration policy {{?RFC8126}} and shall follow the template presented in {{iana-component-param-template}}.
```
richanna, line 1565:
```suggestion
This document defines component identifiers, which can be parameterized in specific circumstances to provide unique modified behavior. IANA is asked to create and maintain a new registry typed "HTTP Signature Component Parameters" to record and maintain the set of parameters for component identifiers, and the methods to produce their associated parameter values. Initial values for this registry are given in {{iana-component-param-contents}}.  Future assignments and modifications to existing assignments are to be made through the Expert Review registration policy {{?RFC8126}} and shall follow the template presented in {{iana-component-param-template}}.
```

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1537,20 +1546,54 @@ Specification document(s):
...
+
+### Registration Template {#iana-component-param-template}
+
+{: vspace="0"}
+Name:
+: A name for the parameter. The name MUST be an ASCII string consisting only of lower-case characters (`"a"` - `"z"`), digits (`"0"` - `"9"`), and hyphens (`"-"`), and SHOULD NOT exceed 20 characters in length.  The name MUST be unique within the context of the registry.
```
richanna, line 1571: Can we reference Structured Fields here, instead of specifying the allowed characters?

draft-ietf-httpbis-message-signatures.md
jricher: @mnot I believe you had a reason that this can't be a MUST, right?

draft-ietf-httpbis-message-signatures.md
jricher: This is actually a smaller list than what's allowed for the `key` production rule in structured fields. I don't remember if there's a compelling reason for this though.

draft-ietf-httpbis-message-signatures.md
mnot: Is it a security concern or an interoperability one?

I'm not strongly against MUST, but wonder why it's a requirement at all - mentioning caching is fine, but making it a requirement is akin to saying 'Messages MUST be well-formed'.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1231,14 +1237,14 @@ JWA algorithm values from the JSON Web Signature and Encryption Algorithms Regis
...
 
-An HTTP message signature MUST use both fields containing the same labels:
-the `Signature` HTTP field contains the signature value, while the `Signature-Input` HTTP field identifies the covered components and parameters that describe how the signature was generated. Each field contains labeled values and MAY contain multiple labeled values, where the labels determine the correlation between the `Signature` and `Signature-Input` fields.
+An HTTP message signature MUST use both fields and each field MUST contain the same labels.
+The Signature-Input field identifies the covered components and parameters that describe how the signature was generated, while the Signature field contains the signature value. Each field contains labeled values and MAY contain multiple labeled values. The result of this constraint is that the Signature-Input and Signature Dictionaries are parallel data structures of each other. Any key found in one field but not in the other is an error.
```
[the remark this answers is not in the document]
jricher: The constraint is that a signature uses both fields. The goal of the language change here is to show that there's a 1:1 mapping, and anything else is an error in the signature headers.

draft-ietf-httpbis-message-signatures.md
jricher: I do think there are some security implications -- if I'm requesting a signed response but I'm getting a cache hit for the signature then I'm going to have a higher level of confidence in the incorrect result than I would have otherwise. It's only a security risk if the cached signature is valid for a different message -- but that's more a case of the signature not covering an appropriately large amount of the message.

But I think this can even be addressed by recommending use of the `created` field on the signature parameters and letting the validator figure it out. I'm fine with it being listed as a consideration but removing the normative requirement here. I'll be the first to admit I'm not an expert on cache semantics and will defer to those who are.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1537,20 +1546,54 @@ Specification document(s):
...
+### Registration Template {#iana-component-param-template}
+
+{: vspace="0"}
+Name:
+: A name for the parameter. The name MUST be an ASCII string consisting only of lower-case characters (`"a"` - `"z"`), digits (`"0"` - `"9"`), and hyphens (`"-"`), and SHOULD NOT exceed 20 characters in length.  The name MUST be unique within the context of the registry.
+
+Description:
+: A description of the parameter's function.
+
+Target:
```
jricher, line 1576:
```suggestion
Status:
: A brief text description of the status of the parameter. The description MUST begin with one of "Active" or "Deprecated", and MAY provide further context or explanation as to the reason for the status.

Target:
```

draft-ietf-httpbis-message-signatures.md
jricher: Ugh, copy-paste mistakes here. Changing it to this instead:

```suggestion
This document defines several kinds of component identifiers, some of which can be parameterized in specific circumstances to provide unique modified behavior. IANA is asked to create and maintain a new registry typed "HTTP Signature Component Parameters" to record and maintain the set of parameters names, the component identifiers they are associated with, and the modifications these parameters make to the component value. Initial values for this registry are given in {{iana-component-param-contents}}.  Future assignments and modifications to existing assignments are to be made through the Expert Review registration policy {{?RFC8126}} and shall follow the template presented in {{iana-component-param-template}}.
```

draft-ietf-httpbis-message-signatures.md
richanna: Looking again, it actually doesn't comply with the requirements for S-F parameter keys, since S-F does not allow parameter keys to start with numbers or hyphens. https://www.rfc-editor.org/rfc/rfc8941#section-3.1.2-3

I think that's a solid argument for referencing S-F here.

draft-ietf-httpbis-message-signatures.md
jricher: Left this text as is, since the text in question was not actually changed from a previous version. We can potentially add a recommendation for how to handle cache in a future PR.

PR #2129, opened 2022-05-26
Editorial tweaks of appx D

martinthomson:
The phrase "uncertainty of doubt" jumped out in this text as probably not being what was intended.

I fixed that and made just a few tweaks to adjacent text.


ioggstream (approved):
LGTM

PR #2132, opened 2022-05-26
reword considered a representation

martinthomson:
I had to read this several times to understand what parts of the sentence were replicated into the
"or" part of the statement.  Duplicating some words makes it a tiny bit more comprehensible, I
think.


ioggstream (approved):
LGTM

PR #2140, opened 2022-05-30
Some changes inspired by James Gruessing

martinthomson:
Moving normative language from Section 6 is the most useful, but there are a bunch of other stuff I noticed as I was reading through trying to implement James' other suggestion.


LPardue (approved):
lgtm

PR #2143, opened 2022-06-01
Clean up signature references and text

### LINKED ISSUE #1183: Multiple Set-Cookie headers are not well supported
The Set-Cookie header can occur multiple times but does not adhere to the list syntax, and thus is not well supported by the header field value concatenation rules.

### LINKED ISSUE #1981: [signatures] Include example of "@query-params" in list of covered components
When composing a list of covered components in the `@signature-params` value, it wasn't clear to me if there should be multiple instances of `"@query-params"`, one for each distinct name.

In [Signature Parameters, 3.](https://github.com/httpwg/http-extensions/blob/main/draft-ietf-httpbis-message-signatures.md#signature-parameters-signature-params), I see

> Serialize the component identifiers of the covered components, including all parameters...

And in the [HTTP Message Components](https://github.com/httpwg/http-extensions/blob/main/draft-ietf-httpbis-message-signatures.md#http-message-components-covered-content) section:

> Component identifiers, including component identifiers with parameters, MUST NOT be repeated within a single list of covered components. Component identifiers with different parameter values MAY be repeated within a single list of covered components.

Both of these sound to me like `"@query-params"` should be repeated as shown below, once for each distinct `name` parameter. For example:

```

("@path" "@authority" "@query-params";name="baz" \
  "@query-params";name="qux" "@query-params";name="param")\
  ;keyid="test-key-rsa-pss";alg="rsa-pss-sha512";\
  created=1618884475;expires=1618884775
```

Is this correct?

If so, would you be open to expanding at least one example to show that use of `"@query-params"` in a list of covered components?

### LINKED ISSUE #2135: Formatting of italicized items in plain text version
In sec. 2.2.9 there are some spurious underscores around component names.

cc: @jricher

### LINKED ISSUE #2137: message-signatures: broken deep links
WARNING: apparently dangling reference to 5.7.2 of HTTP1
WARNING: apparently dangling reference to 5.7.1 of HTTP1
WARNING: apparently dangling reference to 5.7.1 of HTTP1
WARNING: apparently dangling reference to 5.7.1 of HTTP1
WARNING: apparently dangling reference to 5.5 of HTTP1
WARNING: apparently dangling reference to 5.5 of HTTP1
WARNING: apparently dangling reference to 5.5 of HTTP1
WARNING: apparently dangling reference to 5.5 of HTTP1

### LINKED ISSUE #2138: message-signatures: ABNF ref missing
The document uses ABNF, but doesn't cite RFC 5234.

jricher:
Multiple editorial and example cleanups:

- move italics, closes #2135
- clean up ABNF, closes #2138
- fix dangling references, closes #2137
- update request-response example
- add query-param example, closes #1981
- call out weirdness with Set-Cookie, closes #1183


draft-ietf-httpbis-message-signatures.md
```diff
@@ -147,21 +148,23 @@ Additionally, all changes to components not covered by the signature are conside
...
+This document defines several string constructions using {{ABNF}} and uses the following ABNF rules: `VCHAR`, `SP`, `DQUOTE`, `LF`. This document also uses the following ABNF rules from {{STRUCTURED-FIELDS}}: `sf-string`, `inner-list`, `parameters`. This document also uses the following ABNF rules from {{HTTP}}: `field-content`.
 
 In addition to those listed above, this document uses the following terms:
 
 {: vspace="0"}
 HTTP Message Signature:
 : A digital signature or keyed MAC that covers one or more portions of an HTTP message. Note that a given HTTP Message can contain multiple HTTP Message Signatures.
 
-Signer:
+Signer://
```
jricher, line 167:
```suggestion
Signer:
```

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1790,6 +1793,20 @@ From here, the signing process proceeds as usual.
...
 
+## Non-List Field Values {#security-non-list}
+
+When an HTTP field occurs multiple times in a single message, these values need to be combined into a single one-line string value to be included in the HTTP signature base, as described in {{create-sig-input}}. Not all HTTP fields can be combined into a single value and still be a valid value for the field. However, for the purposes of generating the signature base, the message component value is never read back out of the signature base string and used in anyway. Therefore it is considered best practice to treat the signature base generation algorithm separately from processing the field values by the application, particularly for fields that are known to have this property.
```
martinthomson, line 1798: "and used in anyway" ?

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1790,6 +1793,20 @@ From here, the signing process proceeds as usual.
...
+
+For example, the Set-Cookie field {{COOKIE}} defines an internal syntax that does not conform to the List syntax in {{STRUCTURED-FIELDS}}. This field is also typically sent as multiple fields with distinct values when sending multiple cookies. When multiple Set-Cookie fields are sent in the same message, it is not possible to combine these into a single and be able to parse and use the results, as discussed in {{HTTP, Section 5.3}}. Therefore, all the cookies need to be processed from their separate header values, without being combined, while the signature base needs to be processed from the special combined value generated solely for this purpose.
```
martinthomson, line 1800: This misses the attack.

Say you have:

```
Set-Cookie: that, happens, to, contain, a, comma
```

An attacker is now able to construct a variant message that produces the same signature:

```
Set-Cookie: that, happens
Set-Cookie: to, contain, a, comma
```

This will be processed differently, but it will be signed and treated as such.

For me, this is an unacceptable property of a signature scheme.

martinthomson (changes requested):

draft-ietf-httpbis-message-signatures.md
jricher: Thanks, I didn't call out the commas specifically here but I'll tweak the text to address that.

PR #2147, opened 2022-06-08
Mnot/2082

### LINKED ISSUE #2082: Cookie value aliasing
Should we recognise e.g., binary content in cookie values? If so, how?

mnot:
Fixes #2082.


mnot:
... and apparently some other stuff.

PR #2156, opened 2022-06-09
Use the HTTP core RFCs. Twiddle with some other refs.

### LINKED ISSUE #2151: update digest to reference new http core
## I expect

- digest fields to reference http core

LPardue:
All very straightforward, except the reference to obsoleted RFC 2818. mnot suggest we need to point to RFC 9110 but the intent of the text is speaking about TLS records, so lets just point at TLS directly.

Closes #2151


ioggstream (approved):
LGTM

PR #2160, opened 2022-06-14
Tweaks to list intro and item 2

martinthomson:
This should make the whole thing less janky.


rgwilton:
Changes look good to me.  Regards, Rob

PR #2165, opened 2022-06-17
RFC6265bis: Remove note regarding cookie serialization

### LINKED ISSUE #1502: Consider better specifying document.cookie serialization
In #1428 Anne mentions that the current text that discusses cookie serialization isn't quite right.

The current language:
> NOTE: Despite its name, the cookie-string is actually a sequence of octets, not
> a sequence of characters.  To convert the cookie-string (or components
> thereof) into a sequence of characters (e.g., for presentation to the user),
> the user agent might wish to try using the UTF-8 character encoding {{RFC3629}}
> to decode the octet sequence. This decoding might fail, however, because not
> every sequence of octets is valid UTF-8.

Anne says: "This also does not match what actually happens. Only Chrome can have such bytes at the moment and Chrome drops the cookies it cannot serialize, but does serialize the others."

sbingler:
Closes #1502

During the [Sept 2021 Interim meeting](https://notes.ietf.org/notes-httpbis-21-09#Issue-1502---Better-specify-serialization) it was decided that this note should be removed and that another spec (perhaps HTML) can handle it.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -140,7 +140,6 @@ informative:
   RFC2818:
   RFC3986:
   RFC6265:
-  RFC3629:
   RFC4648:
   RFC3864:
   RFC5895:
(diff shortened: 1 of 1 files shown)


sbingler:
FYI @annevk

sbingler:
Ping @mikewest

mikewest:
Hi! Sorry, I'm not sure why I missed this. Looking now. :)

PR #2167, opened 2022-06-17
RFC6265bis: Add case insensitivity note to Set-Cookie Syntax

### LINKED ISSUE #1881: Set-Cookie attribute name case sensitivity
The grammar implies that set-cookie attribute names are case insensitive (because ABNF strings are), and the parsing algorithms say that they're case-insensitive, but I suspect many people may not realise this, because they're so often seen in CamelCase. 

Could we add a small statement to the Set-Cookie syntax section pointing this out?

sbingler:
Closes #1881

Cookie attribute names are case insensitive, this PR adds a note saying as much.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -522,10 +522,9 @@ Note that some of the grammatical terms above reference documents that use
 different grammatical notations than this document (which uses ABNF from
 {{RFC5234}}).
 
-NOTE: The name of an attribute-value pair is not case sensitive (which differs
-from ABNF). So while they are presented here in CamelCase, such as "HttpOnly"
-or "SameSite", any case is accepted. E.x.: "httponly", "Httponly", "hTTPoNLY",
-etc.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -522,6 +522,11 @@ Note that some of the grammatical terms above reference documents that use
 different grammatical notations than this document (which uses ABNF from
 {{RFC5234}}).
 
+NOTE: The name of an attribute-value pair is not case sensitive (which differs
+from ABNF). So while they are presented here in CamelCase, such as "HttpOnly"
```
martinthomson, line 526: I thought that AbNf strings were cafe insensitive, which is why RFC 7405 exists...

sbingler: Yep, you're right! My mistake, I must have misread it.

Khodam2023ak (approved):

mikewest (approved):

PR #2170, opened 2022-06-21
Introduce and use the Date type

### LINKED ISSUE #2162: Date format
Some have pushed back on using integers for dates, see eg https://github.com/ietf-wg-httpapi/deprecation-header/issues/11

As I suggested there, one accommodation would be to define a new structured type for them. That could be done in retrofit, or separately.

mnot:
Fixes #2162.

### DIFF
--- a/draft-ietf-httpbis-retrofit.md
+++ b/draft-ietf-httpbis-retrofit.md
@@ -4,6 +4,7 @@ abbrev: Retrofit Structured Fields
 docname: draft-ietf-httpbis-retrofit-latest
 date: {DATE}
 category: std
+updates: 8941
 
 ipr: trust200902
 keyword:
@@ -48,6 +49,7 @@ This specification nominates a selection of existing HTTP fields as having synta
 
 To accommodate some additional fields whose syntax is not compatible, it also defines mappings of their semantics into new Structured Fields. It does not specify how to negotiate their use.
 
+One of those mappings requires introduction of a new Structured Fields data type, Date.
 
 --- middle
 
@@ -73,6 +75,8 @@ Note that while implementations can parse and serialise compatible fields as Str
 ## Notational Conventions
 
 {::boilerplate bcp14-tagged}
+
+This document uses the date-time, time-offset, and time-secfrac rules from {{!RFC3339}}.
 
 
 # Compatible Fields {#compatible}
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-retrofit.md
```diff
@@ -291,14 +295,14 @@ Cookie attributes map to Parameters on the Inner List, with the parameter name b
...
-| Expires             | Integer             |
+| Expires             | Date                |
 | Max-Age             | Integer             |
 | Path                | String              |
 | Secure              | Boolean             |
 | SameSite            | Token               |
 {:id="cookie-params" title="Set-Cookie Parameter Types"}
 
-The Expires attribute is mapped to an Integer representation of parsed-cookie-date (see {{Section 5.1.1 of COOKIES}}) expressed as a number of seconds delta from the Unix Epoch (00:00:00 UTC on 1 January 1970, excluding leap seconds).
+The Expires attribute is mapped to an Date representation of parsed-cookie-date (see {{Section 5.1.1 of COOKIES}}).
```
MikeBishop, line 305:
```suggestion
The Expires attribute is mapped to a Date representation of parsed-cookie-date (see {{Section 5.1.1 of COOKIES}}).
```

draft-ietf-httpbis-retrofit.md
```diff
@@ -380,4 +384,39 @@ Finally, add a new column to the "Cookie Attribute Registry" established by {{CO
 --- back
 
 
+# The Date Structured Type {#date-type}
+
+This section defines a new Structured Fields data type, Date.
+
+Dates have a data model that is similar to Integers, representing a (possibly negative) delta in seconds from January 1, 1970 00:00:00 UTC, excluding leap seconds.
```
MikeBishop, line 391: Should this definition (converting to an actual date) be embedded in the serializing/parsing instructions, such that the output is actually a date?  Or perhaps just a note that converting an integer to a date representation will be implementation-dependent.

MikeBishop (approved):
I like this direction; thanks.

draft-ietf-httpbis-retrofit.md
mnot: That would require us to define / choose a model for dates. I'd rather just stay with delta to epoch - it's widely used and understood.

reschke:
I still believe this change should end up in the revision of SF.

mnot:
That's #2225

PR #2184, opened 2022-06-28
Fix #2178. Editorial improvements on examples.

### LINKED ISSUE #2178: digest: editorial remarks on Appendix A
I tried to follow the examples but found them slightly confusing.

> The following examples show how representation metadata, payload transformations and method impacts on the message and content. When the content contains non-printable characters (e.g. when it is compressed) it is shown as a Base64-encoded string.

1. [x] payload -> content, right?
2. [x] I think it would be easier to use hex encoding, because it would make it easier to follow what happens with ranges.

> Now the same content conveys a malformed JSON object, because the request does not indicate a content coding.

3. [x]  Maybe: "removing the content coding from the same request causes the content to become malformed" 

> A Range-Request alters the content, conveying a partial representation.

4. [x]  "alters the content" is technically correct, but can easily be read as somehow affecting the server state. Maybe "affects the transferred message content"

> Figure 5: Partial response from a gzip-encoded representation

5. [x]  To which server state does this refer? The one after executing the PUT in figure 3?

> The method can also alter the content.

6. [x]  (see above)

> Finally, the semantics of an HTTP response might decouple the effective request URI from the enclosed representation.

7. [x] We removed "effective request URI" in RFC 9110, see end of https://www.rfc-editor.org/rfc/rfc9110.html#section-7.1)

ioggstream:
Fix: #2178


ioggstream:
Applied suggestions from https://github.com/httpwg/http-extensions/commit/72daa0d8cdbe533445b07720618fd5adbc62622e

draft-ietf-httpbis-digest-headers.md
```diff
@@ -633,29 +633,49 @@ Content-Type: application/json
...
+0x000x00
 ~~~
-{: title="Request containing a gzip-encoded JSON object"}
+{: title="Request containing a gzip-encoded JSON object" #ex-put-gz}
 
-Now the same content conveys a malformed JSON object, because the request does
-not indicate a content coding.
+Removing the content coding from the same request
+causes the content to become malformed.
+The server replies with an error.
```
LPardue, line 654: I guess we can argue it's not the "same request", because the message is different :smile:, so how about something like

```suggestion
Sending the compressed form of the content 
without the correct content coding means that
the content is malformed.
In this case, the server can reply with an error.
```

PR #2186, opened 2022-06-30
Fix: #2182. 4xx and 5xx are both valid responses.

### LINKED ISSUE #2182: digest: example with code 400 for unsupported digest alg
In [C.3.](https://greenbytes.de/tech/webdav/draft-ietf-httpbis-digest-headers-latest.html#rfc.section.C.3) [Server Does Not Support Client Algorithm and Returns an Error](https://greenbytes.de/tech/webdav/draft-ietf-httpbis-digest-headers-latest.html#server-does-not-support-client-algorithm-and-returns-an-error)...

Not sure whether a 4xx code is appropiate here. The request is not broken in any way; shouldn't this be a 5xx, such as https://www.rfc-editor.org/rfc/rfc9110.html#name-501-not-implemented ?

ioggstream:
## This PR

Fix #2182

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1221,6 +1221,10 @@ Content-Type: application/problem+json
 ~~~
 {: title="Response advertising the supported algorithms"}
 
+Note that a 5xx response would have been legitimate too,
+since this specification does not prescribe any behavior
+on status codes.
+
 # Migrating from RFC 3230
 
 HTTP digests are computed by applying a hashing algorithm to input data.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -1221,6 +1221,10 @@ Content-Type: application/problem+json
 ~~~
 {: title="Response advertising the supported algorithms"}
 
+Note that a 5xx response would have been legitimate too,
+since this specification does not prescribe any behavior
+on status codes.
```
LPardue, line 1226: How about we move this to the end of first paragraph of this section? E.g. at L1196 we can say something like 

"Alternatively a server can also reject
the request and return a response with
error status code such as 4xx or 5xx. 
This specification does not prescribe
any requirement on status code selection;
the follow example illustrates one possible
option."

ioggstream: Ok

PR #2189, opened 2022-06-30
Wrapping problematic fields for signatures

### LINKED ISSUE #2166: Wrapped encoding for problematic fields
Some fields, like `Set-Cookie`, have an internal syntax that allows unquoted commas and messes things up when multiple lines are combined for the signature base, leading to the potential case where two semantically different inputs have the same signature base. This means that:

~~~
Set-Cookie: that, happens
Set-Cookie: to, contain, a, comma
~~~

And:

~~~
Set-Cookie: that, happens, to, contain, a, comma
~~~

Both produce the same signature input line:

~~~
"set-cookie": that, happens, to, contain, a, comma
~~~

Even though the two-line and single-line versions are processed differently by the application.

To counter this, we could have a distinct encoding flag that wraps the field values, similar to `sf`. This can be used to protect problematic fields like Set-Cookie so that we get something like this for the multiple line version:

~~~
"set-cookie";b64: :dGhhdCwgaGFwcGVucw==:, :dG8sIGNvbnRhaW4sIGEsIGNvbW1h:
~~~

But you'd get this for the single-line version:

~~~
"set-cookie";b64: :dGhhdCwgaGFwcGVucywgdG8sIGNvbnRhaW4sIGEsIGNvbW1h:
~~~

The background of this has been discussed in #2143 and #1183

jricher:
This adds a binary wrapper for fields that don't follow the List or Dictionary format for multiple values, such as Set-Cookie. The wrapper is selected 
using the `bs` parameter, and the component value is made by wrapping each individual field value in a Byte Sequence and putting them into a List, then 
doing a strict serialization on those values.

Closes #2166

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -850,7 +850,7 @@ The `@status` component identifier MUST NOT be used in a request message.
 When a request message results in a signed response message, the signer can include portions of the request message in the signature base by adding the `req` parameter to the component identifier.
 
 `req`
-: A boolean flag indicatin that the component value is derived from the request that triggered this response message and not from the response message directly.
+: A boolean flag indicating that the component value is derived from the request that triggered this response message and not from the response message directly.
 
 This parameter can be applied to both HTTP fields and derived components that target the request, with the same semantics. The component value for a message component using this parameter is calculated in the same manner as it is normally, but data is pulled from the request message instead of the target response message to which the signature is applied.
 
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -233,7 +233,7 @@ The following sections define component identifier names, their parameters, thei
 
 The component name for an HTTP field is the lowercased form of its field name. While HTTP field names are case-insensitive, implementations MUST use lowercased field names (e.g., `content-type`, `date`, `etag`) when using them as component names.
 
-Unless overridden by additional parameters and rules, the HTTP field value MUST be canonicalized as a single combined value as defined in {{Section 5.2 of HTTP}}. Note that some HTTP fields, such as Set-Cookie {{COOKIE}}, do not follow a syntax that allows for combination of field values in this manner such that the combined output is unambiguous from multiple inputs. However, the canonicalized component value is never parsed by the message signature process, merely used as part of the signature base in {{create-sig-input}}. Even so, caution needs to be taken when including such fields in signatures, see {{security-non-list}} for more discussion of this issue.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-message-signatures.md
```diff
@@ -792,9 +850,9 @@ The `@status` component identifier MUST NOT be used in a request message.
 When a request message results in a signed response message, the signer can include portions of the request message in the signature base by adding the `req` parameter to the component identifier.
 
 `req`
-: Indicates that the component value is derived from the request that triggered this response message and not from the response message directly.
+: A boolean flag indicatin that the component value is derived from the request that triggered this response message and not from the response message directly.
```
jricher, line 853:
```suggestion
: A boolean flag indicating that the component value is derived from the request that triggered this response message and not from the response message directly.
```

draft-ietf-httpbis-message-signatures.md
```diff
@@ -357,6 +366,55 @@ The following example shows canonicalized values for different component identif
 
 Note that the value for `key="c"` has been re-serialized according to the strict `member_value` algorithm.
 
+
+### Binary-wrapped HTTP Fields {#http-header-byte-sequence}
+
+If the value of the the HTTP field in question is known by the application to cause problems with serialization, particularly with combination of multiple values as discussed in {{security-non-list}}, the signer MAY include the `bs` parameter in a component identifier to indicate the values of the fields need to be wrapped as binary structures before being combined.
```
yaronf, line 372: This is not a MAY, it's at least a SHOULD because not using it would result in a security issue.

yaronf, line 372: If we are aware of other common fields that need to be handled this way, let's list them in {{security-non-list}} as well.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -357,6 +366,55 @@ The following example shows canonicalized values for different component identif
...
+
+If this parameter is included with a component identifier, the component value is calculated using the following algorithm:
+
+0. Let the input be the ordered set of values for a field
+1. For each field value in the set:
+    0. Strip leading and trailing whitespace from each item in the list. Note that since HTTP field values are not allowed to contain leading and trailing whitespace, this will be a no-op in a compliant implementation.
+    1. Remove any obsolete line-folding within the line and replace it with a single space (" "), as discussed in {{Section 5.2 of HTTP1}}. Note that this behavior is specific to {{HTTP1}} and does not apply to other versions of the HTTP specification.
+    2. Encode the string as a Byte Sequence
```
yaronf, line 380: What does this mean? Maybe: encode the list as a structured field Byte Sequence using the list's ASCII representation. (This assumes that only ASCII is allowed, but that's really part of the problem.)

yaronf (changes requested):

draft-ietf-httpbis-message-signatures.md
jricher: Set-Cookie is the only common one that I found that fits this, but I didn't want the language to preclude other weirdness out there.

draft-ietf-httpbis-message-signatures.md
jricher: @mnot is this an accurate way to say "take the header value as a binary"? To create the examples I encoded the strings, and field values are already a subset of ASCII characters allowed.

```suggestion
    2. Encode the field value's ASCII representation as a Byte Sequence
```

draft-ietf-httpbis-message-signatures.md
jricher: I'm actually fine with this being a SHOULD for the reasons given.

```suggestion
If the value of the the HTTP field in question is known by the application to cause problems with serialization, particularly with combination of multiple values as discussed in {{security-non-list}}, the signer SHOULD include the `bs` parameter in a component identifier to indicate the values of the fields need to be wrapped as binary structures before being combined.
```

PR #2212, opened 2022-07-06
[6265bis] Changelog Updates

johannhof:
It feels like we've missed a bunch of those recently, so I added the remaining changes to the changelog. I'm not sure I got the versioning right, though.


sbingler (approved):
LGTM

chlily1 (approved):

PR #2214, opened 2022-07-07
[6265bis] Improved Max-Age attribute parsing (fixes #1757)

### LINKED ISSUE #1757: RFC6265bis: Minor improvements to the Max-Age attribute parsing algorithm
The current wording of the Max-Age attribute parsing algorithm could be improved so that `Max-Age=-` is handled better (right now, the instructions would lead to attempting to convert `-` to an integer).

Also, it might be worth mentioning that digits should be converted to an integer using base 10 (so it's clear that `Max-Age=0111` should be the same as `Max-Age=111` and not `Max-Age=73`).

I did some quick testing in recent versions of Chrome, Safari, and Firefox by setting cookies using `document.cookie` in the console, and these cases are all handled as expected.  Implementing the improvements above would be some nice dotting of i's and crossing of t's, though.

johannhof:
Contains the following changes:

- Skip parsing the attribute on empty value.
- Skip parsing the attribute when given Max-Age=-
- Add mention of "base 10" as guidance for implementers parsing integers
  from Max-Age.

Regarding browser interop:
Skip parsing aligns on Safari behavior and against Chrome and Firefox,
which will default to session lifetime when *any* invalid Max-Age
attribute is found (based on manual testing, it's difficult to test this
with WPT at the moment, see #1757).

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1251,24 +1251,26 @@ user agent MUST process the cookie-av as follows.
 If the attribute-name case-insensitively matches the string "Max-Age", the
 user agent MUST process the cookie-av as follows.
 
-1.  If the first character of the attribute-value is not a DIGIT or a "-"
-    character, ignore the cookie-av.
+1.  If the attribute-value is empty, ignore the cookie-av.
 
-2.  If the remainder of attribute-value contains a non-DIGIT character, ignore
+2.  If the first character of the attribute-value is neither a DIGIT, nor a "-"
+    character followed by a DIGIT, ignore the cookie-av.
+
+3.  If the remainder of attribute-value contains a non-DIGIT character, ignore
     the cookie-av.
 
-3.  Let delta-seconds be the attribute-value converted to an integer.
+4.  Let delta-seconds be the attribute-value converted to a base 10 integer.
 
-4.  Let cookie-age-limit be the maximum age of the cookie (which SHOULD be 400 days
+5.  Let cookie-age-limit be the maximum age of the cookie (which SHOULD be 400 days
     or less, see {{attribute-max-age}}).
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1251,24 +1251,26 @@ user agent MUST process the cookie-av as follows.
 If the attribute-name case-insensitively matches the string "Max-Age", the
 user agent MUST process the cookie-av as follows.
 
-1.  If the first character of the attribute-value is not a DIGIT or a "-"
-    character, ignore the cookie-av.
+1.  If the attribute-value is empty, ignore the cookie-av.
 
-2.  If the remainder of attribute-value contains a non-DIGIT character, ignore
+2.  If the first character of the attribute-value is neither a DIGIT, nor a "-"
+    character followed by a DIGIT, ignore the cookie-av.
```
mikewest, line 1257: Should we accept negative numbers? That seems like surprising behavior (and I'm not sure it would meaningfully differ from just ignoring the cookie entirely since it would be expired when we constructed the next `cookie` header).

mikewest (approved):
Looks good to me % a negative nit.

draft-ietf-httpbis-rfc6265bis.md
mikewest: Ok. 🤷  Fine by me if it matches our current behavior.

sbingler: +1 to accepting negative values.

Saying to ignore the cookie is somewhat ambiguous and could result in the cookie never getting set in the first place and therefore not overwriting any existing cookies in order to expire them.

E.x.:
`Set-Cookie: foo=bar; max-age=100000`
_A short time later_
`Set-Cookie: foo=delete_bar; max-age=-5`

I'd expect `foo` to be deleted after the second `Set-Cookie`

PR #2215, opened 2022-07-07
RFC6265bis: Add note not to send invalid cookies due to PSL changes

### LINKED ISSUE #1418: RFC 6265bis does not specify what happens when an existing cookie is no longer valid
This is a more general case of #1385.

It specifies how to store cookies in the `cookie store` in [section 5.4](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-07#section-5.4), but it doesn't specify what happens when a cookie is stored in the `cookie store`, then (at a later point in time, while the cookie is in the `cookie store`) something changes so that the cookie is no longer valid (i.e. would NOT have been stored in the `cookie store` had it been processed now rather than before).

This could happen due to a Public Suffix List change (as in #1385), due to a spec change in the section 5.4 algorithm, some change in things that a user agent MAY and chooses to do, etc.

sbingler:
Closes #1418

This adds a note clarifying that cookies whose domain was added to the PSL should not be sent even if they domain match the URI's host.

For example.
`Set-Cookie: foo=bar; Domain=example.com`
If example.com is added to the PSL then `foo` should not be sent to sub.example.com nor example.com

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1716,7 +1716,7 @@ cookie-string from a given cookie store.
          host of the retrieval's URI domain-matches the cookie's domain.
 
      NOTE: (For user agents configured to reject "public suffixes") It's
-     possible that the public suffic list was changed since a cookie was
+     possible that the public suffix list was changed since a cookie was
      created. If this change results in a cookie's domain becoming a public
      suffix then that cookie is considered invalid as it would have been
      rejected during creation (See {{storage-model}} step 9). User agents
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1715,6 +1715,12 @@ cookie-string from a given cookie store.
      *   The cookie's host-only-flag is false and the canonicalized
          host of the retrieval's URI domain-matches the cookie's domain.
 
+     NOTE: It's possible that the {{PSL}} has changed since a cookie was
+     created, potentially rendering the cookie now invalid if the
```
chlily1, line 1719: I feel like "now invalid" is a bit imprecise. I don't think we define "invalid" anywhere. How about something along the lines of "such that the cookie would be ignored entirely in step 9 of the algorithm in section 5.5, if processed in the context of the updated public suffix list"?

sbingler: How's that?

chlily1: LGTM, thanks!

chlily1 (approved):

mikewest (approved):

PR #2217, opened 2022-07-07
RFC6265bis: Offload Service Worker SFC computation to spec

### LINKED ISSUE #1288: "Site for cookies" handling for service workers
I'm a bit surprised by https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-06#section-5.2.2.2.

As I understand it the SameSite definition cookies use has a recursive parent check to prevent A2 in A1->B->A2 from getting cookies that A1 does get, as B might try to trick it. However, if A2 is entirely generated by a service worker, the above rules would mean it would still get spoofed.

cc @krgovind @jkarlin

sbingler:
Closes #1288

This changes how 6265bis computes the Site for Cookies for Service Workers by referring to their spec instead of spelling it out locally.

It's been some time since this was [last discussed](https://notes.ietf.org/notes-httpbis-21-09#Issue-1288---Service-Workers) so while I think this was the agreed upon change I may be mistaken.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1084,20 +1084,8 @@ Service Workers are more complicated, as they act as a completely separate
 execution context with only tangential relationship to the Document which
 registered them.
 
-Requests which simply pass through a Service Worker will be handled as described
-above: the request's client will be the Document or Worker which initiated the
-request, and its "site for cookies" will be those defined in
-{{document-requests}} and {{dedicated-and-shared-requests}}
-
-Requests which are initiated by the Service Worker itself (via a direct call to
-`fetch()`, for instance), on the other hand, will have a client which is a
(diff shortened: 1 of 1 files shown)


sbingler:
@mikewest ping

mikewest (commented):
This looks reasonable from the perspective of the cookies spec, but I don't think the service worker spec actually says anything explicitly about this topic. Is there a PR underway, or did I miss the explanation?

wanderview:
Can you file a new issue on the service worker spec to add a section discussing SFC computation?

sbingler:
@mikewest 
How does the tracking issue suit you?

sbingler:
@mikewest ping

mikewest:
Hi, sorry. I've been a potato for the recent past, slowly recovering from covid. I'm happy to land this with the issue filed.

PR #2227, opened 2022-08-05
Add QUERY example to introduction

### LINKED ISSUE #2171: Example in introduction
The Introduction has examples of how to do things without QUERY, but no examples of QUERY itself. Having one would help readers understand the method more easily.

mnot:
Fixes #2171


reschke (changes requested):
Yes, but.

If we do this, we'll also need to define the QUERY semantics for this content type, no?

reschke:
> Yes, but.
> 
> If we do this, we'll also need to define the QUERY semantics for this content type, no?

Should I gave that a try?

mnot:
That'd be good -- I'm curious to see how that works :)

sandstrom:
Out of curiosity, what's the status for this (not this issue, but QUERY in general)?

I think QUERY is a great addition to the HTTP spec, and would love to start using it.

reschke (approved):

reschke:
fixed via a2332a521ddb42e0c7d0c3b6f9a636b2684fcecf

PR #2233, opened 2022-08-22
AUTH48 for binary messages

martinthomson:
Still working through the changes, but I need to start referring to some of these changes.


martinthomson:
I've approved these changes with the RPC, your turn @chris-wood.

PR #2237, opened 2022-08-26
added message component context discussion

### LINKED ISSUE #1196: Support expected authority changes
In some cases, the authority of the effective request URI may be expected to change, for example from "public-service-name.example.com" to "service-host-1.public-service-name.example.com".  This is commonly the case for services that are hosted behind a load-balancing gateway, where the client sends requests to a publicly known domain name for the service, and these requests are transformed by the gateway into requests to specific hosts in the service fleet.

One possible way to handle this would be to special-case the Host header field to allow verifier to substitute a known expected value, or a value provided in another header field (e.g., `Via`) when generating the Signature Input, provided that the verifier also recognizes the real value in the `Host` header.  Alternatively, this logic could apply to an `(audience)` content identifier.

jricher:
closes #1196


draft-ietf-httpbis-message-signatures.md
```diff
@@ -1904,6 +1907,12 @@ To counter this, an application needs to validate the content of the fields cove
 
 Multiple applications and protocols could apply HTTP signatures on the same message simultaneously. A naive verifier could become confused in processing multiple signatures, either accepting or rejecting a message based on an unrelated or irrelevant signature. In order to help an application select which signatures apply to its own processing, the application can declare a specific value for the `context` signature parameter. For example, a signature targeting an application gateway could require `context="app-gateway"` as part of the signature parameters for that application.
 
+## Message Component Context {#security-message-component-context}
```
yaronf, line 1910: The section title is very general, but most of the text is about one single use case (a reverse proxy).

jricher: You're totally right -- but @richanna is working on reviewing and expanding this right now. :) I just needed to git it up on GitHub so she could take a pass at it, too.

Issue #2229, opened 2022-08-08
Nameless cookies, client/server inconsistencies


lavish:
I am reporting on some inconsistencies discovered while researching browser/server compliance with rfc6265bis.

rfc6265bis-04 changed the cookie parsing algorithm to support nameless cookies, i.e., `Set-Cookie: token` should create a nameless cookie with value `token`. The commit that introduced the change is https://github.com/httpwg/http-extensions/commit/0178223fb44fb194930f53181c5852954890e233, which followed the discussion at https://github.com/httpwg/http-extensions/issues/159.

The relevant parts are in [Section 5.4](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-10#section-5.4): The Set-Cookie Header Field, point 3.

> If the name-value-pair string lacks a %x3D ("=") character, then
> the name string is empty, and the value string is the value of
> name-value-pair.
> Otherwise, the name string consists of the characters up to, but
> not including, the first %x3D ("=") character, and the (possibly
> empty) value string consists of the characters after the first
> %x3D ("=") character.

And in [Section 5.6.3](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-10#section-5.6.3): Retrieval Algorithm, point 4.

> 1.  If the cookies' name is not empty, output the cookie's name
>     followed by the %x3D ("=") character.
> 2.  If the cookies' value is not empty, output the cookie's
>     value.
> 3.  If there is an unprocessed cookie in the cookie-list, output
>     the characters %x3B and %x20 ("; ").

### Conflict with the BNF syntax
These changes seem to be incompatible with the Set-Cookie BNF syntax in [Section 4.1.1](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-10#section-4.1.1), where the `=` symbol is considered mandatory in the `cookie-pair` definition: `cookie-pair = cookie-name BWS "=" BWS cookie-value`.

### Inconsistent browser behavior
According to my tests, recent versions of Chrome and Firefox follow the latest rfc6265bis, while Safari deviates from the intended behavior. Setting a cookie via `document.cookie = "test"` on `example.com` results into the following:

| Browser | Cookie Name | Cookie Value | `document.cookie` getter | HTTP request header |
| ------- | ----------- | ------------ | ------------------------ | --------------------|
| Chrome  |             | `test`       | `'test'`                 | `Cookie: test`      |
[...]

PR #2251, opened 2022-09-21
RFC6265bis: Prevent nameless cookies with prefixed values

sbingler:
Modify the cookie storage algorithm to reject cookies that:

- Do not have a name
- Have values that look like cookie prefixes

As mentioned in #2229, malicious servers can exploit nameless cookies to impersonate prefix'd cookies.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1671,10 +1671,10 @@ user agent MUST process the cookie as follows:
     true, abort these steps and ignore the cookie:
 
     * the cookie-value begins with a case-insensitive match for the string
-    "__Secure-"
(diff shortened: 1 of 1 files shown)


mikewest (approved):
Given that we accept nameless cookies, this looks like a reasonable approach to me. We would ideally figure out how to stop accepting nameless cookies, but that's a larger effort. :)

Do we have tests for this behavior in WPT?

sbingler:
> Do we have tests for this behavior in WPT?

Almost, I've just put a change up to add some.

sbingler:
> Do we have tests for this behavior in WPT?

Good question, no not yet. I'll add some.

sbingler:
Done

sbingler:
Ping @mikewest

Issue #2146, opened 2022-06-06
update HTTP Semantics reference to RFC 9110


b---c:
I-D.ietf-httpbis-semantics -> RFC 9110

bc-pi:
Also
HTTP/3  ietf-quic-http -> RFC9114 
QPACK ietf-quic-qpack -> RFC9204

PR #2256, opened 2022-10-03
client-cert-field: update refs to now RFCs 9110/9114/9204

bc-pi:
update refs for HTTP Semantics, HTTP/3, and QPACK to point to the now RFCs 9110/9114/9204

for #2146


MikeBishop (changes requested):
HTTP/2 should be RFC9113 now, and these should use the display names in [the style guide](https://github.com/httpwg/admin/blob/main/editors/style-guide.md). #2263 has these changes, since that was easier than using the "suggestions" option in the review.

b---c:
Thanks Mike. That was much easier than using the "suggestions" option. I merged #2263 and then fixed(!?) the YAML header with [79c8b49](https://github.com/httpwg/http-extensions/pull/2256/commits/79c8b493831c83dc4aa78401fd415db1413b1161). I think this one is good to go now?

MikeBishop (approved):

MikeBishop:
Yes, good to go!

Issue #1935, opened 2022-02-04
No defined client cert error mechanism for the origin


davidben:
From https://github.com/httpwg/http-extensions/issues/1927#issuecomment-1030299479, it sounds like the intent with the client certs draft is that the proxy does little to no certificate validation and punts all of it to the origin. In that case, we need to sort out how errors work...

Client cert UX is bad in part due to an impedance mismatch. In a user-facing client (like a browser), presenting a client certificate with a user-facing identity, we typically need to prompt the user to select an identity, reprompt on bad selection, etc. However, a single user-facing "session" covers multiple HTTP requests, which may then create or reuse connections depending on many factors. That means we cannot surface auth to the user at the connection- or request-. TLS session resumption and [single-use sessions](https://datatracker.ietf.org/doc/html/rfc8446#appendix-C.4) further complicate this.

The consequence is clients cache client certificate decisions. (Non user-facing clients tend to have credentials preconfigured, which effectively also remembers the decision.) With a cache, clients need to know when to invalidate it. Otherwise the user can never recover from selecting the wrong identity. As this is TLS-level auth, today this is done by reacting to TLS-level alerts.

draft-ietf-httpbis-client-cert-field breaks this correspondence by introducing a way to move half of TLS-level auth to HTTP. That is incompatible with today's client cert deployment assumptions, so we need to define something here. This is further complicated because the draft is shifting connection-level auth to a per-request scheme. I can see two options, but neither is very satisfying.

**Proxy translates HTTP error to TLS alert**

The most compatible option would be to define an HTTP error that the proxy must translate to a TLS alert on the client. The problem is TLS alerts tear down the connection, so this would be very disruptive in HTTP/2, where we may have other requests multiplexed. (The TLS-level mechanism does not have this problem because it happens in connection setup.)

**Modify clients to interpret HTTP error as TLS error**

The other option is to define an HTTP error and pass it through to the client. However, that means this draft is not only a proxy <-> server protocol, but _also_ changes HTTP clients. We would need to discuss the compatibility implications.

Additionally, by introducing a brand new way to do client cert errors, we introduce some awkward quirks. Suppose the server sends a `text/html` resource with subresources on the same origin. Unlike HTTP auth, client cert auth applies to the whole connection. There would be no way to fetch those subresources without re-triggering the auth prompt. Perhaps we include lots of warnings that the server need to keep all its error page resources inline or on a different origin? Perhaps we say the client should ignore the response and map the HTTP error to some network-level error page?

[...]

PR #2257, opened 2022-10-03
client-cert: Mention that origin server access control decisions can be conveyed with a 403 or by selecting response content

bc-pi:
client-cert: Mention that origin server access control decisions can be conveyed with a 403 or by selecting response content

#1935

### DIFF
--- a/draft-ietf-httpbis-client-cert-field.md
+++ b/draft-ietf-httpbis-client-cert-field.md
@@ -248,11 +248,9 @@ Backend origin servers may then use the `Client-Cert` header field of the
 request to determine if the connection from the client to the TTRP was
 mutually-authenticated and, if so, the certificate thereby presented by the
 client.
-
-Forward proxies and other intermediaries MUST NOT add the `Client-Cert` or
-`Client-Cert-Chain` header fields to requests, or modify an existing
-`Client-Cert` or `Client-Cert-Chain` header field. Similarly, clients MUST NOT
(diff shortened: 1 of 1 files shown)


davidben:
Do you envision that clients would drop their client cert cache in response to this 403? (They currently do not.) If so, how would that interact with subresources on the page?

bc-pi:
I honestly don't envision any change in client behavior. I realize that doesn't address your concern. But I believe doing anything more is firmly beyond the scope of this draft. The sentence added by this PR is only aiming to fulfill Tommy's suggestion at the last virtual interim and in the issue https://github.com/httpwg/http-extensions/issues/1935#issuecomment-1142682468

martinthomson:
You might want to explain that David's concern is a genuine problem that this doesn't solve, something like:

> Note that TLS clients that rely on error indications at the TLS layer for an unacceptable certificate will not receive those signals.

tfpauly (approved):

bc-pi:
Thanks Martin. I'll add that note.

MikeBishop (approved):
This is sufficient.  A similar experience exists today when a client presents a certificate which is valid, but which does not grant access to the requested resource.  While it's not ideal in certain ways -- the client might possess other certificates that would grant access if given an appropriate selector -- the server might also prefer not to share that list.

This draft isn't attempting to create new solutions from scratch -- it's attempting to outline what is currently done in assorted custom ways to help coalesce these into a single format.  Those custom mechanisms don't provide TLS alert translation either, so far as I'm aware.

Issue #1927, opened 2022-02-03
Certificate chain in the same order as it appears in TLS


martinthomson:
There's a bunch of text in the draft about putting the certificate chain in the right order.  This tries to crib from TLS and ends up being somewhat tricky to get right as TLS has a bunch of carefully worded and - frankly - ambiguous rules.  As a simple alternative, how about we just state that the certificates need to appear in the same order as the client sent them in TLS.

bc-pi:
Makes sense. Will look at adjusting the text accordingly.

martinthomson:
It is definitely 1. You do 2 and you don't need this client-cert-chain field at all.

That doesn't mean that a proxy can't be configured to do the work and pass the info along. But one of the two needs to be in the hook for enforcement.

A consequence of 1 is that alerts at the TLS layer won't be used to signal errors, aside from those that the proxy detects (or has to, like if the signature is invalid). That is ok, mostly. The main protocol issue here that might be a problem is validation of signature_scheme_cert. I can't see how failing to validate that would be a significant problem though.

bc-pi:
In [this conversation](https://github.com/httpwg/http-extensions/pull/1874#pullrequestreview-841081689) on the PR that introduced the chain I was aiming for 2. But eventually relented to MT. Some additional considerations about the security implications are likely warranted though.

davidben:
I see. If it's (1), that should probably be clearer in the draft. That also means the draft needs to define an error-signaling scheme. Moreover, if the proxy doesn't then translate that error into TLS alerts, we'd need changes in clients for servers relying on this to work right. I'll file a bug.

b---c:
I ran into real reservations as I looked at making this change. I still don't like pushing the trust chain validation to the origin. For many reasons. The proxy also has no means of signaling to the client which trust anchors the origin is okay with.  
The current text is somewhat ambiguous about what goes in Client-Cert-Chain but I think that's actually better than saying it has to be same as in the TLS handshake.  I dunno, I think I might raise the question of WTF actually goes in Client-Cert-Chain at the next interim or real meeting.

bc-pi:
At the last virtual interim (in May/June yikes!) Martin kindly suggested that I was making this too complicated and to just make the change (paraphrasing). I've tried to do that with changes in PR #2258 .

PR #2258, opened 2022-10-03
client-cert: certificate chain in the same order as it appears in TLS

bc-pi:
attempt @martinthomson's "simple alternative" in #1927 to say the chain is in the same order as it appears in TLS rather than copying the language from TLS


draft-ietf-httpbis-client-cert-field.md
```diff
@@ -183,11 +183,13 @@ multiple times in a request.
...
 In the context of a TLS terminating reverse proxy deployment, the proxy
-MAY make the certificate chain used for validation of the end-entity certificate
+MAY make the certificate chain for validation of the end-entity certificate
 available to the backend application with the Client-Cert-Chain HTTP header
-field. This field contains certificates used by the proxy to validate the
-certificate used by the client in the TLS handshake. These certificates might or
-might not have been provided by the client during the TLS handshake.
+field. This field contains certificates for validation of the end-entity
+certificate provided by the client in the TLS handshake. These certificates might or
+might not have been provided by the client during the TLS handshake, however,
```
martinthomson, line 190: Can you explain your reasoning behind this "might or might not" clause?

bc-pi: That text was from 0a7210f7d644dea399f0fc4d1b5119acb436d1db when @MikeBishop added the Client-Cert-Chain header and I was trying not to step on his work too much. And also to allow for the case that the proxy assembles the chain and conveys it.

martinthomson: This seems problematic as it invites a query about why that might (or might not) be the case.

I would prefer to say that the proxy can add certificates to this field.  It provides the certificates in the order in which they appear in the TLS handshake, unless it has cause to provide other information.

bc-pi: What about even less invitation of questions by making that last sentence this? 

"The order of the certificates is the same as the ordering in TLS (such as described in {{Section 4.4.2 of RFC8446}})."

martinthomson: That would work for me, but I got the impression that you are aware of cases where that is not the case.  I don't think that there is any harm in saying "..., unless the server and proxy agree otherwise." or similar.

MikeBishop: Honestly don't remember what my logic was in writing that; it appears to have been from my first pass at this text and not touched in any of the feedback.

I *think* that was based on TLS 1.3 saying that "...a certificate that specifies a trust anchor MAY be omitted from the chain, provided that supported peers are known to possess any omitted certificates."  If the proxy possesses certificates the origin might not, the client might omit them from the handshake, but the proxy could append them for the origin's information.

I don't think there's harm in allowing for private agreement here, but I don't feel strongly either way.

bc-pi: As I remember things, there was some disagreement when the initial text for `Cient-Cert-Chain` was added/proposed about whether the proxy would construct and validate the chain and convey the constructed chain to the origin or whether the proxy would pass through what the client sent as-is and defer validation and construction to the origin. Some of this discussion is captured in PR #1874. AFAIK most of the time the chain would look the same. But I've tried to adjust the text such that either would be okay. The sentence above about the order being the same was intended to mean only that ordering is the same but not that they are the same.

b---c: [5fb237a](https://github.com/httpwg/http-extensions/pull/2258/commits/5fb237a7dbb030ab323710a1d254d442cd41f35d) is another pass at this that removes the "might or might not" and tries to articulate ordering.

draft-ietf-httpbis-client-cert-field.md
```diff
@@ -183,15 +183,13 @@ multiple times in a request.
...
 available to the backend application with the Client-Cert-Chain HTTP header
-field. This field contains certificates used by the proxy to validate the
-certificate used by the client in the TLS handshake. These certificates might or
-might not have been provided by the client during the TLS handshake.
+field.
 
 Client-Cert-Chain is a List Structured Header {{!RFC8941}}.  Each item in the
 list MUST be a Byte Sequence ({{Section 3.3.5 of RFC8941}}) encoded as described
-in {{encoding}}.
+in {{encoding}}. The order is the same as the ordering in TLS (such as described in {{Section 4.4.2 of RFC8446}})
```
martinthomson, line 192:
```suggestion
in {{encoding}}. The order is the same as the ordering in TLS (such as described in {{Section 4.4.2 of RFC8446}}).
```

draft-ietf-httpbis-client-cert-field.md
```diff
@@ -142,7 +142,7 @@ the reverse proxy from the reverse proxy to the origin server.
 
 Client-Cert-Chain:
 : Conveys the certificate chain used for validation of the end-entity
-certificate used by the client in the TLS handshake from the reverse proxy to the origin server.
+certificate provided by the client in the TLS handshake with the reverse proxy from the proxy to the origin server.
```
martinthomson, line 145: This sentence has gotten a little hard to follow.

There are two things here that might need to be separated to make this more readable:

1. This is the chain (less the EE cert) that the client presents in TLS to the TTRP.
2. The TTRP sends this information to the origin server.

draft-ietf-httpbis-client-cert-field.md
```diff
@@ -506,12 +496,13 @@ latter and ensure wide applicability by not trying to cherry-pick particular
...
-The handshake and validation of the client certificate (chain) of the
-mutually-authenticated TLS connection is performed by the TTRP.  With the
+The validation of the client certificate and chain of the mutually-authenticated
+TLS connection is typically performed by the TTRP during the handshake.  With the
 responsibility of certificate validation falling on the TTRP, the
 end-entity certificate is oftentimes sufficient for the needs of the origin server.
 The separate `Client-Cert-Chain` field can convey the certificate chain for
-deployments that require such information.
+deployments that require such information in order for the origin server to augment
+the chain validation or assume it entirely.
```
martinthomson, line 505:
```suggestion
origin server deployments that require this additional information.
```

Maybe simpler is better.

martinthomson (commented):
This works well enough on the critical points.  Some points about readability though.

draft-ietf-httpbis-client-cert-field.md
b---c: Maybe so.

draft-ietf-httpbis-client-cert-field.md
b---c: I know. I know. Saying it is "a little hard to follow" is being too kind. It's pretty terrible. And I seem to make it worse each time I touch it. I'll take another pass over it though with that separation in mind. Maybe leaning more on the text above to make the second point and leaving it out of the Client-Cert/Client-Cert-Chain items.

Issue #1895, opened 2022-01-20
query: mention security aspects of moving query component into body


reschke:
David Slik (https://lists.w3.org/Archives/Public/ietf-http-wg/2022JanMar/0081.html):

> Moving query parameters from the request URI to the request body improves overall security, given that the request URI is often cached, stored, logged and otherwise potentially disclosed by intermediary systems. As a result, if PII or other sensitive information is included in the query section of an URI, it is at a higher risk when compared to when it is included in the request body.
> 
> A description of this advantage may be worth including.

candoumbe:
The security advantage mentioned here could also be viewed as a debug/development drawback because the server most of the time logs urls to have some insights on what was requested by the 'client'.

asbjornu:
Related to the discussion in #1909. I agree with @candoumbe that if any advice is given, some pros and cons of both alternatives should probably be enumerated.

mnot:
Just my .02 - I'd be wary of making claims about security that are based just on implementation practice (e.g., what logs expose and don't).

reschke:
Ok, let's close this.

MikeBishop:
I'm curious; we already merged #2259 which contains this text. Is that okay, or do we want a separate issue to remove that text?

reschke:
Good point. Did we forget to close the issue?

PR #2259, opened 2022-10-05
Fix: #1895. Mention security advantages.

ioggstream:
## This PR

- briefly mentions security advantages

See #1895


reschke (commented):
I believe this is already covered by the Security Considerations section.

Could you check whether that is sufficient?

ioggstream:
Moved little changes into sec-:

- sensitive information can be in both `path` and `query`;
- mention intermediaries.

MikeBishop (approved):

PR #2270, opened 2022-10-17
Update cookie-value definition to reflect browser behavior

april:
I recently ran into a problem where browsers were allowing cookies to be set containing values that standard libraries were failing to parse. These libraries are mostly following the relatively obsolete published RFC6265 (with tweaks), which limits a lot of characters that browsers allow and frequently appear inside cookies.

As such, this updates section 4.1.1 to reflect the characters that browsers allow to be set inside cookie values: any 8-bit octet except for control characters and semicolons.

For reference as to how browsers handle cookie setting:

* Chromium: [any octet except control characters and semicolons](https://source.chromium.org/chromium/chromium/src/+/main:net/cookies/parsed_cookie.cc;drc=796d9f7be35cc8f766fabc6e84df424046460890;l=495)
* Firefox: [any octet except control characters and semicolons, but allowing 0x09 (htab) and 0x7F (del)](https://searchfox.org/mozilla-central/source/netwerk/cookie/CookieCommons.cpp#210)
* Safari: seems to be any ASCII (7-bit) character except control characters (which includes htab) and semicolons

I was led down this rabbit hole by Golang:
* Golang: [any ASCII (7-bit) character except control characters, double quotes, semicolons, and backslashes](https://cs.opensource.google/go/go/+/master:src/net/http/cookie.go;l=414)
* Python: too many regular expressions for me to read through, but seems to behave the same as Golang

As for servers:
* nginx: lets you set a `Set-Cookie` with the characters listed above, so every browser accepts commas, semicolons, and backslashes, Firefox and Chromium allow unicode, etc.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -2587,6 +2587,9 @@ The "Cookie Attribute Registry" should be created with the registrations below:
 * Update editors and the acknowledgements
   <https://github.com/httpwg/http-extensions/pull/2244>
 
+* Update cookie-value definition to reflect browser behavior
+  <https://github.com/httpwg/http-extensions/pull/2270>
+
 # Acknowledgements
 {:numbered="false"}
 RFC 6265 was written by Adam Barth. This document is an update of RFC 6265,
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -483,10 +483,8 @@ set-cookie-string = BWS cookie-pair *( BWS ";" OWS cookie-av )
 cookie-pair       = cookie-name BWS "=" BWS cookie-value
 cookie-name       = 1*cookie-octet
 cookie-value      = *cookie-octet / ( DQUOTE *cookie-octet DQUOTE )
-cookie-octet      = %x21 / %x23-2B / %x2D-3A / %x3C-5B / %x5D-7E
-                      ; US-ASCII characters excluding CTLs,
-                      ; whitespace DQUOTE, comma, semicolon,
-                      ; and backslash
+cookie-octet      = %x20-3A / %x3C-7E / %x80-FF
+                      ; octets excluding CTLs and semicolon
 
 cookie-av         = expires-av / max-age-av / domain-av /
                     path-av / secure-av / httponly-av /


mikewest:
@sbingler Can you take a look at this, please?

april:
CC'ing @dveditz, as Firefox's behavior is different than Chromium when the stated behavior in the source is to act the same.

sbingler:
Hi,

> I recently ran into a problem where browsers were allowing cookies to be set containing values that standard libraries were failing to parse.

That
Section 4 is meant to be a subset of the characters that the spec asks user agents (browsers) to support. This is almost entirely due to historical reasons and the sub-set was selected to maximize a server's compatibility with all user agents (including more minor UAs). I.e.: If a server stays within Section 4's guidelines then their cookies are the most likely to be accepted by all UAs.

UAs, on the other hand, are meant to implement Section 5 which includes the larger set of characters you've identified.

So it sounds like the behavior you're seeing is not unexpected. Can you provide more information on your problem? That may help us come to a resolution.

april:
Sorry, I'm a bit confused by your response.

> Section 4 is meant to be a subset of the characters that the spec asks user agents (browsers) to support.

Browsers _do_ support what is specified, and quite a bit more, including double quotes and other characters, which is why I've been running into issues. This is an attempt to align section 4.1.1 with what browsers actually accept (and send to servers).

We are in an awkward state at the moment where browsers will accept characters inside cookies (such as dquotes) but then they will fail downstream because standard libraries are very limited in what they accept because they kinda-sorta follow 4.1.1 from the published RFC and then fail silently in a whole bunch of bizarre ways. It would be great if all sides of this equation are aligned on what values are allowed inside cookie values.

The exact issue I'm running into is that a popular piece of vendor software that handles EU cookie laws is setting cookies with a value that contains quotation marks, as it's using JSON. Browsers happily accept it, servers pass it along, and then the standard library (often invisibly) fails to parse it because it is using the rules as defined in 4.1.1.

Note that I just tested nginx and it happily allows me to use `Set-Cookie` with a quotation mark which browsers happily accept, also in violation section 4.1.1. It's rare for anything to follow 4.1.1 except when they follow pieces of it and break things randomly.

april:
To add: even if we updated 4.1.1 to reflect what the most proscriptive browser (Safari) allows, that would still mean that commas, backslashes, and double quotes are allowed inside cookie values.

sbingler:
> Sorry, I'm a bit confused by your response.

Whoops, I may have mangled my reply a bit with some errant and accidental cut/pasting.

> If 4.1.1 is meant to reflect what browsers will accept from servers

It is not.
Section 4 is the "Server Requirements" section, so everything in there is tailored to inform a server how to be the most compatible with every UA. As you've seen not every server follows those recommendations, but they remain good practice nonetheless.

(The definition of servers is a bit hand wavy in the spec. It might be more helpful to think of "servers" as anything that produces cookies.)

Section 5 "User Agent Requirements" details what browsers (and other UAs) should accept. And given that many servers have previously, and still do, deviate from Section 4's recommendation, the UAs are strongly encouraged to accept those deviations in the name of compatibility and dealing with historical baggage. Many of these accepted deviations are footguns or have had issues with consistent processing among different UAs which is why they're discouraged.

As an example: one such troublesome deviation is the ability to create a nameless cookie `Set-Cookie: =foobar`. Trying to remove that ability in the past broke important sites and so many UAs still support it. However it's a giant footgun and so section 4 disallows it. (And I'm hoping to remove it from Section 5 in the not too distant future)

> they kinda-sorta follow 4.1.1

Yes, this has long been a problem :/ Hence section 5.

> The exact issue I'm running into is that a popular piece of vendor software that handles EU cookie laws is setting cookies with a value that contains quotation marks, as it's using JSON. Browsers happily accept it, servers pass it along, and then the standard library (often invisibly) fails to parse it because it is using the rules as defined in 4.1.1.

The spec has mostly operated under the assumption that if a given server can produce a cookie, then that same server must also accept the cookie. Nowadays with SaaS and various frameworks being as popular as they are that isn't always the case. 
But I don't believe this is a problem for the spec to solve, it's not that the spec is giving incorrect information but rather that one of both sides is in noncompliance of the spec. Afaict if both the vendor software and the library implemented 4.1.1 fully this wouldn't be an issue.

Furthermore, expanding the character set in 4.1.1 could result in more servers producing those characters then they would've if we left 4.1.1. as-is. But unless 100% of every other UA/library/software also implements support then that results in more incompatibility across the web ecosystem.

april-dbx:
I guess I don't understand the point is of preventing servers from setting comma, backslash, and double quote in `Set-Cookie` when every single user agent happily accepts them.

Meanwhile, programming languages all over the place are following the guidelines in 4.1.1 for parsing cookies, even though that's not what it's intended for and despite the fact that browsers happily accept and send these cookies into those languages.

The fact that UAs, programming libraries, servers, and websites are all handling these things differently tells me that for whatever the specification is doing, it's not being particularly clear on what the proper behavior should be. In fact, in all of my investigation, I can't find a single anything that actually follows 4.1.1.

april:
Sorry, not at all intending to be combative, if it came off that way.

I'd be more than happy to update the PR to remove only those three characters, if you think that would be a better route?

If not, I can try to make PRs to the various programming languages instead (Golang already indicated they would be amenable), but I had figured that it would be best to start here since so many of them look to 4.1.1 for what they should do.

Thanks.  :)

sbingler:
> when every single user agent happily accepts them.

If we had data indicating this was true I'd be happy to consider expanding what section 4 allows!
Unfortunately I don't know of any. And there are more UAs than just Chrome, Firefox, and Safari: there are smaller browsers, non-browser apps that interact with the web, routers and other network equipment, proxies, etc.

Cookies are used in numerous, and sometimes surprising ways, across the web. Changes are slow to rollout, breakage is sometimes hard to see, and some systems have been in place for literal decades which leads me to be more conservative towards proposed changes.

> it's not being particularly clear on what the proper behavior should be here.

I completely agree, the spec could do a much better job of telling implementers which requirements they should follow. This is something I'd like to improve upon.

> Sorry, not at all intending to be combative, if it came off that way.

Not at all.

> I don't understand what the point is of preventing servers from setting commas, backslashes, and double quotes

I wasn't around for the original discussion that led to these being discouraged, but I can say for the backslash there have been issues/worries around parsing libraries naively interpreting some combinations as escape sequences `Set-Cookie:  foo=ba\r` which could have unintended effects.

> I'd be more than happy to update the PR to remove only those three characters, if you think that would be a better route?

I think we want to drop the entire PR. (Sorry!)

Because we also want to continue discouraging anything `0x80` or higher. In one of the previous IETF meetings it was brought up that characters in that range can cause some older network equipment to choke.

> I can try to make PRs to the various programming languages instead

I think asking the various languages and frameworks to correct themselves is the right approach. Not to imply that I want you to do that work, but if you're up to it that would be appreciated!

Thanks for spending the time investigating. I think your findings are interesting and may still be useful in shaping future decisions.

sbingler:
Section 4 is framed around maximizing compatibility. Changes to what we tell servers to do can affect every web ecosystem component that the server communicates with or along: Major/minor browsers, old/new browsers, CLI utilities, internet infrastructure, etc. Some of these components support different sets of characters from the others. Since we want to encourage a server to be the most compatible across the entire ecosystem we want to take the intersection of all the supported characters among those components which leads to a set that is less-than-or-equal to what a given UA might support. If we have a maximally compatible character set already, any additions must then reduce compatibility. So we should have a good reason to.

(Some servers may deviate from our suggestion and that will exclude some UAs; that's something the server does at its own risk.)

On the other hand: what we tell browsers, and other UAs, to do in section 5 is framed around supporting as many servers as reasonably possible. Since we know that some servers deviate from section 4 for whatever reasons, this necessitates supporting a larger set than section 4.

There’s a recent example of this:

Cookie prefixes, `__Secure-` and `__Host-`, were originally specified to be case sensitive. It turned out that some servers would send `__HOST-` and `__secure-` cookies. So section 5 (UA) was updated to specify that prefixes should be checked case-insensitively in order to still gain the benefits of the prefixes with those servers. However, section 4 continues to specify that servers should send case sensitive prefixes in order to support UAs that don’t do case-insensitive checks.

> Can you help me understand what the criteria is for determining when it's safe to make this change?

Enough benefit to the change such that it outweighs potential compatibility loss.
So the change should be very positive, or we should have some data that the potential compatibility loss is very small.

Security related changes are a good example.

> Is there a test suite where we test against clients to determine when it would be safe to change?

Not that I'm aware of. The closest I know of would be the [web platform tests](https://web-platform-tests.org/), but that is primarily focused on testing web platform features.

> Would it be possible to change Section 4, which seems to indicate that this behavior belongs to both sides

I'm in favor of clarifying section 4.

> should we update Section 5 to forbid user agents from accepting cookies and/or sending cookies with [%x80-FF]

There is some interest in deprecating this historical cruft, so in the longer term hopefully yes.

In the short term though, I think not. While we want to encourage servers not to use them, some of them do. In this case, if the UA is able to receive the character then it obviously was able to make it to the user's device and so it's probably fine to send it back.

april:
> In this case, if the UA is able to receive the character then it obviously was able to make it to the user's device and so it's probably fine to send it back.

Browsers accept those characters both from `Set-Cookie` (where it would probably be fine to reflect as `Cookie`) but also from Javascript. Which the server had nothing to do with, and which breaks things that follow 4.1.1 and is the reason why I opened this ticket in the first place. 😅

sbingler:
After speaking out of band, April and I have agreed to close this PR.

martinthomson:
Can I request that you document the reason for closing an issue when you do so?

sbingler:
Of course, my mistake.

We discussed the reason behind this change (servers' libraries don't like cookies with these characters) and agreed that this change isn't the right way to remedy that problem.

dveditz:
> Browsers accept those characters both from `Set-Cookie` (where it would probably be fine to reflect as `Cookie`) but also from Javascript. Which the server had nothing to do with [...]

From the browser's POV the server had _everything_ to do with that script. The browser won't create any cookies on its own!

It's of course a complication that web documents can be amalgamations of scripts from different servers with different cookie behavior, but even in a pure single-server document (e.g. CSP `default-src 'self'`) this isn't the only way a web application's code can conflict with the server layer it's running on.

I'm not saying this problem is purely the server's fault; I'm pushing back on the implication that it's purely the browser's fault. It's complicated.

PR #2278, opened 2022-10-28
Editorial review on signature

ioggstream:
## This PR

- minor editorial comments on signatures


draft-ietf-httpbis-message-signatures.md
```diff
@@ -191,7 +191,7 @@ HTTP applications may be running in environments that do not provide complete ac
 As mentioned earlier, HTTP explicitly permits and in some cases requires implementations to transform messages in a variety of ways. Implementations are required to tolerate many of these transformations. What follows is a non-normative and non-exhaustive list of transformations that could occur under HTTP, provided as context:
 
 - Re-ordering of header fields with different header field names ({{Section 5.3 of HTTP}}).
-- Combination of header fields with the same field name ({{Section 5.2 of HTTP}}).
+- Combination of fields with the same field name ({{Section 5.2 of HTTP}}).
```
jricher, line 194: @ioggstream -- should everything in this just be "field" and not "header field", in addition to just this line?

jricher: Yes, trailer fields could be a part of signatures, and the signature field could be sent as a trailer itself.

reschke: Q: if a field name appears both in the header and trailer section, should the recipient combine the field values? (see https://www.rfc-editor.org/rfc/rfc9110.html#trailer-fields)

jricher:
Merging this PR and building additional text on top of this based on WG feedback.

PR #2292, opened 2022-11-04
Proposal of server-generated upload URLs

Acconut:
The current draft -00 of the resumable upload protocol uses a client-generated token to identify which procedures belong together. There has been valid criticism of this methodology, especially that the server is not able to control this identifier, which fits badly into the typical request-response schema of HTTP APIs. There seems to be a majority supporting the use of server-generated identifiers over client-generated ones.

That being said, the original intention of these tokens was two-folded and is still desirable:

1. Allow clients to upload small files in a single request, without the need of additional roundtrips.
2. Allow clients to resume an upload before receiving a response for the server by using the client-generated token.

We believe it is possible to achieve these two goals even when using server-generated uploads URLs by utilizing the Idempotency-Key header, whose draft is currently being discussed by the httpapi working group (https://datatracker.ietf.org/doc/draft-ietf-httpapi-idempotency-key-header/).

This PR proposes to move from client-generated tokens to server-generated URLs. The role of Idempotency-Key is not included here and is laid out in a separate issue: https://github.com/httpwg/http-extensions/issues/2293

Any feedback is welcome!

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -75,7 +75,7 @@ The terms byte sequence, Item, string, sf-binary, sf-boolean, sf-integer, sf-str
 
 The terms client and server are imported from {{HTTP}}.
 
-Upload: A sequence of one or more procedures, uniquely identified by a token chosen by a client.
+Upload: A sequence of one or more procedures, uniquely identified by an upload URL chosen by a server.
 
 Procedure: An HTTP message exchange for that can be used for resumable uploads.
 
@@ -88,8 +88,7 @@ The Resumable Uploads Protocol consists of several procedures that rely on HTTP 
 * Upload Appending Procedure ({{upload-appending}})
 * Upload Cancellation Procedure ({{upload-cancellation}})
 
-A single upload is a sequence of one or more procedures. Each upload is uniquely
-identified by a token chosen by a client. The token is carried in the Upload-Token header field; see {{upload-token}}.
+A single upload is a sequence of one or more procedures. Each upload is uniquely identified by an upload URL chosen by a server.
 
 The remainder of this section uses examples of a file upload to illustrate permutations of procedure sequence. Note, however, that HTTP message exchanges use representation data (see {{Section 8.1 of HTTP}}), which means that procedures can apply to many forms of content.
 
@@ -97,21 +96,22 @@ The remainder of this section uses examples of a file upload to illustrate permu
 
 In this example, the client first attempts to upload a file with a known size in a single HTTP request. An interruption occurs and the client then attempts to resume the upload using subsequent HTTP requests.
 
-1) The Upload Creation Procedure ({{upload-creation}}) can be used to notify the server that the client wants to begin an upload. The server should then reserve the required resources to accept the upload from the client. The client also begins transferring the entire file in the request body. The request includes the Upload-Token header, which is used for identifying future requests related to this upload. An informational response can be sent to the client to signal the support of resumable upload on the server.
+1) The Upload Creation Procedure ({{upload-creation}}) can be used to notify the server that the client wants to begin an upload. The server should then reserve the required resources to accept the upload from the client. The client also begins transferring the entire file in the request body. An informational response can be sent to the client to signal the support of resumable upload on the server and transmit the upload URL in the Location header, which is used for identifying future requests related to this upload.
 
 ~~~
 Client                                  Server
 |                                            |
-| POST with Upload-Token                     |
+| POST                                       |
 |------------------------------------------->|
 |                                            |
 |                                            | Reserve resources
-|                                            | for Upload-Token
+|                                            | for upload
 |                                            |------------------
 |                                            |                 |
 |                                            |<-----------------
 |                                            |
 |            104 Upload Resumption Supported |
+|            with upload URL                 |
 |<-------------------------------------------|
 |                                            |
 | Flow Interrupted                           |
@@ -120,12 +120,12 @@ Client                                  Server
 ~~~
 {: #fig-upload-creation-procedure title="Upload Creation Procedure"}
 
-2) If the connection to the server gets interrupted during the Upload Creation Procedure, the client may want to resume the upload. Before this is possible, the client must know the amount of data that the server was able to receive before the connection got interrupted. To achieve this, the client uses the Offset Retrieving Procedure ({{offset-retrieving}}) to obtain the upload's offset.
+2) If the connection to the server gets interrupted during the Upload Creation Procedure, the client may want to resume the upload. Before this is possible, the client must know the amount of data that the server was able to receive before the connection got interrupted. To achieve this, the client uses the Offset Retrieving Procedure ({{offset-retrieving}}) to obtain the upload's offset using the upload URL.
 
 ~~~
 Client                                      Server
 |                                                |
-| HEAD with Upload-Token                         |
+| HEAD to upload URL                             |
 |----------------------------------------------->|
 |                                                |
 |              204 No Content with Upload-Offset |
@@ -139,7 +139,7 @@ Client                                      Server
 ~~~
 Client                                      Server
 |                                                |
-| PATCH with Upload-Token and Upload-Offset      |
+| PATCH to upload URL with Upload-Offset         |
 |----------------------------------------------->|
 |                                                |
 |                      201 Created on completion |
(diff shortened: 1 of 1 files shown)


gstrauss:
> That being said, the original intention of these tokens was two-folded and is still desirable:
>
> *  Allow clients to upload small files in a single request, without the need of additional roundtrips.
> *  Allow clients to resume an upload before receiving a response for the server by using the client-generated token.

For a small file, how are resumable-uploads useful?  If a disconnect occurs, the data can be re-sent with little overhead.
For multiple small files, are HTTP/2 streams somehow insufficient?  Why?
For a large file, why is round-trip delay time a problem for an upload that will presumably take order(s) of magnitude longer than the round-trip delay time to upload?

Acconut:
> For a small file, how are resumable-uploads useful? If a disconnect occurs, the data can be re-sent with little overhead.

Good question! The benefit of resumable uploads (or retryable uploads in this case) for small files is that the client and server can use one API for all uploads, regardless of their size. For example, in our use case, customers can upload files of any size and type to our server. Often, these files are uploaded simultaneously in the same session. It would be helpful if big and small files could use the same mechanism for uploads and therefore reduce the complexity in the our entire uploading infrastructure (clients, proxies, servers etc).

In addition, it makes it possible for the client to decide what a big and small file is. The threshold for when a file is big depends on the available network and is best judged by the client. So if the server supports this hybrid approach, the client can decide if it wants to go the save route and create the upload in a separate request, or try to transfer the file in a single request. For example, a 10MB file might be small if you have a fast fiber connection or big if you are on a spotty mobile network.

Does that make sense?

gstrauss:
> You are correct that the additional request is not a huge concern for big uploads. In these situations, the client would prefer to create the upload in a separate request. Only in cases of small uploads, it makes sense to combine the creation and data transfer in a single request.

It sounds like you are trying to optimize for a case where it should not matter.  In the case of a small file -- as determined by the client -- if upload fails, then send it again.  If your failure rate is high in the common case *for small files*, then that sounds like a different problem.  As you agreed above, when the file is not small, the round-trip latency is not relevant.

Also, HTTP/1.1 pipelining has been around for over 20 years, and HTTP/2 is even better.  For a small file to be created, you send an HTTP request using HTTP.  If you get a 201 or 200 response, you are done.  If you did not, then you may choose to retry.  If you are using resumable-uploads, you have to keep the file around to be able to resume, which is no different than keeping the small file around to resend in total, so I do not see any added value.

If you must upload the file once and only once, then that sounds like a different problem.  You have mentioned Idempotency-Key as one option.  Another is that if you must upload the file once and only once, then that is an application-level requirement and can be solved by application-level communication of a unique token or the like.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -203,11 +203,11 @@ The Upload Creation Procedure is intended for starting a new upload. A limited f
 
 This procedure is designed to be compatible with a regular upload. Therefore all methods are allowed with the exception of `GET`, `HEAD`, `DELETE`, and `OPTIONS`. All response status codes are allowed. The client is RECOMMENDED to use the `POST` method if not otherwise intended. The server MAY only support a limited number of methods.
 
-The request MUST include the `Upload-Token` header field ({{upload-token}}) which uniquely identifies an upload. The client MUST NOT reuse the token for a different upload. The request MUST NOT include the `Upload-Offset` header.
+The request MUST include the `Upload-Incomplete` header field ({{upload-incomplete}}). It MUST be set to true if the end of the request body is not the end of the upload. Otherwise, it MUST be set to false. This header field can be used for request identification by a server. The request MUST NOT include the `Upload-Offset` header.
```
guoye-zhang, line 206: I think of `Upload-Incomplete` as an advanced feature for people who need to do chunked uploads. Using it to identify resumable upload feels a bit wrong. Also `Upload-Incomplete = false` is a double negative which can be confusing. If we go this route, I prefer flipping it to `Upload-Complete` instead. (Also don't forget to update the examples)

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -237,34 +238,34 @@ upload-offset: 100
...
 
 ## Feature Detection
 
-If the client has no knowledge of whether the server supports resumable upload, the Upload Creation Procedure MAY be used with some additional constraints. In particular, the `Upload-Incomplete` header field ({{upload-incomplete}}) MUST NOT be sent in the request if the server support is unclear. This allows the upload to function as if it is a regular upload.
+If the client has no knowledge of whether the server supports resumable upload, the Upload Creation Procedure MAY be used with some additional constraints. In particular, the `Upload-Incomplete` header field ({{upload-incomplete}}) MUST NOT be set to true if the server support is unclear. This allows the upload to function as if it is a regular upload.
 
-If the server detects the Upload Creation Procedure and it supports resumable upload, an informational response with `104 (Upload Resumption Supported)` status MAY be sent to the client while the request body is being uploaded.
+The server SHOULD send the `104 (Upload Resumption Supported)` informational response to the client, to indicate its support for a resumable upload.
```
guoye-zhang, line 260: Is `SHOULD` too strong for feature detection? I used `MAY` since it's optional, but we could update it all to `SHOULD`.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -383,30 +382,18 @@ upload-draft-interop-version: 2
 
 The Upload Creation Procedure ({{upload-creation}}) supports arbitrary methods including `PATCH`, therefore it is not possible to identify the procedure of a request purely by its method. The following algorithm is RECOMMENDED to identify the procedure from a request for a generic implementation:
```
guoye-zhang, line 383: I think we could get rid of this entire section. It's no longer ambiguous what the procedure is since the server determines the URL for appending, retrieving and cancelling.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -429,15 +416,13 @@ The `301 (Moved Permanently)` status code and the `302 (Found)` status code MUST
...
-
-It is OPTIONAL for the server to partition upload tokens based on client identity established through other channels, such as Cookie or TLS client authentication. The client MAY relax the token strength if it is aware of server-side partitioning.
+The upload URL obtained through the Upload Creation Procedure ({{upload-creation}}) is the identifier used for modifying the upload. Without further protection of this upload URL, an attacker may use the upload URL to obtain information about an upload, append data to it, or cancel it. To prevent this, the server SHOULD ensure that only authorized clients can perform the Offset Retrieving Procedure ({{offset-retrieving}}), Upload Appending Procedure ({{upload-appending}}), or Upload Cancellation Procedure ({{upload-cancellation}}) for a given upload URL and otherwise reject the procedure.
```
guoye-zhang, line 419: Let's also recommend the URL to be strong and non-guessable. Even one bit of info leak (whether an upload exists or not) could be detrimental.

guoye-zhang (approved):
Finally got a chance to go through this. Thank you for doing this! Looks good in general

draft-ietf-httpbis-resumable-upload.md
Acconut: Personally, I think SHOULD is better here. Yes, it is optional, but I do think that there is a benefit to using it and hence we should recommend it instead of just providing this mechanism.

draft-ietf-httpbis-resumable-upload.md
Acconut: I get where you are coming from and the relevancy of this section has decreased with this PR, yes. I do think that this section is helpful for people looking to start a server-side implementation. However, I am not sure if such content is best placed in the draft itself.

draft-ietf-httpbis-resumable-upload.md
Acconut: You touch on two good points here. I abuse the Upload-Incomplete header for feature detection here. This may or may not be a smart choice. Maybe we use a more explicit header as mentioned in https://github.com/httpwg/http-extensions/issues/2295 and keep Upload-Incomplete optional.

Removing the double-negative also sounds good but maybe let's do this in a separate PR.

draft-ietf-httpbis-resumable-upload.md
Acconut: Good idea!

Acconut:
@guoye-zhang Thank you for the feedback! I addressed your comments, but will go ahead with merging this. Let me know if you want to discuss some points in more detail.

PR #2296, opened 2022-11-07
Adding an -12 changelog to RFC6265bis.

mikewest:


mikewest:
It's not clear to me why "Update Editor's Copy" is failing. Looks like it started in https://github.com/httpwg/http-extensions/actions/runs/3411913518, which made no change to any draft, but simply pushed a new tag. @reschke, @martinthomson, perhaps y'all have ideas?

reschke:
There was a problem in the resumale uploads markdown causing a duplicate anchor ID in the XML representation. This should be fixed now.

mikewest:
Thanks for looking into it, @reschke! I'll merge this, then. :)

PR #2301, opened 2022-11-08
Address WGLC comments

### LINKED ISSUE #2277: Accept-Signature: if it is a header, it must be stated explicitly
## Question

- if Accept-Signature to be a header (see "Processing Accept-Signature") it should be stated explicitly, or the processing section should replace "header" with "field" 

> The receiver of an Accept-Signature field fulfills that header as follows:
^^ replace header with field

- move normative parts  from Requesting signatures to the Accept-Signature Field

## Note

- it cannot be sent in a trailer section
- it must be header field
- it can be used in requests and responses

@jricher if it's ok, I'll PR on this

### LINKED ISSUE #2279: Addessing missing content-encoding in signature input with content-digest
## Question 

In examples containing content-digest, there's no way to ensure that content-encoding is preserved.

The specific problem is that, for example, `identity` is only allowed in `Accept-Encoding`.

See
https://github.com/httpwg/http-extensions/blob/main/draft-ietf-httpbis-digest-headers.md#usage-in-signatures-usage-in-signatures

Fixing the example is tricky, and the easiest way is to add a content-encoding to the message and to the signature.
The other solution would be to create a specific flow for representation metadata...

cc: @reschke

### LINKED ISSUE #2290: Signatures 13: Problem with test-key-rsa-pss in Browser JS
I am having difficulty getting the [test-key-rsa-pss private key](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-message-signatures-13#appendix-B.1.2) being accepted by the web browsers.

This is a bit of a weird problem that would benefit from help from some JS crypto experts. But it is not a new problem: we had a similar problem a year ago in draft 07, which was the last time I worked on an implementation of the spec. See https://github.com/httpwg/http-extensions/issues/1867 

I am developing a library called bobcats in Scala that compiles both to JS and to Java and runs the same tests.  I have been testing bobcats by developing tests take from the Message Signing spec.

Using the same data, the Java tests pass, but  the JS tests don't  when using the new keys placed  in [object `test-key-rsa-pss` extends TestKeyPair](https://github.com/bblfish/bobcats/blob/26e191d236b27ad6e21ea497238097aa8e947117/core/shared/src/test/scala/bobcats/HttpMessageSignaturesV13.scala#L175).  

When that code is commented out, and instead we use the  `test-key-rsa-pss` from draft 07, as shown in the code above,  then all the tests pass. 

So there is something in the given PEM keys that the JS Web Crypto APIs don't like. I am sure it would help implementations in JS browser land if it was not so difficult to get the PEM key working. There must be some explanation to what the problem is.

Browsers tested are Chrome and Firefox using Selenium. See for example [Chrome log](https://github.com/bblfish/bobcats/actions/runs/3397551264/jobs/5649813316#step:12:65) in PR https://github.com/bblfish/bobcats/pull/7

jricher:
- make `@query` and `@path` use raw values before percent-decoding
- add trailer fields and clarify list-based field combinations (Closes #2277)
- add IANA section for registering new header field names
- add warning note about content digest depending on encoding (Closes #2279)
- add keys in JWK format (Closes #2290)


jricher:
@bblfish : please check these JWK-formatted keys for all the test vectors

bblfish:
You may have already dealt with these typos but I'll just post those I found


* p20:  double "the"
  > The component value of a single named parameter is the the
* 2.2.8 as a language error:
  it looks like "in" is too much or a word is missing on p 25
  > Indicating the baz, qux and param named query parameters in would
  > result in the following @query-param component values:
* 2.1.0 page 15  "are be defined" ("should be..."?)
  > Additional parameters are be defined

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1809,6 +1865,8 @@ From here, the signing process proceeds as usual.
 
 Upon verification, it is important that the verifier validate not only the signature but also the value of the Content-Digest field itself against the actual received content. Unless the verifier performs this step, it would be possible for an attacker to substitute the message content but leave the Content-Digest field value untouched to pass the signature. Since only the field value is covered by the signature directly, checking only the signature is not sufficient protection against such a substitution attack.
 
+As discussed in {{DIGEST}}, the value of the Content-Digest field is dependent on the content encoding of the message. If an intermediary changes the content encoding, the resulting Content-Digest value would change, which would in turn invalidate the signature. Any intermediary performing such an action would need to apply a new signature with the updated Content-Digest field value, similar to the reverse proxy use case discussed in {{signature-multiple}}.
```
ioggstream, line 1868: @jricher Here we are implying that not signing content-encoding means that there is no content coding. @reschke do you think it is a reasonable assertion? In general, this could be a slippery slope...

```suggestion
As discussed in {{DIGEST}}, the value of the Content-Digest field is dependent on the content encoding of the message. 
The above exchange implies that when the Content-Encoding is not included in the signature, then no content coding is applied.
If an intermediary changes the content encoding, the resulting Content-Digest value would change, which would in turn invalidate the signature. Any intermediary performing such an action would need to apply a new signature with the updated Content-Digest field value, similar to the reverse proxy use case discussed in {{signature-multiple}}.
```

ioggstream (commented):
@jricher pls see comments and a minor tweak. Not sure this addresses the content-digest concerns.

draft-ietf-httpbis-message-signatures.md
reschke: Not sure what the question is.

If Content-Digest is signed, and somebody changes the content by changing the content encoding, the signature will be invalid, right? Isn't that what we want here?

draft-ietf-httpbis-message-signatures.md
```diff
@@ -237,22 +237,26 @@ The following sections define component identifier names, their parameters, thei
...
 
-Unless overridden by additional parameters and rules, the HTTP field value MUST be canonicalized as a single combined value as defined in {{Section 5.2 of HTTP}}. Note that some HTTP fields, such as Set-Cookie {{COOKIE}}, do not follow a syntax that allows for combination of field values in this manner such that the combined output is unambiguous from multiple inputs. However, the canonicalized component value is never parsed by the message signature process; instead, it is merely used as part of the signature base in {{create-sig-input}}. Even so, caution needs to be taken when including such fields in signatures, and the `bs` parameter defined in {{http-header-byte-sequence}} provides a method for wrapping such problematic fields. See {{security-non-list}} for more discussion of this issue.
+The component value for an HTTP field is the field value for the named field as defined in {{Section 5.5 of HTTP}}. The field value MUST be taken from the named header field of the target message unless this behavior is overridden by additional parameters and rules, such as the `req` and `tr` flags, below.
```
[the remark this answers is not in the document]
reschke: "header field" is correct here.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -237,22 +237,26 @@ The following sections define component identifier names, their parameters, thei
...
 
-If the combined value is not available for a given field by an implementation, the following algorithm will produce canonicalized results for an implementation:
+Unless overridden by additional parameters and rules, list-based HTTP field values MUST be combined into a single value as defined in {{Section 5.2 of HTTP}}. Specifically, list-based HTTP fields sent as multiple fields MUST be combined using a single comma (",") and a single space (" ") between each item. Note that intermediaries are allowed to separate values of list-based HTTP fields with any amount of whitespace between commas. If this behavior is not accounted for by the verifier, the signature can fail since the addition or removal of spaces between list items will change the signature base. It is RECOMMENDED that signers and verifiers process list-based fields starting with the individual field values based on the strict algorithm below, where possible, to account for possible intermediary behavior.
```
reschke, line 245: I believe the only robust way to sign a field is to actually make sure that only one instance of the field is present in the signed message (thus canonicalize not only the signed value, but the actual message).

draft-ietf-httpbis-message-signatures.md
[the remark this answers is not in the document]
jricher: Yes, with no flags it's specifically a header field of the target message. the "tr" flag means it's a trailer, the "req" flag means it's the related request message.

reschke:
As a general comment: it would be awesome if this PR could be split to address the individual issues separately.

draft-ietf-httpbis-message-signatures.md
jricher: No, that's not the implication -- if Content-Encoding is not included in the signature, it doesn't mean that it's not applied, it only means that the encoding type is not signed. It could still be applied to the message if the field is there. And if it gets changed, the Content-Digest also gets changed, which fails the signature. This is desired behavior but something we want to warn implementors about, which is what the existing text does.

jricher:
@reschke I looked at that (and generally prefer that approach) but the trailer addition was intertwined with a few things so I opted to put everything into one batch instead to avoid conflicts.

draft-ietf-httpbis-message-signatures.md
jricher: @reschke done, please check the new text to see if it aligns with your point

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1886,7 +1944,7 @@ Such verifiers also need to ensure that any differences in message component con
...
+The definition of HTTP field names does not allow for the use of the `@` character anywhere in the name. As such, since all derived component names start with the `@` character, these namespaces should be completely separate. However, some HTTP implementations are not sufficiently strict about the characters accepted in HTTP fields. In such implementations, a sender (or attacker) could inject a header field starting with an `@` character and have it passed through to the application code. These invalid header fields could be used to override a portion of the derived message content and substitute an arbitrary value, providing a potential place for an attacker to mount a [signature collision](#security-collision) attack or other functional substitution attack (such as using the signature from a GET request on a crafted POST request).
```
reschke, line 1947:
```suggestion
The definition of HTTP field names does not allow for the use of the `@` character anywhere in the name. As such, since all derived component names start with the `@` character, these namespaces should be completely separate. However, some HTTP implementations are not sufficiently strict about the characters accepted in HTTP field names. In such implementations, a sender (or attacker) could inject a header field starting with an `@` character and have it passed through to the application code. These invalid header fields could be used to override a portion of the derived message content and substitute an arbitrary value, providing a potential place for an attacker to mount a [signature collision](#security-collision) attack or other functional substitution attack (such as using the signature from a GET request on a crafted POST request).
```

draft-ietf-httpbis-message-signatures.md
reschke: There are stimm multiple issues here:

- this not just about "list based" fields, but any case where there are multiple instances of the same field name - that may soundlike nitpicking, but the combination rule applies to *any* field
- "Note that intermediaries are allowed to separate values of list-based HTTP fields with any amount of whitespace between commas." - that sounds a bit like if this was about uncombining lists, but that's no whay you mean here , right?
- "For robustness, it is RECOMMENDED that messages include only a single value for any field covered under the signature, with the value for list-based fields serialized using the algorithm below." - once it's a single value, the actua algorithm doesn't matter at all.
- "This approach increases the chances of the field value remaining untouched through intermediaries." - it sort of guarantees it, unless the intermediary rewrites values it shouldn't touch (in which case we want the sig to fail)
- "Where that approach is not possible and multiple field lines need to be sent separately, it is RECOMMENDED that signers and verifiers process list-based fields starting with the individual field values based on the strict algorithm below, where possible, to account for possible intermediary behavior." - that makes it sound as if that would help - but if the intermediary combines in a different way (such as no SP), the sig will definitively fail.

reschke: No, intermediaries are not supposed to rewrite field values. In particular, an intermediary cannot assume that a comma is indeed a list delimiter in general -- consider quoted strings. Or ETags.

jricher: I've taken another shot at this -- please take a look at the new text.

draft-ietf-httpbis-message-signatures.md
ioggstream: In the above example we have that:

1. content-digest is signed
2. content-encoding is not present
3. content-encoding is not signed

If I add a content-encoding header, the signature is still valid, since content-encoding is not signed and the client does not know whether:

1. content-encoding in the original request was present but not signed by accident
2. content-encoding was missing and the request was tampered

bblfish:
I have implemented the tests with JWK in the bobcats library in this commit https://github.com/bblfish/bobcats/pull/9/commits/450d29918e25760e8222de1d1fb10413051ce5df

* Java compiles for all except for the ed25519 key, even though the tests work for the PEM key. This is because I am using Nimbus jose JWT, which does not implement the transformation of ed25519 to java.security because java only added that in jdk15, and nimbus wants to be backward compatible to jdk8 in part because of Android. (see [nimbus issue 359](https://bitbucket.org/connect2id/nimbus-jose-jwt/issues/359/add-export-to-java-keypair-from) and others).
* for JS: ll tests pass (if I remove ed25519). BUT I have to first filter out all the private key fields from the JSON you submitted. (search for `filterKeys` in the commit above).

draft-ietf-httpbis-message-signatures.md
jricher: Correct, you cannot sign an absent header. That feature was discussed at length and rejected by the wg. The original message could have a content encoding header, whether or not that's covered by the signature. Absence from the signature does not imply absence from the message, as you claim above. These are independent.

If you add a content encoding header, and this changes the content, the signature is not valid because the digest is no longer valid - while the signature itself would pass the digest would not, as is pointed out in this section. Now this is assuming you change the content and don't just add a meaningless header. If you don't change the content but claim a different encoding, this should fail when the message is processed because the encoding makes no sense. Signature and digest won't fix a badly formed http message.

In all of these cases, I believe the system is behaving as intended and failing where we'd expect it to fail, and how we expect it to fail. So, I am with @reschke here in that I don't see what the issue is with this situation.

jricher:
@bblfish thanks. We explicitly say that the jwk format is for the key pair. Many libraries, including nimbus, have a method of easily extracting a public key from a key pair object, so to me it doesn't make sense to list that separately.

draft-ietf-httpbis-message-signatures.md
LPardue: I'm with Julian and Justin here. Not signing a field is tangential to that field applying to the message.

Signatures doesn't support covering the absence of a field. And it's sad that there's no formal definition that allows "Content-encoding: nothing". But these are things we are stuck with. The rough edge, of signing content-digest and having someone screw up the message isn't nice but will fail closed and will be quickly noticeable. The workaround is gzip (or whatever) the content if your application really can't live this sharp edge, then you have a proper content-encoding to send.

 I don't think Signstures needs to do anything special here. There's lots of ways applications can screw up signing unrelated to content digest.

bblfish:
Ok, I have succeeded in adding support for ED25519 in Java with commit https://github.com/bblfish/bobcats/pull/9/commits/ef158576d44284baca8979edd3eaa1ecc5830dc4 and the tests now all pass. 
https://github.com/bblfish/bobcats/pull/9

jricher:
Merging based on overall positive feedback.

draft-ietf-httpbis-message-signatures.md
ioggstream: I agree that the issue is we can't say `content-encoding: nothing`. The example could use a content-encoding as a workaround.

Issue #2282, opened 2022-11-03
This document aspires to standardize two HTTP header fields


LPardue:
As Yoda might say, do or do not. We can lose the aspiration here and be assertive about the document consensus progress.

bc-pi:
I have a tendency towards that kind of demure writing. But you and Yoda are right. Will change.

PR #2305, opened 2022-11-09
lose the aspiration for #2282

b---c:
and use "describes" rather than "standardize" to better align w/ the Informational intended status


LPardue (approved):
LGTM

MikeBishop (commented):
LGTM, too.

Issue #2281, opened 2022-11-03
Introduction discussion on non-standard fields seems a bit long winded


LPardue:
Intro para 3 has a chunk of text about fields not defined in this doc

> A common way this information is conveyed in practice today is by using non-standard fields to carry the certificate (in some encoding) or individual parts thereof in the HTTP request that is dispatched to the origin server. This solution works but interoperability between independently developed components can be cumbersome or even impossible depending on the implementation choices respectively made (like what field names are used or are configurable, which parts of the certificate are exposed, or how the certificate is encoded). A well-known predictable approach to this commonly occurring functionality could improve and simplify interoperability between independent implementations.

This could be condensed to say that the header fields in this document provide a standardized replacement for custom or proprietary fields that were known to be used prior to this document.

bc-pi:
It is perhaps a bit long winded and admittedly not award winning literature. But I think it's useful context for the reader and prefer to keep it.

LPardue:
Right but in 20 years if the RFC says "a common practice today is using non-standard fields" it could read odd. 

Its clearer if this spec focuses on what it provides  not what people came up with as custom solutions beforehand.

PR #2307, opened 2022-11-09
remove or rephrase the use of the word "today" (for #2281 kinda)

b---c:


draft-ietf-httpbis-client-cert-field.md
```diff
@@ -87,7 +87,7 @@ specific details from the certificate needed also vary with the application
 requirements. In order for these types of application deployments to work in
 practice, the reverse proxy needs to convey information about the client
 certificate to the origin application server. A common way this information is
-conveyed in practice today is by using non-standard fields to carry the
+conveyed in practice at the time of writing is by using non-standard fields to carry the
```
LPardue, line 90: I like this. I might restructure the sentence to start with the statement e.g. "at the time of writing, a common way to convey this information is to use non-atandard fields..."

b---c: I almost did exactly that but opted for a more narrow change. I'm happy to take the suggestion though. 

Overall, I was hopeful that this much smaller change would still alleviate (at least some of) your concern about the text reading odd in the future.

b---c: [095b564](https://github.com/httpwg/http-extensions/pull/2307/commits/095b564a712df2801765b8c04a6317f285782b7b) does that restructure

LPardue (approved):
Thanks, looks good!

PR #2308, opened 2022-11-10
reword the header compression section (for #2287)

### LINKED ISSUE #2287: The subjects of Section 3.1 could be clearer
This section talks about header compression and says

> If the client certificate header field is generated by an intermediary on a connection that compresses fields (e.g., using HPACK [[HPACK](https://httpwg.org/http-extensions/draft-ietf-httpbis-client-cert-field.html#HPACK)] or QPACK [[QPACK](https://httpwg.org/http-extensions/draft-ietf-httpbis-client-cert-field.html#QPACK)]) and more than one client's requests are multiplexed into that connection, it can reduce compression efficiency significantly, due to the typical size of the field value and its variation between clients. Recipients that anticipate connections with these characteristics can mitigate the efficiency loss by increasing the size of the dynamic table. If a recipient does not do so, senders may find it beneficial to always send the field value as a literal, rather than entering it into the dynamic table.

Elsewhere in the document, you've used TTRP and Origin to identify the roles in play. I think it would be clearer to use them here. For instance, I might rewrite this as something like

> If the connection between the TTRP and origin is capable of field compression (e.g. HPACK or QPACK), and the TTRP multiplexes more than one client's requests into that connection, the size and variation of Client-Cert and Client-Cert-Chain field values can reduce compression efficiency significantly. An origin could mitigate the efficiency loss by increasing the size of the dynamic table. If the TTRP determines that the origin dynamic table is not sufficiently large, it may find it beneficial to always send the field value as a literal, rather than entering it into the table.

b---c:
reword the header compression section with more consistent terms for the roles 

(to fix #2287)


draft-ietf-httpbis-client-cert-field.md
```diff
@@ -275,15 +275,14 @@ employ the `Client-Cert` or `Client-Cert-Chain` header field in requests.
...
-compresses fields (e.g., using HPACK {{?HPACK=RFC7541}} or QPACK {{?QPACK=RFC9204}})
-and more than one client's requests are multiplexed into that connection, it can reduce
-compression efficiency significantly, due to the typical size of the field value and
-its variation between clients.
-Recipients that anticipate connections with these characteristics can mitigate the
-efficiency loss by increasing the size of the dynamic table.
-If a recipient does not do so, senders may find it beneficial to always send the
-field value as a literal, rather than entering it into the dynamic table.
+If the connection between the TTRP and origin is capable of field compression
+(e.g., HPACK {{?HPACK=RFC7541}} or QPACK {{?QPACK=RFC9204}}, and the TTRP multiplexes more
```
LPardue, line 279: missing closing bracket
```suggestion
(e.g., HPACK {{?HPACK=RFC7541}} or QPACK {{?QPACK=RFC9204}}), and the TTRP multiplexes more
```

LPardue (approved):

LPardue:
LGTM modulo nit

Issue #2288, opened 2022-11-03
Header Block Size


LPardue:
`3.2 Header Block Size`

Is this a hangover from terminology changes elsewhere? https://www.rfc-editor.org/rfc/rfc9114.html#section-4.2.2 says 

> An HTTP/3 implementation MAY impose a limit on the maximum size of the message header it will accept on an individual HTTP message. A server that receives a larger header section than it is willing to handle can send an HTTP 431 (Request Header Fields Too Large) status code ([[RFC6585](https://www.rfc-editor.org/rfc/rfc9114.html#RFC6585)]).

bc-pi:
Very possible that it's a hangover from terminology changes elsewhere. Very possible. But I'm not quite sure how to update it.  HTTP/3 talks about the "size of the message header" (from your reference) while HTTP/2 talks about block size but now uses field block size https://www.rfc-editor.org/rfc/rfc9113#name-limits-on-field-block-size rather than header block size https://www.rfc-editor.org/rfc/rfc9113#section-4.3-3.1

How about the following though? Assuming I'm even adjusting the terminology you're suggesting. Message header seems more neutral and appropriate. 


> 3.2. Message Header Size
>
> A server in receipt of a larger message header than it is willing to handle can send an HTTP 431 (Request Header Fields Too Large) status code per Section 5 of [RFC6585]. Due to the typical size of the field values containing certificate data, recipients may need to be configured to allow for a larger maximum header size. An intermediary generating client certificate header fields on connections that allow for advertising the maximum acceptable header size (e.g. HTTP/2 [HTTP/2] or HTTP/3 [HTTP/3]) should account for the additional size of the header of the requests it sends vs. requests it receives by advertising a value to its clients that is sufficiently smaller so as to allow for the addition of certificate data.`

b---c:
Makes sense. I, for one, would welcome such expertise.

b---c:
https://github.com/httpwg/http-extensions/pull/2309#pullrequestreview-1201543768 Mike okayed it

PR #2309, opened 2022-11-10
reword 'Header Block Size' section

b---c:
Attempt to reword the 'Header Block Size' section to use more HTTP version neutral terms (for #2288)


MikeBishop (approved):
Yes, this looks reasonable.

Issue #2285, opened 2022-11-03
Style guide - Use of ABNF for structured field


LPardue:
See https://httpwg.org/admin/editors/style-guide#structured-fields, which would affect sections 2.2 and 2.3

bc-pi:
Will update the structured fields parts to align with the style guide.

PR #2311, opened 2022-11-10
align better with the style guide wrt structured fields (for #2285)

b---c:
try to align better with the style guide on structured fields usage
for #2285


LPardue (approved):
LGTM

PR #2323, opened 2022-11-23
message-signatures: "addition" of whitespace of obs-fold IMHO not allowed

reschke:


jricher:
The spec should still be robust against errant addition if it happens, but I'll let others tag in to say if this is something we should avoid calling out. @mnot, @martinthomson -- opinions?

reschke:
In which case it shouldn't be listed below "As mentioned earlier, HTTP explicitly permits and in some cases requires implementations to transform messages in a variety of ways. Implementations are required to tolerate many of these transformations."

reschke:
And, BTW, if "let's repair it when it's broken" is the goal... There are other places where the spec asks for fixes just for HTTP/1.1, although an invalid HTTP/2 message might have the same problem (obs-fold removal).

martinthomson:
I'm not sure what the right answer is here.  The specification appears to be resilient against addition, even if we don't think that addition is a good idea.  Can't we leave as-is, even if this shouldn't happen?

reschke:
I'm ok with resilience if we are consistent about it, and do not present it as something that addresses something HTTP allows.

reschke:
Justin - please consider the "if we are consistent about it, and do not present it as something that addresses something HTTP allows" part...

PR #2325, opened 2022-11-23
message-digest: fix typo

reschke:


draft-ietf-httpbis-message-signatures.md
```diff
@@ -245,7 +245,7 @@ The component name for an HTTP field is the lowercased form of its field name as
...
..." ") between each item. Note that intermediaries are allowed to combine values of HTTP fields with any amount of whitespace between the commas, and if this behavior is not accounted for by the verifier, the signature can fail since the signer and verifier will be see a different component value in their respective signature bases. For robustness, it is RECOMMENDED that signed messages include only a single instance of any field covered under the signature, particularly with the value for any list-based fields serialized using the algorithm below. This approach increases the chances of the field value remaining untouched through intermediaries. Where that approach is not possible and multiple instances of a field need to be sent separately, it is RECOMMENDED that signers and verifiers process any list-based fields by parsing out individual field values and combining them based on the strict algorithm below, to counter possible intermediary behavior. When the field in question is a structured field of style List or Dictionary, this effect can be accomplished more directly by using the strict structured field serialization of the field value, as described in {{http-field-structured}}.
```
jricher, line 248: It was actually supposed to be "type" not "style":

```suggestion
Unless overridden by additional parameters and rules, HTTP field values MUST be combined into a single value as defined in {{Section 5.2 of HTTP}} to create the component value. Specifically, HTTP fields sent as multiple fields MUST be combined using a single comma (",") and a single space (" ") between each item. Note that intermediaries are allowed to combine values of HTTP fields with any amount of whitespace between the commas, and if this behavior is not accounted for by the verifier, the signature can fail since the signer and verifier will be see a different component value in their respective signature bases. For robustness, it is RECOMMENDED that signed messages include only a single instance of any field covered under the signature, particularly with the value for any list-based fields serialized using the algorithm below. This approach increases the chances of the field value remaining untouched through intermediaries. Where that approach is not possible and multiple instances of a field need to be sent separately, it is RECOMMENDED that signers and verifiers process any list-based fields by parsing out individual field values and combining them based on the strict algorithm below, to counter possible intermediary behavior. When the field in question is a structured field of type List or Dictionary, this effect can be accomplished more directly by using the strict structured field serialization of the field value, as described in {{http-field-structured}}.
```

PR #2342, opened 2022-12-03
Move ABNF to an appendix

### LINKED ISSUE #2338: ABNF
As discussed during adoption, we want to reconsider our use of ABNF in this specification.

Uncontroversial was removing the recommendation to define new fields using ABNF; that can be easily accommodated and I'll do that in a commit to `main` shortly.

More discussion is needed about completely removing ABNF from the definitions of the types, and other potential changes to the use of ABNF.

mnot:
Fixes #2338.

### DIFF
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -958,7 +958,7 @@ The serialization algorithm is defined in a way that it is not strictly limited 
 Implementations are allowed to limit the size of different structures, subject to the minimums defined for each type. When a structure exceeds an implementation limit, that structure fails parsing or serialization.
 
 
-# Informative ABNF {#abnf}
+# ABNF {#abnf}
 
 This section uses the Augmented Backus-Naur Form (ABNF) notation {{?RFC5234}} to illustrate expected syntax of Structured Fields. In doing so, it uses the VCHAR, SP, DIGIT, ALPHA, and DQUOTE rules from {{?RFC5234}}. It also includes the tchar and OWS rules from {{!RFC7230}}.
 
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -109,11 +109,9 @@ Note that as a result of this strictness, if a field is appended to by multiple 
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}} when, and only when, they appear in all capitals, as shown here.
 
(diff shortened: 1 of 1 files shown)


reschke (commented):
As stated before, I don't like that change and I believe it makes the spec less readable.

On top of that, we are loosing information (description of sf-binary). It also is a bit strange that we still refer to HTTP's token definition in the introduction for sf-token.

If we move the ABNF *and* the introductions are incomplete / inprecise, all hat is left are the algorithms. That may be nice for people who write parsers/serializers, but is a disservice for everybody else.

mnot:
> On top of that, we are loosing information (description of sf-binary).

Where?

> It also is a bit strange that we still refer to HTTP's token definition in the introduction for sf-token.

That's a convenience for those who are used to that sense of 'token'.

> If we move the ABNF and the introductions are incomplete / inprecise, all hat is left are the algorithms. That may be nice for people who write parsers/serializers, but is a disservice for everybody else.

By definition we're not loosing anything -- the ABNF is being moved, not lost.

reschke:
> >   On top of that, we are loosing information (description of sf-binary).
> Where?

Sorry, missed a sentence.

But for "token", we do not say what a "token" character is (except by pointing to the "token" ABNF in HTTP).

> By definition we're not loosing anything -- the ABNF is being moved, not lost.

Yes, but it turns out that the introductions aren't complete anymore (see the discussion about how list elements are separated).

draft-ietf-httpbis-sfbis.md
```diff
@@ -113,11 +113,9 @@ Note that as a result of this strictness, if a field is appended to by multiple
...
+This document uses algorithms to specify parsing and serialization behaviors. When parsing from HTTP fields, implementations MUST have behavior that is indistinguishable from following the algorithms.
 
-When parsing from HTTP fields, implementations MUST have behavior that is indistinguishable from following the algorithms. If there is disagreement between the parsing algorithms and ABNF, the specified algorithms take precedence.
-
-For serialization to HTTP fields, the ABNF illustrates their expected wire representations, and the algorithms define the recommended way to produce them. Implementations MAY vary from the specified behavior so long as the output is still correctly handled by the parsing algorithm described in {{text-parse}}.
+For serialization to HTTP fields, the algorithms define the recommended way to produce them. Implementations MAY vary from the specified behavior so long as the output is still correctly handled by the parsing algorithm described in {{text-parse}}.
```
martinthomson, line 118: Is this better as a single paragraph?  Ordinarily, I would split the different concepts, but your first sentence says "parsing and serialization", then only parsing is treated further in that paragraph.  Now that you have less stuff here, maybe they can fold up.

martinthomson (approved):

draft-ietf-httpbis-sfbis.md
```diff
@@ -1056,12 +958,64 @@ The serialization algorithm is defined in a way that it is not strictly limited
 Implementations are allowed to limit the size of different structures, subject to the minimums defined for each type. When a structure exceeds an implementation limit, that structure fails parsing or serialization.
 
 
+# Informative ABNF {#abnf}
```
tfpauly, line 961: Does the title need to say "Informative ABNF"? It already says it is non-normative below. How about just "ABNF Examples" or "ABNF Illustrations"?

reschke: ...or just "ABNF".

tfpauly: That too =)

mnot, line 961:
```suggestion
# ABNF {#abnf}
```

The little button up there lets you make suggestions :)

PR #2360, opened 2023-01-05
Remove trailing dot in CNAME examples

### LINKED ISSUE #2359: draft-ietf-httpbis-alias-proxy-status: Omit trailing dots from FQDNs
The draft currently has this example:

```
Proxy-Status: proxy.example.net; next-hop=2001:db8::1
    next-hop-aliases="tracker.example.com.,service1.example-cdn.com."
```

I'd suggest always omitting trailing dots since it's clear that these names will be FQDNs, i.e.:

```
Proxy-Status: proxy.example.net; next-hop=2001:db8::1
    next-hop-aliases="tracker.example.com,service1.example-cdn.com"
```

tfpauly:
Closes #2359

### DIFF
--- a/draft-ietf-httpbis-alias-proxy-status.md
+++ b/draft-ietf-httpbis-alias-proxy-status.md
@@ -79,11 +79,11 @@ For example:
 
 ~~~ example
 Proxy-Status: proxy.example.net; next-hop=2001:db8::1;
-    next-hop-aliases="tracker.example.com.,service1.example-cdn.com."
+    next-hop-aliases="tracker.example.com,service1.example-cdn.com"
 ~~~
 
 indicates that proxy.example.net, which used the IP address "2001:db8::1" as the next hop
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-alias-proxy-status.md
```diff
@@ -79,11 +79,11 @@ For example:
 
 ~~~ example
 Proxy-Status: proxy.example.net; next-hop=2001:db8::1;
-    next-hop-aliases="tracker.example.com.,service1.example-cdn.com."
+    next-hop-aliases="tracker.example.com,service1.example-cdn.com"
 ~~~
 
 indicates that proxy.example.net, which used the IP address "2001:db8::1" as the next hop
-for this request, encountered the CNAMEs "tracker.example.com." and "service1.example-cdn.com"
+for this request, encountered the CNAMEs "tracker.example.com" and "service1.example-cdn.com"
```
gibson042, line 86: I don't think this dot should have been removed, and in fact the whole sentence should be more precise.

PR #2375, opened 2023-01-18
Formalize percent encoding for names in list

### LINKED ISSUE #2357: draft-ietf-httpbis-alias-proxy-status: Percent encoding text is not quite right
The current text says

> Since DNS names can include comma (`,`) characters in them, any commas that appear in a DNS
names MUST be represented using a percent-encoded `%2C` value instead. 

However, domain names can also contain `"` and `%`, so I think the escaping rule has to be a bit more generic than just escaping commas, as indicated in the current text.

tfpauly:
Closes #2357

### DIFF
--- a/draft-ietf-httpbis-alias-proxy-status.md
+++ b/draft-ietf-httpbis-alias-proxy-status.md
@@ -79,12 +79,9 @@ The list of DNS names in `next-hop-aliases` use a comma (",") as a separator bet
 DNS names normally just contain alphanumeric characters and hyphens ("-"), although they
 are allowed to contain any character {{?RFC1035, Section 3.1}}, including a comma. To
 prevent commas or other special characters in names leading to incorrect parsing,
-any URI Reserved Characters {{!RFC3986, Section 2.2}} as well as "%" that appear in names
-in this list MUST be percent-encoded as defined in {{!RFC3986, Section 2.1}}.
-Double quote characters and backslash characters use the escaping defined
-in Structured Field Strings {{!STRUCTUREDFIELDS=RFC8941, Section 3.3.3}}.
-Any names that include characters not allowed in Structured Field Strings, such
-as newlines or tabs, MUST NOT be included in the list.
+any characters that appear in names in this list that do not belong to the set of URI
+Unreserved Characters {{!RFC3986, Section 2.3}} MUST be percent-encoded as
+defined in {{!RFC3986, Section 2.1}}.
 
 For example, consider a proxy "proxy.example.net" that receives the following records when
 performing DNS resolution for the next hop "host.example.com":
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-alias-proxy-status.md
```diff
@@ -75,6 +75,13 @@ in the chain, the first name in the `next-hop-aliases` list would be the value i
 record for the original hostname, and the final name in the `next-hop-aliases` list would
 be the name that ultimately resolved to one or more addresses.
 
+The list of DNS names in `next-hop-aliases` use a comma (",") as a separator between names.
+DNS names normally just contain alphanumeric characters and hyphens ("-"), although they
+are allowed to contain any character {{?RFC1035, Section 3.1}}, including a comma. To
+prevent commas or other special characters in names leading to incorrect parsing,
+any URI Reserved Characters {{!RFC3986, Section 2.2}} as well as "%" that appear in names
+in this list MUST be percent-encoded as defined in {{!RFC3986, Section 2.1}}.
```
tfpauly, line 83:
```suggestion
in this list MUST be percent-encoded as defined in {{!RFC3986, Section 2.1}}.
Double quote characters and backslash characters use the escaping defined
in Structured Field Strings {{!STRUCTUREDFIELDS=RFC8941, Section 3.3.3}}.
Any names that include characters not allowed in Structured Field Strings, such
as newlines or tabs, MUST NOT be included in the list.
```

draft-ietf-httpbis-alias-proxy-status.md
```diff
@@ -75,6 +75,17 @@ in the chain, the first name in the `next-hop-aliases` list would be the value i
...
+Double quote characters and backslash characters use the escaping defined
+in Structured Field Strings {{!STRUCTUREDFIELDS=RFC8941, Section 3.3.3}}.
+Any names that include characters not allowed in Structured Field Strings, such
+as newlines or tabs, MUST NOT be included in the list.
+
```
tfpauly, line 88:
```suggestion
The list of DNS names in `next-hop-aliases` use a comma (",") as a separator between names.
DNS names normally just contain alphanumeric characters and hyphens ("-"), although they
are allowed to contain any character {{?RFC1035, Section 3.1}}, including a comma. To
prevent commas or other special characters in names leading to incorrect parsing,
any characters that appear in names in this list that do not belong to the set of URI
Unreserved Characters {{!RFC3986, Section 2.3}} MUST be percent-encoded as
defined in {{!RFC3986, Section 2.1}}.

```

edmondsfastly (approved):
The unreserved characters are:

```
      unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
```

`.` is a valid interior label character in a wire-format DNS name, which seems like it would allow spoofing a label boundary. Not sure how much of a problem this is.

tfpauly:
Agreed, but I don't see a way around having a "." be allowed there. If a system wants to further escape dots within labels if they know about them, that's OK...

PR #2376, opened 2023-01-19
Add instructions for IANA review

### LINKED ISSUE #2366: Missing expert guidelines
_Comment by @fpalombini_

The new registries do not have any guidance for the designated experts as to what registrations should be accepted or denied (conformance with template, point squatting, availability of reference specification, etc). See [section 4.5](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-4.5) of RFC8126:

The required documentation and review criteria, giving clear guidance to the designated expert, should be provided when defining the registry.  It is particularly important to lay out what should be considered when performing an evaluation and reasons for rejecting a request.

### LINKED ISSUE #2367: Define deprecated
_Comment by @fpalombini_

I would suggest being explicit about the meaning of "active" and especially "deprecated": these are terms everywhere in RFCs (I am aware), but sometimes they are open for interpretation, and we keep discussing their meaning. It would be good to avoid that situation from this doc.

jricher:
Closes #2366
Closes #2367


draft-ietf-httpbis-message-signatures.md
```diff
@@ -1595,11 +1595,11 @@ IANA is asked to update the
...
+The Designated Expert (DE) is expected to ensure that the algorithms referenced by a registered algorithm identifier are fully defined with all parameters (such as salt, hash, required key length, etc) are fixed by the defining text. The DE is expected to ensure that the algorithm definition fully specifies the `HTTP_SIGN` and `HTTP_VERIFY` primitive functions, including all how all defined inputs and outputs map to the underlying cryptographic algorithm. The DE is expected to reject any registrations that are aliases of existing registrations. The DE is expected to ensure all registrations follow the template presented in {{iana-hsa-template}}, including that the length of the name is not excessive while still being unique and recognizable. When setting a registered item's status to "Deprecated", the DE should ensure that a reason for the deprecation is documented, along with instructions for moving away from the deprecated functionality.
```
yaronf, line 1600:
```suggestion
The Designated Expert (DE) is expected to ensure that the algorithms referenced by a registered algorithm identifier are fully defined with all parameters (such as salt, hash, required key length, etc) are fixed by the defining text. The DE is expected to ensure that the algorithm definition fully specifies the `HTTP_SIGN` and `HTTP_VERIFY` primitive functions, including how all defined inputs and outputs map to the underlying cryptographic algorithm. The DE is expected to reject any registrations that are aliases of existing registrations. The DE is expected to ensure all registrations follow the template presented in {{iana-hsa-template}}, including that the length of the name is not excessive while still being unique and recognizable. When setting a registered item's status to "Deprecated", the DE should ensure that a reason for the deprecation is documented, along with instructions for moving away from the deprecated functionality.
```

jricher: Good catch, thanks!

martinthomson: s/with all parameters (...) are fixed by/with all parameters (...) fixed by/ or s/with all parameters (...) are fixed by/with all parameters (...) specified in/

PR #2377, opened 2023-01-19
Update processing of parameters in Accept-Signature

### LINKED ISSUE #2365: Accept clarification question
_Comment by @fpalombini_

[Section 5.1](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-5.1): >algorithm or key identifier. These parameters MUST NOT include parameters that the signer is expected to generate, such as the created parameter.

Why should it not? Couldn't the verifier include these empty parameters to indicate that the signer is expected to generate and send them?

jricher:
Update the semantics of parameters in Accept-Signature to be more fully defined and explicit.

Closes #2365

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -1545,17 +1545,28 @@ The sender of an Accept-Signature field MUST include only identifiers that are a
 
 ## The Accept-Signature Field {#accept-signature-header}
 
-The Accept-Signature field is a Dictionary structured field (defined in {{Section 3.2 of STRUCTURED-FIELDS}}) containing the metadata for one or more requested message signatures to be generated from message components of the target HTTP message. Each member describes a single message signature. The member's name is a label that uniquely identifies the requested message signature within the context of the target HTTP message. The member's value is the serialization of the desired covered components of the target message, including any allowed signature metadata parameters, using the serialization process defined in {{signature-params}}.
+The Accept-Signature field is a Dictionary structured field (defined in {{Section 3.2 of STRUCTURED-FIELDS}}) containing the metadata for one or more requested message signatures to be generated from message components of the target HTTP message. Each member describes a single message signature. The member's name is a label that uniquely identifies the requested message signature within the context of the target HTTP message.
+
+The member's value is the serialization of the desired signature parameters and covered components of the target message using the data structure and serialization process defined in {{signature-params}}.
 
 ~~~ http-message
 NOTE: '\' line wrapping per RFC 8792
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-message-signatures.md
```diff
@@ -1545,35 +1545,50 @@ The sender of an Accept-Signature field MUST include only identifiers that are a
...
+  keyid="test-key-rsa-pss";created;tag="app-123"
 ~~~
 
-The requested signature MAY include parameters, such as a desired algorithm or key identifier. These parameters MUST NOT include parameters that the signer is expected to generate, such as the `created` parameter.
+The list of component identifiers indicates the exact set of component identifiers to be included in the requested signature, including all applicable component parameters.
+
+The signature request MAY include signature metadata parameters that indicate desired behavior for the signer. The following behavior is defined by this specification:
+
+* `created`: The signer is requested to generate and include a creation time. This parameter has no associated value when sent as a signature request.
+* `expires`: The signer is requested to generate and include a creation time. This parameter has no associated value when sent as a signature request.
```
yaronf, line 1565:
```suggestion
* `expires`: The signer is requested to generate and include an expiration time. This parameter has no associated value when sent as a signature request.
```

yaronf:
I think the new text adds complexity and potential security issues. Specifically, the client/signer (which is in some cases the weakest link) now needs to handle a complicated request from the server and resolve potential conflicts. What if `keyid` does not match `alg`?

Some of these flags are not useful, since there's no harm done if the signer _always_ sends `created` and `expires`, and the verifier is free to respect or ignore them. But `alg` is actively harmful -- the algorithm should be determined by `keyid`. (Thinking of well known JWT security issues.)

jricher:
@yaronf the ability to send `alg` and `keyid` is not new with this text -- in fact those were the ones that were previously allowed, the other parameters were prohibited before. There's already a security consideration about choosing algorithms and matching the algorithm to the key ID, both of which the signer has to do :anyway: when signing a message -- this is not added or changed by the Accept-Signature field.

I do think that perhaps it warrants calling out specifically in a security consideration about Accept Signature, though, to hopefully prevent naive processing of Accept-Signature requests.

yaronf:
@jricher Yes, well, the example only included `keyid`.

The new text adds complexity to the client and IMHO does not add any value. The only two parameters that potentially make sense are `keyid` and `nonce`, and I suggest to restrict this header to only these two.

I disagree with @fpalombini: I would say that including `created` and `expired` here adds complexity. From the client's POV, there is no reason to make sending these values conditional and so no benefit in knowing that the verifier expects them. The client might as well _always_ send them.

PR #2379, opened 2023-01-19
AD Review of Signatures (Editorial)

### LINKED ISSUE #2330: message-signatures: names of HTTP versions
It's "HTTP/1.1", not "HTTP 1.1". Similar for the other versions.

### LINKED ISSUE #2333: message-signatures: clarify sentence about proxies
https://www.ietf.org/id/draft-ietf-httpbis-message-signatures-15.html#name-multiple-signatures

> The following non-normative example starts with a signed request from the client. A reverse proxy takes this request validates the client's signature.

Maybe "and" is missing here? Not sure.

### LINKED ISSUE #2364: IANA registry names inconsistencies
_Comment by @fpalombini_

[Section 2.3](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-2.3): >alg: The HTTP message signature algorithm from the HTTP Message Signature Algorithm Registry, as a String value.

[Section 3.2](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-3.2): >If the algorithm is explicitly stated in the signature parameters using a value from the HTTP Message Signatures registry, the verifier will use the referenced algorithm.

[Section 3.3](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-3.3): >through mutual agreement between the signer and verifier. Signature algorithms selected using the alg parameter MUST use values from the HTTP Message Signature Algorithms Registry ([Section 6.2](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-6.2)).

[Section 3.3.7](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-3.3.7): >algorithm values are not registered in the HTTP Message Signature Algorithms Registry ([Section 6.2](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-6.2)), and so the explicit alg signature

[Section 7.3.1](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-7.3.1): >used to sign HTTP messages. The HTTP Message Signatures Algorithm Registry is one source of trusted signature algorithms for applications to apply to their messages.

I believe all of the above should refer to HTTP Signature Algorithms Registry instead (I hope I caught all of them)

[Section 2.2](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-2.2): > Additional derived component names are defined in the HTTP Signatures Derived Component Name Registry. ([Section 6.4](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-6.4))

Should be HTTP Signature Derived Component Names Registry

[Section 2.3](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-2.3): > Additional parameters can be defined in the HTTP Signature Parameters Registry ([Section 6.3.2](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-6.3.2)). Note that there is no general ordering to

Should be HTTP Signature Component Parameters Registry - Also it would be better to reference 6.3 rather than 6.3.2.

### LINKED ISSUE #2368: Wrong ref
_Comment by @fpalombini_

[Section 2.1](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-2.1): > Additional parameters can be defined in the registry established in [Section 6.3](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-6.3)

I believe it should be 6.5. Also it would be good to name the HTTP Signature Component Parameters Registry explicitely.

### LINKED ISSUE #2371: Editorial Nits
_Comment by @fpalombini_

[Section 3.1](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-3.1): >several concrete applications of signing algorithms are defined in in [Section 3.3](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-3.3)

s/in in/in

[Section 3.3](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-3.3): >The following sections contain several common signature algorithms and demonstrates how these cryptographic primitives map to the

s/demonstrates/demonstrate

[Section 3.3.7](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-3.3.7): >other mechanisms common to JOSE implementations. In fact, and JWA algorithm values are not registered in the HTTP Message Signature Algorithms Registry ([Section 6.2](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-6.2)), and so the explicit alg signature

s/and JWA algorithm values/JWA algorithm values

[Section 5.1](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-5.1): >message signature. The member's name is label that uniquely identifies the requested message signature within the context of the

s/is label/is the label

[Section 7.4.3](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-7.4.3): >verifier has access to. Both signers and verifiers need to take carefully consider the source of all information when creating

s/take carefully/carefully

[Section 7.5.2](https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-15.html#section-7.5.2): >The verifier would have have a difficult time finding this class of errors since the value of the

s/have have/have

jricher:
Collects editorial comments of AD-reviewed signatures draft.

Closes #2371
Closes #2364
Closes #2368
Closes #2333
Closes #2330

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -285,16 +285,16 @@ The following example shows the component values for these example header fields
 "example-dict": a=1,    b=2;x=1;y=2,   c=(a   b   c)
 ~~~
 
-Since empty HTTP fields are allowed, they can also be signed when present in a message. The canonicalized value is the empty string. This means that the following empty header, with <SP> indicating a single space character:
+Since empty HTTP fields are allowed, they can also be signed when present in a message. The canonicalized value is the empty string. This means that the following empty header, with (SP) indicating a single trailing space character before the empty field value:
 
 ~~~ http-message
-X-Empty-Header:<SP>
+X-Empty-Header:(SP)
 ~~~
 
 Is serialized by the [signature base generation algorithm](#create-sig-input) with an empty string value following the colon and space added after the content identifier.
 
 ~~~
-"x-empty-header":<SP>
+"x-empty-header":(SP)
 ~~~
 
 Any HTTP field component identifiers MAY have the following parameters in specific circumstances, each described in detail in their own sections:
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -22,6 +22,7 @@ venue:
   mail: ietf-http-wg@w3.org
   arch: https://lists.w3.org/Archives/Public/ietf-http-wg/
   repo: https://github.com/httpwg/http-extensions/labels/signatures
+
 github-issue-label: signatures
 
 author:
@@ -563,7 +564,7 @@ And the following signature base line:
 
 ### Authority {#content-request-authority}
 
-The `@authority` derived component refers to the authority component of the target URI of the HTTP request message, as defined in {{HTTP, Section 7.2}}. In HTTP 1.1, this is usually conveyed using the Host header, while in HTTP 2 and HTTP 3 it is conveyed using the :authority pseudo-header. The value is the fully-qualified authority component of the request, comprised of the host and, optionally, port of the request target, as a string.
+The `@authority` derived component refers to the authority component of the target URI of the HTTP request message, as defined in {{HTTP, Section 7.2}}. In HTTP/1.1, this is usually conveyed using the Host header, while in HTTP/2 and HTTP/3 it is conveyed using the :authority pseudo-header. The value is the fully-qualified authority component of the request, comprised of the host and, optionally, port of the request target, as a string.
 The component value MUST be normalized according to the rules in {{HTTP, Section 4.2.3}}. Namely, the host name is normalized to lowercase and the default port is omitted.
 
 For example, the following request message:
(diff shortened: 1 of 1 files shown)


richanna:
Changes look good to me, but the editor's copy build failed.

jricher:
The failure seems to be in the rfc-http-validate tooling. This isn't failing on any other branch, so I'm not sure what's going on here to be honest. @martinthomson @mnot am I missing something? I thought we could use the `http-message` tag on message fragments, but that seems to be what's breaking from the build log here. It builds fine locally.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -284,26 +285,18 @@ The following example shows the component values for these example header fields
 "example-dict": a=1,    b=2;x=1;y=2,   c=(a   b   c)
 ~~~
 
-Since empty HTTP fields are allowed, they can also be signed when present in a message. The canonicalized value is the empty string. This means that the following empty header:
-
-~~~ http-message
-NOTE: '\' line wrapping per RFC 8792
-
-X-Empty-Header: \
+Since empty HTTP fields are allowed, they can also be signed when present in a message. The canonicalized value is the empty string. This means that the following empty header, with (SP) indicating a single trailing space character before the empty field value:
```
martinthomson, line 288:
```suggestion
Empty HTTP fields can also be signed when present in a message. The canonicalized value is the empty string. This means that the following empty header, with (SP) indicating a single trailing space character before the empty field value:
```

PR #2380, opened 2023-01-19
clarify combination of HTTP fields

### LINKED ISSUE #2324: message-signatures: requirement for field value concatenation
https://www.ietf.org/id/draft-ietf-httpbis-message-signatures-15.html#section-2.1-3

> Unless overridden by additional parameters and rules, HTTP field values MUST be combined into a single value as defined in [Section 5.2](https://rfc-editor.org/rfc/rfc9110#section-5.2) of [[HTTP](https://www.ietf.org/id/draft-ietf-httpbis-message-signatures-15.html#HTTP)] to create the component value. Specifically, HTTP fields sent as multiple fields MUST be combined using a single comma (",") and a single space (" ") between each item.

I don't quite get the reasoning for the second "MUST". Why does it matter? What would break if the client uses a single comma, which is also allowed by HTTP?

IMHO the exact delimiter sequence here does not matter at all, as long as it is allowed per HTTP. Note that HTTP allows *combining* field values, it does not allow splitting them afterwards.

### LINKED ISSUE #2326: message-signatures: requirement to 'separate' values
https://www.ietf.org/id/draft-ietf-httpbis-message-signatures-15.html#section-2.1-5

> 1. Create an ordered list of the field values of each instance of the field in the message, in the order that they occur (or will occur) in the message. If necessary, separate individual values found in a field instance.

This does not work. It's impossible to split multiple values (be it a list-based field or not) without knowledge of the field value's grammar.

For example, what is the canonicalized value for:

~~~
If-None-Match: W/"1,2", "3"
If-None-Match: "bar"
~~~

?

jricher:
Closes #2324
Closes #2326


reschke (approved):
(but I'll note that github's way of displaying diffs in long lines makes it extrremely hard to find out what actually changed)

PR #2383, opened 2023-01-20
Expand on how integrity preferences can be ignored

### LINKED ISSUE #2382: AD review: advice for when a client asks for digest and it doesn't get one
Murray asked:

>    General:

>    * I may be showing some ignorance of this space here, or I may have missed something, but I'll ask anyway: Is there any advice to be given for a situation where a client requests digest(s) but the server provides none, especially when that client has reason to expect that this specific server implements this specification?  i.e., "Hmm, I asked for the digest(s), and they ought to be there but aren't, that's weird..."

on the email thread, I responded with 

> We fall into the general realm of HTTP feature negotiation here, which has some well-trodden problems. If the client and server have some prior knowledge or OOB channel that lets them build an exception for digest, they can define their own rules for how to handle its absence - I don't want to go too far down the path of trying to describe that. I think what is useful is to highlight, with an editorial change, the very real possibility that an endpoint sends "want-*-digest" and the peer ignores it and doesn't provide any "*-digest" at all. Then all we should say is that dealing with this situation is an implementation decision.

To which Murray confirmed it would be a reasonable resolution.

Lets get this fixed as discussed

LPardue:
And how users of digests are the ones responsible for deciding what to
do if something is ignore.

Fixes #2382

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -417,6 +417,14 @@ If `Want-Content-Digest` or `Want-Repr-Digest` are used in a response, it
 indicates that the server would like the client to provide the respective
 Integrity field on future requests.
 
+Integrity preference fields are only a hint. The receiver of the field can
+ignore it and send an Integrity field using any algorithm or omit the field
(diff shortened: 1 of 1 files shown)


ioggstream:
Great @LPardue , thank you!

PR #2391, opened 2023-01-20
Fix HTMLURL reference

### LINKED ISSUE #2370: Reference to HTMLURL
_Comment by @fpalombini_

Thanks to Tommy for detailed shepherd write up, that tells me you have discussed this already. Seems obvious that the HTMLURL reference needs to be normative, so we might need to reference a snapshot of this instead.

See the following text in the IESG statement: https://www.ietf.org/about/groups/iesg/statements/normative-informative-references/

> Note 4: Normative references in RFCs cannot be to "work in progress" documents such as Internet Drafts. Drafts with such references will not be published as RFCs until the references are also published.

I'd like to get this fixed before starting the Last Call.

jricher:
Changes the reference to HTMLURL to use one of the newly-registered permanent anchors. Previously two different section anchors had been used, but I'm not sure how to put 
both of them into the references section. Is there a syntax for multiple target URLs?

Closes #2370

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -75,7 +75,7 @@ normative:
     HTTP1: RFC9112
     URI: RFC3986
     HTMLURL:
-        target: https://url.spec.whatwg.org/
+        target: https://url.spec.whatwg.org/#application/x-www-form-urlencoded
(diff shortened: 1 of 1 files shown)


reschke:
I would know how to do that in RFCXML. Maybe it's good enough to tune in AUTH48.

PR #2392, opened 2023-01-20
Add suggestion for error codes

### LINKED ISSUE #2336: message-signatures: status code when sig in request is invalid
This is not necessarily a spec issue, but I thought it should be recorded and maybe discussed here.

Is there a recommendation how a server should respond when a request is signed but the signature is invalid. It's likely a 400, unless we want to define a more specific code? Maybe a pointer to the HTTP problem format would be useful?

jricher:
This is a non-normative suggestion for applications to define what to do when a signature is invalid.

Closes #2336

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -81,6 +81,7 @@ normative:
     ABNF: RFC5234
 informative:
     RFC7239:
+    RFC7807:
     RFC8792:
     BCP195:
     CLIENT-CERT: I-D.ietf-httpbis-client-cert-field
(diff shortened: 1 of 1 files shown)


asbjornu (approved):

draft-ietf-httpbis-message-signatures.md
```diff
@@ -220,6 +221,7 @@ HTTP Message Signatures are designed to be a general-purpose security mechanism
...
 - A means of determining the context for derivation of message components from an HTTP message and its application context. While this is normally the target HTTP message itself, the context could include additional information known to the application, such as an external host name.
+- The error messages and codes that are returned from the verifier to the signer when the signature is invalid, the key material is inappropriate, the validity time window is out of specification, a component value cannot be calculated, or any other errors in the signature verification process. For example, if a signature is being used as an authentication mechanism, an status code of 401 Unauthorized or 403 Forbidden could be appropriate. If the response is from an HTTP API, a response with an HTTP status code of 400 Bad Request could include details as described in {{RFC7807}}, such as an indicator that the wrong key material was used.
```
tyler-ham, line 224: "an status code" should be "**a** status code" or "**an HTTP** status code"

jricher:
```suggestion
- The error messages and codes that are returned from the verifier to the signer when the signature is invalid, the key material is inappropriate, the validity time window is out of specification, a component value cannot be calculated, or any other errors in the signature verification process. For example, if a signature is being used as an authentication mechanism, an HTTP status code of 401 Unauthorized or 403 Forbidden could be appropriate. If the response is from an HTTP API, a response with an HTTP status code of 400 Bad Request could include details as described in {{RFC7807}}, such as an indicator that the wrong key material was used.
```

PR #2396, opened 2023-02-02
Discuss limitations on types based on field def

### LINKED ISSUE #2393: Date as an extension in 8941-defined fields
As pointed out on list, we need to warn people against just using Date as an extension in a field that was defined using 8941.

mnot:
Fixes #2393.

### DIFF
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -191,7 +191,7 @@ being used.</t>
 </artwork>
 </blockquote>
 
-Note that because the definition of a Strutured Field references a specific RFC for Structured Fields, the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and therefore discarded) by implementations of that specification.
+Note that because the definition of a Structured Field references a specific RFC for Structured Fields, the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and therefore discarded) by implementations of that specification.
 
 Also note that this limitation also applies to future extensions to a field; for example, a field that is defined with reference to RFC 8941 cannot use the Date type, because some recipients might still be using an RFC 8941 parser to process it.
 
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -191,7 +191,7 @@ being used.</t>
 </artwork>
 </blockquote>
 
-Note that a field is defined by referencing a specific RFC for Structured Fields, and the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC8941 cannot, because it will be treated as invalid (and likely discarded) by implementations of that specification.
+Note that a field is defined by referencing a specific RFC for Structured Fields, and the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and likely discarded) by implementations of that specification.
 
 In particular, the extensions available for use in a field are limited to the types allowed by that field's definition, which cannot exceed those defined by the Structured Fields specification it references.
 
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -191,6 +191,9 @@ being used.</t>
 </artwork>
 </blockquote>
 
+Note that a field is defined by referencing a specific RFC for Structured Fields, and its value is limited to the types defined in that RFC. For example, a field whose definition references this document can use the Date type ({{date}}), whereas a field whose definition references RFC8941 cannot. This also limits the kinds of extension values that can be later defined for a given field.
+
+
 # Structured Data Types {#types}
 
 This section defines the abstract types for Structured Fields, and summarises how those types are serialised into textual HTTP fields.
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -197,6 +197,8 @@ Also note that this limitation also applies to future extensions to a field; for
 
 However, this document is designed to be backwards-compatible with RFC 8941; a parser that implements the requirements here can also parse valid Structured Fields whose definitions reference RFC 8941.
 
+The effect of upgrading a Structured Fields implementation is that some field values that were invalid according to RFC 8941 might become valid when processed.  For instance, though it would not be valid for a field defined against RFC 8941, a field might include include a Date value. An RFC 8941 implementation would reject the entire field, whereas an updated Structured Field implementation might permit the value. In some cases, the resulting Date value will be rejected by field-specific logic, but values in fields that are otherwise ignored (such as extension parameters) might not be detected and the field might subsequently be accepted and processed.
+
 
 # Structured Data Types {#types}
 
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -191,7 +191,7 @@ being used.</t>
 </artwork>
 </blockquote>
 
-Note that a field is defined by referencing a specific RFC for Structured Fields, and the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and likely discarded) by implementations of that specification.
+Note that a field is defined by referencing a specific RFC for Structured Fields, and the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and therefore discarded) by implementations of that specification.
 
 In particular, the extensions available for use in a field are limited to the types allowed by that field's definition, which cannot exceed those defined by the Structured Fields specification it references.
 
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -191,7 +191,9 @@ being used.</t>
 </artwork>
 </blockquote>
 
-Note that a field is defined by referencing a specific RFC for Structured Fields, and its value is limited to the types defined in that RFC. For example, a field whose definition references this document can use the Date type ({{date}}), whereas a field whose definition references RFC8941 cannot. This also limits the kinds of extension values that can be later defined for a given field.
+Note that a field is defined by referencing a specific RFC for Structured Fields, and the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC8941 cannot, because it will be treated as invalid (and likely discarded) by implementations of that specification.
+
+In particular, the extensions available for use in a field are limited to the types allowed by that field's definition, which cannot exceed those defined by the Structured Fields specification it references.
 
 
 # Structured Data Types {#types}
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -191,7 +191,9 @@ being used.</t>
 </artwork>
 </blockquote>
 
-Note that a field is defined by referencing a specific RFC for Structured Fields, and the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and therefore discarded) by implementations of that specification.
+Note that because the definition of a Strutured Field references a specific RFC for Structured Fields, the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and therefore discarded) by implementations of that specification.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-sfbis.md
```diff
@@ -191,6 +191,9 @@ being used.</t>
 </artwork>
 </blockquote>
 
+Note that a field is defined by referencing a specific RFC for Structured Fields, and its value is limited to the types defined in that RFC. For example, a field whose definition references this document can use the Date type ({{date}}), whereas a field whose definition references RFC8941 cannot. This also limits the kinds of extension values that can be later defined for a given field.
```
martinthomson, line 194: I might add:

> Otherwise, an extension that uses a type that is permitted by a newer version of the Structured Field specification might be treated as invalid (and discarded) by an implementation that only implements an older version.

I don't know if you want to say that there is a possibility that though an extension like this would be accepted by those implementations that support the newer version, that's not consistently interoperable.

Note also this this doesn't just apply to types.  It might apply if the parsing/validation rules for a type becomes more lenient in a subsequent version.

mnot: See revision

martinthomson (approved):

draft-ietf-httpbis-sfbis.md
```diff
@@ -191,6 +191,11 @@ being used.</t>
 </artwork>
 </blockquote>
 
+Note that a field is defined by referencing a specific RFC for Structured Fields, and the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC8941 cannot, because it will be treated as invalid (and likely discarded) by implementations of that specification.
```
reschke, line 194:
```suggestion
Note that a field is defined by referencing a specific RFC for Structured Fields, and the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and likely discarded) by implementations of that specification.
```

reschke: "likely" is a bit weak; not discarding would be a spec violation, no?

jricher (approved):

mnot:
It's likely because the field is ignored when parsing fails, but 8941 s 2 allows a field to specify other error handling behaivour.

reschke:
How so?

8941, Section 4.2 says:

> If parsing fails -- including when calling another algorithm -- the entire field value MUST be ignored (i.e., treated as if the field were not present in the section). This is intentionally strict, to improve interoperability and safety, and specifications referencing this document are not allowed to loosen this requirement.

mnot:
Section 2 has a bit more nuance:

> When parsing fails, the entire field is ignored (see [Section 4.2](https://httpwg.org/http-extensions/draft-ietf-httpbis-sfbis.html#text-parse)); in most situations, violating field-specific constraints should have the same effect. Thus, if a header is defined as an Item and required to be an Integer, but a String is received, the field will by default be ignored. If the field requires different error handling, this should be explicitly specified

reschke:
You refer to:

> If the field requires different error handling, this should be explicitly specified

?

My reading is that this applies to the preceding text about what happens when parsing does not fail, but additional constraints are not met.

If this is not the case, we should consider this to be an inconsistency in requirements and treat it as spec bug.

reschke (changes requested):
I think "likely discarded" is very misleading; it's a *requirement* to discard.

draft-ietf-httpbis-sfbis.md
```diff
@@ -191,6 +191,11 @@ being used.</t>
 </artwork>
 </blockquote>
 
+Note that a field is defined by referencing a specific RFC for Structured Fields, and the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and likely discarded) by implementations of that specification.
```
reschke, line 194:
```suggestion
Note that a field is defined by referencing a specific RFC for Structured Fields, and the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and therefore discarded) by implementations of that specification.
```

draft-ietf-httpbis-sfbis.md
```diff
@@ -191,6 +191,11 @@ being used.</t>
 </artwork>
 </blockquote>
 
+Note that a field is defined by referencing a specific RFC for Structured Fields, and the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and therefore discarded) by implementations of that specification.
```
mnot, line 194:
```suggestion
Note that because the definition of a Strutured Field references a specific RFC for Structured Fields, the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and therefore discarded) by implementations of that specification.

However, this document is designed to be backwards-compatible with RFC 8941; a parser that implements the requirements here can also parse valid Structured Fields whose definitions reference RFC 8941.
```

draft-ietf-httpbis-sfbis.md
```diff
@@ -191,6 +191,13 @@ being used.</t>
 </artwork>
 </blockquote>
 
+Note that because the definition of a Strutured Field references a specific RFC for Structured Fields, the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and therefore discarded) by implementations of that specification.
+
+However, this document is designed to be backwards-compatible with RFC 8941; a parser that implements the requirements here can also parse valid Structured Fields whose definitions reference RFC 8941.
+
+In particular, the extensions available for use in a field are limited to the types allowed by that field's definition, which cannot exceed those defined by the Structured Fields specification it references.
```
mnot, line 198:
```suggestion
Also note that this limitation also applies to future extensions to a field; for example, a field that is defined with reference to RFC 8941 cannot use the Date type, because some recipients might still be using an RFC 8941 parser to process it.
```

draft-ietf-httpbis-sfbis.md
```diff
@@ -191,6 +191,13 @@ being used.</t>
 </artwork>
 </blockquote>
 
+Note that because the definition of a Strutured Field references a specific RFC for Structured Fields, the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and therefore discarded) by implementations of that specification.
+
+However, this document is designed to be backwards-compatible with RFC 8941; a parser that implements the requirements here can also parse valid Structured Fields whose definitions reference RFC 8941.
+
+Also note that this limitation also applies to future extensions to a field; for example, a field that is defined with reference to RFC 8941 cannot use the Date type, because some recipients might still be using an RFC 8941 parser to process it.
+
```
mnot, line 199:
```suggestion
```

draft-ietf-httpbis-sfbis.md
```diff
@@ -191,6 +191,13 @@ being used.</t>
 </artwork>
 </blockquote>
 
+Note that because the definition of a Strutured Field references a specific RFC for Structured Fields, the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and therefore discarded) by implementations of that specification.
+
+However, this document is designed to be backwards-compatible with RFC 8941; a parser that implements the requirements here can also parse valid Structured Fields whose definitions reference RFC 8941.
+
+Also note that this limitation also applies to future extensions to a field; for example, a field that is defined with reference to RFC 8941 cannot use the Date type, because some recipients might still be using an RFC 8941 parser to process it.
```
mnot, line 198:
```suggestion
Also note that this limitation also applies to future extensions to a field; for example, a field that is defined with reference to RFC 8941 cannot use the Date type, because some recipients might still be using an RFC 8941 parser to process it.

However, this document is designed to be backwards-compatible with RFC 8941; a parser that implements the requirements here can also parse valid Structured Fields whose definitions reference RFC 8941.
```

martinthomson (commented):
A suggestion you might consider...  I think that it would be helpful to include a more complete example here.

There is a hazard here in that any inconsistency how processing occurs creates interesting effects.  That is, the field is destroyed if a Date if observed.  That might lead to [Heisenbug](https://en.wikipedia.org/wiki/Heisenbug)s.

Let's say you have a WAF that uses SF to parse, but it only looks at certain things.  Maybe it doesn't look at parameters.  This might ignore a Date on something like:

```
Example: ?0;t=@4098543601
```

Code that processes the `t` parameter might (rightfully) discard the field if it sees the Date.

draft-ietf-httpbis-sfbis.md
```diff
@@ -191,6 +191,13 @@ being used.</t>
 </artwork>
 </blockquote>
 
+Note that because the definition of a Strutured Field references a specific RFC for Structured Fields, the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and therefore discarded) by implementations of that specification.
+
+Also note that this limitation also applies to future extensions to a field; for example, a field that is defined with reference to RFC 8941 cannot use the Date type, because some recipients might still be using an RFC 8941 parser to process it.
+
+However, this document is designed to be backwards-compatible with RFC 8941; a parser that implements the requirements here can also parse valid Structured Fields whose definitions reference RFC 8941.
+
```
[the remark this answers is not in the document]
mnot: Hm, that's very wordy, and this section is already quite wordy.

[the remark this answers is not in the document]
martinthomson: I know. I don’t like it much for that reason. But maybe this is worth the text. Slapping a section header on might help. “Backwards Compatibility” might be a good title.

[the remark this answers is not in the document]
bsdphk: I cannot in good confidence say that any of this adds clarity, but I will not go as far as to call it surpluage, because there is a point to be made here.

I think what bugs me, is that having a SFbis parser return an unexpected and illegal sf-date is not materially different from having a 8941 parser return sf-decimal where sf-integer was expected.

In the end, the responsibility for validating the header can only fall on the application code.

Maybe we should simply write that, and use the 8941-defined-SFbis-parser-returning-sf-date example to illustrate a detail people might otherwise miss ?

draft-ietf-httpbis-sfbis.md
```diff
@@ -191,6 +191,15 @@ being used.</t>
...
+The effect of upgrading a Structured Fields implementation is that some field values that were invalid according to RFC 8941 might become valid when processed.  For instance, though it would not be valid for a field defined against RFC 8941, a field might include include a Date value. An RFC 8941 implementation would reject the entire field, whereas an updated Structured Field implementation might permit the value. In some cases, the resulting Date value will be rejected by field-specific logic, but values in fields that are otherwise ignored (such as extension parameters) might not be detected and the field might subsequently be accepted and processed.
```
mnot, line 200:
```suggestion
Because of that backwards compatibility, field values that are invalid according to RFC 8941 might become valid when parsed by an implementation of this specification.  For example, a field might be erroneously sent with a Date value. An RFC 8941 implementation would reject the entire field, whereas a parser written to this specification would accept it. Whether that would cause interoperability issues depends on whether the results of parsing are checked appropriately for a type mismatch.```

File: draft-ietf-httpbis-sfbis.md (hunk 8 above)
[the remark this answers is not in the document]
Comment (mnot) [reply draft-ietf-httpbis-sfbis.md:199]: See proposal below.

File: draft-ietf-httpbis-sfbis.md (hunk 9 above)
Comment (bsdphk) [reply draft-ietf-httpbis-sfbis.md:200]: The "for example" grinds me, when that literally is the only possible thing that can happen.

Shorter and more to the point:

However, this document is designed to be backwards-compatible with RFC 8941; a parser that implements the requirements here can also parse valid Structured Fields whose definitions reference RFC 8941, but if the parser output contains any Date values, the caller must reject it as invalid, to remain RFC 8941 compliant.

Comment (mnot) [reply draft-ietf-httpbis-sfbis.md:200]: That's a bit _too_ short; see revision.

**Review by martinthomson — APPROVED** [review]

File: draft-ietf-httpbis-sfbis.md [hunk 10]
```diff
@@ -191,6 +191,15 @@ being used.</t>
 </artwork>
 </blockquote>
 
+Note that because the definition of a Strutured Field references a specific RFC for Structured Fields, the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and therefore discarded) by implementations of that specification.
```
reschke, line 194:
```suggestion
Note that because the definition of a Structured Field references a specific RFC for Structured Fields, the types available for use in its value are limited to those defined in that RFC. For example, a field whose definition references this document can have a value that uses the Date type ({{date}}), whereas a field whose definition references RFC 8941 cannot, because it will be treated as invalid (and therefore discarded) by implementations of that specification.
```

draft-ietf-httpbis-sfbis.md
```diff
@@ -191,6 +191,15 @@ being used.</t>
...
+
+Also note that this limitation also applies to future extensions to a field; for example, a field that is defined with reference to RFC 8941 cannot use the Date type, because some recipients might still be using an RFC 8941 parser to process it.
```
reschke, line 196:
```suggestion
Further note that this limitation also applies to future extensions to a field; for example, a field that is defined with reference to RFC 8941 cannot use the Date type, because some recipients might still be using an RFC 8941 parser to process it.
```

reschke (approved):
(with two nits)

draft-ietf-httpbis-sfbis.md
mnot: I prefer also.

PR #2400, opened 2023-02-06
Explicitly return a type from int/dec

### LINKED ISSUE #2354: Return `type` for Integer/Decimal parsing
[Parsing an Integer or Decimal](https://www.ietf.org/archive/id/draft-ietf-httpbis-sfbis-01.html#section-4.2.4) doesn't return `type`.  Without that, the number itself isn't necessarily enough to determine what the type is (necessary for parsing Date).

mnot:
Fixes #2354

### DIFF
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -816,7 +816,7 @@ Given an ASCII string as input_string, return a key. input_string is modified to
 
 Given an ASCII string as input_string, return an Integer or Decimal. input_string is modified to remove the parsed value.
 
-NOTE: This algorithm parses both Integers ({{integer}}) and Decimals ({{decimal}}), and returns the corresponding structure.
+NOTE: This algorithm parses both Integers ({{integer}}) and Decimals ({{decimal}}), and returns the corresponding structure, along with the type that the value was determined to be.
 
 1. Let type be "integer".
 2. Let sign be 1.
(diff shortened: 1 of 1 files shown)


reschke (approved):

bsdphk:
I'm Ok with this.

draft-ietf-httpbis-sfbis.md
```diff
@@ -839,7 +839,7 @@ NOTE: This algorithm parses both Integers ({{integer}}) and Decimals ({{decimal}
    1. If the final character of input_number is ".", fail parsing.
    2. If the number of characters after "." in input_number is greater than three, fail parsing.
    2. Parse input_number as a decimal number and let output_number be the product of the result and sign.
-0. Return output_number.
+0. Return output_number, type.
```
martinthomson, line 842: There is an alternative framing here.

Change step 8 to be:

```
8. If type is "integer":
    1. Let output_number be an Integer that is the result of parsing input_number as an integer.
9. Otherwise:
    2. If the number of characters after "." in input_number is greater than three, fail parsing.
    3. Let output_number be an Decimal that is the result of parsing input_number as a decimal.
13. Multiply output_number by sign.
14. Return output_number.
```

Not that invasive, but you are saying that output_number has both a type and a value.

You could do that consistently as well, I guess, but this is one of the few places that the output of a function like this matters.

draft-ietf-httpbis-sfbis.md
```diff
@@ -912,8 +912,8 @@ Given an ASCII string as input_string, return a Date. input_string is modified t
 
 1. If the first character of input_string is not "@", fail parsing.
 2. Discard the first character of input_string.
-3. Let output_date be the result of running Parsing an Integer or Decimal ({{parse-number}}) with input_string.
-4. If output_date is a Decimal, fail parsing.
+3. Let output_date, type be the result of running Parsing an Integer or Decimal ({{parse-number}}) with input_string.
+4. If type is "decimal", fail parsing.
```
martinthomson, line 916: Continuing the above, you could revert this.

draft-ietf-httpbis-sfbis.md
mnot: That could work.

mnot:
Closing in favour of #2467.

PR #2401, opened 2023-02-06
message-signatures: relax requirement to use RFC 8941

reschke:
...for serialization - allow use of updated spec.


jricher:
Replaced by #2402

PR #2402, opened 2023-02-06
update language on structured fields

### LINKED ISSUE #2362: signatures: clarify dependency on structured fields
IMHO we need to be clear what the requirements of SF support are. The way the spec currently is written (referencing RFC 8941) means that SF-shaped field values using the new sf-date format are not supported.

I don't think this is what we want, so unless we want to wait for sfbis to be ready, we'll need to add some consideration of how compat with SF extensions/revisions is supposed to work.

jricher:
Closes #2362

Replaces #2401


draft-ietf-httpbis-message-signatures.md
```diff
@@ -323,12 +324,16 @@ Additional parameters can be defined in the HTTP Signature Component Parameters
...
+If the value of an HTTP field is known by the application to be a structured field type (as defined in {{STRUCTURED-FIELDS}} or its extensions or updates), and the expected type of the structured field is known, the signer MAY include the `sf` parameter in the component identifier.
+If this parameter is included with a component identifier, the HTTP field value MUST be serialized using the formal serialization rules specified in {{Section 4 of STRUCTURED-FIELDS}} (or the
+applicable formal serialization section of its extensions or updates) applicable to
+the type of the HTTP field. Note that this process
 will replace any optional internal whitespace with a single space character, among other potential transformations of the value.
 
 If multiple field values occur within a message, these values MUST be combined into a single List or Dictionary structure before serialization.
 
+If the application does not know the type of the field, or the application does not know how to serialize the type of the field, the use of this flag will produce an error. As a consequence, the signer can only reliably sign fields that it knows the verifier's system supports.
```
jricher, line 335:
```suggestion
If the application does not know the type of the field, or the application does not know how to serialize the type of the field, the use of this flag will produce an error. As a consequence, the signer can only reliably sign fields using this flag when the verifier's system knows the type as well.
```

reschke:
Thanks!

PR #2404, opened 2023-02-07
Refine requirement for failure

### LINKED ISSUE #2399: sf: (potentially) clarify strict error handling
https://httpwg.org/http-extensions/draft-ietf-httpbis-sfbis.html#section-2

> When parsing fails, the entire field is ignored (see [Section 4.2](https://httpwg.org/http-extensions/draft-ietf-httpbis-sfbis.html#text-parse)); in most situations, violating field-specific constraints should have the same effect. Thus, if a header is defined as an Item and required to be an Integer, but a String is received, the field will by default be ignored. If the field requires different error handling, this should be explicitly specified.

It appears that some read the last sentence:

> If the field requires different error handling, this should be explicitly specified.

as saying that the generic draconian handling of parsing errors can be relaxed.

mnot:
Fixes #2399


reschke (changes requested):
That's a breaking change from 8941 which essentially ends draconian error handling.

mnot:
... or it's a clarification that aligns the text in the two different sections. Sounds like we should discuss on list.

draft-ietf-httpbis-sfbis.md
```diff
@@ -687,7 +687,7 @@ Example-String: "foo
 Example-String: bar"
 ~~~
 
-If parsing fails -- including when calling another algorithm -- the entire field value MUST be ignored (i.e., treated as if the field were not present in the section). This is intentionally strict, to improve interoperability and safety, and specifications referencing this document are not allowed to loosen this requirement.
+If parsing fails -- including when calling another algorithm -- the entire field value MUST be ignored (i.e., treated as if the field were not present in the section), unless the field definition specifies another behavior. It is RECOMMENDED that fields not modify this requirement, as it enhances interoperability and safety.
```
mnot, line 690:
```suggestion
If parsing fails -- including when calling another algorithm -- either the entire field value MUST be ignored (i.e., treated as if the field were not present in the section), or alternatively the complete HTTP message MUST be treated as malformed. This is intentionally strict to improve interoperability and safety, and field specifications that use Structured Fields are not allowed to loosen this requirement.
```

draft-ietf-httpbis-sfbis.md
```diff
@@ -687,7 +687,7 @@ Example-String: "foo
 Example-String: bar"
 ~~~
 
-If parsing fails -- including when calling another algorithm -- the entire field value MUST be ignored (i.e., treated as if the field were not present in the section). This is intentionally strict, to improve interoperability and safety, and specifications referencing this document are not allowed to loosen this requirement.
+If parsing fails -- including when calling another algorithm -- either the entire field value MUST be ignored (i.e., treated as if the field were not present in the section), or alternatively the complete HTTP message MUST be treated as malformed. This is intentionally strict to improve interoperability and safety, and field specifications that use Structured Fields are not allowed to loosen this requirement.
```
reschke, line 690: I still think that the "..,including when..." needs tuning; will make a proposal later today.

reschke (changes requested):

bsdphk:
I still dont understand why we have some of the "strict" stuff in 1.1 and some in 4.2.

I would move the paragraph from 4.2 to 1.1 to improve the chance people actually read it.

mnot:
Disagree; implementers are more likely to look at the parsing section, not what they perceive as boilerplate at the top. 1.1 covers the philosophy and general approach; 4.2 talks about the implementation details.

bsdphk:
"This is intentionally strict […]" sounded philosophical to me, but whatever :-)

PR #2414, opened 2023-02-19
message-signatures: remove duplicate "the"

reschke:


draft-ietf-httpbis-message-signatures.md
```diff
@@ -388,7 +388,7 @@ Note that the value for `key="c"` has been re-serialized according to the strict
 
 ### Binary-wrapped HTTP Fields {#http-field-byte-sequence}
 
-If the value of the the HTTP field in question is known by the application to cause problems with serialization, particularly with the combination of multiple values into a single line as discussed in {{security-non-list}}, the signer SHOULD include the `bs` parameter in a component identifier to indicate the values of the fields need to be wrapped as binary structures before being combined.
+If the value of the HTTP field in question is known by the application to cause problems with serialization, particularly with the combination of multiple values into a single line as discussed in {{security-non-list}}, the signer SHOULD include the `bs` parameter in a component identifier to indicate the values of the fields need to be wrapped as binary structures before being combined.
```
jricher, line 391:
```suggestion
If the value of the HTTP field in question is known by the application to cause problems with serialization, particularly with the combination of multiple values into a single line as discussed in {{security-non-list}}, the signer SHOULD include the `bs` parameter in a component identifier to indicate the values of the field need to be wrapped as binary structures before being combined.
```

PR #2422, opened 2023-02-24
Fix: #2384. Non-deterministic content coding beyond encryption.

### LINKED ISSUE #2384: 6.5 Usage with Encryption - it isn’t clear what layer of encryption is assumed (representation, content)
James Manger wrote:

> In “6.5 Usage with Encryption” it isn’t clear what layer of encryption is assumed (representation, content). I guess it is assuming a Content-Encoding that encrypts, such as “Content-Encoding: aes128gcm” from RFC 8188. And the security consideration is pointing out that if the encryption is performed multiple times to respond to multiple HTTP requests then the ciphertext (& hence Content-Digest & Repr-Digest) is likely to change each time, as each encryption (of the same plaintext) is likely to use a different nonce and/or key.
> 
>  This issue could occur without encryption. For instance, compression algorithms often have different “levels” (eg 1=fast, 9=best). So the representation could change if the level changed between requests. Maybe that will be rare enough to ignore?
> 
> I thought “6.5 Usage with Encryption” might be warning against including a digest of the plaintext if encryption was applied as, say, a transfer-encoding (not sure if there are encrypting transfer-encodings). That would be insecure.

ioggstream:
## This PR

- some content-coding are "not deterministic" ;)
- not only encryption-related ones.

Fix #2384

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -560,11 +560,18 @@ processing of invalid data.
 Not every hashing algorithm is suitable for use in the trailer section, some may require to pre-process
 the whole payload before sending a message (e.g. see {{?I-D.thomson-http-mice}}).
 
-## Usage with Encryption
+## Content encoding can change between messages
 
-The checksum of an encrypted payload can change between different messages
-depending on the encryption algorithm used; in those cases its value could not be used to provide
-a proof of integrity "at rest" unless the whole (e.g. encoded) content is persisted.
+Some content coding mechanisms can produce different outputs for the same content.
+This happens when they use input parameters that are not conveyed
+in the representation metadata (e.g. the gzip content coding can use different
+compression levels, so the same file has have multiple gzip-encoded versions),
+or when they rely on nonces or timestamps such as
+the aes128gcm content coding defined {{?RFC8188}}.
+
+In those cases, the checksum conveyed by the integrity fields
+could not be used to provide a proof of integrity "at rest"
+unless the whole (e.g. encoded) content is persisted.
 
 ## Algorithm Agility {#sec-agility}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -560,11 +560,18 @@ processing of invalid data.
...
+Some content coding mechanisms can produce different outputs for the same content.
+This happens when they use input parameters that are not conveyed
+in the representation metadata (e.g. the gzip content coding can use different
+compression levels, so the same file has have multiple gzip-encoded versions),
+or when they rely on nonces or timestamps such as
+the aes128gcm content coding defined {{?RFC8188}}.
+
+In those cases, the checksum conveyed by the integrity fields
+could not be used to provide a proof of integrity "at rest"
+unless the whole (e.g. encoded) content is persisted.
```
LPardue, line 574:
```suggestion
Content coding mechanisms can support different encoding parameters, meaning that the same input content can produce different outputs. For example, GZIP supports mulitple compression levels. Such encoding parameters are generally not communicated as representation metadata, for instance different compression levels would all use the same Content-Encoding: gzip field. Other examples include where encoding relies on nonces or timestamps, such as the aes128gcm content coding defined {{?RFC8188}}.

Since it is possible for ther to be variation within content coding, the checksum conveyed by the integrity field cannot be used to provide a proof of integrity "at rest"
unless the whole (e.g. encoded) content is persisted.
```

LPardue (changes requested):
Thanks for this, the core text makes it clearer IMO. I've just proposed a few editorial changes

ioggstream:
@LPardue great! I applied them :) Feel free to merge!

LPardue:
merged, thanks

PR #2423, opened 2023-02-24
use hex-encoding for brotli examples

### LINKED ISSUE #2387: Base-64 encoding bodies is unfortunate
From James Manger

> Base64-encoding non-printable bodies so they can be included in the document is unfortunate. Particularly as there are lots of “real” base64 values (ie all the …-Digest values). Perhaps hex would have been better to display bodies.
>
> If sticking with base64 to display bodies, the “Range: bytes=1-7/18” examples would be better as “Range: bytes=3-10/18”. That way you can visually recognize that the range (“AItWyFwC/6s=”) is a subset of the original (“H4sIAItWyFwC/6tW…”).

LPardue:
Closes #2387

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -895,8 +895,8 @@ The response includes a `Content-Encoding: br` that indicates the selected
 representation is Brotli-encoded. The `Repr-Digest` field-value is therefore
 different compared to the request.
 
-For presentation purposes, the response body is displayed as a Base64-encoded string because it contains
-non-printable characters.
+For presentation purposes, the response body is displayed as a sequence of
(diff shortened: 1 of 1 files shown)


LPardue:
@ioggstream please check my working here

ioggstream:
Great. Verified with

```
import brotli
def check_hex(s):
     decoded_bytes = [  bytes([int(s[i:i+4], 16)]) for i in range(0,len(s),4)]
     return b''.join(decoded_bytes)

s = '0x8B0x080x800x7B0x220x680x650x6C0x6C0x6F0x220x3A0x200x220x770x6F0x720x6C0x640x220x7D0x03'
ret = brotli.decompress(check_hex(s))

assert ret == b'{"hello": "world"}'
```

PR #2425, opened 2023-02-24
client-cert: genart review updates

b---c:
Updates from Genart Last Call review https://lists.w3.org/Archives/Public/ietf-http-wg/2023JanMar/0138.html

### DIFF
--- a/draft-ietf-httpbis-client-cert-field.md
+++ b/draft-ietf-httpbis-client-cert-field.md
@@ -57,7 +57,7 @@ informative:
 
 This document describes HTTP extension header fields that allow a TLS
 terminating reverse proxy to convey the client certificate information of a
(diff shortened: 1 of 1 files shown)


bc-pi:
"Thanks for addressing my concerns. I looked through the diff and it completely solves the points I raised." --  from the reviewer

MikeBishop (approved):
Looks fine.  While we're correcting hyphenation, I think "TLS-terminating reverse proxy" does need a hyphen, but as it wasn't part of this review, it probably doesn't need to be in this PR.  And I expect the RFC editor will have an opinion anyway.

b---c:
Thanks Mike, to your point, I'll go ahead w/ a merge and let the RFC editor correct any hyphenation that needs correcting.

PR #2447, opened 2023-02-25
Clarify that hashing bytes are base-64 encoded and give examples

### LINKED ISSUE #2385: No examples of encoding output any of the 6 "insecure" algorithms
From James Manger:

> There are no examples of any of the 6 “insecure” algorithms that are still listed in the table. This is particularly important as checksums were conveyed in decimal and hex in Digest, but will now be base64-encoded in Content-Digest & Repr-Digest. Do you base64-encode the decimal digits from, say, cksum; or base64-encode the 32-bits (most significant byte first?) they represent?

LPardue:
With Digest, the encoding of the digest value was based on the digest
algorithm. For different algorithms, this varied between base64, hex
encoding or decimal encoding.

In this spec, we removed the variablilty by using SF Byte Sequence,
which is always serialized as a base-64 encoding of the input bytes. All
algorithms are now expected to use this format.

This change tries to make it more clear that we expect the input of the
Byte Sequence to be. For example, we really don't want people accidentally
base-64'ing the textual hex represenation output of cksum - that can
lead to interop pain.

To help implementers further, add a new appendix with some samples that
can be used as test vectors.

Closes #2385

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -257,8 +257,8 @@ where each:
 * key conveys the hashing algorithm (see {{algorithms}})
   used to compute the digest;
 * value is a `Byte Sequence` ({{Section 3.3.5 of STRUCTURED-FIELDS}}), that
-  conveys the base64 encoding of the byte output produced by the digest
-  calculation, delimited by colons.
+  conveys an encoded version of the byte output produced by the digest
+  calculation.
 
 For example:
 
@@ -319,8 +319,8 @@ STRUCTURED-FIELDS}}) where each:
 
 * key conveys the hashing algorithm (see {{algorithms}})
   used to compute the digest;
-* value is a `Byte Sequence`, that conveys the base64 encoding of the byte
-  output produced by the digest calculation, delimited by colons.
+* value is a `Byte Sequence`, that conveys an encoded version of the byte
+  output produced by the digest calculation.
 
 For example:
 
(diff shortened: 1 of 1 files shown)


LPardue:
FWIW it would be nicer to have the examples rendered in a table. But the SHA-512 example is too long and makes the build whinge. I couldn't figure out any kramdown tricks for a multi-line table entry but if somebody else knows, I'd switch over.

LPardue:
Also, I welcome anyone checking my working in generating these values. The old algorithms don't tend to be supported in any decent libraries that are easily available. So I used a hodgepodge of various tools and sources to come up with bytes that would run as input into the SF serialization.

draft-ietf-httpbis-digest-headers.md
```diff
@@ -1291,6 +1321,11 @@ RFC 3230 could never communicate
 the digest of HTTP message content in the Digest field;
 Content-Digest now provides that capability.
 
+RFC 3230 allowed algorithms to define their output encoding format for use with
```
ioggstream, line 1324: +1

ioggstream:
@LPardue I like the example and I will check it.

 I am not sure that we should explain how Binary serialization work (e.g. base64 delimited by `:` ...) because it's explained in SF.

LPardue:
Since we had some feedback that some of the examples were incorrect, and I fixed that up, I'm going to go ahead and assume the rest are ok. The important part here is that we add some coverage over all the algorithms and and I have not heard any pushback to the proposal. So I'm going to merge.

PR #2450, opened 2023-02-28
Add reference and discussion of obs-fold and whitespace and other transformations

### LINKED ISSUE #2410: message-signatures: addition of leading or trailing whitespace to a field value
1.3:

> As mentioned earlier, HTTP explicitly permits and in some cases requires implementations to transform messages in a variety of ways. Implementations are required to tolerate many of these transformations. What follows is a non-normative and non-exhaustive list of transformations that could occur under HTTP, provided as context:
> ...
>  -  Addition or removal of leading or trailing whitespace to a field value.

This is misleading. No HTTP permits *addition* of leading/trailing whitespace. Leading/trailing whitespace in field values by definition is invalid.

### LINKED ISSUE #2411: message-signatures: obs-fold
The spec several times refers to the term obs-fold, but doesn't say where it's defined (-> https://www.rfc-editor.org/rfc/rfc9112.html#section-5.2).

Furthermore, in 1.3 it claims that *addition* of obs-fold could be permitted under certain circumstances. That IMHO is incorrect.

jricher:
- Add reference for obs-fold, closes #2411 
- add note about transforms, closes #2410

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -208,7 +208,7 @@ As mentioned earlier, HTTP explicitly permits and in some cases requires impleme
 - Conversion between different versions of the HTTP protocol (e.g., HTTP/1.x to HTTP/2, or vice-versa).
 - Changes in casing (e.g., "Origin" to "origin") of any case-insensitive components such as field names, request URI scheme, or host.
 - Addition or removal of leading or trailing whitespace to a field value.
-- Addition or removal of `obs-folds` from field values.
+- Addition or removal of `obs-folds` from field values ({{Section 5.2 of HTTP1}}).
 - Changes to the request target and authority that when applied together do not
     result in a change to the message's target URI, as defined in {{Section 7.1 of HTTP}}.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-message-signatures.md
```diff
@@ -207,15 +207,20 @@ As mentioned earlier, HTTP explicitly permits and in some cases requires impleme
 - Addition of fields such as `Via` ({{Section 7.6.3 of HTTP}}) and `Forwarded` ({{Section 4 of RFC7239}}).
 - Conversion between different versions of the HTTP protocol (e.g., HTTP/1.x to HTTP/2, or vice-versa).
 - Changes in casing (e.g., "Origin" to "origin") of any case-insensitive components such as field names, request URI scheme, or host.
-- Addition or removal of leading or trailing whitespace to a field value.
-- Addition or removal of `obs-folds` from field values.
 - Changes to the request target and authority that when applied together do not
     result in a change to the message's target URI, as defined in {{Section 7.1 of HTTP}}.
 
+Additionally, there are some transformations are either deprecated or otherwise not allowed, but still could occur in the wild. These transformations can still be handled without breaking the signature, and include things such as:
```
reschke, line 213:
```suggestion
Additionally, there are some transformations that are either deprecated or otherwise not allowed, but still could occur in the wild. These transformations can still be handled without breaking the signature, and include things such as:
```

PR #2451, opened 2023-03-01
Mention the empty string

### LINKED ISSUE #2420: empty next-hop-aliases?
Since next-hop-aliases is a String parameter, the syntax allows an empty string `""`. In cases where there is no chain, is it advisable to send the empty string to indicate this, rather than omitting the parameter? If so, it might help to add a sentence somewhere that highlights this usage type.

tfpauly:
Closes #2420


LPardue (approved):
This mention is clear. LGTM

PR #2452, opened 2023-03-01
Update advice and examples on signing signatures

jricher:
A signature over another signature does not provide the transitive protection that it seems it should. The examples and advice have been updated to accomodate this.

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -91,6 +91,30 @@ informative:
     CLIENT-CERT: I-D.ietf-httpbis-client-cert-field
     DIGEST: I-D.ietf-httpbis-digest-headers
     COOKIE: RFC6265
+    JACKSON2019:
+        target: https://dennis-jackson.uk/assets/pdfs/signatures.pdf
+        title: "Seems Legit: Automated Analysis of Subtle Attacks on Protocols that Use Signatures"
+        date: November, 2019
+        author:
+        -
+            name: Dennis Jackson
+            ins: D. Jackson
+            org: University of Oxford
+            country: UK
+        -
+            name: Cas Cremers
+            ins: C. Cremers
+            org: CISPA Helmholtz Center for Information Security
+            country: Germany
+        -
+            name: Katriel Cohn-Gordon
+            ins: K. Cohn-Gordon
+            org: Independent Scholar
+        -
+            name: Ralf Sasse
+            ins: R. Sasse
+            org: Department of Computer Science, ETH Zurich
+            country: Switzerland
 
 entity:
   SELF: "RFC nnnn"
@@ -944,7 +968,7 @@ Content-Length: 62
 {"busy": true, "message": "Your call is very important to us"}
 ~~~
 
-To cryptographically link the response to the request, the server signs the response with its own key and includes the method, authority, and the signature value `sig1` from the request in the covered components of the response. The signature base for this example is:
+To cryptographically link the response to the request, the server signs the response with its own key and includes the method, authority, and the signature and signature input labeled `sig1` from the request in the covered components of the response. The signature base for this example is:
 
 ~~~
 NOTE: '\' line wrapping per RFC 8792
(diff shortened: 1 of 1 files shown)


jricher:
@bblfish @yaronf can you please verify the signatures on the updated examples in this PR?

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1470,39 +1499,47 @@ Content-Digest: sha-512=:WZDPaVn/7XgHaAy8pmojAkGWoRx2UFChF41A2svX+T\
...
-The proxy's signature also includes the client's signature value from the original message in its covered components, as a dictionary field member under the label `sig1`. Note that since the client's signature already covers the client's Signature-Input value for `sig1`, this value is transitively covered by the proxy's signature and need not be added explicitly. While the origin server may not be able to directly verify this original signature, it can verify that the proxy has vouched for the signature's validity.
+The proxy's signature also includes the client's signature value and signature input from the original message in its covered components, as dictionary field members under the label `sig1`. The proxy can also sign any unchanged components from the original message, such as the `@method`, `@path`, and `@query` in this example, even if those components were covered by the original signature. See the discussion in {{security-sign-signature}} for more information about signing signature values.
```
martinthomson, line 1509: Does this need to say "client's signature value" given the attacks that Dennis describes?

jricher: @martinthomson not sure I understand the comment -- are you saying the signature value shouldn't be included in the proxy's signature at all, or that where it says "original signature" it should instead read "the client's signature"? If the former, the attack surface is a little bit different here because the proxy is a trusted party that's making a different kind of statement than the request-response-binding non-repudiation. This is mitigated by covering additional portions of the message, including the signature input line and any items that didn't change, along with items that did change at the proxy.

martinthomson: I guess the problem here is that the specific example is hard to follow.  Is it fair to say that the proxy has two goals: 1. it wants to cover the request that arrives at the server and 2. it also wants to transitively attest to what the client requested of it (maybe because the server can also validate that, but this is not entirely clear from context).

The former does not require that the signature or signature input from the client needs to be covered at all.  The latter is more interesting and requires that the proxy not change anything that the client signed, it can only add things that it additionally signs.  But in that case, the proxy can simply sign the same stuff the client did (and anything extra).  The only thing that is left is signing the fact that it knows that the client signed the message or signing to signal that it validated the client's signature (what signing a signature means here between those two, I don't know, maybe the difference is encoded in the difference between signing `signature-input` and signing `signature`).

Either way, the emphasis on including the signature is probably not really advancing a particular goal and should be de-emphasized.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -981,15 +1009,16 @@ Date: Tue, 20 Apr 2021 02:07:56 GMT
 Content-Type: application/json
 Content-Length: 62
 Signature-Input: reqres=("@status" "content-length" "content-type" \
-  "signature";req;key="sig1" "@authority";req "@method";req)\
-  ;created=1618884479;keyid="test-key-ecc-p256"
-Signature: reqres=:mh17P4TbYYBmBwsXPT4nsyVzW4Rp9Fb8WcvnfqKCQLoMvzOB\
-  LD/n32tL/GPW6XE5GAS5bdsg1khK6lBzV1Cx/Q==:
+  "signature";req;key="sig1" "signature-input";req;key="sig1" \
+  "@authority";req "@method";req);created=1618884479\
+  ;keyid="test-key-ecc-p256"
```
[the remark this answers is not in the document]
jricher: The body could be covered by also signing a content-digest header, but we just didn't add that to this example because that's not what the example is trying to showcase and emphasize.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -950,25 +974,29 @@ Content-Length: 62
 {"busy": true, "message": "Your call is very important to us"}
 ~~~
 
-To cryptographically link the response to the request, the server signs the response with its own key and includes the method, authority, and the signature value `sig1` from the request in the covered components of the response. The signature base for this example is:
+To cryptographically link the response to the request, the server signs the response with its own key and includes the method, authority, and the signature and signature input labeled `sig1` from the request in the covered components of the response. The signature base for this example is:
```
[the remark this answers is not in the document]
martinthomson: I'm just looking for plausible.  I'm having a little bit of a hard time suspending disbelief with this one, that's all.

PR #2458, opened 2023-03-03
Message contents has newlines, ensure digest consistency

### LINKED ISSUE #2388: Newlines at the end of JSON bodies
From James Manger:

> No newlines are included at the end of the JSON bodies, though that is never mentioned. “Content-Length: 18” on the first {"hello": "world"} example could indicate that. Not including newlines is okay for 1-line JSON values (even though they have extraneous spaces so they aren’t “compact” JSON). No final newline on the multi-line JSON examples is a bit nasty.

LPardue:
By virtue of the presentation formats available to RFCs, it was ambigous
to readers whether example's content included a trailing LF or not. This was bad
because the content (or even representation) is required for digest
calculation.

This change makes it clear, where needed, what the object bytes are that
we are sticking into the digest calculation. And since that now includes
LF, we need to recompute all the example digest values.

Fixes #2388

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -1417,36 +1417,31 @@ account their specific formatting rules.
 import base64, json, hashlib, brotli, logging
 log = logging.getLogger()
 
-def encode_item(item, encoding=lambda x: x):
-    indent = 2 if isinstance(item, dict) and len(item) > 1 else None
-    json_bytes = json.dumps(item, indent=indent).encode() + b'\n'
-    return encoding(json_bytes)
-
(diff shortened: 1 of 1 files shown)


LPardue:
Please check and keep me honest here.

ioggstream:
Hi Lucas, thanks for the support! 

If the problem is only related to JSON content, the most sensible choice is to use text/plain.

LPardue:
I think this PR strikes a decent balance between providing examples that are easily legible and correctly encoded.

ioggstream:
Thanks @LPardue for taking care of this. I suggested #2461 so people are instructed not to add \n to the actual json encoding.

PR #2461, opened 2023-03-06
Align example code with the new editorial changes (\n at the end of content).

ioggstream:
## This PR

- clarify that we are not altering json serialization.


LPardue:
Thanks, I think this is clearer anyhow!

Issue #2441, opened 2022-10-13
Require TLS 1.3 or TLS 1.2 with Extended Master Secret


DavidSchinazi:
From Ilari Liusvaara on the [list](https://lists.w3.org/Archives/Public/ietf-http-wg/2022OctDec/0027.html):

> I do not see requirement for TLS 1.3 or Extended Master Secret anywhere. It is not safe to use TLS Exporters for authentication otherwise.

PR #2462, opened 2023-03-06
Require modern TLS

jhoyla:
Add requirement for unprompted auth to use TLS 1.2 with extended master secret or later.
Fixes httpwg/http-extensions#2441

### DIFF
--- a/draft-ietf-httpbis-unprompted-auth.md
+++ b/draft-ietf-httpbis-unprompted-auth.md
@@ -109,10 +109,18 @@ contexts used by each scheme). The TLS keying material exporter is used to
 generate a 32-byte key which is then used as a nonce.
 
 Because the TLS keying material exporter is only secure for authentication when
-it is uniquely bound to the TLS session {{!RFC7627}}, unprompted auth is only
-defined for versions of TLS with this property, i.e. TLS 1.2 with extended
-master secret {{!RFC7627}}, TLS 1.3 {{!TLS}}, and later. Use of unprompted auth
-with prior versions of TLS is not allowed.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -108,6 +108,12 @@ material exporter uses a label that starts with the characters
 contexts used by each scheme). The TLS keying material exporter is used to
 generate a 32-byte key which is then used as a nonce.
 
+Because the TLS keying material exporter is only secure for authentication when
+it is uniquely bound to the TLS session {{!RFC7627}}, unprompted auth is only
+defined for versions of TLS with this property, i.e. TLS 1.2 with extended
+master secret {{!RFC7627}}, TLS 1.3 {{!TLS}}, and later. Use of unprompted auth
+with prior versions of TLS is not allowed.
```
DavidSchinazi, line 115:
```suggestion
Because the TLS keying material exporter is only secure for authentication when
it is uniquely bound to the TLS session {{!RFC7627}}, the Signature and HMAC
authentication schemes require either one of the following properties:

* The TLS version in use is greater or equal to 1.3 {{TLS}}.
* The TLS version in use is greater or equal to 1.2 and the Extended
Master Secret extension {{RFC7627}} has been negotiated.

Clients MUST NOT use the Signature and HMAC authentication
schemes on connections that do not meet one of the two properties
above. If a server receives a request that uses these authentication
schemes on a connection that meets neither of the above properties,
the server MUST treat the request as malformed.
```

DavidSchinazi:
Please mark this PR as fixing the corresponding issue

DavidSchinazi (approved):

Issue #2438, opened 2022-10-13
Discuss security properties of keys


DavidSchinazi:
From @nharper on the [list](https://lists.w3.org/Archives/Public/ietf-http-wg/2022OctDec/0028.html):

> The draft is very light on details about where the user's key (private or symmetric) comes from. I assume that key generation/distribution/registration/etc is out of scope for this draft, but the draft should address the security properties expected of said key, e.g. can a key be used for Unprompted Authentication and another use?

jhoyla:
The conservative thing to do here would be to insist that the key be unique to the protocol. 
We have domain separation in our outputs, so we don't need to worry about unprompted auth messages being used in cross-protocol attacks, but a different protocol (that doesn't use context strings) could possibly be induced to generate unprompted auth messages. We also need to worry about key compromise.

Imagine a protocol that leaks the private / secret key, that shares a key with unprompted auth. This doesn't even need to be a bug, it could be a commit-before-reveal style protocol. (Obviously that's a pathological example, but it shows we can't assume composability with all protocols) 

If we want to do a formal analysis (which I do) I think we need to be conservative here.

DavidSchinazi:
SGTM

PR #2463, opened 2023-03-06
Require unique key material.

jhoyla:
Fixes httpwg/http-extensions#2438.

### DIFF
--- a/draft-ietf-httpbis-unprompted-auth.md
+++ b/draft-ietf-httpbis-unprompted-auth.md
@@ -232,6 +232,10 @@ the browser such that it could access another context's memory, the attacker
 might also be able to access the corresponding key, so binding authentication to
 requests would not provide much benefit in practice.
 
+Key material used for authentication in unprompted authentication, whether
+symmetric or asymmetric MUST NOT be reused in other protocols. Doing so can
+undermine the security guarantees of the authentication.
+
 # IANA Considerations {#iana}
 
 ## Unprompted-Authentication Header Field {#iana-header}


DavidSchinazi:
Please mark this PR as fixing the corresponding issue

draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -232,6 +232,10 @@ the browser such that it could access another context's memory, the attacker
 might also be able to access the corresponding key, so binding authentication to
 requests would not provide much benefit in practice.
 
+Key material used for authentication in unprompted auth, whether symmetric or
```
DavidSchinazi, line 235: Please don't use "unprompted auth", that abbreviation is not used in the document

DavidSchinazi: Additionally, keys are used for the new authentication schemes more than unprompted auth so we should just mention those instead.

draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -232,6 +232,10 @@ the browser such that it could access another context's memory, the attacker
...
+asymmetric MUST NOT be reused in other protocols. Doing so invalidates the
```
DavidSchinazi, line 236:
```suggestion
asymmetric MUST NOT be reused in other protocols. Doing so can undermine the
```

DavidSchinazi (changes requested):

DavidSchinazi (approved):

PR #2470, opened 2023-03-13
Unprompted Auth: Update to use cases and concerns about super-cookies

DavidSchinazi:
Fixes #2426 
Fixes #2427

PR #2471, opened 2023-03-13
Add note about linking keys

### LINKED ISSUE #2426: Make sure u= is not a tracking vector

DavidSchinazi:
Based on #2470 
Fixes #2426

### DIFF
--- a/draft-ietf-httpbis-unprompted-auth.md
+++ b/draft-ietf-httpbis-unprompted-auth.md
@@ -270,6 +270,11 @@ Key material used for authentication in unprompted authentication, whether
 symmetric or asymmetric MUST NOT be reused in other protocols. Doing so can
 undermine the security guarantees of the authentication.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -270,6 +270,11 @@ Key material used for authentication in unprompted authentication, whether
 symmetric or asymmetric MUST NOT be reused in other protocols. Doing so can
 undermine the security guarantees of the authentication.
 
+Sites offering Unprompted Authentication are able to link requests that use
+the same key for the Authentication Schemes provided. However, requests are
+not linkable across other sites if the keys used are private to the individual
```
martinthomson, line 275: "private" is not really the right word to use in this context.  I think that "unique" would be better, or you could use the opposite: provided that the keys are not used for requests to other sites.

Note that the term "site" probably needs to be defined here.  Origin is probably an easier term to lean on.

PR #2480, opened 2023-03-22
RFC6265bis: Use navigables concept

### LINKED ISSUE #2372: Fix browsing context integration in 6265bis
6265bis has a bunch of references to browsing context and its active document. There were some big changes to HTML recently  in https://github.com/whatwg/html/pull/6315 that should have broken some of those.

Note that a lot of this could be solved through adding more accurate layering as described in #2084, but if we wanted to publish a new cookies RFC before that change happens, it may still be worth adding the correct terminology and references. 

cc @annevk @domenic @sbingler

sbingler:
Closes #2372

The HTML spec recommends avoiding using "browser contexts" and instead recommends "navigables".

This PR refactors rfc6265bis to use "navigables".

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -2676,15 +2676,17 @@ The "Cookie Attribute Registry" should be created with the registrations below:
 * Prevent nameless cookies with prefixed values
   <https://github.com/httpwg/http-extensions/pull/2251>
 
+## draft-ietf-httpbis-rfc6265bis-12
+
 * Advise the reader which section to implement
   <https://github.com/httpwg/http-extensions/pull/2478>
 
 * Define top-level navigation
   <https://github.com/httpwg/http-extensions/pull/2481>
 
-## draft-ietf-httpbis-rfc6265bis-12
+* Use navigables concept
+  <https://github.com/httpwg/http-extensions/pull/2478>
 
-* None. Yet!
 
 # Acknowledgements
 {:numbered="false"}
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -323,11 +323,11 @@ only if they are equivalent under the i;ascii-casemap collation defined in
 
 The term string means a sequence of non-NUL octets.
 
-The terms "active document", "ancestor navigables", "container document",
-"content navigable", "dedicated worker", "Document", "inclusive ancestor
-navigables", "navigable", "opaque origin", "sandboxed origin browsing context
-flag", "shared worker", "the worker's Documents", "top-level traversable", and
-"WorkerGlobalScope" are defined in {{HTML}}.
+The terms "active browsing context", "active document", "ancestor navigables",
+"container document", "content navigable", "dedicated worker", "Document",
+"inclusive ancestor navigables", "navigable", "opaque origin", "sandboxed
+origin browsing context flag", "shared worker", "the worker's Documents",
+"top-level traversable", and "WorkerGlobalScope" are defined in {{HTML}}.
 
 "Service Workers" are defined in the Service Workers specification
 {{SERVICE-WORKERS}}.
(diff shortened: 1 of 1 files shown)


sbingler:
@johannhof PTAL

@jakearchibald If you have any time. These concepts look like they should be drop in replacements but I'd appreciate your feedback just in case.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1711,9 +1710,9 @@ user agent MUST process the cookie as follows:
         processing the cookie.
 
     3.  If the cookie was received from a request which is navigating a
-        top-level browsing context {{HTML}} (e.g. if the request's "reserved
+        top-level traversable {{HTML}} (e.g. if the request's "reserved
         client" is either `null` or an environment whose "target browsing
-        context" is a top-level browing context), skip the remaining substeps
+        context" is in a top-level traversable), skip the remaining substeps
```
sbingler, line 1715: `target browsing context` is defined by FETCH and it doesn't look like it's been changed (yet?)

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1093,15 +1092,15 @@ cookies" is an origin set to an opaque origin.
...
+1.  Let `top-document` be the active document in `document`'s navigable's
+    top-level traversable.
 
 2.  Let `top-origin` be the origin of `top-document`'s URI if `top-document`'s
     sandboxed origin browsing context flag is set, and `top-document`'s origin
     otherwise.
 
 3.  Let `documents` be a list containing `document` and each of `document`'s
-    ancestor browsing contexts' active documents.
+    ancestor navigables' active documents.
```
jakearchibald, line 1103: Similar to above, I'd reword this to avoid assuming that a document is its navigable's active document. Or, if you're sure that this algorithm only runs when the chain of documents are active, add a line that asserts that.

jakearchibald (approved):
Couple of nits, but LGTM

johannhof:
Thanks @sbingler, this looks conceptually right to me but I'll defer to @jakearchibald for detailed inspection :)

draft-ietf-httpbis-rfc6265bis.md
domenic: This is https://github.com/whatwg/html/issues/8463 . In the meantime you can use "target browsing context's navigable is a top-level traversable".

draft-ietf-httpbis-rfc6265bis.md
domenic: I think you want to use https://html.spec.whatwg.org/#inclusive-ancestor-navigables . See e.g. this call site for something that does exactly what you're doing: https://html.spec.whatwg.org/#the-autofocus-attribute:inclusive-ancestor-navigables

You kind of need to care about activeness because you need a non-ambiguous way of going from navigables (which are in a tree) to documents (which are not). Each navigable has a single active document and potentially many inactive documents; the latter are stored in bfcache. You want to pick out the active document.

E.g. given:

```
parent navigable: A --> [B] --> C
                         |
child navigable:      [child]
```

where `[]` denotes the currently-active document, you want the set to contain B, but not A or C.

sbingler: Thanks for all the feedback, I used the inclusive ancestor navigables concept. I also kept the note that all docs should be active, @mikewest do you have a preference to remove/keep it?

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1894,7 +1897,7 @@ cookie-string from a given cookie store.
      * The same-site-flag is "Lax" or "Default".
      * The HTTP request associated with the retrieval uses a "safe" method.
      * The target browsing context of the HTTP request associated with the
-       retrieval is a top-level browsing context.
+       retrieval is within a top-level traversable.
```
sbingler, line 1900: @domenic Can I get your thoughts on this bullet point?

domenic: I guess the most proper way would be "is the [active browsing context](https://html.spec.whatwg.org/#nav-bc) or a top-level traversable". But hopefully we will fix https://github.com/whatwg/html/issues/8463 at some point soon so you don't need these circumlocutions...

sbingler: Thanks,

Is there an ETA for that issue? I'm fine with holding off on this change until that lands

domenic: No specific ETA, sorry.

sbingler:
@mikewest ping

mikewest (approved):
Yup. LGTM.

Thanks!

sbingler:
Thanks Mike, conflicts should be resolved

PR #2484, opened 2023-03-28
Add implementation considerations for CNAME chain access

### LINKED ISSUE #2358: draft-ietf-httpbis-alias-proxy-status: Refer to AI_CANONNAME
I think the alias draft should probably have a (non-normative) reference to [getaddrinfo's AI_CANONNAME functionality](https://pubs.opengroup.org/onlinepubs/9699919799/functions/freeaddrinfo.html), in particular noting that:

* `next-hop-aliases` can be populated from the output of `getaddrinfo`.
* `next-hop-aliases` can be used to emulate or substitute for `getaddrinfo`.

tfpauly:
Closes #2358

### DIFF
--- a/draft-ietf-httpbis-alias-proxy-status.md
+++ b/draft-ietf-httpbis-alias-proxy-status.md
@@ -120,8 +120,9 @@ of alias names and canonical names received in CNAME records.
 Implementations ought to note that the full chain of names might not available in common DNS
 resolution APIs, such as `getaddrinfo`. `getaddrinfo` does have an option for `AI_CANONNAME`,
 but this will only return the last name in the chain (the canonical name), not the alias
-names. Implementations without access to the full chain MAY include incomplete information
-(such as only the canonical name).
+names.
+
+An implementation MAY include incomplete information in the `next-hop-aliases` parameter to accommodate cases where it is unable to include the full chain, such as only including the canonical name if the implementation can only use `getaddrinfo` as described above.
 
 # Security Considerations {#sec-considerations}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-alias-proxy-status.md
```diff
@@ -112,6 +112,17 @@ in which this parameter is not present.
...
+# Implementation Considerations
+
+In order to include the `next-hop-aliases` parameter, a proxy needs to have access to the chain
+of alias names and canonical names received in CNAME records.
+
+Implementations ought to note that the full chain of names might not available in common DNS
+resolution APIs, such as `getaddrinfo`. `getaddrinfo` does have an option for `AI_CANONNAME`,
+but this will only return the last name in the chain (the canonical name), not the alias
+names. Implementations without access to the full chain MAY include incomplete information
+(such as only the canonical name).
```
LPardue, line 124: Is this the only circumstance under which the incomplete chain might be sent? In theory, there might be other circumstances where truncating the list might be useful. Maybe we could split off this `MAY` as a general allowance ?

tfpauly, line 124:
```suggestion
names.

An implementation MAY include incomplete information in the `next-hop-aliases` parameter to accommodate cases where it is unable to include the full chain, such as only including the canonical name if the implementation can only use `getaddrinfo` as described above.
```

tfpauly: @LPardue I made the MAY more general — how does it look now?

PR #2489, opened 2023-03-28
Reframe signing signatures

jricher:
Reframe use cases that had previously signed signature values to avoid problems with this approach.

### DIFF
--- a/draft-ietf-httpbis-message-signatures.md
+++ b/draft-ietf-httpbis-message-signatures.md
@@ -261,7 +261,8 @@ HTTP Message Signatures are designed to be a general-purpose security mechanism 
 - The set of allowable signature algorithms to be used by signers and accepted by verifiers.
 - A means of determining that the signature algorithm used to verify the signature is appropriate for the key material and context of the message. For example, the process could use the `alg` parameter of the signature parameters ({{signature-params}}) to state the algorithm explicitly, derive the algorithm from the key material, or use some pre-configured algorithm agreed upon by the signer and verifier.
 - A means of determining that a given key and algorithm used for a signature are appropriate for the context of the message. For example, a server expecting only ECDSA signatures should know to reject any RSA signatures, or a server expecting asymmetric cryptography should know to reject any symmetric cryptography.
-- A means of determining the context for derivation of message components from an HTTP message and its application context. While this is normally the target HTTP message itself, the context could include additional information known to the application, such as an external host name.
+- A means of determining the context for derivation of message components from an HTTP message and its application context. While this is normally the target HTTP message itself, the context could include additional information known to the application through configuration, such as an external host name.
+- If binding between a request and response is needed using the mechanism in {{content-request-response}}, all elements of the request and response message that would be required to provide properties of such a binding.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-message-signatures.md
```diff
@@ -1008,25 +998,99 @@ HTTP/1.1 503 Service Unavailable
...
 
-Applications needing this type of binding have to sign sufficient portion the request to ensure that it is uniquely tied to the response. Signing the signature value of a signed request alone does not provide sufficient coverage in most cases, as discussed in {{security-sign-signature}}.
+While it is syntactically possible to sign another signature on a message, applications SHOULD NOT sign the Signature and Signature-Input fields of the request, as shown in this example. Doing so does not provide transitive coverage of covered components and is susceptible to attack as discussed in {{security-sign-signature}}. In some applications, signing these fields of a request can indicate to the client that the server has processed an input signature, but such applications need to carefully specify such usage in addition to ensuring sufficient coverage of the messages.
```
dennisjackson, line 1089:
```suggestion
While it is syntactically possible to include a signature component from a request as an input to a response signature, this is NOT RECOMMENDED. This is because signatures of signatures do not provide the transitive coverage of covered components as one might expect and is susceptible to various attacks as discussed in {{security-sign-signature}}.
```

Rationale: I think NOT RECOMMENDED is a bit clearer than SHOULD NOT in this case. I worry the last sentence muddies the water a little bit so I think the text is stronger without it.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -1555,27 +1617,31 @@ Host: origin.host.internal.example
...
+    wiyUpZ10KHgOEe11sx3G2gxI8S0jnxQB+Pu68U9vVcasqOWAEObtNKKZd8tSFu7\
+    LB5YAv0RAGhB8tmpv7sFnIm9y+7X5kXQfi8NMaZaA8i2ZHwpBdg7a6CMfwnnrtf\
+    lzvZdXAsD3LH2TwevU+/PBPv0B6NMNk93wUs/vfJvye+YuI87HU38lZHowtznbL\
+    Vdp770I6VHR6WfgS9ddzirrswsE1w5o0LV/g==:
+
 
 {"hello": "world"}
 ~~~
 
+While the proxy could additionally include the client's Signature value and Signature-Input fields from the original message in the new signature's covered components, in this example the proxy does not do this due to known problems in signing signature values as discussed in {{security-sign-signature}}. In some applications, the proxy is a highly trusted component and signing an input signature can be used as a way for the proxy to indicate to the origin server that the proxy has read and verified the client's signature in the original message context. While the origin server may not be able to directly verify this original signature, it can verify that the proxy has vouched for the signature's validity. Use of this approach should be subject to careful analysis of the trust relationship between the different components of the application.
```
dennisjackson, line 1643:
```suggestion
While the proxy could additionally include the client's Signature value and Signature-Input fields from the original message in the new signature's covered components, in this example the proxy does not do this due to the weaknesses of signing signature values as discussed in {{security-sign-signature}}. If the proxy needs to convey to the origin that it has successfully checked the client's signature, the application SHOULD define a header to indicate so and ensure that the header and all relevant components from the client's request are included within the proxy's signature.    
```

Rationale: I think this is a little clearer on the right path for future adoption, but still compatible with the approach taken by existing applications.

dennisjackson (commented):
Thanks Justin. This is looking heaps better.

dennisjackson (approved):
This looks good to me! Summary of the changes: 

- Explicitly flagging that the application / implementer must specify how request-response binding is to be achieved (if desired). 
- Removed all references to request-response binding in the main body. 
- Removed all signatures of signatures from the examples
- Added guidance that applications SHOULD sign all signed components of a request when generating a signed response. 
- Added guidance that generating signatures of signatures is NOT RECOMMENDED when generating responses. 
- Added guidance that generating signatures of signatures is NOT RECOMMENDED when acting as a proxy between a request / response.

PR #2494, opened 2023-03-31
Rough in Display Strings

### LINKED ISSUE #2343: explore solution space for non-ASCII characters in Strings
RFC 8941 does not support non-ASCII characters in sf-string (but recommends to use sf-binary when non-ASCII characters are needed).

The solution space is roughly:

1. No support at all
2. No support, but addings hints about workarounds, such as
2.1 Using sf-binary (approach in RFC 8941)
2.2 Using RFC8187 parameter encoding (works only for parameters at it uses the "*" marker)
2.3 Using an escaping mechanism inside sf-string (with out-of-band info to indicate that it is in use) - currently proposed for the "Problem" header field (see https://github.com/ietf-wg-httpapi/rfc7807bis/commit/ebf9465504713876e6cd507629511313bdbce39c)
3. Direct support
3.1 By extending sf-string to allow native UTF-8
3.2 By extending sf-string to support a new escaping mechanism (for instance, by introducing new backslash escapes)
3.3 By adding a new type ("sf-ustring") that somehow supports non-ASCII codepoints (NOTE: this used to say "characters")

Note that the question whether non-ASCII octets appear in the field value (obs-text) is somehat orthogonal, expect if 3.1 would be chosen.

mnot:
Fixes #2343.

### DIFF
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -692,7 +692,7 @@ Given a string of Unicode characters as input_string, return an ASCII string sui
 1. Let encoded_string be an empty string.
 2. For each byte in byte_array:
   1. If byte is %x25 ("%"), append "%25" to encoded_string.
-  2. If byte is in the ranges x00-1f or %x7f-ff, append the result of applying the percent-encoding defined in {{Section 2.1 of URI}} to encoded_string.
+  2. If byte is in the ranges %x00-1f or %x7f-ff, apply the percent-encoding defined in {{Section 2.1 of URI}} to byte and append the result to encoded_string.
   3. Otherwise, decode byte as an ASCII character and append the result to encoded_string.
 3. Let formatted_string be the result of running Serialising a String ({{ser-string}}) with encoded_string.
 4. Return the character "%" followed by formatted_string.
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -399,7 +399,7 @@ Note that the serialization algorithm ({{ser-decimal}}) rounds input with more t
 
 Strings are zero or more printable ASCII {{!RFC0020}} characters (i.e., the range %x20 to %x7E). Note that this excludes tabs, newlines, carriage returns, etc.
 
-Unicode is not directly supported in Strings, because it causes a number of interoperability issues, and -- with few exceptions -- field values do not require it.
+Non-ASCII characters are not directly supported in Strings, because it causes a number of interoperability issues, and -- with few exceptions -- field values do not require it.
 
 When it is necessary for a field value to convey non-ASCII content, a Display String ({{displaystring}}) can be specified.
 
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -48,7 +48,7 @@ informative:
       DOI:  10.1109/IEEESTD.2019.8766229
       ISBN: 978-1-5044-5924-2
 
-  STD63:
+  UTF8:
     title: UTF-8, a transformation format of ISO 10646
     author:
     - ins: F. Yergeau
@@ -62,6 +62,7 @@ informative:
   RFC9113:
     display: HTTP/2
   HPACK: RFC7541
+  URI: RFC3986
 
 venue:
   group: HTTP
@@ -391,7 +392,7 @@ Strings are zero or more printable ASCII {{!RFC0020}} characters (i.e., the rang
 
 Unicode is not directly supported in Strings, because it causes a number of interoperability issues, and -- with few exceptions -- field values do not require it.
 
-When it is necessary for a field value to convey non-ASCII content, a Byte Sequence ({{binary}}) can be specified, along with a character encoding (preferably UTF-8 {{STD63}}).
+When it is necessary for a field value to convey non-ASCII content, a Display String ({{displaystring}}) can be specified.
 
 When serialised in a textual HTTP field, Strings are delimited with double quotes, using a backslash ("\\") to escape double quotes and backslashes. For example:
 
@@ -454,6 +455,22 @@ For example:
 ~~~ http-message-new
 Example-Date: @1659578233
 ~~~
+
+### Display Strings {#displaystring}
+
+Display Strings are similar to Strings, in that they consist of zero or more characters, but they allow unicode content, unlike Strings.
+
+Display Strings are intended for use in cases where a value is displayed to end users, and therefore may need to carry non-ASCII content. It is NOT RECOMMENDED that they be used in situations where a String ({{string}}) or Token ({{token}}) would be adequate, because unicode has processing considerations (e.g., normalization) and security considerations (e.g., homograph attacks) that make it more difficult to handle correctly.
+
+Note that Display Strings do not indicate the language used in the value; that can be done separately if necessary (e.g., with a parameter).
+
+For example:
+
+~~~ http-message-new
+Example-DisplayString: %"This is intended for display to %C3%BCsers."
+~~~
+
+
 
 
 # Working with Structured Fields in HTTP {#text}
@@ -563,7 +580,8 @@ Given an Item as input_item, return an ASCII string suitable for use in an HTTP 
 5. If input_item is a Byte Sequence, return the result of running Serializing a Byte Sequence ({{ser-binary}}) with input_item.
 6. If input_item is a Boolean, return the result of running Serializing a Boolean ({{ser-boolean}}) with input_item.
 7. If input_item is a Date, return the result of running Serializing a Date ({{ser-date}}) with input_item.
-8. Otherwise, fail serialization.
+8. If input_item is a Display String, return the result of running Serializing a Display String ({{ser-display}}) with input_item.
+9. Otherwise, fail serialization.
 
 
 ### Serializing an Integer {#ser-integer}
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-sfbis.md
```diff
@@ -464,6 +465,22 @@ For example:
 Example-Date: @1659578233
 ~~~
 
+### Display Strings {#displaystring}
+
+Display Strings are similar to Strings, in that they consist of zero or more characters, but they allow unicode content, unlike Strings.
```
martinthomson, line 470:
```suggestion
Display Strings are similar to Strings, in that they consist of zero or more characters, but the content of a Display String is UTF-8-encoded Unicode, rather than the ASCII carried by a String.
```

martinthomson (approved):
Just being pedantic.  This is a pretty small change, but a big and important feature.

draft-ietf-httpbis-sfbis.md
mnot: No - at this level (the data model), they're unicode strings; utf-8 is a characteristic of how they're serialised on the wire.

bsdphk:
I am, as amply communicated previously, very firmly against this very idea.

I am against it in principle because the last two decades have seen the HTTP protocol systematically move away from being UX-adjecent to being pure IPC messaging, to the extent that "HTTP is the new TCP" has become a meme.

This is why H2 and H3 wisely, and explicitly, dropped the "reason phrase" from the response status message, to avoid wading into the I18N sump at the protocol handling level.

Adding a Yet Another Escape Mechanism, when we already have a well defined sf-blob to safely transfer arbitrary byte-sequences is also a blatant violation of Getty's first rule:

»Do not add new functionality unless an implementor cannot complete a real application without it.«

Precisely what is it, that who cannot do, using sf-blob to transfer these byte-sequences ?!

Second, the specific proposal contains the kind of half-assery which time and time again has come back to haunt us as security issues.

Is the contents UTF8 sequence or not ?

Who can tell ?

The proposed text makes this entirely up to the sender, and makes no requirement for them declare it in the message.

We're supposed to take a byte sequence and "present it to users", but we're not told if it is UTF8, BAUDOT or EBCDIC ?

And dont forget that validating Unicode requires an up-to-date copy of the Unicode registry.

How can that possibly go wrong...

And reusing %xx encoding in a place where it has never before been used ?

How many pieces of code will unintentionally interpret %xx encoding and emit random byte-values where they are not supposed to go ?

You can bet your sweet arse that there are places out there which do %xx substitution on the entire HTTP message because the URI field needs it:  Now add a couple of %0d%0a in a HTTP message and you have a smugling attack.

If we want to provide a mechanism for "strings to be presented to users" in HTTP fields, we should do it properly.

As the absolute minimum, that requires nailing down the encoding and a negotiation mechanism for cultural choices of language and dialect.

But in my world, that does certainly not belong at the "HTTP is the new TCP" level of the stack.

Summary:

* Needless addition of complexity, all uses cases are already covered by sf-blob.
* Half-assed specification invites security issues down the road.
* Wrong level of the protocol stack to be doing this.

draft-ietf-httpbis-sfbis.md
```diff
@@ -400,7 +401,7 @@ Strings are zero or more printable ASCII {{!RFC0020}} characters (i.e., the rang
 
 Unicode is not directly supported in Strings, because it causes a number of interoperability issues, and -- with few exceptions -- field values do not require it.
```
reschke, line 402:
```suggestion
Non-ASCII characters are not directly supported in Strings, because it causes a number of interoperability issues, and -- with few exceptions -- field values do not require it.
```

draft-ietf-httpbis-sfbis.md
```diff
@@ -464,6 +465,22 @@ For example:
 Example-Date: @1659578233
 ~~~
 
+### Display Strings {#displaystring}
+
+Display Strings are similar to Strings, in that they consist of zero or more characters, but they allow unicode content, unlike Strings.
+
+Display Strings are intended for use in cases where a value is displayed to end users, and therefore may need to carry non-ASCII content. It is NOT RECOMMENDED that they be used in situations where a String ({{string}}) or Token ({{token}}) would be adequate, because unicode has processing considerations (e.g., normalization) and security considerations (e.g., homograph attacks) that make it more difficult to handle correctly.
```
reschke, line 472:
```suggestion
Display Strings are intended for use in cases where a value is displayed to end users, and therefore may need to carry non-ASCII content. It is NOT RECOMMENDED that they be used in situations where a String ({{string}}) or Token ({{token}}) would be adequate, because Unicode has processing considerations (e.g., normalization) and security considerations (e.g., homograph attacks) that make it more difficult to handle correctly.
```

draft-ietf-httpbis-sfbis.md
```diff
@@ -464,6 +465,22 @@ For example:
...
+
+Note that Display Strings do not indicate the language used in the value; that can be done separately if necessary (e.g., with a parameter).
+
+For example:
+
+~~~ http-message-new
+Example-DisplayString: %"This is intended for display to %C3%BCsers."
+~~~
+
+
```
reschke, line 482: The prose should say what the display string in the example is

reschke (commented):
There would be other approaches we could try, but in the end which one we choose doesn't matter a lot.

One thing we need to check is that we are clear whether the \ escapces from Strings can be mixed into Display Strings (the parsing seems to imply that).

We may also want to discuss NUL.

bsdphk:
> @bsdphk - but then why didn't you reject SF-Date? This extension falls into a very very similar category.

Because SF-Date provides the biggest performance gains you can make at this level of the stack.

reschke:
> Because SF-Date provides the biggest performance gains you can make at this level of the stack, without opening a can of worms of new security issues, the way this proposal does.

Care to elaborate? How is SF-Date better than SF-Integer?

draft-ietf-httpbis-sfbis.md
```diff
@@ -398,9 +399,9 @@ Note that the serialization algorithm ({{ser-decimal}}) rounds input with more t
 
 Strings are zero or more printable ASCII {{!RFC0020}} characters (i.e., the range %x20 to %x7E). Note that this excludes tabs, newlines, carriage returns, etc.
 
-Unicode is not directly supported in Strings, because it causes a number of interoperability issues, and -- with few exceptions -- field values do not require it.
+Non-ASCII characters are not directly supported in Strings, because it causes a number of interoperability issues, and -- with few exceptions -- field values do not require it.
```
mnot, line 402:
```suggestion
Non-ASCII characters are not directly supported in Strings, because they cause a number of interoperability issues, and -- with few exceptions -- field values do not require them.
```

draft-ietf-httpbis-sfbis.md
martinthomson: Oh, I see.  This is fine.  Capitalize "Unicode".

martinthomson:
I am not concerned about encodings (it is UTF-8, always, and a failure to convert to a string or validate as UTF-8 should be treated as a parse error).  We can note that this can encode any byte sequence, but that non-UTF-8 sequences are invalid.  We don't need to validate that the characters are valid or in any particular form in order for them to be useful; UTF-8 decoding is simple, easy, and sufficient at this layer.

I am not concerned about NUL, though we should definitely note that it is possible to encode `%00`, which might be surprising.

I'm not concerned about display requirements.  Software always needs to be careful with unsanitized or untrustworthy inputs, this is no different.  Encoding strings as binary and interpreting them that way has exactly the same problem.

I'm not concerned about smuggling.  Or rather, I think we should note the possibility and say "don't serialize display strings back into messages without escaping them properly".

So I think that PHK has pointed out where the security considerations on this change are insufficient, but that insufficiency only exists to the extent that this change isn't highlighting the risks involved.

In addition to those, we should also note that language tagging (it is often a requirement when handling Unicode data that language choice determines how that handling is performed) is not directly provided.  We can point out that parameters could be used for this purpose, if necessary, though there are obviously other ways to obtain that information as well (like `Content-Language`).

davidben:
(unimportant bikeshed) Would Unicode Strings be a better name than Display Strings? I could see folks wanting to transmit Unicode not immediately for display. E.g. maybe I'm sending my name to the server to save somewhere. The value will, eventually, be displayed back but that happens after this HTTP transaction. Or perhaps it's a filename a la Content-Disposition, which has both storage and display purposes.

(Of course, sending a name for storage probably will go in a request body, and what filenames are is OS-dependent and Complicated. But nevermind that. :-) )

mnot:
@davidben I avoided that name purposefully, because developers have been taught to 'always use unicode', which often isn't appropriate for fields.

mnot:
@martinthomson all good points. Suggestions for text welcome :)

mnot:
@martinthomson I've added some straw-man text to security considerations; I'm sure you can find improvements.

reschke (approved):

reschke:
FWIW: I changed my review to "approved" so that we can this change in; remaining minor issues IMHO should be handled in separate, more fine-grained tickets.

draft-ietf-httpbis-sfbis.md
```diff
@@ -665,6 +684,20 @@ Given a Date as input_integer, return an ASCII string suitable for use in an HTT
...
 
+### Serializing a Display String {#ser-display}
+
+Given a string of Unicode characters as input_string, return an ASCII string suitable for use in an HTTP field value.
+
+0. Let byte_array be the result of applying UTF-8 encoding {{UTF8}} to input_string. If there is an error in doing so, fail parsing.
+1. Let encoded_string be an empty string.
+2. For each byte in byte_array:
+  1. If byte is %x25 ("%"), append "%25" to encoded_string.
+  2. If byte is in the ranges x00-1f or %x7f-ff, append the result of applying the percent-encoding defined in {{Section 2.1 of URI}} to encoded_string.
```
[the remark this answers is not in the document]
mnot: How so? %x25 isn't in those ranges.

duerst:
I looked at the diffs for this, and I'm missing any mention of the quotes. Decoding goes directly from the leading '%' to the actual payload, without mentioning that you have to check/skip for the starting quote, and without mentioning that you have to stop at the final quote. The quotes appear correctly in the ABNF, but unless I have missed something, they are not mentioned anywhere else.

mnot:
@duerst that happens when parse / serialise a string are called.

PR #2503, opened 2023-04-06
Rewrite Signature Collision Security Considerations

dennisjackson:
This change expands the guidance for avoiding signature replays and clarifies the risks from signature collisions by rewriting the section to be more general. It also brings together the new signature collisions section and the section on signature binding so they're consecutive.


yaronf:
Is this not virtually the same issue as [key commitment](https://eprint.iacr.org/2020/1153.pdf) in symmetric keys? If so, we could add a `@fingerprint` signed component (a hash of the public key used for signing) to be used in the signature-of-signature use case.

PR #2505, opened 2023-04-12
Signature component encodings

### LINKED ISSUE #2415: message-signatures: ASCII restriction in field-value to byte sequence
in 2.1.3:

> Encode the bytes of the resulting field value's ASCII representation as a Byte Sequence.

What if the field value contains non-ASCII characters (from obs-text)?

Of course that could be considered an error, but I believe the intention is to be able to represent anything here. So the text should refer to the field value as byte sequence, in which case the encoding step would not be needed.

### LINKED ISSUE #2416: message-signatures: query parameters vs encoding
In 2.2.8, it might be good to mention that the referenced algorithm only suppors query parameters using (percent-escaped) UTF-8 encoding (this came as a surprise to me because I assumed it works the same as the algorithm used for forms decoding).

### LINKED ISSUE #2417: message-signatures: format of signature base
2.5:

> The signature base is a US-ASCII string containing the canonicalized HTTP message components covered by the signature.

This seems to be inconsistent with Section 2.2.8 ("Query Parameters"), which could create non-ASCII component values.

jricher:
Addresses encodings of signature component values for query parameters and field values.

Closes #2417
Closes #2416
Closes #2415


draft-ietf-httpbis-message-signatures.md
```diff
@@ -287,7 +287,7 @@ The following sections define component identifier names, their parameters, thei
 
 The component name for an HTTP field is the lowercased form of its field name as defined in {{Section 5.1 of HTTP}}. While HTTP field names are case-insensitive, implementations MUST use lowercased field names (e.g., `content-type`, `date`, `etag`) when using them as component names.
 
-The component value for an HTTP field is the field value for the named field as defined in {{Section 5.5 of HTTP}}. The field value MUST be taken from the named header field of the target message unless this behavior is overridden by additional parameters and rules, such as the `req` and `tr` flags, below.
+The component value for an HTTP field is the field value for the named field as defined in {{Section 5.5 of HTTP}}. The field value MUST be taken from the named header field of the target message unless this behavior is overridden by additional parameters and rules, such as the `req` and `tr` flags, below. For most fields, the field value is an ASCII string. Other encodings could exist, but extra care needs to be taken when dealing with these values to prevent inclusion if illegal octets (such as a newline character).
```
reschke, line 290: I don't think "extra care" is sufficient here. If the field value contains an octet >= 128, it can't eb placed into the signature base, which is said use ASCII. The spec needs to exactly tell implementers what to do here.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -434,7 +434,7 @@ If this parameter is included with a component identifier, the component value M
 2. For each field value in the set:
     0. Strip leading and trailing whitespace from the field value. Note that since HTTP field values are not allowed to contain leading and trailing whitespace, this will be a no-op in a compliant implementation.
     1. Remove any obsolete line-folding within the line and replace it with a single space (" "), as discussed in {{Section 5.2 of HTTP1}}. Note that this behavior is specific to {{HTTP1}} and does not apply to other versions of the HTTP specification.
-    2. Encode the bytes of the resulting field value's ASCII representation as a Byte Sequence.
+    2. Encode the bytes of the resulting field value's bytes as a Byte Sequence. Note that most fields are restricted to ASCII characters, but other octets could be included and need to be treated with care.
```
reschke, line 437: What does "encode as a byte sequence" mean here? (Sounds like an identity transformation).

reschke (changes requested):

draft-ietf-httpbis-message-signatures.md
jricher: Byte Sequence is a data type from structured fields, so this is saying take the bytes and put them in that data structure.

draft-ietf-httpbis-message-signatures.md
jricher: What would you suggest? The obvious thing is that if an octet is not in the ASCII range then it needs to be encoded, likely percent-encoded first.

draft-ietf-httpbis-message-signatures.md
reschke: But that's a breaking change which also invalidates all examples, no?

reschke: Hu? https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-16.html#section-2.5-12 has:

~~~
"@method": POST
"@authority": example.com
"@path": /foo
"content-digest": sha-512=:WZDPaVn/7XgHaAy8pmojAkGWoRx2UFChF41A2svX\
  +TaPm+AbwAgBWnrIiYllu7BNNyealdVLvRwEmTHWXvJwew==:
"content-length": 18
"@signature-params": ("@method" "@authority" "@path" \
  "content-digest" "content-length" "content-type")\
  ;created=1618884473;keyid="test-key-rsa-pss"~~~

draft-ietf-httpbis-message-signatures.md
reschke: percent encoding works for me, but you need to make sure that then "%" is percent-escaped as well.

draft-ietf-httpbis-message-signatures.md
jricher: None of those use the `bs` flag which is the subject of this section.

reschke: Well, in that case you need to clearly point out thar field values that use non-ASCII can only be used *with* the "bs" flag (otherwise what happens? an error?)

draft-ietf-httpbis-message-signatures.md
reschke: I still don't think "extra care" is sufficient here. Just say what the implementer needs to do.

reschke (changes requested):

draft-ietf-httpbis-message-signatures.md
```diff
@@ -844,19 +849,44 @@ Indicating the `baz`, `qux` and `param` named query parameters would result in t
...
+"@query-param";name="qux":(SP)
+"@query-param";name="param": value
+~~~
+
+This derived component has some limitations. Specifically, the algorithm in {{HTMLURL}} only supports query parameters using percent-escaped UTF-8 encoding. Other encodings are not supported.
+Additionally, multiple instances of a named parameter are not reliably supported in the wild. If a parameter name occurs multiple times in a request, the named query parameter MUST NOT be included. If multiple parameters are common within an application, it is RECOMMENDED to sign the entire query string using the `@query` component identifier defined in {{content-request-query}}.
+
+The encoding process allows query parameters that include newlines or other problematic characters in their values, or with alternative encodings such as using the plus character to represent spaces. For the query parameters in this message:
+
+~~~ htto-message
```
yaronf, line 865:
```suggestion
~~~ http-message
```

draft-ietf-httpbis-message-signatures.md
jricher: That is incorrect. They are encoded with percent encoding as you suggested above.

draft-ietf-httpbis-message-signatures.md
```diff
@@ -287,7 +287,7 @@ The following sections define component identifier names, their parameters, thei
...
 
-The component value for an HTTP field is the field value for the named field as defined in {{Section 5.5 of HTTP}}. The field value MUST be taken from the named header field of the target message unless this behavior is overridden by additional parameters and rules, such as the `req` and `tr` flags, below.
+The component value for an HTTP field is the field value for the named field as defined in {{Section 5.5 of HTTP}}. The field value MUST be taken from the named header field of the target message unless this behavior is overridden by additional parameters and rules, such as the `req` and `tr` flags, below. For most fields, the field value is an ASCII string as recommended by {{HTTP}}, and the component value is exactly that string. Other encodings could exist in some implementations, and any characters outside the ASCII range MUST be encoded using the percent-encoding mechanism defined in {{Section 2.1 of URI}}.
```
reschke, line 290: If you use percent-encoding, you'll have to encode "%" as well.

PR #2507, opened 2023-04-13
Editorial clarifications of Appendix A

### LINKED ISSUE #2506: Digest compression example inflates payload
@reschke reported on the list 

> draft-ietf-httpbis-digest-headers-11, "Appendix A. Resource Representation and Representation Data"

> Compression is an efficient way to reduce the content length.
> 
> But then, the example payload is indeed longer as in the previous example.
> 
> Maybe it would be good to use the same JSON payload in both examples, in
> order to avoid potential confusion? (Also, adding Content-Length might
> clarify things).

The way the test plays out is confusing and distracting. We don't really care whether the representation is smaller or not. The example should be succinct for understanding, changing the "plaintext" to something that compresses well is not going to improve comprehension IMO. 

I'll prepare an editorial PR to make this read better without changing the section significantly.

LPardue:
I tried doing it the way I suggested in #2506 and it wasn't ggod. So I ended up with something closer to what I believe @reschke suggested.

Closes #2506

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -672,50 +672,61 @@ instead.
 
 # Resource Representation and Representation Data {#resource-representation}
 
-The following examples show how representation metadata, content transformations,
-and method impacts on the message and content. When the content
-contains non-printable characters (e.g., when it is compressed) it is shown as
-a sequence of hex-encoded bytes.
+This section following examples show how representation metadata, content
+transformations, and method impacts on the message and content. These examples a
+not exhaustive.
+
+Unless otherwise indicated, the examples are based on the JSON object `{"hello":
+"world"}` followed by an LF. When the content contains non-printable characters
+(e.g., when it is compressed) it is shown as a sequence of hex-encoded bytes.
+
+Consider a client that wishes to upload a JSON object using the PUT method. It
+could do this using the application/json content type without any content
+coding.
 
 ~~~ http-message
 PUT /entries/1234 HTTP/1.1
 Host: foo.example
 Content-Type: application/json
+Content-Length: 19
 
 {"hello": "world"}
 ~~~
 {: title="Request containing a JSON object without any content coding"}
 
-Compression is an efficient way to reduce
-the content length.
+However, the use of content coding is quite common. The client could also upload
+the same data with a gzip coding ({{Section 8.4.1.3 of RFC9110}}). Note that in
+this case, the `Content-Length` contains a larger value due to the coding
+overheads.
 
 ~~~ http-message
 PUT /entries/1234 HTTP/1.1
 Host: foo.example
 Content-Type: application/json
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-digest-headers.md
```diff
@@ -725,31 +736,38 @@ HTTP/1.1 400 Bad Request
 ~~~
 {: title="An error response for a malformed content"}
 
-A Range-Request affects the transferred message content,
-conveying a partial representation of the JSON object
-in {{ex-put-gz}}.
+A Range-Request affects the transferred message content. In this example, the
+client is accessing the resource at `/entires/1234`, which is the JSON object
+`{"hello": "world"}` followed by an LF. However, the client has indicates a
```
reschke, line 741:
```suggestion
`{"hello": "world"}` followed by an LF. However, the client has indicated a
``

**Review by reschke — COMMENTED** [review]
I didn't check the payloads, but the direction is clearly good.

**Review by martinthomson — COMMENTED** [review]
Do you really need so many repetitions of messages *without* digest fields?

**Comment by LPardue** [comment]
Fair question. We heard al9ng the way people struggle with rhus aspect of HTTP semantics and without such an appendix, people think digest 8s adding something on top of what semantics can already do. 

Possibly we could drop the last example in appendix but then not sure it helps much.

File: draft-ietf-httpbis-digest-headers.md [hunk 2]
```diff
@@ -672,50 +672,61 @@ instead.
...
-and method impacts on the message and content. When the content
-contains non-printable characters (e.g., when it is compressed) it is shown as
-a sequence of hex-encoded bytes.
+This section following examples show how representation metadata, content
+transformations, and method impacts on the message and content. These examples a
+not exhaustive.
+
+Unless otherwise indicated, the examples are based on the JSON object `{"hello":
+"world"}` followed by an LF. When the content contains non-printable characters
+(e.g., when it is compressed) it is shown as a sequence of hex-encoded bytes.
```
ioggstream, line 681:
```suggestion
(e.g., when it is encoded) it is shown as a sequence of hex-encoded bytes.
```

We don't care if the encoding compresses or not ;)

draft-ietf-httpbis-digest-headers.md
```diff
@@ -725,31 +736,39 @@ HTTP/1.1 400 Bad Request
 ~~~
 {: title="An error response for a malformed content"}
 
-A Range-Request affects the transferred message content,
-conveying a partial representation of the JSON object
-in {{ex-put-gz}}.
+A Range-Request affects the transferred message content. In this example, the
+client is accessing the resource at `/entires/1234`, which is the JSON object
+`{"hello": "world"}` followed by an LF. However, the client has indicated a
+``
```
ioggstream, line 742:
```suggestion
```

ioggstream (commented):
See minor suggestions.

PR #2518, opened 2023-04-17
sfbis: Fix ABNF to include sf-date in bare-item

gibson042:
Right now the `sf-date` nonterminal is (presumably unintentionally) orphaned.


mnot:
Thanks!

PR #2519, opened 2023-04-17
sfbis: Uniformly use American spellings

gibson042:
There are currently some uses of "serialise" in place of the far more common "serialize", including one case that breaks the pattern of header names ("Seriali**z**ing a List", "Seriali**z**ing a Dictionary", "Seriali**z**ing an Item", …, "Seriali**s**ing a Date").

This PR also includes an analogous update of "summarises" to "summarizes" to increase the general pattern of American spellings.

### DIFF
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -202,7 +202,7 @@ The effect of upgrading a Structured Fields implementation is that some field va
 
 # Structured Data Types {#types}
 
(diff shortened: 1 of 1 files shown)


bsdphk:
Speaking Danglish myself, I am disqualified from having an opinion on this subject matter :-)

mnot:
Despite my Australianised sensibilities, I see that 8941 was normalised to American spelling by the RFC Editor, and changing things now would introduce more diffs. Thanks for the PR.

PR #2523, opened 2023-04-25
call out query parameter handling oddities

jricher:
Point out that many frameworks smush together form parameters and query parameters into one bucket, and that can cause unexpected shadowing.

Some examples changed to avoid ambiguity.


adeinega:
This is great.

PR #2527, opened 2023-04-26
Be more explicit about content processing and trailer

### LINKED ISSUE #2526: Digest: Section 6.4: Usage in Trailer Fields
On the list, Rob Sayre asked:

> I'm a little confused by the last paragraph of Section 6.4: Usage in Trailer Fields.
> 
> > Not every hashing algorithm is suitable for use in the trailer section, some may
> > require to preprocess the whole content before sending a message (e.g., see [I-D.thomson-http-mice]).
> 
> Firstly, there's a small grammar problem with "require to...". But that's not my main concern.
> 
> Isn't the point of using a trailer to work around those algorithms that require processing all of the content?
> 
> For example, the code samples for Python's "hashlib" show SHA-256:
> https://docs.python.org/3/library/hashlib.html
> 
> Presumably, you'd call the update() method as you put bits on the wire, and then put the result of digest() in the trailer.
> 
> draft-thomson-http-mice-02 would seem to address issues with "eager processing" that the document covered in the previous paragraph.
> 
> Anyway, I think there's something to fix here, even if I am confused about the technical issue.

I agree it seems we can improve the text here a little.

Maybe @ioggstream or @martinthomson can elaborate as to why a digest trailer is not suitable in this case?

LPardue:
Fixes #2526

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -579,8 +579,14 @@ When Integrity fields are used in a trailer section, the field-values are receiv
 Eager processing of content before the trailer section prevents digest validation, possibly leading to
 processing of invalid data.
 
-Not every hashing algorithm is suitable for use in the trailer section, some may require to preprocess
-the whole content before sending a message (e.g., see {{?I-D.thomson-http-mice}}).
+One of the benefits of using Integrity fields in a trailer section is that it
+allows hashing of bytes as they are sent or received. However, it is possible to
+design a hashing algorithm that requires processing of content in such a way
+that would negate these benefits. For example, Merkle Integrity Content Encoding
+{{?I-D.thomson-http-mice}} requires content to be processed in reverse order.
+This means the complete data needs to be available, which means there is
+negligible processing difference in sending an Integrity field in a header or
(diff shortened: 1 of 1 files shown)


ioggstream:
LGTM

draft-ietf-httpbis-digest-headers.md
```diff
@@ -579,8 +579,14 @@ When Integrity fields are used in a trailer section, the field-values are receiv
 Eager processing of content before the trailer section prevents digest validation, possibly leading to
 processing of invalid data.
 
-Not every hashing algorithm is suitable for use in the trailer section, some may require to preprocess
-the whole content before sending a message (e.g., see {{?I-D.thomson-http-mice}}).
+One of the benefits of using Integrity fields in a trailer section is that it
+allows hashing of bytes as they are sent or received. However, it is possible to
```
sayrer, line 583: I don't think you want "or received" here. If the digest is in a header, you can still hash as you're receiving the content.

LPardue, line 583:
```suggestion
allows hashing of bytes as they are sent. However, it is possible to
```

PR #2529, opened 2023-05-01
Explain encoding for dots in labels

### LINKED ISSUE #2486: Next-hop aliases: Escaping "."
The "[unreserved characters](https://www.rfc-editor.org/rfc/rfc3986#section-2.3)" set includes ".", but we actually need to escape "." to avoid [Period Injection](https://arxiv.org/pdf/2205.05439.pdf) attacks.

tfpauly:
Closes #2486

### DIFF
--- a/draft-ietf-httpbis-alias-proxy-status.md
+++ b/draft-ietf-httpbis-alias-proxy-status.md
@@ -75,13 +75,9 @@ in the chain, the first name in the `next-hop-aliases` list would be the value i
 record for the original hostname, and the final name in the `next-hop-aliases` list would
 be the name that ultimately resolved to one or more addresses.
 
-The list of DNS names in `next-hop-aliases` use a comma (",") as a separator between names.
-DNS names normally just contain alphanumeric characters and hyphens ("-"), although they
-are allowed to contain any character {{?RFC1035, Section 3.1}}, including a comma. To
-prevent commas or other special characters in names leading to incorrect parsing,
-any characters that appear in names in this list that do not belong to the set of URI
-Unreserved Characters {{!RFC3986, Section 2.3}} MUST be percent-encoded as
-defined in {{!RFC3986, Section 2.1}}.
+The list of DNS names in `next-hop-aliases` uses a comma (",") as a separator between names.
+Note that if a comma is included in a name itself, the comma must be encoded as described in
+{{encoding}}.
 
 For example, consider a proxy "proxy.example.net" that receives the following records when
 performing DNS resolution for the next hop "host.example.com":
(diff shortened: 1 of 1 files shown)


bemasc (approved):
With the suggested change, I think this is probably good enough.  Strictly speaking, I'm not sure it fully specifies a decoding algorithm, and ideally there would be some test vectors, but I think implementors would be able to figure it out from this much.

draft-ietf-httpbis-alias-proxy-status.md
```diff
@@ -112,6 +108,36 @@ in which this parameter is not present.
...
+"dot\.label.example.com" would be encoded within a `next-hop-aliases` parameter as follows:
+
+~~~ example
+Proxy-Status: proxy.example.net; next-hop=2001:db8::1;
+    next-hop-aliases="dot%5C.label.example.com,service1.example-cdn.com"
+~~~
+
+Upon parsing this name, "dot%5C.label" MUST be treated as a single label.
+
+Similarly the "\" character in a label MUST be escaped as "\\".  Other uses of "\" MUST NOT appear in the label after percent-decoding.
```
bemasc, line 140: Hooray escaping.

```suggestion
as "\\.". Since the "\\" character itself will be percent-encoded, the name
"dot\.label.example.com" would be encoded within a `next-hop-aliases` parameter as follows:

~~~ example
Proxy-Status: proxy.example.net; next-hop=2001:db8::1;
    next-hop-aliases="dot%5C.label.example.com,service1.example-cdn.com"
~~~

Upon parsing this name, "dot%5C.label" MUST be treated as a single label.

Similarly the "\\" character in a label MUST be escaped as "\\\\".  Other uses of "\\" MUST NOT appear in the label after percent-decoding.
```

draft-ietf-httpbis-alias-proxy-status.md
```diff
@@ -112,6 +108,36 @@ in which this parameter is not present.
...
+"dot\\.label.example.com" would be encoded within a `next-hop-aliases` parameter as follows:
+
+~~~ example
+Proxy-Status: proxy.example.net; next-hop=2001:db8::1;
+    next-hop-aliases="dot%5C.label.example.com,service1.example-cdn.com"
+~~~
+
+Upon parsing this name, "dot%5C.label" MUST be treated as a single label.
+
+Similarly the "\\" character in a label MUST be escaped as "\\\\".  Other uses of "\\" MUST NOT appear in the label after percent-decoding.
```
tfpauly, line 140:
```suggestion
Similarly the "\\" character in a label MUST be escaped as "\\\\".  Other uses of "\\" MUST NOT appear in the label after percent-decoding.

```

PR #2545, opened 2023-05-22
Link to HTTP Section 8.1 when talking about representation data

LPardue:
In a few places we link to HTTP Section 3.2, which doesn't do the best
job of defining representation data. 8.1 is a better section for that.


LPardue:
cc @martinduke

reschke (approved):

PR #2546, opened 2023-05-24
Be more explicit in the abstract

LPardue:
Just a suggestion, based on frustration that I had to read all of the abstract and Introduction until I see what the new parameter is actually named.


tfpauly (approved):

tfpauly:
Actually this would overlap with #2551. Let's follow up there?

PR #2556, opened 2023-05-24
Remove reserved status and tweak Ref's description

### LINKED ISSUE #2555: Remove "reserved" status from IANA considerations
John Scudder raised during IESG review:

> In your description of the Status template field, you have ""reserved" - for
algorithms that use a reserved token value that cannot be expressed in
Structured Fields". This is a well-formed sentence but I have no idea what it
means. I made a desultory attempt to suss it out by searching the document for
"token" and this was the only occurrence. If people who will actually be making
use of the registry can be expected to make sense of it, then feel free to
disregard my comment, of course.

This reminded me at some point, I'd sketched out a design for the "Hash Algorithms for HTTP Digest Fields" registry that worked differently with what we ended up with now. In that design, it would have been possible to register an algorithm in the table under some friendly name and then separately register some Structured-Fields identifier that it related to. This would have made it possible to register an algorithm but reserve it from being used, in order to (for example) try and maintain some parity with old digest.

Needless to say, the original sketch design had flaws and we didn't go that way. Instead, the registry requires entries to be identified using a Structured Fields key, which makes the idea of a reserved status due to a key that can't be used rather strange.

TL;DR lets remove the "reserved" status because it isn't used by anything already and doesn't have much logic behind how it would be used in future.

LPardue:
Fixes #2555

### DIFF
--- a/draft-ietf-httpbis-digest-headers.md
+++ b/draft-ietf-httpbis-digest-headers.md
@@ -489,10 +489,9 @@ should use the following template:
      - "standard" - for standardized algorithms without known problems,
      - "provisional" - for non-standard or unproven algorithms,
      - "insecure" - for insecure algorithms,
-     - "reserved"  - for algorithms that use a reserved token value that cannot be expressed in Structured Fields
  - Description: a short description of the algorithm
- - Reference(s): pointer(s) to the primary document(s) defining the technical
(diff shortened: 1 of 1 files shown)


ioggstream:
SGTM. Feel free to merge if you are sure we won't need this.

Kind regards,
R

PR #2561, opened 2023-06-01
Adjust ignored preferences text.

sayrer:
Apologies in advance.


LPardue:
No apology needed. I stuttered over this same sentence yesterday when fixing an unrelated nit. I almost changed it then but didn't want to mix concerns.

Given that somebody else has raised a comment and provided an alternative I agree with, leads me to believe its worth fixing. Thanks.

mnot:
Much cleaner, +1

LPardue (approved):

PR #2569, opened 2023-06-16
IPv6 next hop examples need to be strings

### LINKED ISSUE #2564: SF syntax violations in examples
#2562 exposed an issue in the use of Structured Fields; IPv6 addresses start with a number, but contain non-numeric characters, so they fail parsing. Fix is to wrap them in quotes (to make them a String).

This affects all three examples.

tfpauly:
Closes #2564


tfpauly:
Woo it passes!

PR #2571, opened 2023-06-23
Remove HMAC and switch to Authorization

### LINKED ISSUE #2432: New header field vs existing Authorization header
Is there a technical reason why the `WWW-Authenticate` header was not chosen for this draft? That can be sent without a corresponding challenge (in an unprompted way), as we note in the [HTTP authentication scheme for Privacy Pass](https://ietf-wg-privacypass.github.io/base-drafts/draft-ietf-privacypass-auth-scheme.html). It seems like that would work in this context, too.

### LINKED ISSUE #2440: Signature hash algorithm is ambiguous
From Ilari Liusvaara on the [list](https://lists.w3.org/Archives/Public/ietf-http-wg/2022OctDec/0027.html):

> There is no requirement to include hash algorithm in signatures. There are TLS signature algorithms that mean totally different things depending on hash function, and more of those could appear in the future. E.g, signatures 7 and 8 already have double meaning (EdDSA [hash 8] and some Chinese stuff [hash 7]).

From @nharper on the [list](https://lists.w3.org/Archives/Public/ietf-http-wg/2022OctDec/0028.html):

> I agree with Ilari - do not use the TLS SignatureAlgorithm and HashAlgorithm registries that were orphaned by RFC 8447. For (asymmetric) signatures, you could use the 16-bit TLS SignatureScheme registry (https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-signaturescheme).

DavidSchinazi:
Fixes #2440 
Fixes #2432

### DIFF
--- a/draft-ietf-httpbis-unprompted-auth.md
+++ b/draft-ietf-httpbis-unprompted-auth.md
@@ -126,8 +126,9 @@ it is uniquely bound to the TLS session {{!RFC7627}}, the Signature
 authentication scheme requires either one of the following properties:
 
 * The TLS version in use is greater or equal to 1.3 {{TLS}}.
-* The TLS version in use is greater or equal to 1.2 and the Extended
-Master Secret extension {{RFC7627}} has been negotiated.
+
+* The TLS version in use is 1.2 and the Extended Master Secret extension
+  {{RFC7627}} has been negotiated.
 
 Clients MUST NOT use the Signature authentication scheme on connections that do
 not meet one of the two properties above. If a server receives a request that
@@ -223,15 +224,16 @@ clear-text TLS Client Hello extensions.
 
 The authentication proofs described in this document are not bound to
 individual HTTP requests; if the key is used for authentication proofs on
-multiple requests they will all be identical. This allows for better
-compression when sending over the wire, but implies that client implementations
-that multiplex different security contexts over a single HTTP connection need
-to ensure that those contexts cannot read each other's header fields.
-Otherwise, one context would be able to replay the Authorization header field
-of another. This constraint is met by modern Web browsers. If an attacker were
-to compromise the browser such that it could access another context's memory,
-the attacker might also be able to access the corresponding key, so binding
-authentication to requests would not provide much benefit in practice.
+multiple requests on the same connection, they will all be identical. This
+allows for better compression when sending over the wire, but implies that
+client implementations that multiplex different security contexts over a single
+HTTP connection need to ensure that those contexts cannot read each other's
+header fields. Otherwise, one context would be able to replay the Authorization
+header field of another. This constraint is met by modern Web browsers. If an
+attacker were to compromise the browser such that it could access another
+context's memory, the attacker might also be able to access the corresponding
+key, so binding authentication to requests would not provide much benefit in
+practice.
 
 Key material used for the Signature HTTP authentication scheme MUST NOT be
 reused in other protocols. Doing so can undermine the security guarantees of
--- a/draft-ietf-httpbis-unprompted-auth.md
+++ b/draft-ietf-httpbis-unprompted-auth.md
@@ -239,7 +239,7 @@ Key material used for the Signature HTTP authentication scheme MUST NOT be
 reused in other protocols. Doing so can undermine the security guarantees of
 the authentication.
 
-Origins offering this scheme are able to link requests that use the same key.
+Origins offering this scheme can link requests that use the same key.
 However, requests are not linkable across origins if the keys used are specific
 to the individual origins using this scheme.
 
--- a/draft-ietf-httpbis-unprompted-auth.md
+++ b/draft-ietf-httpbis-unprompted-auth.md
@@ -145,7 +145,9 @@ then transmitted to the server using the Authorization field.
 
 # Authentication Parameters
 
-This specification defines the following authentication parameters:
+This specification defines the following authentication parameters. These
+parameters use structured fields ({{STRUCTURED-FIELDS}}) in their definition,
+even though the Authorization field itself does not use structured fields.
 
 ## The k Parameter {#parameter-k}
 
--- a/draft-ietf-httpbis-unprompted-auth.md
+++ b/draft-ietf-httpbis-unprompted-auth.md
@@ -1,5 +1,5 @@
 ---
-title: HTTP Unprompted Authentication
+title: The Signature HTTP Authentication Scheme
 docname: draft-ietf-httpbis-unprompted-auth-latest
 submissiontype: IETF
 number:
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -55,47 +55,47 @@ informative:
 
 --- abstract
 
-Existing HTTP authentication mechanisms are probeable in the sense that it is
+Existing HTTP authentication schemes are probeable in the sense that it is
```
martinthomson, line 58: Is your abstract really just a repetition of the first paragraph of the introduction?

Maybe you can instead say what the document does (the last sentence, basically).

draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -109,129 +109,73 @@ This document uses the following terminology from {{Section 3 of
...
+authentication and key exchange mechanism {{?QUIC-TLS=RFC9001}}.
 
 Because the TLS keying material exporter is only secure for authentication when
-it is uniquely bound to the TLS session {{!RFC7627}}, the Signature and HMAC
-authentication schemes require either one of the following properties:
+it is uniquely bound to the TLS session {{!RFC7627}}, the Signature
+authentication scheme requires either one of the following properties:
 
 * The TLS version in use is greater or equal to 1.3 {{TLS}}.
 * The TLS version in use is greater or equal to 1.2 and the Extended
```
martinthomson, line 129:
```suggestion
* The TLS version in use is 1.2 and the Extended
```

draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -109,129 +109,73 @@ This document uses the following terminology from {{Section 3 of
...
 
-The "Unprompted-Authentication" header field allows a user agent to authenticate
-with an origin server. The authentication is scoped to the HTTP request
-associated with this header field. The value of the Unprompted-Authentication
-header field is a credentials object, as defined in {{Section 11.4 of HTTP}}.
-Credentials contain an authentication scheme followed by optional authentication
-parameters.
+The user agent leverages a TLS keying material exporter {{!KEY-EXPORT=RFC5705}}
+with the label "EXPORTER-HTTP-Signature-Authentication" to generate a 32-byte
+symmetric key. That symmetric key is then used as nonce which can be signed
```
martinthomson, line 141: I might strike "used as [a] nonce" here.  The signature will be usable for multiple requests, so it isn't strictly a "number that is used once".

The key property this relies on is the contributory nature of TLS exporter output.  The client (or any other entity, for that matter) is unable to predict the value that the exporter might produce for a given connection.

draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -109,129 +109,73 @@ This document uses the following terminology from {{Section 3 of
...
+using the client's chosen asymmetric private key. The resulting signature is
+then transmitted to the server using the Authorization field.
 
 # Authentication Parameters
 
-This specification defines the following authentication parameters, they can be
-used by the authentication schemes defined in {{schemes}}.
+This specification defines the following authentication parameters:
```
martinthomson, line 147: You are using structured fields for the definitions of these, even though Authorization doesn't use SF.  That's good, but I might take this opportunity to make this point.

draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -253,10 +197,9 @@ verification is not observable.
...
-Since the Signature and HMAC HTTP Authentication Schemes leverage TLS keying
-material exporters, their output cannot be transparently forwarded by HTTP
-intermediaries. HTTP intermediaries that support this specification have two
-options:
+Since the Signature HTTP authentication scheme leverages TLS keying material
+exporters, its output cannot be transparently forwarded by HTTP intermediaries.
+HTTP intermediaries that support this specification have two options:
 
 * The intermediary can validate the authentication received from the client,
 then inform the upstream HTTP server of the presence of valid authentication.
```
martinthomson, line 205: I just noticed below that both of these options depend on the intermediary being trusted to represent the status or value faithfully.  That makes the second option far less appealing, unless the intermediary doesn't want to perform the validation (but an intermediary is probably in a better position to cache the per-connection status so that extra work can be avoided, so...).

draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -270,87 +213,64 @@ upstream HTTP server is out of scope for this document.
...
+of another. This constraint is met by modern Web browsers. If an attacker were
+to compromise the browser such that it could access another context's memory,
+the attacker might also be able to access the corresponding key, so binding
+authentication to requests would not provide much benefit in practice.
+
+Key material used for the Signature HTTP authentication scheme MUST NOT be
+reused in other protocols. Doing so can undermine the security guarantees of
+the authentication.
+
+Origins offering this scheme are able to link requests that use the same key.
```
martinthomson, line 240: But these use the same connection anyway, so is this really a problem?  I guess if your goal is to break linkability, but only OHTTP does that (and this makes no real sense to do for the outer request).

draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -270,87 +213,64 @@ upstream HTTP server is out of scope for this document.
...
-symmetric or asymmetric MUST NOT be reused in other protocols. Doing so can
-undermine the security guarantees of the authentication.
-
-Origins offering Unprompted Authentication are able to link requests that use
-the same key for the Authentication Schemes provided. However, requests are
-not linkable across origins if the keys used are specific to the individual
-origins using Unprompted Authentication.
+The authentication proofs described in this document are not bound to
+individual HTTP requests; if the key is used for authentication proofs on
+multiple requests they will all be identical. This allows for better
```
martinthomson, line 226:
```suggestion
multiple requests on the same connection they will all be identical. This allows for better
```

draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -270,87 +213,64 @@ upstream HTTP server is out of scope for this document.
...
-Key material used for authentication in unprompted authentication, whether
-symmetric or asymmetric MUST NOT be reused in other protocols. Doing so can
-undermine the security guarantees of the authentication.
-
-Origins offering Unprompted Authentication are able to link requests that use
-the same key for the Authentication Schemes provided. However, requests are
-not linkable across origins if the keys used are specific to the individual
-origins using Unprompted Authentication.
+The authentication proofs described in this document are not bound to
+individual HTTP requests; if the key is used for authentication proofs on
```
martinthomson, line 225:
```suggestion
individual HTTP requests; if the key used for authentication proofs on
```

draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -270,87 +213,64 @@ upstream HTTP server is out of scope for this document.
...
+multiple requests they will all be identical. This allows for better
+compression when sending over the wire, but implies that client implementations
+that multiplex different security contexts over a single HTTP connection need
+to ensure that those contexts cannot read each other's header fields.
+Otherwise, one context would be able to replay the Authorization header field
+of another. This constraint is met by modern Web browsers. If an attacker were
```
martinthomson, line 231: Not clear what "this constraint" applies to at this point.  Maybe instead "Modern web browsers restrict visibility of the Authorization field, see [FETCH]."  (A citation here isn't cheap - you'd need to ask WHATWG to make the anchor stable.  They are good about that, but it's annoying to have to.)

draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -270,87 +213,64 @@ upstream HTTP server is out of scope for this document.
...
+to compromise the browser such that it could access another context's memory,
+the attacker might also be able to access the corresponding key, so binding
+authentication to requests would not provide much benefit in practice.
```
martinthomson, line 234: I'd drop this last sentence.  It's a good explanation, but almost too web-specific.

martinthomson (approved):

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: Yeah that was somewhat lazy. I plan on fixing that once the draft gets a bit more stable

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: Agree with you, but we use the term nonce to refer to this in other parts of the draft. I'm happy to replace nonce with something else though - thoughts on something less wordy than "the data that is signed"?

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: Agreed, done.

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: I think I agree with the idea but I'm not sure I see what you mean. Do you have thoughts on how best to make said point?

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: Yeah, the number of dragons therein is what prompted me to punt this out of scope. To be fair since the intermediary has the TLS cert private key, this isn't the worst thing it can do.

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: Done

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: I think we want the "is" here

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: Done

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: It's more about site isolation, JS for a given origin can't see the headers for another origin even if they're on the same connection. Thoughts on a best non-painful reference for that?

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: Isn't the whole paragraph web-specific?

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: Not if you're using h1 though. I guess this is just a "sticking eye in fork hurts" sign, but it could be removed

draft-ietf-httpbis-unprompted-auth.md
jhoyla: The 'correct' term is channel binding, but some people inexplicably think that phrase is too scary.

chris-wood (approved):

draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -270,87 +214,65 @@ upstream HTTP server is out of scope for this document.
...
+attacker were to compromise the browser such that it could access another
+context's memory, the attacker might also be able to access the corresponding
+key, so binding authentication to requests would not provide much benefit in
+practice.
+
+Key material used for the Signature HTTP authentication scheme MUST NOT be
+reused in other protocols. Doing so can undermine the security guarantees of
+the authentication.
+
+Origins offering this scheme are able to link requests that use the same key.
```
chris-wood, line 242:
```suggestion
Origins offering this scheme can link requests that use the same key.
```

draft-ietf-httpbis-unprompted-auth.md
martinthomson: "These parameters use SF [RFC8941 (or -bis)] in their definition, though the Authorization field does not use SF."

draft-ietf-httpbis-unprompted-auth.md
martinthomson: Maybe open an issue and make that assumption explicit.

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: Done

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: Agreed, done via #2576

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: Filed #2577 to keep track of this

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: Filed #2578 to track this

PR #2573, opened 2023-06-23
Avoid issues with signatures that cover any nonce

### LINKED ISSUE #2490: Unprompted Auth: Freshness & Signature Binding
In the current draft, the header is composed as: 
```
{key_id}, {sig_alg}, {hash_alg}, {signature_over_nonce}
```
The server computes the nonce from the TLS key material exporter and verifies the signature. 

This doesn't actually guarantee freshness, because its possible for parties to construct signatures which are valid for any message/nonce. This means that a party which does not have access to the TLS Export material can still inject a valid Auth header. I don't think that's a major problem, but it is a bit unintuitive and its hard to predict how drafts like this will eventually get deployed. 

A simple fix would be in to add an additional parameter, say `v` which is `hash(seperator || nonce)` or similar and have servers check its been correctly computed. This ensures that only parties that can access TLS key material can inject valid headers. 

Behind the scenes, this attack / fix works because hash functions have to be collision resistant for all inputs, whereas signatures don't have to satisfy **any** properties if the associated public key was generated maliciously. [Seems Legit](https://dennis-jackson.uk/assets/pdfs/signatures.pdf) has a longer discussion.

DavidSchinazi:
Fixes #2490

### DIFF
--- a/draft-ietf-httpbis-unprompted-auth.md
+++ b/draft-ietf-httpbis-unprompted-auth.md
@@ -233,7 +233,10 @@ number of bytes necessary.
 
 ## Key Exporter Output {#output}
 
-The TLS key exporter output is described in {{fig-output}}:
+The key exporter output is 48 bytes long. Of those, the first 32 bytes are
+input to the signature and the next 16 bytes are sent alongside the signature.
+This allows the recipient to confirm that the exporter produces the right
+values. This is described in {{fig-output}}:
 
 ~~~
   Nonce (256),
--- a/draft-ietf-httpbis-unprompted-auth.md
+++ b/draft-ietf-httpbis-unprompted-auth.md
@@ -52,6 +52,32 @@ informative:
   H3:
     =: RFC9114
     display: HTTP/3
+  SEEMS-LEGIT:
+    title: "Seems Legit: Automated Analysis of Subtle Attacks on Protocols That Use Signatures"
+    author:
+      -
+        initials: D.
+        surname: Jackson
+        name: Dennis Jackson
+      -
+        initials: C.
+        surname: Cremers
+        name: Cas Cremers
+      -
+        initials: K.
+        surname: Cohn-Gordon
+        name: Katriel Cohn-Gordon
+      -
+        initials: R.
+        surname: Sasse
+        name: Ralf Sasse
+    date: 2019
+    refcontent:
+      - "CCS '19: Proceedings of the 2019 ACM SIGSAC Conference on Computer and Communications Security"
+      - "pp. 2165–2180"
+    seriesinfo:
+      DOI: 10.1145/3319535.3339813
+
 
 --- abstract
 
@@ -144,7 +170,7 @@ exporter {{!KEY-EXPORT=RFC5705}} with the following parameters:
 
 * the context is set to the structure described in {{context}}
 
-* the exporter output length is set to 32 bytes (see {{output}})
+* the exporter output length is set to 48 bytes (see {{output}})
 
 ## Key Exporter Context {#context}
 
@@ -205,10 +231,26 @@ of bytes necessary.
 
 ## Key Exporter Output {#output}
 
-The output of the exporter is a 32-byte symmetric key. That symmetric key is
-then used as nonce which can be signed using the client's chosen asymmetric
-private key. The resulting signature is then transmitted to the server using
-the Authorization field.
+The TLS key exporter output is described in {{fig-output}}:
+
+~~~
+  Nonce (256),
+  Verification (128),
+~~~
+{: #fig-output title="Key Exporter Output Format"}
+
+The key exporter context contains the following fields:
+
+Nonce:
+
+: The nonce is to be signed using the client's chosen asymmetric private key.
+The resulting signature is then transmitted to the server using the
+p Parameter (see {{parameter-p}}).
+
+Verification:
+
+: The verification is transmitted to the server using the v Parameter (see
+{{parameter-v}}).
 
 # Authentication Parameters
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -205,10 +231,26 @@ of bytes necessary.
 
 ## Key Exporter Output {#output}
 
-The output of the exporter is a 32-byte symmetric key. That symmetric key is
-then used as nonce which can be signed using the client's chosen asymmetric
-private key. The resulting signature is then transmitted to the server using
-the Authorization field.
+The TLS key exporter output is described in {{fig-output}}:
```
martinthomson, line 234: I think that you want to be clearer about what is going on here.  In prose.  That is, 48 bytes is output, then the first 32 bytes is input to the signature and the next 16 bytes are sent along with the signature (to allow the recipient to confirm that the exporter produces the right values).

@dennisjackson, could you just transmit half the nonce instead of extending the KDF output?  The other half should be hard to guess, but confirming that way would save an extra invocation of HMAC.  (I'm not particularly concerned about the cost; I'm just asking to improve my understanding.)

draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -205,10 +231,26 @@ of bytes necessary.
...
+
+~~~
+  Nonce (256),
+  Verification (128),
+~~~
+{: #fig-output title="Key Exporter Output Format"}
+
+The key exporter context contains the following fields:
+
+Nonce:
```
martinthomson, line 244: I don't like this word, as we've previously discussed.  "Signature Input" would work.

draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -234,12 +276,21 @@ Its value is an integer between 0 and 65535 inclusive from the IANA "TLS
 SignatureScheme" registry maintained at
 <[](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-signaturescheme)>.
 
+## The v Parameter {#parameter-v}
```
chris-wood, line 279: Why not simply send the nonce on the wire? The server could then check that (1) the nonce is as expected and (2) the signature is valid. I realize that would result in more bytes on the wire, but splitting up the nonce/verification just seems odd to me. The `verification` value is like a proxy for the actual message being signed, so I thought to myself, why not just send the actual message that's signed?

chris-wood (changes requested):

draft-ietf-httpbis-unprompted-auth.md
chris-wood: Why not just send the entire nonce?

DavidSchinazi: If we send the entire nonce, then a lazy implementer could use that as the value for signature checking instead of properly computing the key exporter. The current design (or MT's proposal to only send part of the nonce) avoid that risk.

My gut feeling is that using another part of the key exporter would make security analysis easier than using part of the nonce, but I'll let @dennisjackson comment on that

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: Signature Input sounds good to me. I'll address this as an editorial change after I land the current train of PRs to avoid having to deal with 17 merge conflicts

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: See other comment thread

draft-ietf-httpbis-unprompted-auth.md
chris-wood: I'm not sure I buy the lazy implementer argument, but I also think sending half the nonce would be sufficient here.

DavidSchinazi: I agree that sending half the nonce would work, but what's the benefit? We need to transmit some sort of verification bit regardless, so the only benefit I see to reusing the start of the nonce instead of getting some more key exporter output is that it saves an HKDP operation. I'd rather take the minimal CPU hit to avoid bit reuse if there's no other benefit

chris-wood: It just seems like the most obvious thing to do: send the message that is being signed.

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: Filed #2578 to track this

draft-ietf-httpbis-unprompted-auth.md
dennisjackson: Sorry for the slow response. 

Short answer: @chris-wood is right that from a formal analysis perspective sending the entire message or a hash of the entire message would also be fine and this is the generic fix to this kind of signature attack. @martinthomson's suggestion of sending half the signature input also works in this specific case. However, for pragmatic reasons I think @DavidSchinazi's solution is nicest overall. 

Longer answer: 

From the perspective of a formal analysis, using the full or half signature input is fine. The positive property we're looking for is that the sender is committed to a single message input, so the full message obviously achieves that. The half input is only going to match with prob 2^128 so also fine. Thinking about negative consequences, we're going from transmitting the signature of the authenticator to the authenticator itself. In a formal analysis, this isn't ever going to be a problem because signature schemes are assumed to be unforgeable (EUF-CMA) but nothing more, so an attacker in 'proof world' can do things like take a signature and reverse it to learn the message (even though that's typically hard in the real world).   

Pragmatically though, I like @DavidSchinazi's solution best. It lessens the risk of a lazy implementer using the sender's value by mistake in the signature verification or trying to save a `memcmp` by skipping the second check of the signature input. 

Another *very* minor benefit that in some signature schemes like Ed25519, the attacker can't reverse it in practice as they only learn `h(pk || m)` and in this case `m` is high entropy. So in the case the implementer uses Ed25519, by using the extra KDF output you make it intractable for an attacker who has witnessed an unprompted auth header to extract the authenticator value and make their own valid header with their key.

PR #2574, opened 2023-06-24
Contextualize the signature itself

### LINKED ISSUE #2439: Consider contextualizing signatures
From Ilari Liusvaara on the [list](https://lists.w3.org/Archives/Public/ietf-http-wg/2022OctDec/0027.html):

> The signatures do not appear to be contextualized in any way, which is questionable. For example, one could use the same contextualization mechanism that TLS 1.3 uses (which prepends 64 spaces, a context label and NUL [one zero octet]).

DavidSchinazi:
Fixes #2439

### DIFF
--- a/draft-ietf-httpbis-unprompted-auth.md
+++ b/draft-ietf-httpbis-unprompted-auth.md
@@ -261,7 +261,7 @@ issues caused by key reuse. The signature is computed over the concatenation of:
 
 * The context string "HTTP Signature Authentication"
 
-* A single 0 byte which serves as the separator
+* A single 0 byte which serves as a separator
 
 * The nonce extracted from the key exporter output (see {{output}})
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -243,15 +243,45 @@ The key exporter context contains the following fields:
...
+2020202020202020202020202020202020202020202020202020202020202020
+48545450205369676E61747572652041757468656E7469636174696F6E
+00
+0101010101010101010101010101010101010101010101010101010101010101
+~~~
+
+This constructions mirrors that of the TLS 1.3 CertificateVerify message
+defined in {{Section 4.4.3 of TLS}}.
+
+The resulting signature is then transmitted to the server using the p Parameter
```
martinthomson, line 282:
```suggestion
The resulting signature is then transmitted to the server using the `p` Parameter
```

martinthomson (approved):

PR #2594, opened 2023-07-11
All percent encoding

### LINKED ISSUE #2575: DisplayString pick-your-own-escaping is an invitation for smuggling attacks
As currently defined, the two serialized DisplayStrings

    %"foo\"bla%22bar"

and

    %"foo%22bla\"bar"

are semantically identical.

We should either use %-escapes or \\-escapes, mixing them this way is asking for smuggling-attacks.

mnot:
Fixes #2575.

### DIFF
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -689,16 +689,17 @@ Given a Date as input_date, return an ASCII string suitable for use in an HTTP f
 
 Given a string of Unicode characters as input_string, return an ASCII string suitable for use in an HTTP field value.
 
-0. Let byte_array be the result of applying UTF-8 encoding {{UTF8}} to input_string. If there is an error in doing so, fail parsing.
-1. Let encoded_string be an empty string.
-2. For each byte in byte_array:
+0. If input_string is not a string of Unicode characters, fail parsing.
+1. Let byte_array be the result of applying UTF-8 encoding {{UTF8}} to input_string.
+2. Let encoded_string be an empty string.
+3. For each byte in byte_array:
    1. If byte is %x25 ("%"), append "%25" to encoded_string.
    2. If byte is in the ranges %x00-1f or %x7f-ff, apply the percent-encoding defined in {{Section 2.1 of URI}} to byte and append the result to encoded_string.
    3. Otherwise, decode byte as an ASCII character and append the result to encoded_string.
-3. Let output be a string containing %x25 ("%") followed by DQUOTE.
-4. Append encoded_string to output.
-5. Append DQUOTE to output.
-6. Return output.
+4. Let output be a string containing %x25 ("%") followed by DQUOTE.
+5. Append encoded_string to output.
+6. Append DQUOTE to output.
+7. Return output.
 
 
 ## Parsing Structured Fields {#text-parse}
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -696,7 +696,7 @@ Given a string of Unicode characters as input_string, return an ASCII string sui
    1. If byte is %x25 ("%"), append "%25" to encoded_string.
    2. If byte is in the ranges %x00-1f or %x7f-ff, apply the percent-encoding defined in {{Section 2.1 of URI}} to byte and append the result to encoded_string.
    3. Otherwise, decode byte as an ASCII character and append the result to encoded_string.
-4. Let output be a string containing %x25 ("%") followed by DQUOTE.
+4. Let output be a string containing "%" followed by DQUOTE.
 5. Append encoded_string to output.
 6. Append DQUOTE to output.
 7. Return output.
@@ -969,20 +969,20 @@ Given an ASCII string as input_string, return a Date. input_string is modified t
 
 Given an ASCII string as input_string, return a string of Unicode characters. input_string is modified to remove the parsed value.
 
-0. If the first two characters of input_string are not  %x25 ("%") followed by DQUOTE, fail parsing.
+0. If the first two characters of input_string are not  "%" followed by DQUOTE, fail parsing.
 1. Discard the first two characters of input_string.
 2. Let byte_array be an empty byte array.
 3. While input_string is not empty:
    1. Let char be the result of consuming the first character of input_string.
    2. If char is in the range %x00-1f or %x7f-ff (i.e., it is not in VCHAR or SP), fail parsing.
-   3. If char is %x25 ("%"):
+   3. If char is "%":
       1. Let octet_hex be the two bytes after char in input_string. If there are not two bytes, fail parsing.
       2. Let octet be the result of decoding octet_hex as hexidecimal, in a case-insensitive fashion.
       3. Append octet to byte_array.
    2. If char is DQUOTE:
       1. Let unicode_string be the result of decoding byte_array as a UTF-8 string {{UTF8}}. Fail parsing if decoding fails.
       2. Return unicode_string.
-   3. Otherwise, if char is not %x25 or DQUOTE:
+   3. Otherwise, if char is not "%" or DQUOTE:
       1. Let byte be the result of applying ASCII encoding to char.
       2. Append byte to byte_array.
 4. Reached the end of input_string without finding a closing DQUOTE; fail parsing.
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -61,7 +61,6 @@ informative:
   RFC9113:
     display: HTTP/2
   HPACK: RFC7541
-  ENCODING: RFC4648
 
 venue:
   group: HTTP
@@ -695,7 +694,7 @@ Given a string of Unicode characters as input_string, return an ASCII string sui
 3. For each byte in byte_array:
    1. If byte is %x25 ("%"), %x22 (DQUOTE), or in the ranges %x00-1f or %x7f-ff:
       1. Append "%" to encoded_string.
-      2. Let encoded_byte be the result of applying base16 encoding ({{Section 8 of ENCODING}}) to byte.
+      2. Let encoded_byte be the result of applying base16 encoding ({{Section 8 of !RFC4648}}) to byte.
       3. Append encoded_byte to encoded_string.
    2. Otherwise, decode byte as an ASCII character and append the result to encoded_string.
 4. Let output be a string containing "%" followed by DQUOTE.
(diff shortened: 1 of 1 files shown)


reschke (commented):
> 0. Let byte_array be the result of applying UTF-8 encoding {{UTF8}} to input_string. If there is an error in doing so, fail parsing.

How can that fail?

draft-ietf-httpbis-sfbis.md
```diff
@@ -695,8 +695,10 @@ Given a string of Unicode characters as input_string, return an ASCII string sui
    1. If byte is %x25 ("%"), append "%25" to encoded_string.
    2. If byte is in the ranges %x00-1f or %x7f-ff, apply the percent-encoding defined in {{Section 2.1 of URI}} to byte and append the result to encoded_string.
    3. Otherwise, decode byte as an ASCII character and append the result to encoded_string.
-3. Let formatted_string be the result of running Serialising a String ({{ser-string}}) with encoded_string.
-4. Return the character "%" followed by formatted_string.
+3. Let output be a string containing %x25 ("%") followed by DQUOTE.
```
reschke, line 698: If we use the definition from the URI spe, shouldn't we mandate either upper or lowercase?

draft-ietf-httpbis-sfbis.md
```diff
@@ -966,16 +968,23 @@ Given an ASCII string as input_string, return a Date. input_string is modified t
...
-6. Return unicode_string.
+0. If the first two characters of input_string are not  %x25 ("%") followed by DQUOTE, fail parsing.
+1. Discard the first two characters of input_string.
+2. Let byte_array be an empty byte array.
+3. While input_string is not empty:
+   1. Let char be the result of consuming the first character of input_string.
+   2. If char is in the range %x00-1f or %x7f-ff (i.e., it is not in VCHAR or SP), fail parsing.
+   3. If char is %x25 ("%"):
+      1. Let octet_hex be the two bytes after char in input_string. If there are not two bytes, fail parsing.
+      2. Let octet be the result of decoding octet_hex as hexidecimal, in a case-insensitive fashion.
```
reschke, line 979: that needs a failure condition, no?

draft-ietf-httpbis-sfbis.md
```diff
@@ -966,16 +968,23 @@ Given an ASCII string as input_string, return a Date. input_string is modified t
...
-1. Discard the first character of input_string.
-2. Let parsed_string be the result of running Parsing a String ({{parse-string}}) with input_string.
-3. Let byte_array be the result of applying ASCII encoding to input_string.
-4. For each sigil_byte in byte_array which is %25 ("%"):
-   1. Let octet_hex be the two bytes after sigil_byte in byte_string. If there are not two bytes, fail parsing.
-   2. Let octet be the result of decoding octet_hex as hexidecimal, in a case-insensitive fashion.
-   3. Replace sigil_byte and octet_hex in byte_array with octet.
-5. Let unicode_string be the result of decoding byte_array as a UTF-8 string {{UTF8}}. Fail parsing if decoding fails.
-6. Return unicode_string.
+0. If the first two characters of input_string are not  %x25 ("%") followed by DQUOTE, fail parsing.
```
reschke, line 971: Maybe define "PERCENT" upfront and use it throughout.

draft-ietf-httpbis-sfbis.md
```diff
@@ -966,16 +968,23 @@ Given an ASCII string as input_string, return a Date. input_string is modified t
...
+   2. If char is in the range %x00-1f or %x7f-ff (i.e., it is not in VCHAR or SP), fail parsing.
+   3. If char is %x25 ("%"):
+      1. Let octet_hex be the two bytes after char in input_string. If there are not two bytes, fail parsing.
+      2. Let octet be the result of decoding octet_hex as hexidecimal, in a case-insensitive fashion.
+      3. Append octet to byte_array.
+   2. If char is DQUOTE:
+      1. Let unicode_string be the result of decoding byte_array as a UTF-8 string {{UTF8}}. Fail parsing if decoding fails.
+      2. Return unicode_string.
+   3. Otherwise, if char is not %x25 or DQUOTE:
+      1. Let byte be the result of applying ASCII encoding to char.
```
reschke, line 985:
```suggestion
      1. Let byte be the result of applying ASCII decoding to char.
```

reschke:
Looks good except for minor details and potentially a bug (encoding vs decoding).

I'd also like to see the spec enforce a consistent form of percent encoding (lower vs upper)

draft-ietf-httpbis-sfbis.md
mnot: I've changed to just using "%", to align with other usage.

draft-ietf-httpbis-sfbis.md
mnot: Added.

draft-ietf-httpbis-sfbis.md
mnot: Converting a character to a byte is encoding it, not decoding it.

draft-ietf-httpbis-sfbis.md
mnot: Should parsing fail if that case isn't seen?

draft-ietf-httpbis-sfbis.md
reschke: True. The sentence still reads weird and might be hard to parse for many (I know that "ASCII encoding" of a char actually is a type cast, but how many readers do?)

draft-ietf-httpbis-sfbis.md
reschke: which case?

mnot: whichever we specify...

reschke: Maybe talk about different things? Parsing percent-decoded as per URI spec can fail, and we need to say something about that, no?

mnot: I thought you were asking if we could mandate a particular casing here (in the serialisation algorithm); I was asking if that casing should be enforced by the parsing algorithm.

reschke: Ah. Yes, it should.

mnot: I'm a bit concerned that some producers might not be able to control the casing of their output. While they could run it through an additional step of processing, that would increase overhead for them.

Do we expect anyone to be doing something (eg security scanning) that would be expecting a particular casing, and wouldn't be able to adapt to two possible casings? My initial reaction is that I'd rather not increasing efficiency for those parties by decreasing efficiency for others...

reschke: It's mostly a matter of consistency with the remainder of the spec. For values, there's exactly one value to serialize them (or maybe that's not the case because of Decimal???).

mnot: We don't require sytems to error when parsing eg `0001`. Byte sequences don't require parsers to fail on padding irregularities. I appreciate that we specify error handling very carefully in this document, but we do so when there's a point to it, not merely for consistency.

draft-ietf-httpbis-sfbis.md
```diff
@@ -689,14 +689,18 @@ Given a Date as input_date, return an ASCII string suitable for use in an HTTP f
...
-0. Let byte_array be the result of applying UTF-8 encoding {{UTF8}} to input_string. If there is an error in doing so, fail parsing.
-1. Let encoded_string be an empty string.
-2. For each byte in byte_array:
+0. If input_string is not a string of Unicode characters, fail parsing.
+1. Let byte_array be the result of applying UTF-8 encoding {{UTF8}} to input_string.
+2. Let encoded_string be an empty string.
+3. For each byte in byte_array:
    1. If byte is %x25 ("%"), append "%25" to encoded_string.
+   2. If byte is %x22 (DQUOTE), append "%22" to encoded_string.
    2. If byte is in the ranges %x00-1f or %x7f-ff, apply the percent-encoding defined in {{Section 2.1 of URI}} to byte and append the result to encoded_string.
```
martinthomson, line 698: Aside from the counting strangeness (2x2), perhaps it would be better if you spelled this out rather than doing it by reference.  That is, append % (0x25).  Encode the byte value as a two character, uppercase hexadecimal string {{cite Section 8 of RFC 4648 if you like}} and append the UTF-8 encoding of those characters.

(Ulterior motive: I would much prefer that we avoid relying on RFC 3986 for this.)

draft-ietf-httpbis-sfbis.md
```diff
@@ -966,16 +970,23 @@ Given an ASCII string as input_string, return a Date. input_string is modified t
...
-6. Return unicode_string.
+0. If the first two characters of input_string are not  "%" followed by DQUOTE, fail parsing.
+1. Discard the first two characters of input_string.
+2. Let byte_array be an empty byte array.
+3. While input_string is not empty:
+   1. Let char be the result of consuming the first character of input_string.
+   2. If char is in the range %x00-1f or %x7f-ff (i.e., it is not in VCHAR or SP), fail parsing.
+   3. If char is "%":
+      1. Let octet_hex be the two bytes after char in input_string. If there are not two bytes, fail parsing.
+      2. Let octet be the result of decoding octet_hex as hexidecimal, in a case-insensitive fashion. If decoding fails, fail parsing.
```
martinthomson, line 981: I don't think that we need this to be case insensitive.

martinthomson, line 981:
```suggestion
      2. Let octet be the result of decoding octet_hex as hexadecimal, in a case-insensitive fashion. If decoding fails, fail parsing.
```

martinthomson (approved):
WFM, though I think I prefer to go all uppercase (because then we can use RFC 4648 instead) rather than be case-variation-tolerant.

draft-ietf-httpbis-sfbis.md
mnot: If by "2x2" you mean the fact that there are two "2." items, markdown will sort it out.

I _think_ what you're suggesting is roughly:

n. If byte is in the range %x00-1f, %x22, %x25, or in the range %x7f-ff:
    1. Append "%" to encoded_string.
    2. Append the percent encoding to byte...

Is that roughly what you have in mind?

Re: 4648 vs 3986 -- I strongly suspect that people will be reaching for their URI encoding libraries for this, not their Base16 implementations. That's one of the reasons I'm a bit nervous about specifying case -- it's not sensitive in URLs.

reschke:
I am sceptical wrt "use the URI percent encoder". The reason being:

- percent encoding rules depend on what part of the URI needs encoding (and many APIs do not get that)
- calling the API for a single character multiple times usually requires converting a char to a string, and that might not be good for perf if we really care about that

mnot:
I've removed the reliance on URI for percent encoding; was clean as @martinthomson said it would be.

Regarding case -- @martinthomson RFC4648 Section 8 is explicitly case-insensitive. Given that and the discussion in the group yesterday, I've left this case-insensitive. Say if you still disagree.

draft-ietf-httpbis-sfbis.md
```diff
@@ -61,7 +61,7 @@ informative:
   RFC9113:
     display: HTTP/2
   HPACK: RFC7541
-  URI: RFC3986
+  ENCODING: RFC4648
```
reschke, line 64: Using "ENCODING" is a bit misleading. If we don't have a catchy name for the spec, I'd suggest just saying RFC4648.

reschke (commented):
I don't believe that relying on a different RFC for the hex encoding makes this much better (so I sort of disagree with Martin's proposal).

Just define the escaping exactly inline, and then we can choose upper/lower.

martinthomson:
@reschke , I was suggesting an inline definition. The fact that uppercase  is base16 would just be an observation (and not a particularly useful one). Lowercase tends to be the default mode in several places.

mnot:
@martinthomson it wasn't at all clear that's what you were suggesting. I'll make an attempt.

mnot:
Looking at it, I think I prefer referring to 4648. This this addresses the original issue, I'll merge; if someone wants to propose an alternate way to specify this, please feel free to open a PR.

Issue #2544, opened 2023-05-17
Server should reject upload resumptions with inconsistent Content-Length


guoye-zhang:
When the server receives a series of appends:

```
PATCH /upload
Upload-Offset: 10
Content-Length: 90

<interrupted after 10 bytes>
```

```
PATCH /upload
Upload-Offset: 20
Content-Length: 30

409 Conflict
<10+90 != 20+30>
```

If the offset+length is inconsistent, the server should reject the upload.

smatsson:
To clarify; Is what you are saying here that the `Content-Length` provided during the upload creation procedure should act as the total size of the file, similar to Upload-Length in tus1x?

Acconut:
I am not sure if we want this. I ran into this exact problem because your NIOResumableUpload implementation uses this logic (see https://lists.w3.org/Archives/Public/ietf-http-wg/2023JulSep/0024.html).

We currently have the `Upload-Incomplete` header to indicate the completeness of the upload. Based on this information, the server can determine if an upload is complete or not without relying on `Content-Length`. Adding `Content-Length` to the upload logic would mean that we have two indicators for upload completeness, which could potentially disagree with each other.

Personally, I think we should not use `Content-Length` to determine the upload's completeness and only determine it based on whether the client set `Upload-Incomplete: ?0` in a successful request. For each Upload Creation Procedure or Upload Appending Procedure, the `Content-Length` should be set to the number of bytes that the client wants to send in this individual procedure.

For example, if you want to transmit 100 bytes in the first POST request, set `Content-Length: 100` (or not if you use `Transfer-Encoding: chunked`). If you want to transmit 200 bytes in the following PATCH request, set `Content-Length: 200`. If you want to finish the upload then, set `Upload-Incomplete: ?0` and the upload is considered complete with its 300 bytes of content.

guoye-zhang:
The issue in NIOResumableUpload was a bug, not how this is supposed to work. Offset + Content-Length = final upload size when the upload is complete, and if the client has told the server about the final size, it shouldn't be allowed to change its mind.

jrflat:
> Personally, I think we should not use Content-Length to determine the upload's completeness and only determine it based on whether the client set Upload-Incomplete: ?0 in a successful request. For each Upload Creation Procedure or Upload Appending Procedure, the Content-Length should be set to the number of bytes that the client wants to send in this individual procedure.

[...]

Issue #2500, opened 2023-04-05
Reuse of the Upload-Complete header field on requests and responses


guoye-zhang:
@martinthomson brought up that during the WG meeting that reusing Upload-Complete on requests and responses can be confusing. They don’t mean exactly the same thing: `Upload-Complete: ?1` on the server means it has received the full upload, while `Upload-Complete: ?1` on the client means the body it will send is a complete upload.

guoye-zhang:
We discussed this internally and prepared a response:

Upload-Complete is used on both requests and responses because it is symmetrical. Upon a typical successful transaction, an `Upload-Complete: ?0` request results in an `Upload-Complete: ?0` response, and an `Upload-Complete: ?1` request results in an `Upload-Complete: ?1` response. A server implementation can simply echo the field.

However, they are also meanings that we can take advantage of. The presence on the request means it’s a resumable request, and the presence on the response is also a useful signal. Since we are upgrading regular uploads, there are cases where 4xx/5xx responses are intentional and we can't just eat it by attempting to resume. We would be much more confident that the response came from the real server rather than a middlebox which shouldn’t randomly add `Upload-Complete: ?1`.

The server can also use `Upload-Complete: ?1` to terminate a request early if it does not care about the body, similar to the early response mechanism in HTTP/3.

Acconut:
I am having second thoughts about this as well. While debugging a server implementation, the naming of the headers also confused me for a moment (even though I wrote parts of the draft). The reason is that `Upload-Complete` in a response (or `Upload-Incomplete` how it is currently called) refers to the current state of the upload, while `Upload-Complete` in a request refers to the potential state in the future, if the request completed successfully.

I don't think that we have a general issue with the concept of `Upload-Complete` and have to change much. But maybe a simple name change can help here already to alleviate the confusion.

We could use `Upload-Complete` for the response header to indicate that it represents the current state. And `Upload-Is-Completing` for request headers to indicate that process the request successfully will complete the upload.

We loose the symmetry of the names in request and response, but maybe this makes their relation even clearer.

jrflat:
I think `Upload-Complete` for both client and server works because it represents their own perspectives of the upload.

For the client, sending `Upload-Complete: ?1` means "the upload is complete on my end, I've included all the data to complete it." The client isn't making assumptions about whether the server has received all the data; it's only expressing to the server that from its perspective, the upload data supplied is now whole.

[...]

PR #2595, opened 2023-07-19
Clarify handling of Content-Length and Upload-Complete

Acconut:
Closes https://github.com/httpwg/http-extensions/issues/2544

I am not super sure about the wording here, but this is a start. The sentences seem quite long and maybe the rules could also be expressed in fewer words.

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -214,7 +214,7 @@ If the request completes successfully and the entire upload is complete, the ser
 
 If the request completes successfully but the entire upload is not yet complete indicated by the `Upload-Incomplete` header, the server MUST acknowledge it by responding with the `201 (Created)` status code, the `Upload-Incomplete` header set to true.
 
-If the request includes the `Upload-Incomplete: ?0` header field and a valid `Content-Length` header field, the client attempts to upload a fixed-length resource in one request. In this case, the upload's final size is the value of the `Content-Length` header field and the server MUST record the upload's final size to ensure its consistency with future Upload Appending Procedures.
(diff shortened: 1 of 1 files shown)


jrflat:
If the client acknowledged its final upload size at any point (`Upload-Complete: ?1` + `Content-Length`), it should still know it in the future, so I don't think we need a way for the server to echo it in the offset retrieving response. If the server disagrees with the client, then someone's data representation has changed, and we shouldn't continue the upload anyway.

guoye-zhang:
Looks good. Wondering if these could be reduced to SHOULDs though. It wouldn't really affect interop if the server didn't implement this.

Acconut:
> If the client acknowledged its final upload size at any point (`Upload-Complete: ?1` + `Content-Length`), it should still know it in the future, so I don't think we need a way for the server to echo it in the offset retrieving response. If the server disagrees with the client, then someone's data representation has changed, and we shouldn't continue the upload anyway.

Fair point, let's leave such functionality out of the draft.

> Wondering if these could be reduced to SHOULDs though. It wouldn't really affect interop if the server didn't implement this.

While it is true that clients usually won't really notice if the server does or does not support this, I think we can be stricter here and require this additional safety catch here. It will help to create a more standard landscape of resumable upload servers. Unless there is something against requiring this, we can keep it a MUST IMO.

Acconut:
I just updated this PR to use the Upload-Complete header as agreed on in #2500

guoye-zhang:
The arguments for SHOULD are:

1. This check requires additional code and complexity. MUSTs are situations that a minimal implementation has to handle anyway and we are just strict about how to handle them.
2. For people using chunked uploads, this check is only useful if the last chunk is interrupted which is sort of an edge case. Maybe we can generalize the check if we adopt ranged PATCH.

I don't have a strong opinion on this, so either way is OK with me.

guoye-zhang (approved):

PR #2598, opened 2023-07-21
AUTH48 for Client-Cert / RFC 9440

MikeBishop:
This PR contains the changes which occurred during AUTH48.


b---c (approved):

b---c:
Did you manually apply the changes from AUTH48 to the markdown?

MikeBishop:
Yes, with a few rounds of `diff`ing the output and correcting any divergences. There are a few XML things that are harder to reconcile (it doesn't like a bulleted list of contacts, for example), but I think all the text changes are there.

bc-pi (approved):

PR #2603, opened 2023-07-24
Unprompted: add public key to context and header

### LINKED ISSUE #2599: Add public key to context
Based on discussions with @jhoyla, @chris-wood, and others, there are potential key confusion attacks when the server is checking the signature using the wrong public key. I think the simplest solution is to add the public key to the TLS key exporter. That should solve the issue without increasing the amount of bytes sent on the wire

DavidSchinazi:
Fixes #2599

### DIFF
--- a/draft-ietf-httpbis-unprompted-auth.md
+++ b/draft-ietf-httpbis-unprompted-auth.md
@@ -47,6 +47,14 @@ author:
 
 normative:
   FOLDING: RFC8792
+  X.690:
+    title: "Information technology - ASN.1 encoding Rules: Specification of Basic Encoding Rules (BER), Canonical Encoding Rules (CER) and Distinguished Encoding Rules (DER)"
+    date: 2021-02
+    author:
+      org: ITU-T
+    seriesinfo:
+      ISO/IEC 8824-1:2021
+
 informative:
   H2:
     =: RFC9113
@@ -206,7 +214,7 @@ Key ID:
 Public Key:
 
 : The public key used by the server to validate the signature provided by the
-client.
+client (the encoding is described below).
 
 Scheme:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -201,6 +203,11 @@ Key ID:
 
 : The key ID sent in the `k` Parameter (see {{parameter-k}}).
 
+Public Key:
+
+: The public key used by the server to validate the signature provided by the
+client.
```
chris-wood, line 209: What is the encoding of this public key? We probably need to be clear about that in order for this to work.

chris-wood (changes requested):
Looks good! The only thing I suggest we include are the specifics of how the public key is encoded.

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: Good point. Fixed in 9dc88eef6543e47dd32c84427a1df73424e4edff

draft-ietf-httpbis-unprompted-auth.md
```diff
@@ -304,6 +340,13 @@ The REQUIRED "k" (key ID) parameter is a byte sequence that identifies which key
 the user agent wishes to use to authenticate. This can for example be used to
 point to an entry into a server-side database of known keys.
 
+## The a Parameter {#parameter-a}
+
+The REQUIRED "a" (public key) parameter is a byte sequence that contains the
+public key used by the server to validate the signature provided by the client.
+This avoids key confusion issues (see {{SEEMS-LEGIT}}). The encoding of the
+public key is described in {{context}}.
```
chris-wood, line 348: Should we add text which describes what to do if the encoding of this parameter is invalid, or is that already covered elsewhere?

chris-wood (approved):

draft-ietf-httpbis-unprompted-auth.md
DavidSchinazi: I think that's covered by the `mismatch between key ID and provided public key` text (you don't need to check for invalid encoding, just `memcmp` with your known valid encoding), but this is making me want to make that normative. Will do in followup PR.

chris-wood: Sounds good!

PR #2607, opened 2023-07-31
Allow original name in alias list

### LINKED ISSUE #2600: `next-hop-aliases` is missing requested/configured hostname
> The items in the list include all alias names and canonical names received in CNAME records [[DNS](https://httpwg.org/http-extensions/draft-ietf-httpbis-alias-proxy-status.html#DNS)] during the course of resolving the next hop's hostname using DNS, ***not including the original requested hostname itself***.

I understand that the current draft is specifically targeting forward proxies, but as currently described, the `next-hop-aliases` is missing some information when used by reverse proxies, and it's not self-contained anymore (i.e. the requested/configured next hop is not part of the `Proxy-Status` field), making debugging much harder.

> For example, consider a proxy "proxy.example.net" that receives the following records when performing DNS resolution for the next hop "host.example.com":
>
>     host.example.com.           CNAME   tracker.example.com.
>     tracker.example.com.        CNAME   service1.example.com.
>     service1.example.com.       AAAA    2001:db8::1
>
> The proxy could include the following proxy status in its response:
>
>     Proxy-Status: proxy.example.net; next-hop="2001:db8::1";
>         next-hop-aliases="tracker.example.com,service1.example.com"

I find the fact that `host.example.com` is not included here a bit confusing.

While operators or clients connecting directly to `proxy.example.net` might know that it's a forward proxy and could infer that `host.example.com` was the requested hostname, the clients that are a few hops away from it, won't know this, and might be surprised how the `tracker.example.com` and other aliases got in there.

Furthermore, in a similar scenario with a client connecting to the reverse proxy `proxy.example.net` and requesting `www.example.com`, which is configured to use `host.example.com` as the backend, that response is missing the most critical piece of information, and instead of helping, it adds more confusion.

As such, I'd suggest that the requested/configured hostname should be included as the first entry in `next-hop-aliases`, especially when `next-hop` points to the resolved IP address.

cc @mnot @tfpauly

tfpauly:
Allow the original name of the next hop to be the start of the list, and add an example.

Closes #2600

### DIFF
--- a/draft-ietf-httpbis-alias-proxy-status.md
+++ b/draft-ietf-httpbis-alias-proxy-status.md
@@ -70,7 +70,7 @@ by including the list of names in a new `next-hop-aliases` Proxy-Status paramete
 The `next-hop-aliases` parameter's value is a String {{!STRUCTURED-FIELDS=RFC8941}} that contains
 one or more DNS names in a comma-separated list. The items in the list include all alias names and
 canonical names received in CNAME records {{DNS}} during the course of resolving the next hop's
-hostname using DNS, not including the original requested hostname itself. The names SHOULD
+hostname using DNS, and MAY include the original requested hostname itself. The names SHOULD
 appear in the order in which they were received in DNS. If there are multiple CNAME records
 in the chain, the first name in the `next-hop-aliases` list would be the value in the CNAME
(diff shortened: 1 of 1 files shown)


tfpauly:
@PiotrSikora @mnot please review!

mnot:
LGTM

PR #2608, opened 2023-08-02
[CONNECT-TCP] Rephrase guidance related to early data and false start

### LINKED ISSUE #2606: Describe 0-RTT "delay" behavior as a subcategory of "accept"
Within TLS 1.3, the only options for 0-RTT are "accept" or "reject".  "accept but delay until after the handshake" is not visible within the TLS 1.3 protocol.  The current draft language doesn't align nicely with this, and should be adjusted.

Credit: @knekritz

bemasc:
Fixes #2606

### DIFF
--- a/draft-ietf-httpbis-connect-tcp.md
+++ b/draft-ietf-httpbis-connect-tcp.md
@@ -128,9 +128,9 @@ This section discusses some behaviors that are permitted or recommended in order
 
 ## Latency optimizations
 
-When using this specification in HTTP/2 or HTTP/3, clients MAY start sending TCP stream content without waiting for an HTTP response.  Proxies MUST buffer this "false start" content until the TCP stream becomes writable, and discard it if the TCP connection fails.  (This "false start" behavior is not permitted in HTTP/1.1 because it would prevent reuse of the connection after an error response such as 407 (Proxy Authentication Required).)
+When using this specification in HTTP/2 or HTTP/3, clients MAY start sending TCP stream content optimistically, without waiting for an HTTP response, up to the proxy's initial flow-control window size ({{!RFC9113, Section 6.9.2}})({{!RFC9000, Section 4.1}}).  Proxies MUST buffer this "optimistic" content until the TCP stream becomes writable, and discard it if the TCP connection fails.  (This "optimistic" behavior is not permitted in HTTP/1.1 because it would prevent reuse of the connection after an error response such as "407 (Proxy Authentication Required)".)
 
-Servers that host a proxy under this specification MAY offer support for TLS early data in accordance with {{!RFC8470}}.  Clients MAY send "connect-tcp" requests in early data, and MAY include "false start" content in early data (in HTTP/2 and HTTP/3).  Proxies MAY accept, reject, or delay processing of this early data.  For example, a proxy with limited anti-replay defenses might choose to perform DNS resolution of the `target_host` when a request arrives in early data, but delay the TCP connection until the TLS handshake completes.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-connect-tcp.md
```diff
@@ -128,9 +128,9 @@ This section discusses some behaviors that are permitted or recommended in order
...
+Servers that host a proxy under this specification MAY offer support for TLS early data in accordance with {{!RFC8470}}.  Clients MAY send "connect-tcp" requests in early data, and MAY include "optimistic" TCP content in early data (in HTTP/2 and HTTP/3).  At the TLS layer, proxies MAY ignore, reject, or accept the `early_data` extension ({{!RFC8446, Section 4.2.10}}).  At the HTTP layer, proxies MAY process the request immediately, return a "425 (Too Early)" response ({{!RFC8470, Section 5.2}}), or delay some or all processing of the request until the handshake completes.  For example, a proxy with limited anti-replay defenses might choose to perform DNS resolution of the `target_host` when a request arrives in early data, but delay the TCP connection until the TLS handshake completes.
```
knekritz, line 133: nit: ignore and reject are the same thing at the TLS layer

draft-ietf-httpbis-connect-tcp.md
```diff
@@ -128,9 +128,9 @@ This section discusses some behaviors that are permitted or recommended in order
 
 ## Latency optimizations
 
-When using this specification in HTTP/2 or HTTP/3, clients MAY start sending TCP stream content without waiting for an HTTP response.  Proxies MUST buffer this "false start" content until the TCP stream becomes writable, and discard it if the TCP connection fails.  (This "false start" behavior is not permitted in HTTP/1.1 because it would prevent reuse of the connection after an error response such as 407 (Proxy Authentication Required).)
+When using this specification in HTTP/2 or HTTP/3, clients MAY start sending TCP stream content optimistically, without waiting for an HTTP response, up to the proxy's initial flow-control window size ({{!RFC9113, Section 6.9.2}})({{!RFC9000, Section 4.1}}).  Proxies MUST buffer this "optimistic" content until the TCP stream becomes writable, and discard it if the TCP connection fails.  (This "optimistic" behavior is not permitted in HTTP/1.1 because it would prevent reuse of the connection after an error response such as "407 (Proxy Authentication Required)".)
```
knekritz, line 131: Any reason to restrict to the initial flow control window? The client must respect flow control limits, but if the server wants to open it up further before sending the response, I'm not sure why the spec should restrict that?

draft-ietf-httpbis-connect-tcp.md
bemasc: RFC 8446 says ["A server which receives an "early_data" extension MUST behave in one of three ways..."](https://datatracker.ietf.org/doc/html/rfc8446#:~:text=A%20server%20which%20receives%20an%20%22early_data%22%20extension%20MUST%20behave%20in%20one%0A%20%20%20of%20three%20ways%3A), which I've summarized as "ignore, reject, and accept".  "ignore" means "ServerHello without early_data".  "reject" means "HRR without early_data".

It does seem a bit excessive to document all of this here, several layers up the stack, but I wanted to be clear that this specification doesn't constrain the proxy's TLS behavior.  Maybe there's a better way to say that...

draft-ietf-httpbis-connect-tcp.md
bemasc: Thanks, that's a bug in my text.  Fixed!

bemasc: Thanks, that's a bug in my text.  Fixed!

draft-ietf-httpbis-connect-tcp.md
knekritz: FWIW, my interpretation is that both options involve rejecting and ignoring the early data. Sending an HRR additionally means you are rejecting the ClientHello (this is getting into bikeshedding territory though).

knekritz (approved):

draft-ietf-httpbis-connect-tcp.md
bemasc: I agree that it's not perfectly precise, but in the absence of a better phrasing I guess it's good enough.

PR #2621, opened 2023-08-28
Mnot/2620

### LINKED ISSUE #2620: ABNF issues
In the current ABNF, Display Strings are built upon:

~~~ abnf
uchr             = chr / uescaped
~~~

However, `chr` is:

~~~ abnf
chr       = unescaped / escaped
~~~

I think it would be better to remove the `chr` and `uchr` abstractions and just put the alternations directly in `sf-string` and `sf-displaystring`. 

It'd probably also be good to change `escaped` to `backslash-escaped` and `uescaped` (which is super confusing) to `percent-escaped`.

Also, `\` (`%x5C`) _is_ allowed in display strings.

And HEXDIG is not case sensitive.

Yes, I've been using a fuzzer.

mnot:
Fixes #2620.

### DIFF
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -1083,7 +1083,7 @@ Implementations are allowed to limit the size of different structures, subject t
 
 # ABNF {#abnf}
 
-This section uses the Augmented Backus-Naur Form (ABNF) notation {{?RFC5234}} to illustrate expected syntax of Structured Fields.
+This section uses the Augmented Backus-Naur Form (ABNF) notation {{?RFC5234}} to illustrate expected syntax of Structured Fields. However, it cannot be used to validate their syntax, because it does not capture all requirements.
 
 This section is non-normative. If there is disagreement between the parsing algorithms and ABNF, the specified algorithms take precedence.
 
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -1102,35 +1102,30 @@ key           = ( lcalpha / "*" )
 lcalpha       = %x61-7A ; a-z
 param-value   = bare-item
 
-sf-dictionary  = dict-member *( OWS "," OWS dict-member )
-dict-member    = member-key ( parameters / ( "=" member-value ))
-member-key     = key
-member-value   = sf-item / inner-list
+sf-dictionary = dict-member *( OWS "," OWS dict-member )
+dict-member   = member-key ( parameters / ( "=" member-value ))
+member-key    = key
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-sfbis.md
```diff
@@ -1102,37 +1102,31 @@ key           = ( lcalpha / "*" )
...
-sf-boolean = "?" boolean
-boolean    = "0" / "1"
-
-sf-date = "@" sf-integer
-
-sf-displaystring = "%" DQUOTE *uchr DQUOTE
-uchr             = chr / uescaped
-uescaped         = "%" HEXDIG HEXDIG
+pct-encoded  = "%" lc-hexdig lc-hexdig
+lc-hexdig = DIGIT / "a" / "b" / "c" / "d" / "e" / "f"
```
martinthomson, line 1129: IMPORTANT: "a" means "a" or "A".  You really need to use either `%x61-66` or the RFC 7405 extension of `%s"a" / %s"b" ` and so on.

mnot: Thanks for the reminder.

draft-ietf-httpbis-sfbis.md
```diff
@@ -1102,37 +1102,31 @@ key           = ( lcalpha / "*" )
...
+sf-decimal       = ["-"] 1*12DIGIT "." 1*3DIGIT
+sf-string        = DQUOTE *( unescaped / "%" / bs-escaped ) DQUOTE
+sf-token         = ( ALPHA / "*" ) *( tchar / ":" / "/" )
+sf-binary        = ":" *base64 ":"
+sf-boolean       = "?" ( "0" / "1" )
+sf-date          = "@" sf-integer
+sf-displaystring = "%" DQUOTE *( unescaped / "\" / pct-encoded ) DQUOTE
 
-sf-decimal  = ["-"] 1*12DIGIT "." 1*3DIGIT
+base64       = ALPHA / DIGIT / "+" / "/" / "="
```
martinthomson, line 1123: base64 is really:

```abnf
base64 = *(ALPHA / DIGIT / "+" / "/") *"="
```

mnot: Yeah, I was wondering how far to go down that rabbit hole. I think that's a nice balance.

martinthomson: Yup.  `*3"="` is one step too many perhaps.

mnot: In particular, I don't want to imply that we're validating base64 here -- same as the percent encoding.

PR #2642, opened 2023-09-27
Remove some cruft from the Makefile

martinthomson:
This stuff isn't doing anything right now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,4 @@
 LIBDIR := lib
-DISABLE_RIBBON := true
 INDEX_FORMAT := md
 GHPAGES_COMMIT_TTL = 7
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,5 +1,4 @@
 LIBDIR := lib
-DISABLE_RIBBON := true
 INDEX_FORMAT := md
 GHPAGES_COMMIT_TTL = 7
 GHPAGES_BRANCH_TTL = 2
```
martinthomson, line 4: I didn't cut the use of Lucas' mirror for reference XML.

mnot: Should we?

martinthomson: Maybe.  But it is likely that the mirror is still much faster, so I'd want to test it.

Issue #2631, opened 2023-09-19
hash agility


martinthomson:
The draft has a whole thing where you negotiate what hash function to use.  That seems overwrought.  Yes, you need to plan for the possibility that your chosen hash function is no longer considered good.  

For this use case, even a broken SHA-256 would still likely be fine, so I would instead advocate for a different strategy: if SHA-256 cannot be used, we'll build a new set of fields.

Then you can drop the whole negotiation mechanism and simplify things.

PR #2645, opened 2023-10-09
Remove hash negotiation and force sha-256

pmeenan:
This removes the `hashes` negotiation and changes the `Available-Dictionary` request header to explicitly use sha-256 as discussed in #2631. If sha-256 becomes insufficient, the hash negotiation can then be added (or different mechanisms can be used).

### DIFF
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -77,33 +77,56 @@ This document uses the line folding strategies described in [FOLDING].
 ## Use-As-Dictionary
 
 When responding to a HTTP Request, a server can advertise that the response can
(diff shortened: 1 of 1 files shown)


martinthomson (commented):
WFM, though I would also add a paragraph somewhere about what you might do to upgrade SHA-256 if there was ever cause to do so (and maybe why that is unlikely, as we are not relying on any of the properties of the function that are security-relevant).

pmeenan:
> WFM, though I would also add a paragraph somewhere about what you might do to upgrade SHA-256 if there was ever cause to do so (and maybe why that is unlikely, as we are not relying on any of the properties of the function that are security-relevant).

Thanks. I added a paragraph to the security section about what the hash is used for and what it would look like if it was decided that SHA-256 needed to be replaced.

Issue #2639, opened 2023-09-20
Wildcard matching - Raised by Martin Thomson and Roy Fielding


pmeenan:
From @martinthomson:

For what it's worth, I like Roy's idea of prefix+suffix matching in place of arbitrary wildcards. Pattern matching is more expensive than simple prefix/suffix matches.

Also, "*" is a valid path character in HTTP. Maybe that was unwise, but that's what RFC 3986 says. Probably not a big deal (you can avoid using it if you plan to use this feature), but still.

Moved from private ID [issue #1](https://github.com/pmeenan/i-d-compression-dictionary/issues/1)

pmeenan:
[Original listserv response](https://lists.w3.org/Archives/Public/ietf-http-wg/2023JulSep/0130.html) from @royfielding:

Using * as a wildcard in arbitrary URL references is a foot gun. It would make more sense to have two attributes, prefix and suffix, and have them only match within the URL path (i.e., exclude the origin and query portions, preventing matches on full URIs or user-supplied query parameters). That is far more likely to get right than allowing things like "//example.com/////****"

pmeenan:
The origin is already excluded from being configurable. There is some discussion about only supporting relative paths but allowing for full URLs just made it easier to reference the existing URL RFC without having to re-define just the parts we need to support.

Query params can't necessarily be excluded and some sites are going to want to allow for either fixed query param matching or wildcard (and maybe for both the static and dynamic use case). Allowing for * allows for some flexibility in site URL structure while still keeping the matching relatively simple and without the complexity of [URLPattern](https://github.com/WICG/urlpattern/blob/main/mdn-drafts/QUICK-REFERENCE.md)

[listserv response](https://lists.w3.org/Archives/Public/ietf-http-wg/2023JulSep/0131.html)

pmeenan:
From @royfielding:

I get really nervous when I see shell syntax mixed with URI syntax as a form of pattern matching. We usually don't do that on servers. Regular expression syntax is an even bigger foot gun, but we already have that implemented. Alternatively, the URI Template syntax is a form of pattern matching that is more specific than a wildcard.

If prefix/suffix is insufficient, then maybe "contains" could be there as well, though I'd seriously question why a site would want this much flexibility.

pmeenan:
[...]

Issue #2638, opened 2023-09-20
Match URLs should not allow full URLs - Raised by Ilari Liusvaara


pmeenan:
Allowing absolute URLs in match is a footgun, since dictionaries are restricted to same-origin. I don't think any of the usual URI productions are suitable here.

I think most suitable would be 'segment *( "/" segment )', where segment is the production from RFC 3986.

[Original listserv response](https://lists.w3.org/Archives/Public/ietf-http-wg/2023AprJun/0309.html)

Moved from private ID [issue #2](https://github.com/pmeenan/i-d-compression-dictionary/issues/2)

pmeenan:
Even though the URLs must be same-origin, there's some benefit to allowing arbitrary URL formatting to allow for the clean use of relative URLs. They are all going to be expanded to absolute URLs anyway and it wouldn't break anything if someone chose to include the protocol and host parts (though there's no benefit to sending them).

[Original listserv response](https://lists.w3.org/Archives/Public/ietf-http-wg/2023AprJun/0312.html)

pmeenan:
There is talk about using [URLPattern](https://wicg.github.io/urlpattern/) with an implicit base URL.

The current plan would be to use the version that takes a patternString and have the baseUrl be set automatically to the URL of the request (and to limit the supported tokens).  That mode would still allow for full URLs.

There is also the possibility to break up the `path` and `search` components into different `use-as-dictionary` keys in which case it would prevent full URLs. There's some discussion of what the default `search` pattern behavior should be if one isn't specified though (match all or only match an empty search string).

royfielding:
It is really hard to imagine why we would go to all the trouble of optimizing data compression and yet not care about repeating the origin data in the header fields (even though the client is already known to have that data and is required to ignore it if provided). If we don't need it, don't send it on the wire.

It is a foot gun because a lazy implementer (the kind of person who never reads an RFC) will use the full URL provided for the match because it's easier than remembering (or finding) the request context. I know it's silly, but I have seen a lot worse.

pmeenan:
Closing this now that we have switched to using URLPattern and the path and query components are explicitly split (and the only settable components).

pmeenan:
Re-opening this because the [URLPattern integration changes](https://github.com/httpwg/http-extensions/pull/2689) don't allow for explicitly preventing the origin from being included in the match string.

pmeenan:
[...]

Issue #2625, opened 2023-09-11
Support matching on fetch destination


pmeenan:
In addition to matching dictionaries based on URL, it would be beneficial to be able to specify an optional [fetch destination](https://fetch.spec.whatwg.org/#concept-request-destination).

Not 100% sure it is necessary since a site-wide dictionary could still be used without specifying that it should only be advertised for documents as long as the origin only applies them on appropriate responses but it could be useful. A site-wide HTML dictionary was the use case referenced (which would only really be useful to be used on `document` requests).

Maybe as a `match-dest` field on the `Use-As-Dictionary` response.

PR #2646, opened 2023-10-09
Switch to URLPattern for request matching

pmeenan:
This switches the custom wildcard matching to use [URLPattern](https://wicg.github.io/urlpattern/) with regexp support disabled. URLPattern is being used by the service worker routing API's and is what the web side is standardizing on for anything that pattern-matches URLs.

This splits the match pattern into explicit path and search (query param) components and forces the origin to be inherited from the dictionary request URL.

Also included here is a new `Use-As-Dictionary` field for `match-dest` to allow for matching on request destinations. The biggest use case for that will be for allowing site-wide HTML dictionaries to do something like `Use-As-Dictionary: match-path="/*", match-dest="document"` and for the dictionary to not be sent for every request across the site (images, etc).

See issues #2639, #2638, #2625

### DIFF
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -77,33 +77,56 @@ This document uses the line folding strategies described in [FOLDING].
 ## Use-As-Dictionary
 
 When responding to a HTTP Request, a server can advertise that the response can
-be used as a dictionary for future requests for URLs that match the pattern
+be used as a dictionary for future requests for URLs that match the rules
 specified in the Use-As-Dictionary response header.
 
 The Use-As-Dictionary response header is a Structured Field
-{{STRUCTURED-FIELDS}} sf-dictionary with values for "match", "ttl", "type" and
-"hashes".
+{{STRUCTURED-FIELDS}} sf-dictionary with values for "match-path",
+"match-search", "match-dest", "ttl", "type" and "hashes".
 
-### match
+### match-path
 
-The "match" value of the Use-As-Dictionary header is a sf-string value that
-provides an URL-matching pattern for requests where the dictionary can be used.
+The "match-path" value of the Use-As-Dictionary header is a sf-string value
+that provides the "pathname" of a URLPattern
+(https://wicg.github.io/urlpattern/#dictdef-urlpatterninit).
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -77,33 +77,56 @@ This document uses the line folding strategies described in [FOLDING].
...
+"match-search", "match-dest", "ttl", "type" and "hashes".
 
-### match
+### match-path
 
-The "match" value of the Use-As-Dictionary header is a sf-string value that
-provides an URL-matching pattern for requests where the dictionary can be used.
+The "match-path" value of the Use-As-Dictionary header is a sf-string value
+that provides the "pathname" of a URLPattern
+(https://wicg.github.io/urlpattern/#dictdef-urlpatterninit).
```
martinthomson, line 91: Citing a WICG spec gives me the willies.  A normative reference to a CG draft is sketchy, but WICG is in a class of its own.

It might be worth talking to @wanderview about whether this has any hope of turning into a REC, hopefully through a working group.  It might be possible to get this particular piece of work done in the IETF, which might be easier than trying to find a home for it in W3C-land.

pmeenan: Oh yeah, I don't disagree: https://github.com/WICG/urlpattern/issues/180

It's absolutely on-path to move out of wicg and into a published spec - currently somewhere in w3c or whatwg though. I'll see if I can get a committment on exactly where it is destined to land and the timelines. @wanderview has moved on to other projects but there is a team within Chrome that has taken ownership and is going to be doing the work to complete standardization.

In the meantime, I added an editor's note to the top of the PR calling out the fact that this PR depends on a draft spec and that this ID can't be ratified until that spec is complete.

pmeenan: Looks like [this](https://github.com/whatwg/sg/issues/215) is the whatwh issue tracking the move (and consensus) from wicg to whatwg.

pmeenan: URLPattern is a whatwg standard now (https://urlpattern.spec.whatwg.org/) so I updated the references to point to the new doc.

Issue #2635, opened 2023-09-20
Should not use Content-encoding - Raised by Roy T. Fielding


pmeenan:
I think implementation of such through content-codings is fundamentally misguided because it changes the resource itself and impacts all caching along the chain of requests in ways that are non-recoverable. That is due to the lost metadata and variance on whatever request field is used to indicate that some downstream client can grok some possible dictionary.

In short, it looks like an easy solution for a browser, but will wreak havoc with the larger architecture of the Web.

The right way to do this is to implement it as a transfer encoding that can be decoded without loss or confusion with the unencoded resource, which would require extending h2 and h3 to support that feature of HTTP/1.1.

[Original listserv response](https://lists.w3.org/Archives/Public/ietf-http-wg/2023JulSep/0130.html)
cc @royfielding 

Moved from private ID [issue #6](https://github.com/pmeenan/i-d-compression-dictionary/issues/6)

pmeenan:
The decoded version of the resource is unchanged. It's not fundamentally different than brotli which happens to include a default dictionary and the caching is guaranteed to be maintained in a consistent way as long as "Vary" works on "Accept-Encoding" as well as whatever header negotiates the dictionary.

Even without the dictionary, if something in the middle doesn't know how to process one of the content-encodings (and needs to be able to access the content) then the accept-encoding should be modified to only include encodings that it knows how to work with.

This isn't really notably different than "br" or "zstd".

Moving the compression down into the transport layer is what we tried before but failed to navigate the browser security issues because the transport layer doesn't have the context of which responses need to be opaque, which responses are partitioned across document or frame boundaries, etc and that the dictionary compression could be used to perform oracle attacks across those boundaries.

[listserv response](https://lists.w3.org/Archives/Public/ietf-http-wg/2023JulSep/0131.html)

pmeenan:
@royfielding if the response includes the hash of the dictionary that was used for the compression so that the response is self-contained and independent of the request, would that address your concerns?

That is being discussed in #2636 and would allow for other ways of negotiating dictionaries (but would require a new header to identify the used dictionary).

pmeenan:
For the scale of improvement (for the resources impacted), we see results that are 60-90% smaller than brotli which is orders of magnitude bigger than we get from deflate -> brotli (for example). A few bytes added to headers or elsewhere aren't going to make a difference and, honestly, if it does then those resources are probably not appropriate for dictionary compression in the first place.

## Variants
[...]

PR #2648, opened 2023-10-12
Improve consistency of request variations

pmeenan:
This set of changes should reduce the number of variants that caches end up seeing from clients requesting dictionary-compressed resources with `Vary: accept-encoding, available-dictionary".

The default `ttl` is reduced to 7 days and the language is updated to refresh the expiration time whenever the dictionary is loaded from a local cache. That way the dictionary will default to being usable for up to 7 days after the last time a given client loaded the dictionary (as part of a page load or otherwise). This is independent of when it was last sent as "available" so if a page rotates dictionaries or JS libraries, the old one that is being used as a dictionary by clients won't be "loaded" any more, even if it isn't replaced so it will age out.

The `Accept-Encoding` header was updated to require `br` when `br-d` and `zstd` when `zstd-d` are advertised as well as specifying that the order be sorted alphabetically. This is an attempt to inject some cross-client consistency in the Accept-Encoding header when the same compression schemes are supported.

There will still be some fan-out for clients that support dictionaries but different sets of encodings but it should be a better baseline than we have today where they are not normalized.

See #2635

### DIFF
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -120,7 +120,11 @@ limit the life of a dictionary in cases where the dictionary is updated
 frequently which can help limit the number of possible incoming dictionary
 variations.
 
-The "ttl" value is optional and defaults to 31536000 (1 year).
+When a resource is loaded from a local cache in response to a client fetch with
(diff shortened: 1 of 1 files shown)


pmeenan:
Discussion in IETF 118 was to not specify anything around normalizing the Accept-Encoding (and that the ttl was not a concern). Closing

PR #2652, opened 2023-10-13
Add Token caveat

### LINKED ISSUE #2628: Strings, Tokens, and Display Strings
Display Strings are a superset of Strings, which are a superset of Tokens. This makes APIs more complex, and makes designers have to think about which structure to use.

A couple of thoughts:

1) Should we define best practice, that e.g., you should specify Token or String when you specify String -- similar to HTTP?

2) Should we define a new TokenOrString type to encourage this pattern (both in header definitions and APIs)?

mnot:
Fixes #2628.


bsdphk:
OK with me.

rashworld-max (approved):

PR #2655, opened 2023-10-18
Upgrade resumable diagrams to SVG

LPardue:
@martinthomson this works but generates warnings. Am I holding it wrong?

> /home/lucas/workspace/http-extensions/draft-ietf-httpbis-resumable-upload.xml(92): Warning: Found SVG with width or height specified, which will make the artwork not scale.  Specify a viewBox only to let the artwork scale.
/home/lucas/workspace/http-extensions/draft-ietf-httpbis-resumable-upload.xml(152): Warning: Found SVG with width or height specified, which will make the artwork not scale.  Specify a viewBox only to let the artwork scale.
/home/lucas/workspace/http-extensions/draft-ietf-httpbis-resumable-upload.xml(191): Warning: Found SVG with width or height specified, which will make the artwork not scale.  Specify a viewBox only to let the artwork scale.
/home/lucas/workspace/http-extensions/draft-ietf-httpbis-resumable-upload.xml(231): Warning: Found SVG with width or height specified, which will make the artwork not scale.  Specify a viewBox only to let the artwork scale.
/home/lucas/workspace/http-extensions/draft-ietf-httpbis-resumable-upload.xml(275): Warning: Found SVG with width or height specified, which will make the artwork not scale.  Specify a viewBox only to let the artwork scale.
/home/lucas/workspace/http-extensions/draft-ietf-httpbis-resumable-upload.xml(318): Warning: Found SVG with width or height specified, which will make the artwork not scale.  Specify a viewBox only to let the artwork scale.

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -88,7 +88,7 @@ In this example, the client first attempts to upload a file with a known size in
 
 An informational response can be sent to the client to signal the support of resumable upload on the server and transmit the upload resource URL in the Location header.
 
-~~~
+~~~ aasvg
 Client                                  Server
 |                                            |
 | POST                                       |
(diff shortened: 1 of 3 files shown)


martinthomson:
Those warnings are silly.

At some point, someone decided that because SVG diagrams scale, that all of them had to scale.  Consequently, xml2rfc has a warning when you choose not to allow the SVG to scale to the full width of the page.

This is silly because not all diagrams are the full width of the page.  Also, the size of the page varies, so if the size your diagram is anchored on the size of text (as aasvg does), and if you let the scale as xml2rfc would have it, then you end up with wildly different sizes for your diagrams.

(I should probably just tell xml2rfc to suppress that warning in a blanket rule.)

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -88,7 +88,7 @@ In this example, the client first attempts to upload a file with a known size in
 
 An informational response can be sent to the client to signal the support of resumable upload on the server and transmit the upload resource URL in the Location header.
 
-~~~
+~~~ aasvg
```
martinthomson, line 91: If you do this, you should also add `package.json` to the repository.  You might also need to add `package-lock.json` to the `.gitignore` file.  Otherwise, people without aasvg installed will have a hard time.

That's a once-only thing that you have the privilege of having to do.

package.json
```diff
@@ -0,0 +1,5 @@
+{
+    "dependencies": {
```
martinthomson, line 2: Consider adding `package-lock.json` to `.gitignore` as well.

guoye-zhang (approved):

Acconut (approved):
Thanks, I am curious to see how it will look in the end

LPardue:
> Thanks, I am curious to see how it will look in the end

Oh you can see that here in the HTML rendering - https://httpwg.org/http-extensions/lucas/resumable-svgs/draft-ietf-httpbis-resumable-upload.html

e.g.

image

PR #2666, opened 2023-10-25
Add example for  encoding`\\`

### LINKED ISSUE #2665: Provide an example for escaping "\\"
From the list:

At the end of Section 2.1 in https://www.ietf.org/archive/id/draft-ietf-httpbis-alias-proxy-status-05.html, there is this:

> It is also possible for a DNS name to include a period character (".") within a label, instead of as a label separator. In this case, the period character MUST be first escaped as "\.". Since the "\" character itself will be percent-encoded, the name "dot\.[label.example.com](http://label.example.com/)" would be encoded within a next-hop-aliases parameter as follows:
> 
> Proxy-Status: [proxy.example.net](http://proxy.example.net/); next-hop="2001:db8::1";
>     next-hop-aliases="dot%[5C.label.example.com](http://5c.label.example.com/),[service1.example.com](http://service1.example.com/)"
> Upon parsing this name, "dot%5C.label" MUST be treated as a single label.
> 
> Similarly the "\" character in a label MUST be escaped as "\\". Other uses of "\" MUST NOT appear in the label after percent-decoding.

Can that last line be expounded on and clarified? Will a "\" character in a label be 'eventually' encoded as "\\" or "%5C%5C" or some combination of them?

tfpauly:
Closes #2665

### DIFF
--- a/draft-ietf-httpbis-alias-proxy-status.md
+++ b/draft-ietf-httpbis-alias-proxy-status.md
@@ -160,7 +160,7 @@ Upon parsing this name, "dot%5C.label" MUST be treated as a single label.
 Similarly the "\\" character in a label MUST be escaped as "\\\\" and then percent-encoded.
 Other uses of "\\" MUST NOT appear in the label after percent-decoding. For example,
 if there is a DNS name "backslash\\name.example.com", it would first be escaped as
-"backslash\\\\name.example.com", and the percent-encoded as follows:
+"backslash\\\\name.example.com", and then percent-encoded as follows:
 
 ~~~ http-message
 Proxy-Status: proxy.example.net; next-hop="2001:db8::1";
--- a/draft-ietf-httpbis-alias-proxy-status.md
+++ b/draft-ietf-httpbis-alias-proxy-status.md
@@ -157,7 +157,15 @@ Proxy-Status: proxy.example.net; next-hop="2001:db8::1";
 
 Upon parsing this name, "dot%5C.label" MUST be treated as a single label.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-alias-proxy-status.md
```diff
@@ -157,7 +157,15 @@ Proxy-Status: proxy.example.net; next-hop="2001:db8::1";
 
 Upon parsing this name, "dot%5C.label" MUST be treated as a single label.
 
-Similarly the "\\" character in a label MUST be escaped as "\\\\".  Other uses of "\\" MUST NOT appear in the label after percent-decoding.
+Similarly the "\\" character in a label MUST be escaped as "\\\\" and then percent-encoded.
+Other uses of "\\" MUST NOT appear in the label after percent-decoding. For example,
+if there is a DNS name `backslash\\name.example.com`, it would first be escaped as
```
tfpauly, line 162:
```suggestion
if there is a DNS name "backslash\\name.example.com", it would first be escaped as
```

draft-ietf-httpbis-alias-proxy-status.md
```diff
@@ -157,7 +157,15 @@ Proxy-Status: proxy.example.net; next-hop="2001:db8::1";
...
+`backslash\\\\name.example.com`, and the percent-encoded as follows:
```
tfpauly, line 163:
```suggestion
"backslash\\\\name.example.com", and the percent-encoded as follows:
```

PR #2669, opened 2023-10-29
[message-signatures] fix wrong Content-Digest field value in example

takemar:
Hello,

In section B.2, the value of the Content-Digest field in "test-response message" does not match what is actually calculated from the content body.

The same problem in section B.2.4 was fixed in commit b3fa85b44f417b0b29101e3713bd8c22bbdc0893, but not here at the beginning of section B.2.

I am not sure if it is appropriate to submit a pull request since this document seems to be in the final stages of the publication process, but I would like to let it know anyway.


jricher:
Thank you for noticing this. The document is with the RFC editor right now and so we can't take the pull request, however we should be able to correct the example during the final editing phases before the document goes to RFC.

takemar:
I understand the status of the document and the measures being considered. Thank you for the reply!

mnot:
Signatures is published; closing.

PR #2670, opened 2023-11-01
Update resumable uploads IANA considerations

LPardue:
IANA early review highlighted that we were attempting to register to the old table. Lets fix that and do a little more tidyup.


guoye-zhang (approved):

Acconut (approved):

Acconut:
Thank you!

Issue #2636, opened 2023-09-20
Response should include dictionary used for encoding - raised by Martin Thomson


pmeenan:
I don't understand how a client determines which dictionary was used by the server. It seems like the response depends on the request, which is fragile (and doesn't allow clients to advertise multiple dictionaries, which seems likely with Use-As-Dictionary as it is).

[Original listserv response](https://lists.w3.org/Archives/Public/ietf-http-wg/2023JulSep/0102.html)
cc @martinthomson 

Moved from private ID [issue #3](https://github.com/pmeenan/i-d-compression-dictionary/issues/3)

pmeenan:
For identifying the dictionary, it's worth noting that the spec requires that the client advertise a SINGLE dictionary that it supports but yes, the response depends on the dictionary identified in the request. This also helps with minimizing the variants that caches would need to store since the response needs to be varied on the request's available dictionary (and multiple would explode the permutations). It wouldn't hurt to echo the dictionary hash in the response which would allow for other mechanisms of advertising dictionaries but that also carries the cost of complicating the vary logic (would need to vary based on whatever request header was used to negotiate the dictionary that ended up being used).

[Listserv response](https://lists.w3.org/Archives/Public/ietf-http-wg/2023JulSep/0104.html)

PR #2680, opened 2023-11-10
Add the dictionary hash to the HTTP response

pmeenan:
Adds a new "Content-Dictionary" with the hash of the dictionary used when encoding the HTTP response.

For #2636

### DIFF
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -267,7 +267,7 @@ fetched dictionary takes precedence.
 ## Content-Dictionary
 
 When a HTTP server responds with a resource that is encoded with a dictionary
-the server SHOULD send the hash of the dictionary that was used in the
+the server MUST send the hash of the dictionary that was used in the
 "Content-Dictionary" response header.
 
 The "Content-Dictionary" response header is a lowercase Base16-encoded
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -264,6 +264,37 @@ longest "match-query" takes precedence.
 1. Given equivalent destination, path and search precedence, the most recently
 fetched dictionary takes precedence.
 
+## Content-Dictionary
+
+When a HTTP server responds with a resource that is encoded with a dictionary
+the server SHOULD send the hash of the dictionary that was used in the
+"Content-Dictionary" response header.
+
+The "Content-Dictionary" response header is a lowercase Base16-encoded
+{{RFC4648}} {{SHA-256}} hash of the contents of the dictionary that was used
+to encode the response.
+
+If the HTTP response contains a "Content-Dictionary" response header with the
+hash of a dictionary that the client does not have available then the client
+MUST NOT attempt to decode or use the HTTP response.
+
+Its syntax is defined by the following {{ABNF}}:
+
+~~~ abnf
+Content-Dictionary = hvalue
+hvalue             = 1*hchar
+hchar              = DIGIT / "a" / "b" / "c" / "d" / "e" / "f"
+~~~
+
+For example:
+
+~~~ http-message
+NOTE: '\' line wrapping per RFC 8792
+
+Content-Dictionary: \
+  a591a6d40bf420404a011733cfb7b190d62c65bf0bcda32b57b277d9ad9f146e
+~~~
+
 # Negotiating the compression algorithm
 
 When a compression dictionary is available for use for a given request, the
@@ -335,6 +366,7 @@ IANA is asked to update the
 |----------------------|-----------|-------------------------------------------|
 | Use-As-Dictionary    | permanent | {{use-as-dictionary}} of this document    |
 | Available-Dictionary | permanent | {{available-dictionary}} of this document |
+| Content-Dictionary   | permanent | {{content-dictionary}} of this document   |
 |----------------------|-----------|-------------------------------------------|
 
 # Compatibility Considerations
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -202,7 +202,7 @@ Its syntax is defined by the following {{ABNF}}:
 ~~~ abnf
 Available-Dictionary = hvalue
 hvalue               = 1*hchar
-hchar                = DIGIT / "a" / "b" / "c" / "d" / "e" / "f"
+hchar                = DIGIT / %x61-66 ; 0-9, a-f
 ~~~
 
 The client MUST only send a single "Available-Dictionary" request header
@@ -283,7 +283,7 @@ Its syntax is defined by the following {{ABNF}}:
 ~~~ abnf
 Content-Dictionary = hvalue
 hvalue             = 1*hchar
-hchar              = DIGIT / "a" / "b" / "c" / "d" / "e" / "f"
+hchar              = DIGIT / %x61-66 ; 0-9, a-f
 ~~~
 
 For example:
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -276,7 +276,7 @@ to encode the response.
 
 If the HTTP response contains a "Content-Dictionary" response header with the
 hash of a dictionary that the client does not have available then the client
-MUST NOT attempt to decode or use the HTTP response.
+cannot decode or use the HTTP response.
 
 Its syntax is defined by the following {{ABNF}}:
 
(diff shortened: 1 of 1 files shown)


pmeenan:
@martinthomson this adds the dictionary hash to the HTTP response when used.  Right now it is a `SHOULD`. Let me know if you think it should be a `MUST` instead.

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -264,6 +264,37 @@ longest "match-query" takes precedence.
...
+If the HTTP response contains a "Content-Dictionary" response header with the
+hash of a dictionary that the client does not have available then the client
+MUST NOT attempt to decode or use the HTTP response.
+
+Its syntax is defined by the following {{ABNF}}:
+
+~~~ abnf
+Content-Dictionary = hvalue
+hvalue             = 1*hchar
+hchar              = DIGIT / "a" / "b" / "c" / "d" / "e" / "f"
```
martinthomson, line 286: This isn't lowercase.  Strings in ABNF are case-insensitive.  You need to use `%x61-66`.

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -264,6 +264,37 @@ longest "match-query" takes precedence.
...
+the server SHOULD send the hash of the dictionary that was used in the
+"Content-Dictionary" response header.
+
+The "Content-Dictionary" response header is a lowercase Base16-encoded
+{{RFC4648}} {{SHA-256}} hash of the contents of the dictionary that was used
+to encode the response.
+
+If the HTTP response contains a "Content-Dictionary" response header with the
+hash of a dictionary that the client does not have available then the client
+MUST NOT attempt to decode or use the HTTP response.
```
martinthomson, line 279: A "MUST NOT" here doesn't really help.  "cannot" is probably the word you want.

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -264,6 +264,37 @@ longest "match-query" takes precedence.
 1. Given equivalent destination, path and search precedence, the most recently
 fetched dictionary takes precedence.
 
+## Content-Dictionary
+
+When a HTTP server responds with a resource that is encoded with a dictionary
+the server SHOULD send the hash of the dictionary that was used in the
+"Content-Dictionary" response header.
```
martinthomson, line 271: Usually when we use "SHOULD", there is a discussion about what sort of reasons might lead to the recommendation being ignored.  What might cause a server to choose not to include this field?

martinthomson (commented):
If this field carries binary data, why aren't you using structured fields?

pmeenan:
> If this field carries binary data, why aren't you using structured fields?

Mostly for developer ergonomics to make it easier to deploy. In the case of static resources using previous versions as dictionaries (v1 to v2 of main.js), one method for enabling that is to compress the new version using the old version on the cli as part of the push to production and store the dictionary-compressed artifact as a file on the server next to the full file.  Then at serving time, compare the value of the `Available-Dictionary` request header with the file names on the server for the given URL to see if a delta-compressed artifact is available.

Structured fields encode the binary data as base-64 encoded strings which aren't filesystem-friendly so there would need to be some steps at serving time to decode the header, convert it to ascii and then check. It's not insurmountable but it's a bit more processing and a straight string match could be handled directly with templating in server config.

Lowercase hex was picked since that's what all of the cli tools for generating hashes emit.

draft-ietf-httpbis-compression-dictionary.md
pmeenan: Thanks. Updated.

draft-ietf-httpbis-compression-dictionary.md
pmeenan: I changed it to MUST since there isn't really a good case for when you wouldn't send the dictionary used in the response.

draft-ietf-httpbis-compression-dictionary.md
pmeenan: Is "cannot" the right way to phrase it when it is technically possible but bad things could happen? i.e. You can decompress a resource using a dictionary other than the one that was used to compress it but you will get the wrong output (level of brokenness depends on how much the dictionaries varied).  So it's "possible" to do but you absolutely must not do it or you'll get corrupt data.

pmeenan:
Is `Content-Dictionary` a reasonable name for the response header?  I'd actually prefer `Dictionary` but it felt like that might be over-stepping since there could be some other generic need for associating a dictionary with a HTTP-response but maybe I'm being overly-cautious.

draft-ietf-httpbis-compression-dictionary.md
MikeBishop: "cannot decode or use" would probably be correct. The client could _attempt_ to decode, but the decoding will not be successful, and a content-digest would reveal that.

pmeenan: Thanks.  Changed to cannot.  I was just worried that it was a technical barrier that people might try to rely on (decode and trust that it would fail if the dictionary didn't match) instead of failing silently by generating garbage.

martinthomson:
I don't think that authoring convenience outweighs the advantages of having a unified grammar here.  It's not like base64 is some esoteric function.  If you are talking about sha256sum, here's a toy: https://github.com/martinthomson/shabase64

reschke:
I agree with @martinthomson - don't. To make the sf value filesystem friendly, you just need to strip a few characters and remap "/".

davidben:
> Lowercase hex was picked since that's what all of the cli tools for generating hashes emit.

Keep in mind that hashes are not hex strings. A SHA-256 hash is 32 bytes of binary data. The hex thing is just what some tools happen to output. But anything consuming this programmatically to, e.g., compare the hash against something would be best served by the underlying bytes.

Using base64 is also slightly more compact on the wire. And if binary structured fields gets off the ground, it can be even more compact.

davidben:
Having a finally consistent grammar for HTTP is the precursor to get consistent tooling. It may be more convenient to do something else in the short term, but in the long term we'd just be adding to the mess of inconsistency in HTTP. In the long term, I would hope that these config language DSLs would be able to process structured fields natively. (Although, it may be on this WG to demonstrate this being possible, since the data model for structured fields is slightly interesting...)

I also don't think we should be encouraging, much less designing for, folks trying to process HTTP headers with regexps. There's a well-defined grammar and inventing other parsers is a great way to not quite get things right. We already have a pretty hard time with HTTP extensibility due to folks parsing it wrong.

The path-unfriendliness is a little unfortunate. Possibly we should have picked the URL-safe version of base64, but too late for that. But since byte strings and text strings aren't the same type, I imagine any sf-binary-aware config language DSL would also have hex and base64 functions available. Indeed taking attacker-supplied input and mapping it to a path, without running it through some very restrictive encoding, seems a pretty bad idea!

At the end of the day, SHA-256 hashes are byte strings, and sf-binary is the way HTTP decided to say byte string.

We could say that, although SHA-256 hashes are byte strings, dictionary names are ASCII strings computed by hex(sha256). Then these headers should be sf-string (with quotes). But that would be less efficient on the wire, and our [mandate](https://www.w3.org/TR/design-principles/#priority-of-constituencies) on the web is to put user needs (bandwidth, in this case) first, so I think this is the wrong tradeoff.

That said, I see `Available-Dictionary` also uses this format. It wouldn't make sense for `Available-Dictionary` and `Content-Dictionary` to use different formats, so perhaps switching away from the less consistent and less efficient format should probably be handled separately from this PR?

pmeenan:
I switched `Available-Dictionary` and `Content-Dictionary` both to sf-binary.  The plan is to update the ID and then send out a summary of changes for discussion on the listserv and since this PR is about the dictionary in the headers anyway it should be fine to update both of them at the same time.  Happy to split it out if it helps with history tracking.

yoavweiss:
I think this change adds significant complexity, and I'm not sure regarding its benefits.

I very recently played around with prototyping a compression dictionaries deployment and was very happy with the simplicity of it:
* You generate diff files at build time with some naming convention
* Then at serving time, you configure your routing logic (similarly to how @pmeenan [pointed out](https://github.com/httpwg/http-extensions/pull/2680#issuecomment-1810759792)) to those diffs

That can be done today in almost any deployed routing layer.

Once this change lands in implementations, a similar deployment would require me to add custom logic to transform that header value to something that is file system friendly. 
It doesn't matter if that transformation is to base64 decode the binary data and then to hex encode it, or to 'just.. strip a few characters and remap "/"'. It is an operation that is not necessarily supported in many layers that developers currently operate in.

To take just one example - if I were to implement this as a [Cloudflare transform rule](https://developers.cloudflare.com/rules/transform/url-rewrite/), I'm not even sure it's feasible to do that. (although it might be,  with a creative [`regex_replace`](https://developers.cloudflare.com/ruleset-engine/rules-language/functions/#function-regex_replace).

Can you elaborate on the advantages of moving to base64, beyond theoretical purity?

Issue #2633, opened 2023-09-20
Add dictionary URL to sec-available-dictionary requests - Raised by Vlad Krasnov


pmeenan:
I think it would make the proposal easier to support from a CDN POV if the Sec-Available-Dictionary included the dictionary URL and Etag, otherwise it requires too much state keeping.

Given a URL and an Etag it is pretty easy to have a best effort dictionary compression.

[Original listserv response](https://lists.w3.org/Archives/Public/ietf-http-wg/2023JulSep/0091.html)
cc @vkrasnov

Moved from private ID [issue #9](https://github.com/pmeenan/i-d-compression-dictionary/issues/9)

pmeenan:
That would necessarily couple the use of a dictionary with the negotiation of the storage. They way it is currently set up, "use-as-dictionary" is one way to populate the dictionary but the dictionaries could also be pre-loaded into a client or otherwise negotiated.

As far as state, if you're passively observing traffic to compress and not actively managing the dictionaries, doesn't something like this work?

See response with "use-as-dictionary" response header
Store dictionary in key-value store, using the hash of the payload as
the key
See request with "sec-available-dictionary" request header
For supporting static, offline resource compression:

If a version of the resource is in cache compressed with the requested dictionary, serve it
If not, kick off an async task to compress the cached full response with the dictionary (if one is available in KV)
For dynamic responses:

If the dictionary is in KV, use it to compress the response
The main "state" that needs to be managed is the storage of the dictionary keys, indexed by the appropriate hash.

If the CDN is generating the dictionaries (or adding the headers) there's not much more state than that either.

I'm sure I'm probably hand-waving a lot of things that need to be done at scale when handling thousands of dictionaries and thousands of simultaneous requests for the same dictionary but I'd hate to add complexity to the negotiation itself if it isn't absolutely necessary.

[listserv response](https://lists.w3.org/Archives/Public/ietf-http-wg/2023JulSep/0092.html)

pmeenan:
One area I could see benefitting from reducing some overhead would be around the hash of the dictionary contents since it could be expensive to calculate.

One thing we could do there is have an additional (optional) field in the use-as-dictionary response where the origin could populate the hash as "hint" so caches or clients could use the advertised hash to determine if it needs to store the dictionary or if it already has it.

The language would have to be careful to require that the actual hash still be calculated before storing (or use) to be sure that the has represents the contents accurately but the hint could make the processing lighter weight, particularly in the CDN flow.

[...]

PR #2687, opened 2023-11-27
Add support for a server-provided dictionary ID

pmeenan:
This adds support for a server-provided sf-string "id" of up to 1024 characters that is echoed back in a "Dictionary-ID" request header when the dictionary is advertised as being available.

This allows for a server to use something other than the hash to retrieve the dictionary (i.e. a cache key). This should be more flexible than the client sending the original request URL and will allow for servers to store arbitrary context information with the dictionary.

It doesn't open up any additional privacy or security concerns because the dictionary hash was already being treated as if it was user-identifiable information.

For #2633


vkrasnov:
Yes, looks great! Thanks!

PR #2688, opened 2023-11-27
Change the URLPattern integration to use URLPatternInit

pmeenan:
This cleans up the URLPattern construction logic to make the init steps clear.


pmeenan:
@jeremyroman @domenic this is a tweak to the existing split method of initializing URLPattern using URLPatternInit (much like the [JSON integration](https://whatpr.org/urlpattern/199.html#other-specs-json) of the in-flight PR).

I can switch it to use a single string-based construction which would simplify this part of the matching algorithm but I'm less clear about what the language should look like when validating the pattern to recommend path-relative patterns and reject patterns that aren't same-origin as the base URL.

Not necessarily looking to land this as-is but figured this could be a good place to work through what the language should look like.

pmeenan:
I have a single-string variant option in #2689

pmeenan:
Closing this in favor of iterating on #2689

PR #2693, opened 2023-12-02
Digests AUTH48

LPardue:
This is just the initial tranche of trivial changes. There's more to come as we work through some of the more complicated stuff. Will remain in draft state until the RFC is published.

- Remove code samples and changelog
- Section title nits
- nits in section 1
- nits in section 2+3
- nits in section 4
- nits in section 5
- nits in section 6
- nits in section 7
- nit in refs
- nits in examples
- nits in remaining appendicies
- nit: one more 3230 ref


LPardue:
The diff is available at https://author-tools.ietf.org/api/iddiff?url_1=https://www.rfc-editor.org/authors/rfc9530.txt&url_2=https://httpwg.github.io/http-extensions/draft-ietf-httpbis-digest-headers.txt

The diff should be as minimal as we can get it  (some things are simply can't get the same, but they don't matter)

PR #2706, opened 2023-12-17
compression-dictionary: remome abbreviated title

reschke:
1. Not needed
2. Would need to be at title, not label-like


pmeenan (approved):
Thanks

PR #2709, opened 2024-01-16
Removed ttl from dictionary response

### LINKED ISSUE #2649: Compression dictionary lifetime and cache lifetime
Why does the dictionary lifetime need to be independent of the cache lifetime?  Having an object that is cached with a separate lifetime seems unnecessary.

I would have said that clients can cache a dictionary for as long as they want (as always), but no longer than the lifetime stated in the cache-control field.

Now, you might say that a server might want a resource to be usable in its own right (i.e., as an ordinary resource) for one duration, with a separate lifetime for its use as a dictionary.  I see two reasons that this isn't much of a compelling argument:

* A server can forget about a dictionary and not provide delta compression at any time, so there is no real need for the dictionary lifetime to be shorter than the cache lifetime.   Client caches won't want to retain something for longer than its validity, so I don't see much chance that a resource will be kept past its cache lifetime.

* Resources like this are likely exclusively used as a dictionary.  There isn't much value for those resources having a separate lifetime in that case.

pmeenan:
This removes the dictionary-specific ttl and switches to using the dictionary resource freshness as the mechanism for expiring dictionaries.

Fix #2649

### DIFF
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -199,7 +199,7 @@ Available-Dictionary: :pZGm1Av0IEBKARczz7exkNYsZb8LzaMrV7J32a2fFG4=:
 ### Dictionary freshness requirement
 
 To be considered as a match, the dictionary resource MUST be either fresh
-[HTTP1.1] or within the "stale-while-revalidate" window defined in [RFC5861].
+[HTTP-CACHING] or allowed to be served stale (see eg [RFC5861]).
 
 ### Dictionary URL matching
 
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -80,7 +80,7 @@ be used as a dictionary for future requests for URLs that match the rules
 specified in the Use-As-Dictionary response header.
 
 The Use-As-Dictionary response header is a Structured Field
-{{STRUCTURED-FIELDS}} sf-dictionary with values for "match", "match-dest",
+{{STRUCTURED-FIELDS}} with values for "match", "match-dest",
 "id", and "type".
 
 ### match
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -199,7 +199,7 @@ Available-Dictionary: :pZGm1Av0IEBKARczz7exkNYsZb8LzaMrV7J32a2fFG4=:
 ### Dictionary freshness requirement
 
 To be considered as a match, the dictionary resource MUST be either fresh
-[HTTP1.1]. Or within the "stale-while-revalidate" window defined in [RFC5861].
+[HTTP1.1] or within the "stale-while-revalidate" window defined in [RFC5861].
 
 ### Dictionary URL matching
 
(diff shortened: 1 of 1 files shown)


pmeenan:
@martinthomson could you please take a look and see if this looks ok?

@mnot could you look at the language around the freshness of the resource relative to cache-control and stale-while-revalidate and make sure I expressed it properly (and it is something we're ok with leveraging)?

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -214,12 +198,8 @@ Available-Dictionary: :pZGm1Av0IEBKARczz7exkNYsZb8LzaMrV7J32a2fFG4=:
...
 ### Dictionary freshness requirement
 
-To be considered as a match, the dictionary must not yet be expired as a
-dictionary. When iterating through dictionaries looking for a match, the
-expiration time of the dictionary is calculated by taking the last time the
-dictionary was fetched and adding the "ttl" seconds from the
-"Use-As-Dictionary" response. If the current time is beyond the expiration time
-of the dictionary, it MUST be ignored.
+To be considered as a match, the dictionary resource MUST be either fresh
+[HTTP1.1]. Or within the "stale-while-revalidate" window defined in [RFC5861].
```
martinthomson, line 202:
```suggestion
[HTTP1.1] or within the "stale-while-revalidate" window defined in [RFC5861].
```

martinthomson (approved):
I like red patches.  Thanks.

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -39,6 +39,8 @@ author:
 normative:
   FOLDING: RFC8792
   HTTP: RFC9110
+  HTTP1.1: RFC2616
```
mnot, line 42:
```suggestion
  HTTP-CACHING: RFC9111
```

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -79,7 +81,7 @@ specified in the Use-As-Dictionary response header.
 
 The Use-As-Dictionary response header is a Structured Field
 {{STRUCTURED-FIELDS}} sf-dictionary with values for "match", "match-dest",
```
mnot, line 83:
```suggestion
{{STRUCTURED-FIELDS}} with values for "match", "match-dest",
```

Currently practice is to NOT use the ABNF.

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -214,12 +198,8 @@ Available-Dictionary: :pZGm1Av0IEBKARczz7exkNYsZb8LzaMrV7J32a2fFG4=:
...
 ### Dictionary freshness requirement
 
-To be considered as a match, the dictionary must not yet be expired as a
-dictionary. When iterating through dictionaries looking for a match, the
-expiration time of the dictionary is calculated by taking the last time the
-dictionary was fetched and adding the "ttl" seconds from the
-"Use-As-Dictionary" response. If the current time is beyond the expiration time
-of the dictionary, it MUST be ignored.
+To be considered as a match, the dictionary resource MUST be either fresh
+[HTTP1.1] or within the "stale-while-revalidate" window defined in [RFC5861].
```
mnot, line 202:
```suggestion
To be considered as a match, the dictionary resource MUST be either fresh
[HTTP-CACHING] or allowed to be served stale (see eg [RFC5861]).
```

PR #2712, opened 2024-01-22
Specify unicode encoding failure handling.

### LINKED ISSUE #2672: sfbis: clarify what display strings can contain
[4.1.11. ](https://www.ietf.org/archive/id/draft-ietf-httpbis-sfbis-03.html#section-4.1.11)[Serializing a Display String](https://www.ietf.org/archive/id/draft-ietf-httpbis-sfbis-03.html#name-serializing-a-display-strin) currently does not consider what happens when UTF8 serialization fails, although this could happen with surrogate points. I believe it should explicitly say that.

Alternatively, the introduction of display strings could say that only unicode scalars (https://www.ietf.org/archive/id/draft-bray-unichars-07.html#unicode-scalars) are allowed (but even then I'd add the extra precaution in the serializer).

Finally, we may want to add an example because all of this is far from obvious.

mnot:
Fixes #2672.

### DIFF
--- a/draft-ietf-httpbis-sfbis.md
+++ b/draft-ietf-httpbis-sfbis.md
@@ -720,7 +720,7 @@ Given a Date as input_date, return an ASCII string suitable for use in an HTTP f
 Given a sequence of Unicode codepoints as input_sequence, return an ASCII string suitable for use in an HTTP field value.
 
 0. If input_sequence is not a sequence of Unicode codepoints, fail serialization.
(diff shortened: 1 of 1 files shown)


reschke (approved):

mnot:
... in a swamp.

clelland:
Is it worth dealing with the fact that 3629 doesn't actually define how to encode a sequence of characters, but we're calling it with one? We'd need to add something to this text, like "for each character (or code point?) in input_sequence, append the result of applying UTF-8 encoding to character to byte_array".

PR #2719, opened 2024-01-25
[connect-tcp] Define syntax more precisely

### LINKED ISSUE #2714: connect-tcp: target_host is under-defined
The target_host variable is only described vaguely. It should be defined more rigorously. For example, it doesn't specify whether IPv6 scope IDs are allowed or not. I'd recommend reusing the text from RFC 9298.

bemasc:
The language here is adapted from RFC 9298.

Fixes #2714

### DIFF
--- a/draft-ietf-httpbis-connect-tcp.md
+++ b/draft-ietf-httpbis-connect-tcp.md
@@ -80,20 +80,34 @@ host-value = ip-addr / reg-name
 
 The "target_host" variable is permitted to contain either a `host-value` or a list of `ip-addr`.  The value of the "tcp_port" variable MUST match the `port` rule from {{!RFC3986}}.
 
-Note that IPv6 scoped addressing zone identifiers are not supported, and the colons in IPv6 addresses will normally be percent-encoded by the URI Template expansion process (see {{substitution-example}}).
+Note that IPv6 scoped addressing zone identifiers are not supported, and the colons in IPv6 addresses will normally be percent-encoded by the URI Template expansion process (see {{substitution-examples}}).
 
 ~~~~
-URI template:
-    https://a.example/{?target_host,tcp_port}
+Example URI Templates:
+  A: https://a.example{/target_host,tcp_port}
+  B: https://b.example/{?target_host*,tcp_port}
 
-Variable assignments in RFC 6570 syntax:
-    target_host := ("192.0.2.1", "2001:db8::1")
+Example variable assignments in RFC 6570 syntax:
+  1:
+    target_host := "one.example"
     tcp_port    := "443"
+  2:
+    target_host := ("192.0.2.2", "2001:db8::2")
+    tcp_port    := "80"
 
-Resulting URI:
-https://a.example/?target_host=192.0.2.1,2001%3Adb8%3A%3A1&tcp_port=443
+
+Resulting URIs:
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-connect-tcp.md
```diff
@@ -66,7 +66,23 @@ This specification describes an alternative mechanism for proxying TCP in HTTP.
...
+
+The "tcp_port" variable MUST represent a single integer between 1 and 65535 inclusive.  Both the "target_host" and "tcp_port" values MUST NOT be empty or undefined when using the template for TCP proxying.
+
+Using the terms `IPv6address`, `IPv4address`, `reg-name`, and `port` from {{!RFC3986}} and notation from {{!RFC5234}} and {{!RFC6570}}, the "target_host" and "tcp_port" values MUST adhere to the format in {{abnf}}:
+
+~~~~ abnf
+ip-addr = IPv6address / IPv4address
+single-host-str = DQUOTE ( ip-addr / reg-name ) DQUOTE
+ip-addr-str = DQUOTE ip-addr DQUOTE
+ip-addr-list = "(" ( ip-addr-str ", " )* ip-addr-str ")"
```
DavidSchinazi, line 79: This ABNF doesn't match the example, it includes DQUOTES and parentheses that aren't in the example. Why not simply reuse the syntax from 9298 with just the tweak for the list?

bemasc: Yes, the quotes and parentheses are from RFC 6570's value syntax.  They're needed in order to distinguish lists from strings, since ABNF has no concept of a list.

It sounds like you'd prefer to use ABNF only for the individual list elements, and leave the description of the list itself in English?

DavidSchinazi: I guess I'm confused, are the double quotes and parentheses required? If so, can you add them to the examples so I can visualize the syntax better?

DavidSchinazi: I'm even more confused by 5f5e1ac8645bd8352bbd1c372bf6178782d3e538. I assumed that the ABNF covered what would end up in the path, not in the "Variable assignments in RFC 6570 syntax". What am I missing?

bemasc: I agree that it's confusing.  ABNF is not really able to do what we need here.  I've moved more of the description to English to reduce the complexity.

DavidSchinazi: Why not? Doesn't this work?
```
ip-addr = IPv6address / IPv4address
ip-addr-list = *( ip-addr "," ) ip-addr
target_host = reg-name / ip-addr-list
```

bemasc: That's an option, but it would result in a change in the behavior because it would not respect the explode modifier.  I've added more examples that I hope make this clear.

RFC 6570 does not really define a canonical string representation for lists, so we can't specify them using ABNF.

DavidSchinazi: I see. If it were up to me, I would remove the explode modifier and only use the ABNF - that way we only have a single way to convey the information, and it's formally defined.

bemasc:
Closing in favor of #2736

PR #2720, opened 2024-01-25
[connect-tcp] Discuss limits of multiprotocol templates

### LINKED ISSUE #2713: connect-tcp: use target_port instead of tcp_port
The draft currently defines the "target_host" and "tcp_port" variables. We should replace "tcp_port" with "target_port" for consistency with connect-udp. This consistency isn't just cosmetic, it simplifies reusing the same URI template for both protocols.

bemasc:
Fixes #2713

### DIFF
--- a/draft-ietf-httpbis-connect-tcp.md
+++ b/draft-ietf-httpbis-connect-tcp.md
@@ -195,7 +195,7 @@ If this request succeeds, the client SHOULD record a preference for "connect-tcp
 
 ## Multi-purpose proxies
 
-The names of the variables in the URI Template uniquely identify the capabilities of the proxy.  Undefined variables are permitted in URI Templates, so a single template can be used for multiple purposes.
+The names of the variables in the URI Template allow the client to infer certain capabilities of the proxy.  Undefined variables are permitted in URI Templates, so a single template can be used for multiple purposes.
 
 Multipurpose templates can be useful when a single client may benefit from access to multiple complementary services (e.g., TCP and UDP), or when the proxy is used by a variety of clients with different needs.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-connect-tcp.md
```diff
@@ -195,15 +195,17 @@ If this request succeeds, the client SHOULD record a preference for "connect-tcp
...
 Multipurpose templates can be useful when a single client may benefit from access to multiple complementary services (e.g., TCP and UDP), or when the proxy is used by a variety of clients with different needs.
 
 ~~~
 https://proxy.example/{?target_host,tcp_port,target_port,
                         target,ipproto,dns}
 ~~~
-{: title="Example multipurpose template for a combined TCP, UDP, and IP proxy and DoH server"}
+{: title="Example multipurpose template that allows the client to infer support for TCP, UDP, and IP proxying, and for DNS over HTTPS"}
+
+To allow correct inference of the proxy's capabilities, multipurpose templates for general use MUST only contain variable names specified by the proxy's supported standards.  Note that the "connect-ip" protocol has no mandatory variable names, so a proxy's support for "connect-ip" cannot always be inferred from the template's contents.
```
DavidSchinazi, line 208: This statement requires updating RFC 9298, because it adds a new requirement on an implementation that only supports RFC 9298 and 9484. But more importantly, I'm not sure that requirement is a good idea. I'd rather find capabilities through discovery (e.g., draft-schwartz-masque-access-descriptions or draft-pauly-intarea-proxy-config-pvd) rather than inference.

tfpauly: Yeah — the feedback in the PvD document was to make the types more explicit and not just rely on the implicit interpretation of the template. Worth more discussion, I think.

bemasc:
Closing in favor of #2736

PR #2729, opened 2024-01-31
Treat empty dictionary ID's as "no ID".

### LINKED ISSUE #2721: Prohibit an empty string Dictionary-ID
Dictionary-ID was introduced to the Compression Dictionary Transport specification by #2687.

However, it is unclear whether empty string dictionary IDs are allowed.
I don't think there is any use case for an empty string dictionary ID.
So how about explicitly disallowing empty string dictionary IDs in the spec?
@pmeenan

pmeenan:
This treats empty dictionary strings as the same as there being no dictionary and reqires all ID's have a non-zero length for the Dictionary-ID header to be sent.

Fix #2721


pmeenan:
@horo-t could you PTAL and see if this handles the case you were concerned about?

horo-t (approved):
Thanks.
lgtm.

PR #2730, opened 2024-02-01
Change dictionary match-dest to a list of strings

### LINKED ISSUE #2722: Support a list of match-dest
`match-dest` option was introduced in the Compression Dictionary Transport spec by #2646.
But if my understanding is correct, currently it is not allowed to setting multiple types of destinations.
So if I want to use a dictionary for HTML files including top frame, `<frame>` and `<iframe>`, I need to download the dictionary three times.

@pmeenan 
How about allowing setting multiple types of destinations?
Example:
```
Use-As-Dictionary: match="/", match-dest=("document" "frame" "iframe")
```

### LINKED ISSUE #2723: Handling empty destination
The spec of compression dictionary transport says:
```
The "match-dest" value of the Use-As-Dictionary header is a sf-string value that provides a request destination (https://fetch.spec.whatwg.org/#concept-request-destination).

An empty string for "match-dest" MUST match all destinations.
``` 

But an empty [request destination](https://fetch.spec.whatwg.org/#concept-request-destination) "" is used for some features such as `navigator.sendBeacon()`, `EventSource`,  `fetch()`, `XMLHTTPRequest`, `WebSocket`, Cache API, Download, prefetch and prerender. (See this [table](https://fetch.spec.whatwg.org/#destination-table)).

So currently there is no way to set up a dictionary for such destinations.
For example, I want to use a dictionary for fetch() API, but I don't want to use the dictionary for HTML documents.

@pmeenan 
How about using a string "empty" for such empty destination?
We are already using "empty" for `Sec-Fetch-Dest` header of empty destination request.
https://w3c.github.io/webappsec-fetch-metadata/#abstract-opdef-set-dest

Also if we don't need to use an empty string "" for `match-dest`, I think `match-dest` should be a [sf-token](https://datatracker.ietf.org/doc/html/rfc8941#section-3.3.4) instead of a [sf-string](https://datatracker.ietf.org/doc/html/rfc8941#section-3.3.3). We don't need to use double quotes.

pmeenan:
This changes the `match-dest` field to be a list of destination strings to allow for matching multiple destinations (i.e. document and frame).

This allows for matching an empty destination explicitly using an empty string. e.g. `("script" "")`

Fix #2722
Fix #2723

### DIFF
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -80,8 +80,7 @@ be used as a dictionary for future requests for URLs that match the rules
 specified in the Use-As-Dictionary response header.
 
 The Use-As-Dictionary response header is a Structured Field
-{{STRUCTURED-FIELDS}} with values for "match", "match-dest",
-"id", and "type".
+{{STRUCTURED-FIELDS}} with values for "match", "match-dest", "id", and "type".
 
 ### match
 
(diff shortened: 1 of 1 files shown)


pmeenan:
@horo-t could you PTAL and see if this looks OK for the destination matching?

horo-t (commented):
lgtm

Sorry for the late reply.

PR #2733, opened 2024-02-05
Switch away from using ABNF names for field types

### LINKED ISSUE #2731: Change dictionary `type` to token instead of string

pmeenan:
Switched from using the ABNF field types to the human names (i.e. sf-dictionary to Dictionary) for all of the Structured Field header values.

Switched dictionary "type" to a Token (from String).

Fix #2731

### DIFF
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -84,8 +84,8 @@ The Use-As-Dictionary response header is a Structured Field
 
 ### match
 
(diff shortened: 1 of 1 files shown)


pmeenan:
@mnot, could you PTAL at the change in referencing the header field types?  Specifically, I wasn't sure if I should capitalize the types or not (I did since that's how the structured fields RFC referred to them).

mnot:
Yep, this looks good. It's common to add something to you Notational Conventions section (or whatever) that says roughly "This specification uses the following terms from [ref to SF]: String, Byte Sequence, Token, ..."

PR #2736, opened 2024-02-14
[connect-tcp] Adjust rules for target_host and target_port

### LINKED ISSUE #2713: connect-tcp: use target_port instead of tcp_port

bemasc:
Changes:

* s/tcp_port/target_port/ (aligning with connect-udp)
* Remove support for multiple IPs

Fixes #2713 and #2714


DavidSchinazi (approved):

MikeBishop (approved):

bemasc:
@MikeBishop The discussion at IETF 119 leaned toward removing the multiple-IP mode entirely.  Are you suggesting that we should keep it?

MikeBishop:
This PR contains two changes; I'm not making a suggestion with regard to multiple-IP mode, but I think aligning the variable names in the template makes sense. Did both get dropped, or is that being discarded because it's in the same PR?

bemasc:
OK, I've updated this PR to also drop the multiple-IP support entirely.

draft-ietf-httpbis-connect-tcp.md
```diff
@@ -52,8 +52,6 @@ HTTP clients can be configured to use proxies by selecting a proxy hostname, a p
 
 The absence of an explicit origin for the proxy also rules out the usual defenses against server port misdirection attacks (see {{Section 7.4 of ?RFC9110}}) and creates ambiguity about the use of origin-scoped response header fields (e.g., "Alt-Svc" {{?RFC7838}}, "Strict-Transport-Security" {{?RFC6797}}).
 
-Classic HTTP CONNECT proxies can be used to reach a target host that is specified as a domain name or an IP address.  However, because only a single target host can be specified, proxy-driven Happy Eyeballs and cross-IP fallback can only be used when the host is a domain name.  For IP-targeted requests to succeed, the client must know which address families are supported by the proxy via some out-of-band mechanism, or open multiple independent CONNECT requests and abandon any that prove unnecessary.
-
 ## Overview
```
MikeBishop, line 55: Given that `target_host` can contain either an IP address or a DNS name, this text still seems relevant to the overall doc. Is the removal deliberate, or was the text relocated?

DavidSchinazi (approved):
Thanks for the updates based on the discussion at 119

draft-ietf-httpbis-connect-tcp.md
```diff
@@ -187,21 +183,20 @@ If the client infers that classic HTTP CONNECT is not supported, it SHOULD retry
...
+A single URI Template can be used for multiple purposes if it includes all the relevant variables.  Variables that are not relevant to a particular usage are left undefined, which is permitted when expanding a URI Template.
 
-Multipurpose templates can be useful when a single client may benefit from access to multiple complementary services (e.g., TCP and UDP), or when the proxy is used by a variety of clients with different needs.
+Multipurpose templates can be useful when a single client may benefit from access to multiple complementary services (e.g., TCP and UDP), or when the proxy is used by a variety of clients with different needs.  The contents of the URI Template are not necessarily sufficient to determine its purpose, so clients must determine this in some other way, such as by probing or via a separate usage indication.
 
 ~~~
-https://proxy.example/{?target_host,tcp_port,target_port,
-                        target,ipproto,dns}
+https://proxy.example/{?target_host,target_port,target,ipproto,dns}
```
tfpauly, line 199: @bemasc with the rest of the updates in this PR (which I think are good and reflect the discussion), I'm thinking that it is best to remove this multi-purpose proxies section. I don't think it makes sense to have in this particular document. It doesn't seem specific to connect-tcp, and it explicitly mentions that it isn't sufficient to determine the purpose.

bemasc: OK, I've removed this section.

Issue #2728, opened 2024-01-30
Expand on text related to content-length:0 


LPardue:
The field definitions allow for a client to send 0-length content. Sometimes this is useful, other times it it might not be. For example, 

1. A client could perform upload creation by making a request content-length: 0 and with upload-complete:?0. This would allow it to test creation wihout having to commit to an upload in the same message.
2. A client could attempt to append to append zero bytes. Is this useful or should we prohibit it? This is a cheap way to keep an upload "active". However, that  might be a vector for abuse by allowing a client to cause server-side state commintment. Depending on what we decide, we might need some text in the security considerations.

smatsson:
Just to add to the two bullet points:

1. This can also be used for clients/servers that do not support 1xx response by sending a 0 byte upload-complete: ?0 request to get the unique resource URL. This is for instance how [tus-js-client](https://github.com/tus/draft-example/tree/main/clients/tus-js) does it.
2. For cases where multiple upload-complete: ?0 requests are made, a final 0 byte upload-complete: ?1 request can be used to notify the server that the upload is complete. This can be useful in cases where the client wants to provide all the data before finalizing the upload to reduce the risk of losing the final response. I don't really see a use case for sending _multiple_ 0 byte upload-complete: ?0 but there might be some?

PR #2742, opened 2024-02-23
Resumable Upload: Explain empty requests

Acconut:
This PR adds explanations helping readers to understand when empty requests for upload creation or appending might be useful and why server should support them.

Closes https://github.com/httpwg/http-extensions/issues/2728.

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -219,6 +219,8 @@ If the request includes an `Upload-Complete` field value set to true and a valid
 
 The request content MAY be empty. If the `Upload-Complete` header field is then set to true, the client intends to upload an empty entity. An `Upload-Complete` header field is set to false is also valid. This can be used to create an upload resource URL before transferring data, which can save client or server resources. Since informational responses are optional, this technique provides another mechanism to learn the URL, at the cost of an additional round-trip before data upload can commence.
 
+The following example shows an upload creation. The client transfers the entire 100 bytes in the first request. The server generates two informational responses to transmit the upload resource's URL and progress information, and one final response to acknowledge the completed upload:
+
 ~~~ http-message
 POST /upload HTTP/1.1
 Host: example.com
@@ -242,6 +244,8 @@ HTTP/1.1 201 Created
 Location: https://example.com/upload/b530ce8ff
 Upload-Offset: 100
 ~~~
+
+The next example shows an upload creation, where only the first 25 bytes are transferred. The server acknowledges the received data and that the upload is not complete yet:
 
 ~~~ http-message
 POST /upload HTTP/1.1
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -208,6 +208,8 @@ If the request completes successfully but the entire upload is not yet complete,
 
 If the request includes an `Upload-Complete` field value set to true and a valid `Content-Length` header field, the client attempts to upload a fixed-length resource in one request. In this case, the upload's final size is the `Content-Length` field value and the server MUST record it to ensure its consistency.
 
+The request content MAY be empty. If the `Upload-Complete` header field is then set to true, the client intends to upload an empty entity. If the `Upload-Complete` header field is then set to false, the client wants to obtain the upload resource's URL before transferring data. This is useful when the client is not able to retrieve the resource URL in an informational response.
```
LPardue, line 211: While the statement is correct, I don't think we want to promote the idea that 1xx can fail, how about something like

```suggestion
The request content MAY be empty. If the `Upload-Complete` header field is then set to true, the client intends to upload an empty entity. An `Upload-Complete` header field is set to false is also valid. This can be used to create an upload resource URL before transferring data, which can save client or server resources. Since informational responses are optional, this technique provides another mechanism to learn the URL, at the cost of an additional round-trip before data upload can commence.
```

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -406,6 +410,8 @@ The upload resource URL is the identifier used for modifying the upload. Without
 
 Some servers or intermediaries provide scanning of content uploaded by clients. Any scanning mechanism that relies on receiving a complete file in a single request message can be defeated by resumable uploads because content can be split across multiple messages. Servers or intermediaries wishing to perform content scanning SHOULD consider how resumable uploads can circumvent scanning and take appropriate measures. Possible strategies include waiting for the upload to complete before scanning a full file, or disabling resumable uploads.
 
+A malicious client may create upload resources and keep them alive by regularly sending `PATCH` requests with no or small content to the upload resources. The server SHOULD protect against such resource exhaustion attacks by limiting the lifetime of upload resources or the number of incomplete upload resources per client.
```
LPardue, line 413: This is good. Its reminds me a bit of the slow loris treatment in https://datatracker.ietf.org/doc/html/rfc9000#section-21.6, so let's ~steal~ borrow some of that

We should cite [SLOWLORIS ](https://web.archive.org/web/20150315054838/http://ha.ckers.org/slowloris/) but I'm too lazy right now and in a web editor.

```suggestion
Resumable uploads are vulnerable to Slowloris-style attacks [SLOWLORIS]. A malicious client may create upload resources and keep them alive by regularly sending `PATCH` requests with no or small content to the upload resources. This could be abused to exhaust server resources by creating and holding open uploads indefinately with minimal work.

Servers SHOULD provide mitigations for Slowloris attacks, such as increasing the maximum number of clients the server will allow, limiting the number of uploads a single client is allowed to make, imposing restrictions on the minimum transfer speed an uploads is allowed to have, and restricting the length of time an upload resource can exist.
```

LPardue (commented):
Thanks for this. A couple of suggestions to build on what you've put in.

guoye-zhang (approved):

draft-ietf-httpbis-resumable-upload.md
smatsson: Is it common to refer to specific attacks in a spec like this? It's probably not the case with Slowloris but I imagine it being possible that a referenced attack is no longer applicable due to other reasons?

draft-ietf-httpbis-resumable-upload.md
Acconut: Good suggestion, thank you! I didn't want to imply that the transmission of a 1xx can fail, but rather wanted to highlight situations where the client is incapable of receiving them because the HTTP interface does not expose them. But your change expresses that better.

draft-ietf-httpbis-resumable-upload.md
Acconut: It's probably better to name and link to similar attacks to help the reader understand the topic and related issues better. So it doesn't do any harm, I guess.

Issue #2610, opened 2023-08-03
Media type for Upload Appending Procedure


Acconut:
During IETF 117, it was noted that PATCH requests need to specify some media type, so our Upload Appending Procedure needs one as well. This could either be one from the byte range draft (#2501) or a custom one. For example, in the [tus uploading protocol](https://tus.io/protocols/resumable-upload) we use the content type `application/offset+octet-stream` to indicate that it is an undefined byte sequence that should be applied at a specific offset (as defined in the request).

Acconut:
I looked through the list of registered media types (https://www.iana.org/assignments/media-types/media-types.xhtml) to find if there is a media type that represents a file chunk without any additional metadata.

There is `message/partial`, but as far as I understand, it is meant for chunks of e-mail messages specifically, meaning it not suitable for chunks of files, in general.

Is there anybody aware of another existing media type? If not, we can propose registering `application/offset+octet-stream` or just going with `application/octet-stream`.

Acconut:
> Is there anybody aware of another existing media type? If not, we can propose registering `application/offset+octet-stream` or just going with `application/octet-stream`.

`application/octet-stream` is too generic and does not describe the patch document properly.
`application/offset+octet-stream` does not fit because the suffix (delimited by the `+` has a specific meaning for MIME types).

How about a dedicated `application/upload-chunk` or `application/partial-upload` media type? It would mean that the corresponding content is a consecutive, partial chunk of an upload. In the draft we can then define that an upload resource can apply such a patch document by appending the data to the file on the server.

The IANA registry of media types already includes a similar media type in the vendor tree from Adobe for use in one of their web applications: [application/vnd.adobe.partial-upload](https://www.iana.org/assignments/media-types/application/vnd.adobe.partial-upload)

smatsson:
My two cents: I would avoid the word "chunk" as it's been a source of confusion in tus.  There is ambiguity if it refers to part of the upload, a whole request (containing parts of the file and not the whole thing), part of the content of the request etc.

PR #2743, opened 2024-03-01
Resumable Uploads: `application/partial-upload` media type

Acconut:
As discussed in https://github.com/httpwg/http-extensions/issues/2501 and https://github.com/httpwg/http-extensions/issues/2610, the PATCH requests needs a suiting media type to describe the message content.

Since there does not seem to be an existing, suitable media type, [we proposed the `application/partial-upload` media type on the media-types mailing list](https://mailarchive.ietf.org/arch/msg/media-types/XoF-_vBDWWS4KjLFjLlrbnuO8bw/). This PR is a first, rough attempt at including such a media type in the draft.


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -324,7 +324,7 @@ The client SHOULD NOT automatically retry if a client error status code between
 
 Upload appending is used for resuming an existing upload.
 
-The request MUST use the `PATCH` method and be sent to the upload resource. The `Upload-Offset` field value ({{upload-offset}}) MUST be set to the resumption offset.
+The request MUST use the `PATCH` method and `application/partial-upload` media type and be sent to the upload resource. The `Upload-Offset` field value ({{upload-offset}}) MUST be set to the resumption offset.
```
smatsson, line 327: I feel that this sentence is a bit off, probably due to language barriers on my end. 

Suggested edit:
>  The request MUST use the `PATCH` method with `application/partial-upload` as content type and be sent to the upload resource. The `Upload-Offset` field value ({{upload-offset}}) MUST be set to the resumption offset.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -336,6 +336,8 @@ The client MUST NOT perform multiple upload transfers for the same upload resour
 
 If the offset indicated by the `Upload-Offset` field value does not match the offset provided by the immediate previous offset retrieval ({{offset-retrieving}}), or the end offset of the immediate previous incomplete successful transfer, the server MUST respond with a `409 (Conflict)` status code.
 
+The server applies the patch document of the `application/partial-upload` media type by appending the unmodified request content to the targeted upload resource.
```
smatsson, line 339: To me "unmodified" reads as "take the entire thing and apply it without changes" which is probably not what is intended here? The server is not allowed to pick bytes that it appends, but I think we need to clarify that the entire request must not complete for the server to append data. I.e. if the connection breaks down after 100 bytes then as much as possible of those 100 bytes _should_ be appended and what is appended _must_ be appended without modification.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -406,6 +409,10 @@ The `Upload-Complete` request and response header field indicates whether the co
 
 The `Upload-Complete` header field MUST only be used if support by the resource is known to the client ({{feature-detection}}).
 
+# Media Type `application/partial-upload`
+
+The `application/partial-upload` media type describes a contiguous block of data that should be uploaded to a resource. There is no minimum size on the block of data and it might be empty. The start and end of block of data might align with the start and end of the file that should be uploaded, but are not required to do so.
+
```
smatsson, line 415: Minor nit: 

> The start and end of *THE* block of data might align with the start and end of the file that should be uploaded, but are not required to do so.

or 

> The start and end of *THE* data might align with the start and end of the file that should be uploaded, but are not required to do so.


Should we avoid using "file" as the thing to be uploaded and use something more generic like "resource that should be uploaded"?

draft-ietf-httpbis-resumable-upload.md
Acconut: Good idea, thank you.

guoye-zhang (approved):

Issue #2674, opened 2023-11-07
Restrict use of Content-Encoding for upload append


Acconut:
The initial POST request to create the upload resource might contain Content-Encoding, e.g. `Content-Encoding: gzip`. If the upload gets interrupted, it might be resumed using a HEAD and PATCH request. This subsequent PATCH request should then not contain a conflicting Content-Encoding header field, because the server would not be able to know what Content-Encoding has to be used once the upload is complete.

Maybe we should disallow including Content-Encoding at all when appending to an upload. Because the upload might have been interrupted at an arbitrary position, neither the partial content received before the interruption nor the partial content transferred after the interruption are guaranteed to be valid gzip data, for example.

A similar argument could also be made to restrict Content-Language in these PATCH requests.

LPardue:
I don't follow.

The content-encoding is tied to the representation. So an upload using gzip that gets interrupted halfway through, should be able to be patched with the last half of the gzip representation.

Maybe the concern is

1. Uploading with one encoding and resuming with another
2. A server converting the encoding "on the fly" as it receives bytes I.e. gunzipping and writing bytes to storage
3. That the resumed bytes are encoded differently I.e. gzip only half the last bytes in a way that is not compatible and can't be cat'tted to the first half

?

Each of those seems like things we should accomodate and explain but not prohibit. HTTP representations are our friend here.

Acconut:
> Maybe the concern is
> 
>     1. Uploading with one encoding and resuming with another
> 3. That the resumed bytes are encoded differently I.e. gzip only half the last bytes in a way that is not compatible and can't be cat'tted to the first half

Yes, those two are the concerns behind this issue. I am not sure if there are use cases where the client would like to change the encoding when resuming the upload.

I will dig more into the HTTP representations to see how we can accommodate this better in the draft.

MikeBishop:
Even though real-world servers do perform compression on-the-fly, as a matter of HTTP, Content-Encoding is a property of the representation.  If something is uploaded with `Content-Encoding: gzip`, that indicates the entire body, once collected, is a valid gzipped resource.

If a client wanted to gzip the bytes being appended using compression local only to that message, that would be a Transfer-Encoding. However, T-Es have been removed from HTTP/2 and later, so they're not likely to be useful here.

More generally, I think what we want to say is that:

- The initial request to the main resource contains fields which describe the entire resource which the client proposes to upload (length, content-encoding, media type, etc.)
[...]

PR #2746, opened 2024-03-04
Resumable Upload: Add details about Transfer-Encoding and Content-Encoding

### LINKED ISSUE #2679: Speak about considerations of byte counting and transfer-encoding
Ted Hardie asked:

> How does the upload offset calculation interact with transfer encoding/compression? I'm reporting how many bytes came through the network, is that right?

We probably would benefit from some consideration text to mention how HTTP/1.1 transfer-encoding might be a footgun if the server counted those bytes naiively.

Acconut:
With this PR, I attempt to clarify the role of content and transfer encoding in resumable upload that is:
- the bytes for offsets are counted on a representation, where Content-Encoding might be applied, but not Transfer-Encoding because that is a property of the transfer and not the representation itself.
- an upload retain the same content coding throughout its lifecycle, meaning that requests to resume the upload should target the same representation with the same content coding. Otherwise, data might end up invalid on the server if the content coding is switched during the upload. 

Closes #2679 and #2674.


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -266,7 +268,7 @@ Upload-Offset: 25
...
+File metadata can affect how servers might act on the uploaded file. Clients can send representation metadata (see {{Section 8.3 of HTTP}}) in the request that starts an upload. Servers MAY interpret this metadata or MAY ignore it. The `Content-Type` header field ({{Section 8.3 of HTTP}}) can be used to indicate the MIME type of the file. The content coding of the representation is specified using the `Content-Encoding` header field and is retained throughout the entire upload. When resuming an interrupted upload, the same content coding is used for appending to the upload, producing a representation of the upload resource with one consistent content coding. The `Content-Disposition` header field ({{!RFC6266}}) can be used to transmit a filename; if included, the parameters SHOULD be either `filename`, `filename*` or `boundary`.
```
smatsson, line 271: Minor nit, but is "coding" and "encoding" the same thing? Google doesn't give a straight answer and English is not my first language.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -423,6 +425,10 @@ The `Upload-Complete` header field MUST only be used if support by the resource
...
 
+# Transfer and Content Codings
+
+A message might have a content coding, indicated by the `Content-Encoding` header field, and/or a transfer coding, indicated by the `Transfer-Encoding` header field ({{Section 6.1 of RFC9112}}), applied, which modify the representation of uploaded data in a message. For correct interoperability, the client and server must share the same logic when counting bytes for the upload offset. From the client's perspective, the offset is counted after content coding but before transfer coding is applied. From the server's perspective, the offset is counted after the content's transfer coding is reversed but before the content coding is reversed.
```
smatsson, line 430: Good addition! :) Same with coding vs encoding here

draft-ietf-httpbis-resumable-upload.md
Acconut: I took the language from [RFC 9110](https://httpwg.org/specs/rfc9110.html#field.content-encoding):
> The "Content-Encoding" header field indicates what content codings have been applied to the representation

So Content-Encoding is the header field name and its value identifies the used content codings, if I understood it correctly.

draft-ietf-httpbis-resumable-upload.md
Acconut: Same as above, but from RFC 9112: https://datatracker.ietf.org/doc/html/rfc9112#name-transfer-codings

draft-ietf-httpbis-resumable-upload.md
smatsson: Thanks for looking it up! Seems reasonable. Ignore my minor nit :)

guoye-zhang (approved):

PR #2750, opened 2024-03-07
Remove consideration of same-site redirect chain

### LINKED ISSUE #2104: Same-Site cookies and redirects
We noticed a difference between Firefox & Chrome’s behavior for same-site cookies.
Specifically, if a web page is requesting a resource whose final redirect target is same-site with the web page, [Chrome will treat it as same-site (crbug 1221316)](https://bugs.chromium.org/p/chromium/issues/detail?id=1221316). Firefox follows what was originally decided in https://github.com/httpwg/http-extensions/pull/1348 and specified in recent [drafts of rfc6265bis](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-10#section-5.2) i.e., Firefox looks at the whole redirect chain.

For context: Firefox shipped SameSite=Lax by default this January and had to backpedal due to this (and some other webcompat issues). Firefox does not ship SameSite=Lax as the default setting for new cookies as of now.
While we’re still working on our own metrics, we’ve noticed that Chrome is seeing about 1% of page loads potentially affected, which seems prohibitively high.  At the same time, we would prefer not to weaken our implementation of explicit SameSite cookies, given that the redirect check was specifically added to prevent CSRF abuse scenarios.

With these things in mind, we’ve been exploring the idea that we might treat cookies differently depending on whether the cookie was set to lax by default or is an explicit samesite cookie.

@miketaylr, @sbingler: it would be nice to hear some thoughts about this idea and also @johnwilander how other browsers may have implemented #1348.

CC: @miketaylr, @sbingler, @dveditz, @annevk, @evilpie, @johnwilander

sbingler:
Closes #2104

This PR removes the requirement of checking the request's redirect chain during the computation of same-site-ness.

This is being done because RFC6265bis is blocked by this work but we have yet to find a way to implement it in a web compatible way. In the interest of moving RFC6265bis forward the requirement is being removed.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -2690,6 +2690,7 @@ The "Cookie Attribute Registry" should be created with the registrations below:
 ## draft-ietf-httpbis-rfc6265bis-14
 
 * Remove consideration of same-site redirect chain
+  <https://github.com/httpwg/http-extensions/pull/2750>
 
 
 # Acknowledgements
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -17,7 +17,7 @@ venue:
   type: Working Group
   home: https://httpwg.org/
   mail: ietf-http-wg@w3.org
-  arch: https://lists.w3.org/Archives/Public/ietf-http-wg/
+  arch: https://lists.w3.org/Archives/Publi5.7.1. The Cookie Header Fieldc/ietf-http-wg/
   repo: https://github.com/httpwg/http-extensions/labels/6265bis
 github-issue-label: 6265bis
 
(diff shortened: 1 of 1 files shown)


miketaylr (approved):
Not an editor, but LGTM (given that it reflects reality).

mikewest (commented):
Huh. I guess this obviates the comment I made in the previous PR.

It surprises me we didn't implement this. I though we did. :/

If it's the case that we didn't, and that no one else did, then yeah, removing LGTM.

miketaylr:
> It surprises me we didn't implement this. I though we did. :/

Yeah, we did implement this - but we had to back it out (and re-spin stable, IIRC?) because it broke too many sites. :(

sbingler:
Correct, Chrome had to disable the change. Firefox had a similar experience.

mikewest (approved):
LGTM, given the context.

PR #2752, opened 2024-03-07
Correction of URLPattern URL

MikeBishop:
A very ironic URL to be incorrect, but the anchor as given doesn't exist in the document.


pmeenan (approved):
LGTM, thanks. The anchor urls within URLPattern have broken a few times which worries me a bit but at least the spec itself is unlikely to move.

reschke:
This:

  https://whatwg.org/working-mode#anchors

may be helpful.

PR #2753, opened 2024-03-07
Refactor cookie header text

### LINKED ISSUE #2541: Multiple Cookie: headers vs RFC 9113
In the original 6265 as well as in the latest 6265bis draft, this paragraph exists:

> When the user agent generates an HTTP request, the user agent MUST NOT attach more than one Cookie header field.

This is directly contradicted by RFC 7540 and then [RFC 9113 section 8.2.3](https://www.rfc-editor.org/rfc/rfc9113.html#name-compressing-the-cookie-head):

> To allow for better compression efficiency, the Cookie header field MAY be split into separate header fields, each with one or more cookie-pairs

I personally suspect that trying to split the header up into multiples is a recipe for disaster and interop problems even if done only over HTTP/2, but maybe I'm just a pessimist.

Would it be sensible to maybe address this contradiction (better) in the bis document?

sbingler:
Fixes #2541

Technologies such as HTTP/2 will split the `cookie` header into multiple fields. This PR modifies the spec to no longer expressly forbid this and instructs servers to be aware and capable of handling multiple incoming `cookie` headers.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -862,7 +862,7 @@ cookie        = cookie-string
 cookie-string = cookie-pair *( ";" SP cookie-pair )
 ~~~
 
-Servers SHOULD be tolerant to potentially receiving multiple cookie
+Servers MUST be tolerant to potentially receiving multiple cookie
 headers. For example, if an HTTP/2 {{RFC9113}} or HTTP/3 {{RFC9114}} connection
 splits a cookie header to improve compression.
 
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -862,9 +862,9 @@ cookie        = cookie-string
 cookie-string = cookie-pair *( ";" SP cookie-pair )
 ~~~
 
-Servers MUST be tolerant to potentially receiving multiple cookie
-headers. For example, if an HTTP/2 {{RFC9113}} or HTTP/3 {{RFC9114}} connection
-splits a cookie header to improve compression.
+Servers MUST be tolerant of multiple cookie headers. For example, an HTTP/2
+{{RFC9113}} or HTTP/3 {{RFC9114}} connection might split a cookie header to
+improve compression.
 
 ### Semantics
 
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -2695,6 +2695,11 @@ The "Cookie Attribute Registry" should be created with the registrations below:
 * Use navigables concept
   <https://github.com/httpwg/http-extensions/pull/2478>
 
+## draft-ietf-httpbis-rfc6265bis-14
+
+* Refactor cookie header text
+  <https://github.com/httpwg/http-extensions/pull/2753>
+
 
 # Acknowledgements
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -860,6 +862,10 @@ cookie        = cookie-string
 cookie-string = cookie-pair *( ";" SP cookie-pair )
 ~~~
 
+Servers MUST be tolerant to potentially receiving multiple cookie
```
sbingler, line 865: "to" or "of"?

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1821,18 +1827,16 @@ are defined below depending on the type of retrieval.
...
 A user agent MAY omit the Cookie header field in its entirety.  For example, the
 user agent might wish to block sending cookies during "third-party" requests
 from setting cookies (see {{third-party-cookies}}).
 
 If the user agent does attach a Cookie header field to an HTTP request, the
-user agent MUST compute the cookie-string following the algorithm defined in
-{{retrieval-algorithm}}, where the retrieval's URI is the request-uri, the
-retrieval's same-site status is computed for the HTTP request as defined in
-{{same-site-requests}}, and the retrieval's type is "HTTP".
+user agent MUST generate a single cookie-string and the user agent MUST compute
```
sbingler, line 1835:
```
user agent MUST generate a single cookie-string and the user agent MUST compute
```
My intention here is that the cookie code within the UA is still only trying to generate a single string (to maximize compatibility), but once that string is generated it may be split by by another layer of the networking stack.

MikeBishop: I agree with that. Clients MUST produce a single value. Servers MUST tolerate them being split, either by actually addressing that case or by using lower-level components that guarantee they'll be properly recombined before it sees the header value.

draft-ietf-httpbis-rfc6265bis.md
mikewest: I'd write "of". And I'd probably drop "potentially receiving", since it seems clear from context.

That said, there's nothing wrong with this, just personal preference.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -860,6 +862,10 @@ cookie        = cookie-string
 cookie-string = cookie-pair *( ";" SP cookie-pair )
 ~~~
 
+Servers MUST be tolerant to potentially receiving multiple cookie
+headers. For example, if an HTTP/2 {{RFC9113}} or HTTP/3 {{RFC9114}} connection
+splits a cookie header to improve compression.
```
mikewest, line 867: Nit: The following reads slightly better to me.

```suggestion
headers. For example, an HTTP/2 {{RFC9113}} or HTTP/3 {{RFC9114}} connection might
split a cookie header to improve compression.
```

mikewest (approved):
LGTM % tiny nits.

draft-ietf-httpbis-rfc6265bis.md
sbingler: SG

draft-ietf-httpbis-rfc6265bis.md
sbingler: Done

PR #2758, opened 2024-03-12
Add additional developer warnings for SameSite cookies

### LINKED ISSUE #2644: [rfc6265bis] SameSite=Strict cookie isolation on cross-site windows
We are reporting on the behavior of `samesite=strict` cookies on cross-site windows opened through `window.open` that goes against our assumptions (implied by the current specification) on when and how `samesite=strict` cookies should be isolated. Breaking these assumptions can lead to security issues, allowing the bypass of `samesite=strict` cookies as a defense mechanism against CSRF and XS-Leak attacks.

## 1. Overview

The following table shows  when `samesite=strict` cookies are attached to requests associated with `iframe`s and pages opened through `window.open` (where :x: stands for *not attached*/*not accessible* and :white_check_mark: for *attached*/*accessible*):

| Action/Element        | iframe | window.open        | 
| --------------------- | ------ | ------------------ |
| page load             | :x:    | :x:                |
| document.cookie       | :x:    | :white_check_mark: |
| subresource load      | :x:    | :white_check_mark: |
| user-initiated reload | :x:    | :x:                |
| JavaScript reload     | :x:    | :white_check_mark: |

The behavior of `samesite=strict` cookies regarding page reloads for pages opened through `window.open` from a cross-site position is evident from the [specification](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-12#section-8.8.5):

> **Requests issued for reloads triggered through user interface elements** (such as a refresh button on a toolbar) **are same-site only if the reloaded document was originally navigated to via a same-site request**. This **differs from the handling of other reload navigations, which are always same-site if top-level**, since the source navigable's active document is precisely the document being reloaded.
> This special handling of reloads triggered through a user interface element **avoids sending `SameSite` cookies on user-initiated reloads if they were withheld on the original navigation** (i.e., if the initial navigation were cross-site). 

Currently, the specification mandates blocking cookies on user-initiated page reloads but not on reloads performed through JavaScript. We think that `samesite=strict` cookies should also be blocked in this scenario, as this interaction between JavaScript navigations and `samesite=strict` cookies allows attackers to bypass the restrictions imposed by `samesite=strict` (see 2.1). Additionally, we verified that `samesite=strict` cookies, even when not sent in the top-level request that loads a page, are still attached to requests that load subresources for that page. This behavior is not explicitly defined in the specification. However, its counterpart (setting a cookie) is defined as follows:

> The "SameSite" attribute affects cookie creation as well as delivery. Cookies which assert "samesite=lax" or "samesite=strict" cannot be set in responses to cross-site subresource requests, or cross-site nested navigations.

Although these requests are not cross-site, they are performed implicitly by a page navigated from a cross-site context, effectively enabling an attacker to perform authenticated cross-site requests (see 2.1), or cookie tossing attacks.

> **If the reload navigation were instead considered same-site, and sent all the initially withheld `SameSite` cookies, the security benefits of withholding the cookies in the first place would be nullified**

Following this argument, when considering that user-initiated page reloads are deemed cross-site and therefore do not attach `samesite=strict` cookies, it seems reasonable to assume both JavaScript navigations and subresource loads should also not attach `samesite=strict` cookies, as they are also transparent to the user.

### 2.1 samesite=strict cookies bypass and CSRF

It is possible to abuse the previously described behavior to bypass CSRF protections An attacker can forge same-site requests starting from a cross-site position by abusing gadgets that result in secondary same-site requests. An example of one of these gadgets is a JavaScript-based redirect that performs a redirection using attacker-controlled data. Here is a possible attack flow:

1. Victim visits https://good.com/set, setting the cookie `foo=bar; samesite=strict` cookie for domain `good.com`.
2. Victim then visits https://attacker.com/, which executes `window.open("https://good.com/redir?q=/")`
	1. The request to https://good.com/redir?q=/ is cross-site, so it does not attach `samesite=strict` cookies
3. https://good.com/redir?q=/ performs a navigation to the value of the URL parameter `q` through `document.location.href = '${parsedURL.query['q']}'`
	1. This triggers a request to https://good.com/, which attaches `samesite=strict` cookies, allowing an attacker to make authenticated requests from a cross-site position

A similar scenario was discussed in this [post](https://portswigger.net/web-security/csrf/bypassing-samesite-restrictions#bypassing-samesite-restrictions-using-on-site-gadgets).

Besides JavaScript redirections, other requests, such as subresource loads, could also be exploited to subvert the security assumptions granted by `samesite=strict` cookies.

### 2.2 XS-Leaks 

`samesite=lax` cookies effectively defend against certain classes of XS-Leaks attacks. However, `window.open`-based attacks are still possible since `samesite=lax`cookies are attached to top-level cross-site requests. Since `samesite=strict` cookies are not attached to top-level cross-site requests,  they have been deemed an effective mitigation against `window.open`-based XS-Leaks. 

As discussed in 2.1, gadgets that allow attackers to forge same-site requests from a cross-site position can be leveraged for XS-Leaks. An example would be a page that performs a stateful subresource load, which modifies the page DOM. If an attacker opens this page through `window.open`, `samesite=strict` cookies will not be attached, but following subresource loads in the new page will attach these cookies, leading to state-dependent changes to the DOM, which could be measured by the attacker (through frame counting for example).

Best regards

sbingler:
Closes #2644

Adds some additional notes and warnings to developers on how to more carefully use SameSite cookies.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -2714,6 +2714,9 @@ The "Cookie Attribute Registry" should be created with the registrations below:
 * Refactor cookie header text
   <https://github.com/httpwg/http-extensions/pull/2753>
 
+* Add additional developer warnings for SameSite cookies
+  <https://github.com/httpwg/http-extensions/pull/2758>
+
 
 # Acknowledgements
 {:numbered="false"}
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -2276,7 +2276,16 @@ security properties required by applications.
 strict mode, and when supported by the client. It is, however, prudent to ensure
 that this designation is not the extent of a site's defense against CSRF, as
 same-site navigations and submissions can certainly be executed in conjunction
-with other attack vectors such as cross-site scripting.
+with other attack vectors such as cross-site scripting or abuse of page
+redirections.
+
+Understanding how and when a request is considered same-site is also important
+in order to properly design a site for SameSite cookies. For example, if a
+top-level request is made to a sensitive page that request will be considered
+cross-site and SameSite cookies won’t be sent; that page’s sub-resources
+requests, however, are same-site and would receive SameSite cookies. Sites can
+avoid inadvertently allowing access to these sub-resources by returning an error
+for the initial page request if it doesn’t include the appropriate cookies.
 
 Developers are strongly encouraged to deploy the usual server-side defenses
 (CSRF tokens, ensuring that "safe" HTTP methods are idempotent, etc) to mitigate
@@ -2365,6 +2374,11 @@ cookie would defeat the purpose of withholding it in the first place, as the
 reload navigation triggered through the user interface may replay the original
 (potentially malicious) request. Thus, the reload request should be considered
 cross-site, like the request that initially navigated to the page.
+
+Because requests issued for, non-user initiated, reloads attach all SameSite
+cookies, developers should be careful and thoughtful about when to initiate a
+reload in order to avoid a CSRF attack. For example, the page could only
+initiate a reload if a CSRF token is present on the initial request.
 
 ### Top-level requests with "unsafe" methods {#unsafe-top-level-requests}
 
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -2720,7 +2720,6 @@ The "Cookie Attribute Registry" should be created with the registrations below:
 * Add additional developer warnings for SameSite cookies
   <https://github.com/httpwg/http-extensions/pull/2758>
 
-
 # Acknowledgements
 {:numbered="false"}
 RFC 6265 was written by Adam Barth. This document is an update of RFC 6265,


draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -2276,7 +2276,16 @@ security properties required by applications.
 strict mode, and when supported by the client. It is, however, prudent to ensure
 that this designation is not the extent of a site's defense against CSRF, as
 same-site navigations and submissions can certainly be executed in conjunction
-with other attack vectors such as cross-site scripting.
+with other attack vectors such as cross-site scripting or abuse of page
+redirections.
```
mikewest, line 2280: Is the redirection part accurate? My recollection is that we treated A->B->A navigation as cross-site as a mitigation for exactly this risk.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -2276,7 +2276,16 @@ security properties required by applications.
...
+
+Understanding how and when a request is considered same-site is also important
+in order to properly design a site for SameSite cookies. For example, if a
+top-level request is made to a sensitive page that request will be considered
```
mikewest, line 2284: Nit: There's nuance here, insofar as "top-level request" includes user-initiated navigation from the address bar or right-click menu weirdness.

mikewest (approved):
LGTM % nits.

draft-ietf-httpbis-rfc6265bis.md
sbingler: Of course there is :)

Changing it to

`For example, if a user clicks a link and a top-level request is made....`

Should sidestep that problem, wdyt?

draft-ietf-httpbis-rfc6265bis.md
sbingler: Obsoleted by #2750

The initial issue also [links to an attack](https://portswigger.net/web-security/csrf/bypassing-samesite-restrictions#bypassing-samesite-restrictions-using-on-site-gadgets) that makes use of this.

draft-ietf-httpbis-rfc6265bis.md
mikewest: 👍

mikewest (approved):
Still LGTM.

Issue #2605, opened 2023-07-27
[rfc6265bis] Inconsistent browser behavior with secure and prefix cookies (__Host- and __Secure-) on localhost


Pedro-Bernardo:
We are reporting inconsistent browser behavior regarding `Secure` cookies and `__Host-` and `__Secure-` prefixed cookies on `http://localhost`.

[rfc6265bis-12](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-12) delegates the decision of what is a "secure protocol" to the user agent when describing how to handle cookies with the `Secure` attribute. 
[Section 5.6](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis/#section-5.6-3.13)
> 13. If the scheme component of the request-URI does not denote a "secure" protocol (as defined by the user agent), and the cookie's secure-only-flag is true, then abort these steps and ignore the cookie entirely.

[Section 5.7.3](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-12#name-retrieval-algorithm)
> If the cookie's secure-only-flag is true, then the retrieval's URI's scheme must denote a "secure" protocol (as defined by the user agent).
> 
> NOTE: The notion of a "secure" protocol is not defined by this document. Typically, user agents consider a protocol secure if the protocol makes use of transport-layer security, such as SSL or TLS. For example, most user agents consider "https" to be a scheme that denotes a secure protocol.

The `secure-only-flag` affects both cookies with the `Secure` attribute and cookies with name prefixes (`__Host-` and `__Secure-`). rfc6265bis-12 states the following ([Section 5.6](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-12#name-storage-model)):

> 20. If the cookie-name begins with a case-insensitive match for the string "\_\_Secure-", abort these steps and ignore the cookie entirely unless the cookie's secure-only-flag is true.
>
> 21. If the cookie-name begins with a case-insensitive match for the string "\_\_Host-", abort these steps and ignore the cookie entirely unless the cookie meets all the following criteria:
>     1. The cookie's secure-only-flag is true.
>     2. The cookie's host-only-flag is true.
>     3. The cookie-attribute-list contains an attribute with an attribute-name of "Path", and the cookie's path is `/`.

### Inconsistent Browser Behavior
According to our tests:
[...]

PR #2759, opened 2024-03-12
Support potentially trustworthy origins

sbingler:
Rephrases the requirements for the `Secure` attribute to be slightly broader in order to support [potentially trustworthy origin](https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-origin).

Currently the requirements specifically ask for a secure protocol, with this change they'll ask for a secure connection (as defined by the user agent). This allows UAs to support `Secure` cookie access on http://localhost


mikewest (approved):
This is a good change, thanks.

I don't think we have WPT in place for this (and I'm not entirely sure that we can?), but have you looked into the support for this change across browser engines? I'm generally in favor of landing it as long as it's reflecting (near) reality.

sbingler:
No WPT that I can see.

According to #2605, Firefox fully supports localhost, Chrome supports `Secure` but not cookie prefixes, and Safari doesn't have support but does have an open bug.

PR #2761, opened 2024-03-16
Remove recursive references

LPardue:
There were a few cases of the section linking to itself, which I find distracting, so lets remove them.


guoye-zhang (approved):

Acconut (approved):
Good catch, thank you!

PR #2769, opened 2024-04-03
Specify the compression window sizes for zstd-d and br-d

### LINKED ISSUE #2754: Window Size for Shared Dictionary Content Encodings
Context: https://datatracker.ietf.org/doc/draft-jaju-httpbis-zstd-window-size/ aims to clarify the window size limit for the `zstd` Content Encoding token as it wasn't clear in RFC8878.

Since the Compression Dictionary Transport draft hasn't been published yet, we should also add some window size considerations to it for `br-d` and `zstd-d` to maintain interoperability.
I think in Chrome we ended up using [max(dictionary size, 8 MB)](https://source.chromium.org/chromium/chromium/src/+/main:net/filter/zstd_source_stream.cc;l=49-57;drc=f88f2c49794bb22673f9048f7ee0d0e8279582c7).

@pmeenan

pmeenan:
This change makes it explicit that the `zstd-d` content encoding is `Zstandard with up to a 8 MB compression window` (same as `zstd`) and `br-d` is `Brotli with up to a 16 MB compression window` (same as `br`).

See the discussion in #2754 for why larger ZStandard values were not selected.

This carries some nuance for what encoding to use when compressing for maximum benefit as the resource size crosses beyond 8 MB but that is more of an editorial discussion about the individual compression algorithms and probably better laid out in blog posts or explainers than in the spec.

Fix #2754

### DIFF
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -300,12 +300,12 @@ negotiating content encoding in HTTP.
 
 This document introduces two new content encoding algorithms:
 
-|------------------|----------------------------------------------------|
-| Content-Encoding | Description                                        |
-|------------------|----------------------------------------------------|
-| br-d             | Brotli using an external compression dictionary    |
-| zstd-d           | Zstandard using an external compression dictionary |
-|------------------|----------------------------------------------------|
+|------------------|--------------------------------------------------------------------------------------------|
+| Content-Encoding | Description                                                                                |
+|------------------|--------------------------------------------------------------------------------------------|
+| br-d             | Brotli using an external compression dictionary and a compression window of up to 16 MB.   |
+| zstd-d           | Zstandard using an external compression dictionary and a compression window of up to 8 MB. |
+|------------------|--------------------------------------------------------------------------------------------|
 
 The dictionary to use is negotiated separately and advertised in the
 "Available-Dictionary" request header.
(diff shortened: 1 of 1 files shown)


pmeenan:
@nidhijaju could you PTAL and make sure the language matches what you're working on for the `zstd` update?

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -345,12 +345,12 @@ Vary: accept-encoding, available-dictionary
...
-| Name   | Description                                                                           | Reference   |
-|--------|---------------------------------------------------------------------------------------|-------------|
-| br-d   | A stream of bytes compressed using the Brotli protocol with an external dictionary    | {{RFC7932}} |
-| zstd-d | A stream of bytes compressed using the Zstandard protocol with an external dictionary | {{RFC8878}} |
-|--------|---------------------------------------------------------------------------------------|-------------|
+|--------|--------------------------------------------------------------------------------------------------------------------------------|-------------|
+| Name   | Description                                                                                                                    | Reference   |
+|--------|--------------------------------------------------------------------------------------------------------------------------------|-------------|
+| br-d   | A stream of bytes compressed using the Brotli protocol with an external dictionary and a compression window of up to 16 MB.    | {{RFC7932}} |
```
martinthomson, line 351: It might be better to avoid tables for this now.

(As a point on style, I find that if you need to define something like this, it is better to add an entire (short) section that includes citations and whatnot.)

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -300,12 +300,12 @@ negotiating content encoding in HTTP.
...
-|------------------|----------------------------------------------------|
-| br-d             | Brotli using an external compression dictionary    |
-| zstd-d           | Zstandard using an external compression dictionary |
-|------------------|----------------------------------------------------|
+|------------------|--------------------------------------------------------------------------------------------|
+| Content-Encoding | Description                                                                                |
+|------------------|--------------------------------------------------------------------------------------------|
+| br-d             | Brotli using an external compression dictionary and a compression window of up to 16 MB.   |
+| zstd-d           | Zstandard using an external compression dictionary and a compression window of up to 8 MB. |
+|------------------|--------------------------------------------------------------------------------------------|
```
martinthomson, line 308: A list might be easier to consume here.  This is also the place where you should include citations.  (though you might also include those citations at the first place that Brotli and Zstandard are mentioned in the intro.

martinthomson (approved):

pmeenan:
Thanks.  I switched the content-encodings definitions and IANA registrations to a list with references back to the algorithm RFCs.  For the IANA registrations I removed the window size from the text and linked it to this document (where the window sizes are defined) instead of to the compression algorithms directly.

Let me know if a more verbose description of the need to specify the window sizes would be useful or if just being explicit about what they are is enough.

martinthomson (approved):

nidhijaju (approved):
lgtm other than the comment

PR #2773, opened 2024-04-08
Fixed the URLPattern citations.

### LINKED ISSUE #2757: Citation format
For example: 

> provides the URLPattern to use for request matching (https://urlpattern.spec.whatwg.org/).

Usually, we add something like URLPattern to the references, then `[CITE]` that.

pmeenan:
Changed it to a proper normative reference.

For #2757

### DIFF
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -41,6 +41,10 @@ normative:
   HTTP: RFC9110
   HTTP-CACHING: RFC9111
   RFC5861: # Stale-While-Revalidate
+  URLPattern:
+    title: URL Pattern Standard
+    date: 18 March 2024
+    target: https://urlpattern.spec.whatwg.org/
 
 informative:
   Origin: RFC6454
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -99,20 +103,19 @@ and "type".
 ### match
 
 The "match" value of the Use-As-Dictionary header is a String value that
-provides the URLPattern to use for request matching
-(https://urlpattern.spec.whatwg.org/).
+provides the {{URLPattern}} to use for request matching.
```
martinthomson, line 106: "[URLPattern]" is not a word, it's a citation.  You want:

```suggestion
provides the URL Pattern {{URLPattern}} to use for request matching.
```

pmeenan: Thanks. I cleaned that up and changed the reference in the CORS section to point to the matching section of the doc instead of directly to URLPattern.

I'm assuming it is cleaner to not cite every reference to URL Pattern in the doc and doing it in the sections where it is directly used is the way to go but let me know if that's not the case and I can add the citations to the 2 `URL Pattern` references that I removed it from.

pmeenan:
Merging this. Can always clean it up more if it is still a concern.

PR #2777, opened 2024-04-15
[connect-tcp] Note the dependency on Extended CONNECT

### LINKED ISSUE #2774: No reference for "extended CONNECT", or requirement for its setting
The CONNECT-TCP draft makes informative references to other drafts which use "extended CONNECT" (CONNECT-UDP, CONNECT-IP), but does not have a normative reference to the draft which defines this use of the CONNECT method. It references the "extended CONNECT" method without supplying a reference for this term.

This draft spells out the required values for each pseudo-header, but omits the requirement to set SETTINGS_ENABLE_CONNECT_PROTOCOL. A reference to RFCs 8441/9220 is probably needed, and a reminder that the method in this document can only be used if the server sends this setting first.

bemasc:
Fixes #2774

### DIFF
--- a/draft-ietf-httpbis-connect-tcp.md
+++ b/draft-ietf-httpbis-connect-tcp.md
@@ -106,7 +106,7 @@ Upgrade: connect-tcp
 
 ## In HTTP/2 and HTTP/3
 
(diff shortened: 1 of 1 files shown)


MikeBishop (approved):
That'll do.

PR #2780, opened 2024-04-17
[compression-dictionary] Allow for larger ZStandard compression windows.

### LINKED ISSUE #2754: Window Size for Shared Dictionary Content Encodings

pmeenan:
This changes the maximum ZStandard compression window to be the larger of 8 MB and the size of the dictionary. This allows for clients to limit the maximum window size to as low as 8 MB if they like (by not advertising `zstd-d` support) while also allowing for clients to use larger dictionaries with ZStandard than they would be able to if the global limit was 8 MB.

Fix #2754

### DIFF
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -305,8 +305,45 @@ The dictionary to use is negotiated separately and advertised in the
 ## Compression Algorithms
 This document introduces two new content encoding algorithms:
 
-- br-d: Brotli {{RFC7932}} using an external compression dictionary and a compression window of not more than 16 MB.
-- zstd-d: Zstandard {{RFC8878}} using an external compression dictionary and a compression window of not more than 8 MB.
+- br-d: Brotli {{RFC7932}} using an external compression dictionary.
+- zstd-d: Zstandard {{RFC8878}} using an external compression dictionary.
+
+The client and server need to agree on a maximum size to allow for the
+compression window before a request is made so that the server can be sure that
+the client will be able to decompress the response.
+
+Dictionary-based compression benefits from having the full dictionary available
+whenever possible, particularly for cases where a previous version of a
+resource is being used as a dictionary for the newer version (delta
+compression).
+
+### Brotli Compression Window
+
+Brotli {{RFC7932}} makes the full dictionary available independent of the
+compression window so the br-d content encoding allows for the same 16 MB
+maximum compression window as the br content encoding.
+
+Clients that announce support for br-d content encoding MUST be able to
+decompress resources that were compressed with a window size of up to 16 MB.
+
+### ZStandard Compression Window
+
+ZStandard {{RFC8878}} makes the full dictionary available as long as the input
+is smaller than the compression window. Once the input grows beyond the
+compression window the dictionary becomes unavailable.
+
+In order to allow for large resources to be dictionary-compressed while still
+giving clients control over how much memory they will consume, the zstd-d
+content encoding allows for a compression window of up to the larger of 8 MB
+and the size of the dictionary.
+
+For dictionaries that exceed the amount of memory that a client is willing to
+allocate for decompression, the client can choose to not announce support for
+zstd-d content encoding.
+
+Clients that announce support for br-d content encoding MUST be able to
+decompress resources that were compressed with a window size of up to the
+larger of 8 MB and the size of the dictionary.
 
 ## Accept-Encoding
 
@@ -340,17 +377,22 @@ Vary: accept-encoding, available-dictionary
 
 ## Content Encoding
 
-IANA is asked to enter the following into the "HTTP Content Coding Registry" registry ({{HTTP}}):
+IANA is asked to enter the following into the "HTTP Content Coding Registry"
+registry ({{HTTP}}):
 
 - Name: br-d
-- Description: A stream of bytes compressed using the Brotli protocol with an external dictionary of not more than 16 MB.
+- Description: A stream of bytes compressed using the Brotli protocol with an
+external dictionary of not more than 16 MB.
 - Reference: This document
 - Notes: {{compression-algorithms}}
 
-IANA is asked to enter the following into the "HTTP Content Coding Registry" registry ({{HTTP}}):
+IANA is asked to enter the following into the "HTTP Content Coding Registry"
+registry ({{HTTP}}):
 
 - Name: zstd-d
-- Description: A stream of bytes compressed using the Zstandard protocol with an external dictionary of not more than 8 MB.
+- Description: A stream of bytes compressed using the Zstandard protocol with
+an external dictionary of not more than the larger of 8 MB and the size of the
+dictionary.
 - Reference: This document
 - Notes: {{compression-algorithms}}
 
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -317,13 +317,15 @@ The "zstd-d" content encoding identifies a resource that is a
 "Dictionary-Compressed Zstandard" stream.
 
 A "Dictionary-Compressed Zstandard" stream is a Zstandard {{RFC8878}} stream
-for a response that has been compressed with an external dictionary using a
-compression window not larger than the larger of 8 MB and 1.25 times the size
-of the dictionary, not to exceed 128 MB.
+for a response that has been compressed with an external dictionary.
 
 Clients that announce support for zstd-d content encoding MUST be able to
-decompress resources that were compressed with a window size of up to the
-larger of 8 MB and 1.25 times the size of the dictionary, not to exceed 128 MB.
+decompress resources that were compressed with a window size of at least 8 MB
+or 1.25 times the size of the dictionary, which ever is greater, up to a
+maximum of 128 MB.
+
+If 1.25 times the size of the dictionary results in a fractional byte window
+size, the window size should be rounded down to the nearest byte.
 
 With Zstandard compression, the full dictionary is available during compression
 and decompression until the size of the input exceeds the compression window.
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -324,8 +324,9 @@ decompress resources that were compressed with a window size of at least 8 MB
 or 1.25 times the size of the dictionary, which ever is greater, up to a
 maximum of 128 MB.
 
-If 1.25 times the size of the dictionary results in a fractional byte window
-size, the window size should be rounded down to the nearest byte.
+The window size used will be encoded in the content (currently, this can be expressed
+in powers of two only) and it MUST be lower than this limit. An implementation MAY
+treat a window size that exceeds the limit as a decoding error.
 
 With Zstandard compression, the full dictionary is available during compression
 and decompression until the size of the input exceeds the compression window.
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -326,9 +326,9 @@ maximum compression window as the br content encoding.
 Clients that announce support for br-d content encoding MUST be able to
 decompress resources that were compressed with a window size of up to 16 MB.
 
-### ZStandard Compression Window
+### Zstandard Compression Window
 
-ZStandard {{RFC8878}} makes the full dictionary available as long as the input
+Zstandard {{RFC8878}} makes the full dictionary available as long as the input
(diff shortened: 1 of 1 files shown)


pmeenan:
@felixhandte and @nidhijaju could you also take a look and see if the language looks ok?

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -340,17 +377,22 @@ Vary: accept-encoding, available-dictionary
...
 
-IANA is asked to enter the following into the "HTTP Content Coding Registry" registry ({{HTTP}}):
+IANA is asked to enter the following into the "HTTP Content Coding Registry"
+registry ({{HTTP}}):
 
 - Name: zstd-d
-- Description: A stream of bytes compressed using the Zstandard protocol with an external dictionary of not more than 8 MB.
+- Description: A stream of bytes compressed using the Zstandard protocol with
+an external dictionary of not more than the larger of 8 MB and the size of the
+dictionary.
```
felixhandte, line 395: If we want to keep the discussion of the window size here, I would propose phrasing this as:

> A stream of bytes compressed using the Zstandard protocol using a negotiated external dictionary. Unlike the `zstd` Content Coding, the Zstandard frame(s) making up the stream may use a `Window_Size` up to and including the size of the external dictionary or 8 MB, whichever is larger.

This makes it clear that the constraint is on the window, and not the size of the dictionary: "an external dictionary of not more than [...] the size of the dictionary" is tautological.

If we take @Cyan4973's suggestion in [this comment](https://github.com/httpwg/http-extensions/issues/2754#issuecomment-2061998009) (and I agree wiggle room would be nice to have here) I guess we'll have to complicate this a little further.

But since you are actually specifying this above, maybe it doesn't even need to be mentioned here.

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -305,8 +305,45 @@ The dictionary to use is negotiated separately and advertised in the
...
+In order to allow for large resources to be dictionary-compressed while still
+giving clients control over how much memory they will consume, the zstd-d
+content encoding allows for a compression window of up to the larger of 8 MB
+and the size of the dictionary.
+
+For dictionaries that exceed the amount of memory that a client is willing to
+allocate for decompression, the client can choose to not announce support for
+zstd-d content encoding.
+
+Clients that announce support for br-d content encoding MUST be able to
```
felixhandte, line 344: nit: `br-d` -> `zstd-d`.

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -305,8 +305,45 @@ The dictionary to use is negotiated separately and advertised in the
...
+### Brotli Compression Window
+
+Brotli {{RFC7932}} makes the full dictionary available independent of the
+compression window so the br-d content encoding allows for the same 16 MB
+maximum compression window as the br content encoding.
+
+Clients that announce support for br-d content encoding MUST be able to
+decompress resources that were compressed with a window size of up to 16 MB.
+
+### ZStandard Compression Window
```
felixhandte, line 329: nit: here and below, the capitalization is "Zstandard" rather than "ZStandard".

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -340,17 +377,22 @@ Vary: accept-encoding, available-dictionary
...
 ## Content Encoding
 
-IANA is asked to enter the following into the "HTTP Content Coding Registry" registry ({{HTTP}}):
+IANA is asked to enter the following into the "HTTP Content Coding Registry"
+registry ({{HTTP}}):
 
 - Name: br-d
-- Description: A stream of bytes compressed using the Brotli protocol with an external dictionary of not more than 16 MB.
+- Description: A stream of bytes compressed using the Brotli protocol with an
+external dictionary of not more than 16 MB.
```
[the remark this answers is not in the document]
pmeenan: Whoops. yes, thanks. Copy/paste problem where I accidentally cropped out too much of the description last time.

draft-ietf-httpbis-compression-dictionary.md
pmeenan: The main reason to mention it here is that this chunk of text is what is going to be in the IANA registry for the content encodings.  I fixed the language around it being the window size and not the dictionary, thanks.

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -340,17 +377,22 @@ Vary: accept-encoding, available-dictionary
...
 ## Content Encoding
 
-IANA is asked to enter the following into the "HTTP Content Coding Registry" registry ({{HTTP}}):
+IANA is asked to enter the following into the "HTTP Content Coding Registry"
+registry ({{HTTP}}):
 
 - Name: br-d
-- Description: A stream of bytes compressed using the Brotli protocol with an external dictionary of not more than 16 MB.
+- Description: A stream of bytes compressed using the Brotli protocol with an
+external dictionary and a compression window of not more than 16 MB.
```
martinthomson, line 385: Now that the window size is more complex, I would drop this additional text.  Instead, reference the definition of the window size above.  (I would structure this differently and have sections for each algorithm, that includes a discussion of window size, rather than sections for the window size.)

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -305,8 +305,45 @@ The dictionary to use is negotiated separately and advertised in the
...
+compression).
+
+### Brotli Compression Window
+
+Brotli {{RFC7932}} makes the full dictionary available independent of the
+compression window so the br-d content encoding allows for the same 16 MB
+maximum compression window as the br content encoding.
+
+Clients that announce support for br-d content encoding MUST be able to
+decompress resources that were compressed with a window size of up to 16 MB.
```
nidhijaju, line 327:
```suggestion
decompress resources that were compressed with a window size of not more than 16 MB.
```

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -305,8 +305,45 @@ The dictionary to use is negotiated separately and advertised in the
...
+
+### Zstandard Compression Window
+
+Zstandard {{RFC8878}} makes the full dictionary available as long as the input
+is smaller than the compression window. Once the input grows beyond the
+compression window the dictionary becomes unavailable.
+
+In order to allow for large resources to be dictionary-compressed while still
+giving clients control over how much memory they will consume, the zstd-d
+content encoding allows for a compression window of up to the larger of 8 MB
```
nidhijaju, line 337:
```suggestion
content encoding allows for a compression window of not more than the larger of 8 MB
```

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -305,8 +305,45 @@ The dictionary to use is negotiated separately and advertised in the
...
+and the size of the dictionary.
+
+For dictionaries that exceed the amount of memory that a client is willing to
+allocate for decompression, the client can choose to not announce support for
+zstd-d content encoding.
+
+Clients that announce support for zstd-d content encoding MUST be able to
+decompress resources that were compressed with a window size of up to the
```
nidhijaju, line 345:
```suggestion
decompress resources that were compressed with a window size of not more than the
```

draft-ietf-httpbis-compression-dictionary.md
nidhijaju: It seems nice to keep the requirement itself in the IANA registry, and have the longer discussion and reasoning up above.

nidhijaju (approved):
lgtm with nits, thank you!

draft-ietf-httpbis-compression-dictionary.md
pmeenan: I changed it to be a named format instead of trying to include all of the requirements and settings directly in the IANA registry.  They're now referenced as `Dictionary-Compressed Brotli` and `Dictionary-Compressed Zstandard`

felixhandte (approved):
LGTM!

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -293,24 +295,57 @@ For example:
...
+A "Dictionary-Compressed Zstandard" stream is a Zstandard {{RFC8878}} stream
+for a response that has been compressed with an external dictionary using a
+compression window not larger than the larger of 8 MB and 1.25 times the size
+of the dictionary, not to exceed 128 MB.
 
-- br-d: Brotli {{RFC7932}} using an external compression dictionary and a compression window of not more than 16 MB.
-- zstd-d: Zstandard {{RFC8878}} using an external compression dictionary and a compression window of not more than 8 MB.
+Clients that announce support for zstd-d content encoding MUST be able to
+decompress resources that were compressed with a window size of up to the
+larger of 8 MB and 1.25 times the size of the dictionary, not to exceed 128 MB.
```
martinthomson, line 326: Do you really need to repeat this?  I'd keep this and drop the earlier version.

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -293,24 +295,57 @@ For example:
...
+
+The "zstd-d" content encoding identifies a resource that is a
+"Dictionary-Compressed Zstandard" stream.
 
-## Compression Algorithms
-This document introduces two new content encoding algorithms:
+A "Dictionary-Compressed Zstandard" stream is a Zstandard {{RFC8878}} stream
+for a response that has been compressed with an external dictionary using a
+compression window not larger than the larger of 8 MB and 1.25 times the size
+of the dictionary, not to exceed 128 MB.
```
martinthomson, line 322:
```suggestion
for a response that has been compressed with an external dictionary.
```

draft-ietf-httpbis-compression-dictionary.md
martinthomson, line 326:
```suggestion
decompress resources that were compressed with a window size of at least 8 MB
or 1.25 times the size of the dictionary, which ever is greater, up to a maximum of 128 MB.
```

martinthomson: Or, with a bonus text.

```suggestion
decompress resources that were compressed with a window size of at least 8 MB
or 1.25 times the size of the dictionary, which ever is greater, up to a maximum of 128 MB.
Clients that are not willing to accept a larger window need to avoid advertising
any dictionary that would require a larger window.
```

What needs to be said about rounding?  12345 * 1.25 = what?  Are we rounding up or down, to a power of two?

pmeenan: On the bonus language, I think it is a bit more subtle because they could advertise a dictionary but only `br-d` if they want to keep the window under 16 MB (and if they support both).

As far as rounding goes, as far as I understand, the window representation in the stream is where the power of two (rounding down) happens so that part of it isn't something we need to spec since accepting the larger non-rounded window is still acceptable, there is just no way to represent it in the stream.

That said, it's worth mentioning rounding on the actual 1.25 times calculation to avoid fractional bytes. I added some language to round it down (effectively what you would get with integer math on most languages).

MikeBishop (approved):

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -293,24 +295,59 @@ For example:
...
 
-- br-d: Brotli {{RFC7932}} using an external compression dictionary and a compression window of not more than 16 MB.
-- zstd-d: Zstandard {{RFC8878}} using an external compression dictionary and a compression window of not more than 8 MB.
+Clients that announce support for zstd-d content encoding MUST be able to
+decompress resources that were compressed with a window size of at least 8 MB
+or 1.25 times the size of the dictionary, which ever is greater, up to a
+maximum of 128 MB.
+
+If 1.25 times the size of the dictionary results in a fractional byte window
+size, the window size should be rounded down to the nearest byte.
```
martinthomson, line 328: Given that the stream will include a window size as a power of two, maybe you should just say:

```suggestion
The window size used will be encoded in the content (currently, this can be expressed
in powers of two only) and it MUST be lower than this limit. An implementation MAY
treat a window size that exceeds the limit as a decoding error.
```

PR #2783, opened 2024-04-30
[compression-dictionary] Added link relation

### LINKED ISSUE #2772: Add link relation to trigger compression dictionary fetches
The explainer calls for a link relation of `dictionary` to let clients know of stand-alone dictionaries that should be fetched for future use. The ID should include the link relation definition and registration.

`dictionary` is too generic for this use case to use so we will have to come up with alternatives. `compression-dictionary` and `comp-dict` have been floated. Multiple references to the same dictionary will likely be compressed away after the first one by HPACK/QPACK so the length may not be a significant issue (if it is, `cdict` might be worth considering).

Some other things to include:

There needs to be some language about the risk of abuse as a tracking vector. Practically, this means it should be treated the same as `preload` and, for browsers, be limited to the main document resource.

The response needs to be CORS-readable. The explainer also calls for it being fetched with credentials omitted to discourage private information from being included in the dictionary contents but that's more to prevent a footgun since the resources are CORS-readable anyway. Forcing a new connection for the cases where it will commonly be used (documents, scripts, etc) for what will likely be a single request on the separate connection might be more cost than the footgun-prevention justifies.

pmeenan:
This adds the definition for the `compression-dictionary` link relation type and relevant IANA registration information.

The link relation provides a way to trigger the fetch of a stand-alone dictionary (e.g. for compressing HTML). The actual dictionary processing follows the standard path and headers once the fetch has been initiated.

There is discussion in issue #2772 as well as a [related issue](https://github.com/whatwg/html/issues/10162) in the HTML spec.

In Chrome's current origin trial and the initial explainers this is the `dictionary` relation type but that term is too generic for this specific case.

Fix #2772

### DIFF
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -46,6 +46,7 @@ normative:
     title: URL Pattern Standard
     date: 18 March 2024
     target: https://urlpattern.spec.whatwg.org/
+  WEB-LINKING: RFC8288
 
 informative:
   Origin: RFC6454
@@ -295,6 +296,38 @@ For example:
 Content-Dictionary: :pZGm1Av0IEBKARczz7exkNYsZb8LzaMrV7J32a2fFG4=:
 ~~~
 
+# The 'compression-dictionary' Link Relation Type
+
+This specification defines the 'compression-dictionary' link relation type
+{{WEB-LINKING}} that provides a mechanism for a HTTP response to provide a URL
+for a compression dictionary that is related to, but not directly used by the
+current HTTP response.
+
+The 'compression-dictionary' link relation type indicates that fetching and
+caching the specified resource is likely to be beneficial as future requests
+are likely to support using the provided compression dictionary for encoding
+responses.
+
+Clients SHOULD fetch the provided resource at a time that they determine would
+be appropriate.
+
+The response to the fetch for the compression dictionary is responsible for
+including the appropriate "Use-As-Dictionary" and caching response headers. The
+link relation only provides the mechanism for triggering the fetch of the
+dictionary.
+
+## Example
+For example:
+
+~~~ http-message
+Link: <https://example.org/dictionary.dat>;
+      rel="compression-dictionary"
+~~~
+
+indicates that the resource at https://example.org/dictionary.dat is a
+compression dictionary that is likely to be beneficial to use for future
+requests related to the current one.
+
 # Dictionary-Compressed Brotli
 
 The "br-d" content encoding identifies a resource that is a
@@ -408,6 +441,15 @@ IANA is asked to update the
 | Dictionary-ID        | permanent | {{dictionary-id}} of this document        |
 | Content-Dictionary   | permanent | {{content-dictionary}} of this document   |
 |----------------------|-----------|-------------------------------------------|
+
+## Link Relation Registration
+
+IANA is asked to update the "Link Relation Type Registry" registry
+({{WEB-LINKING}}):
+
+- Relation Name: compression-dictionary
+- Description: Refers to a compression dictionary used for content encoding.
+- Reference: This document, {{the-compression-dictionary-link-relation-type}}
 
 # Compatibility Considerations
 
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -34,8 +34,8 @@ author:
     ins: Y. Weiss
     name: Yoav Weiss
     role: editor
-    organization: Google LLC
-    email: yoavweiss@google.com
+    organization: Shopify Inc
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -295,6 +296,38 @@ For example:
...
+# The 'compression-dictionary' Link Relation Type
+
+This specification defines the 'compression-dictionary' link relation type
+{{WEB-LINKING}} that provides a mechanism for a HTTP response to provide a URL
+for a compression dictionary that is related to, but not directly used by the
+current HTTP response.
+
+The 'compression-dictionary' link relation type indicates that fetching and
+caching the specified resource is likely to be beneficial as future requests
+are likely to support using the provided compression dictionary for encoding
```
yoavweiss, line 308: "as future requested resources are likely able to use the specified resource as a compression dictionary for encoding their responses"?

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -295,6 +296,38 @@ For example:
...
+Clients SHOULD fetch the provided resource at a time that they determine would
+be appropriate.
+
+The response to the fetch for the compression dictionary is responsible for
+including the appropriate "Use-As-Dictionary" and caching response headers. The
+link relation only provides the mechanism for triggering the fetch of the
+dictionary.
+
+## Example
+For example:
```
yoavweiss, line 320: "For example" feels spurious

yoavweiss (approved):
LGTM % nits

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -295,6 +296,38 @@ For example:
...
+
+This specification defines the 'compression-dictionary' link relation type
+{{WEB-LINKING}} that provides a mechanism for a HTTP response to provide a URL
+for a compression dictionary that is related to, but not directly used by the
+current HTTP response.
+
+The 'compression-dictionary' link relation type indicates that fetching and
+caching the specified resource is likely to be beneficial as future requests
+are likely to support using the provided compression dictionary for encoding
+responses.
```
martinthomson, line 309:
```suggestion
The 'compression-dictionary' link relation type indicates that fetching and
caching the specified resource is likely to be beneficial for future requests.
The response to some of those future requests are likely to be able to use
the indicated resource as a compression dictionary.
```

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -295,6 +296,38 @@ For example:
...
+
+Clients SHOULD fetch the provided resource at a time that they determine would
+be appropriate.
```
martinthomson, line 312: In general, leave 2119 language for cases that have interoperability consequences.

```suggestion
Clients can fetch the provided resource at a time that they determine would
be appropriate.
```

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -295,6 +296,38 @@ For example:
...
+
+The response to the fetch for the compression dictionary is responsible for
+including the appropriate "Use-As-Dictionary" and caching response headers. The
```
martinthomson, line 315:
```suggestion
The response to the fetch for the compression dictionary needs to include a
"Use-As-Dictionary" and caching response headers for it to be usable as a
compression dictionary. The
```

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -295,6 +296,38 @@ For example:
...
+link relation only provides the mechanism for triggering the fetch of the
+dictionary.
+
+## Example
+For example:
+
+~~~ http-message
+Link: <https://example.org/dictionary.dat>;
+      rel="compression-dictionary"
```
martinthomson, line 324:
```suggestion
Link: <https://example.org/dictionary.dat>; rel="compression-dictionary"
```

I don't think that this blows the line limit.  If it does, shorten the URL before wrapping.

draft-ietf-httpbis-compression-dictionary.md
martinthomson: As does the title.  I'd put the description of the example above the example and avoid the section heading.

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -295,6 +296,38 @@ For example:
...
+~~~
+
+indicates that the resource at https://example.org/dictionary.dat is a
+compression dictionary that is likely to be beneficial to use for future
+requests related to the current one.
```
martinthomson, line 329: Move above, but...

```suggestion
The following example shows a link to a resource at https://example.org/dictionary.dat that
is expected to produce a compression dictionary:
```

pmeenan:
Thanks. Flows better with the suggested changes (PR updated).

PR #2784, opened 2024-05-09
[compression-dictionary] Moved the dictionary hash into the body

### LINKED ISSUE #2770: Essential content coding metadata: header or body?
This is a discussion we've had several times when defining content codings, but it seems like there is never really a single answer.

Should the content coding be self-describing, or can it rely on metadata in fields?

The compression dictionary work uses header fields to identify which compression dictionary is in use.  Originally, the client would indicate a dictionary and the server would indicate use of that dictionary by choosing the content coding.  This made interpreting the body of the response quite challenging in that you needed to have a request in order to make sense of it.

More recently, the specification has changed to having the client list available dictionaries, with the server echoing the one it chooses.  Both use header fields.

There is a third option, which is to embed the dictionary identification (which is a hash of the dictionary) ahead of the compressed content.  This has some real advantages:

1. Requests could use delta encoding.  There is - of course - a real question about availability of the dictionary, but we have that problem with encrypted content codings and keys.  That is a problem we solve for responses with a field.  It might be solved using the same technique, combined with the client-initiated content coding advice.  Or applications can use their own systems.
2. A field costs a lot more bits.  This would not benefit from header compression, so it might be a net loss in some cases, but in many cases it would be a distinct win because 32 bytes of hash in the body of a request is far less than a field name and field value containing a base64 encoding of those same 32 bytes.

It also comes with disadvantages:

1. Header compression could remove a lot of the cost in bytes.
2. It might be ever-so-slightly more complex for encoding and decoding to have to split the first 32 bytes off the body of a message.

pmeenan:
This eliminates the `Content-Dictionary` response header and moves the hash of the dictionary into the payload of the response.

Specifically, it adds 36-byte header in front of the compressed stream which is a 4 or 8-byte signature followed by the 32-byte SHA-256 hash digest.

The signature is different for the two different compressions:
* Dictionary-Compressed Brotli: `\xFF\x44\x43\x42` (0xFF followed by `DCB`)
* Dictionary-Compressed Zstandard: `\x5e\x2a\x4d\x18\x20\x00\x00\x00` (32-byte skippable Zstandard frame)

This also changes the content-encoding names for the different schemes to make it clear that they are a new format and not a raw brotli or Zstandard stream (`dcb` and `dcz`).

From the cli, the stream is equivalent to:

```sh
echo -en '\xffDCB' > data.txt.dcb && \
openssl dgst -sha256 -binary dictionary.txt >> data.txt.dcb && \
brotli --stdout -D dictionary.txt data.txt >> data.txt.dcb
```

or 

```sh
echo -en '\x5e\x2a\x4d\x18\x20\x00\x00\x00' > data.txt.dcz && \
openssl dgst -sha256 -binary dictionary.txt >> data.txt.dcz && \
zstd -D dictionary.txt -f -o tmp.zstd data.txt && \
cat tmp.zstd >> data.txt.dcz
```

Fix #2770

### DIFF
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -53,6 +53,10 @@ informative:
   Origin: RFC6454
   STRUCTURED-FIELDS: RFC8941
   RFC7932:  # Brotli
+  SHARED-BROTLI:
+    title: Shared Brotli Compressed Data Format
+    date: 28 September 2022
+    target: https://datatracker.ietf.org/doc/draft-vandevenne-shared-brotli-format/
   RFC8878:  # Zstandard
 
 --- abstract
@@ -310,10 +314,10 @@ The "dcb" content encoding identifies a resource that is a
 "Dictionary-Compressed Brotli" stream.
 
 A "Dictionary-Compressed Brotli" stream has a fixed header that is followed by
-a Brotli {{RFC7932}} stream. The header consists of a fixed 4 byte sequence
-and a 32 byte hash of the external dictionary that was used.  The Brotli stream
-is created using the referenced external dictionary and a compression window
-that is at most 16 MB in size.
+a Shared Brotli {{SHARED-BROTLI}} stream. The header consists of a fixed 4 byte
+sequence and a 32 byte hash of the external dictionary that was used.  The
+Shared Brotli stream is created using the referenced external dictionary and a
+compression window that is at most 16 MB in size.
 
 The 36-byte fixed header is as follows:
 
@@ -455,9 +459,9 @@ be used in secure contexts (HTTPS).
 
 # Security Considerations
 
-The security considerations for Brotli {{RFC7932}} and Zstandard
-{{RFC8878}} apply to the dictionary-based versions of the respective
-algorithms.
+The security considerations for Brotli {{RFC7932}}, Shared Brotli
+{{SHARED-BROTLI}} and Zstandard {{RFC8878}} apply to the
+dictionary-based versions of the respective algorithms.
 
 ## Changing content
 
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -340,18 +340,22 @@ The "dcz" content encoding identifies a resource that is a
 "Dictionary-Compressed Zstandard" stream.
 
 A "Dictionary-Compressed Zstandard" stream is a binary stream that starts with a
-36-byte fixed header and is followed by a Zstandard {{RFC8878}} stream
+40-byte fixed header and is followed by a Zstandard {{RFC8878}} stream
 of the response that has been compressed with an external dictionary.
 
-The 36-byte stream header consists of a 4-byte signature followed by the
+The 40-byte header consists of a fixed 8-byte sequence followed by the
 32-byte SHA-256 hash of the external dictionary that was used to compress the
 resource:
 
 Magic_Number:
-: 4 fixed bytes: 0xff, 0x44, 0x45, 0x5a.
+: 8 fixed bytes: 0x5e, 0x2a, 0x4d, 0x18, 0x20, 0x00, 0x00, 0x00.
 
 SHA_256_Hash:
 : 32 Bytes. SHA-256 hash digest of the dictionary {{SHA-256}}.
+
+The 40-byte header is a Zstandard skippable frame (little-endian 0x184D2A5E)
+with a 32-byte length (little-endian 0x00000020) that is compatible with existing
+Zstandard decoders.
 
 Clients that announce support for dcz content encoding MUST be able to
 decompress resources that were compressed with a window size of at least 8 MB
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -309,22 +309,19 @@ Link: <https://example.org/dict.dat>; rel="compression-dictionary"
 The "dcb" content encoding identifies a resource that is a
 "Dictionary-Compressed Brotli" stream.
 
-A "Dictionary-Compressed Brotli" stream is a binary stream that starts with a
-36-byte header and is followed by a Brotli {{RFC7932}} stream of the response
-that has been compressed with an external dictionary using a compression window
-not larger than 16 MB.
+A "Dictionary-Compressed Brotli" stream has a fixed header that is followed by
+a Brotli {{RFC7932}} stream. The header consists of a fixed 4 byte sequence
+and a 32 byte hash of the external dictionary that was used.  The Brotli stream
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -326,14 +306,27 @@ Link: <https://example.org/dict.dat>; rel="compression-dictionary"
...
+32-byte SHA-256 hash of the external dictionary that was used to compress the
+resource:
+
+|  Field             | Size       |
+|--------------------|------------|
+|  Magic_Number      | 4 bytes    |
+|  SHA_256_Hash      | 32 bytes   |
 
-Clients that announce support for br-d content encoding MUST be able to
+Magic_Number: 4 bytes, little-endian format.  Value: 0x424344FF
```
martinthomson, line 326:
```suggestion
Magic_Number: 4 fixed bytes: 0xff, 0x44, 0x43, 0x42.
```

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -342,13 +335,26 @@ delta-compression of resources larger than the compression window.
...
+The 36-byte stream header consists of a 4-byte signature followed by the
+32-byte SHA-256 hash of the external dictionary that was used to compress the
+resource:
+
+|  Field             | Size       |
+|--------------------|------------|
+|  Magic_Number      | 4 bytes    |
+|  SHA_256_Hash      | 32 bytes   |
+
+Magic_Number: 4 bytes, little-endian format.  Value: 0x5A4344FF
```
martinthomson, line 354:
```suggestion
Magic_Number: 4 fixed bytes: 0xff, 0x44, 0x45, 0x5a.
```

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -326,14 +306,27 @@ Link: <https://example.org/dict.dat>; rel="compression-dictionary"
...
+not larger than 16 MB.
+
+The 36-byte stream header consists of a 4-byte signature followed by the
+32-byte SHA-256 hash of the external dictionary that was used to compress the
+resource:
+
+|  Field             | Size       |
+|--------------------|------------|
+|  Magic_Number      | 4 bytes    |
+|  SHA_256_Hash      | 32 bytes   |
```
martinthomson, line 324: I'm not sure if the table really gets you much.

You might instead consider using a definition list, as I suggest below.

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -326,14 +306,27 @@ Link: <https://example.org/dict.dat>; rel="compression-dictionary"
...
 
-Clients that announce support for br-d content encoding MUST be able to
+Magic_Number: 4 bytes, little-endian format.  Value: 0x424344FF
+SHA_256_Hash: 32 Bytes. SHA-256 hash digest of the dictionary {{SHA-256}}.
```
martinthomson, line 327:
```suggestion
Magic_Number: 
: 4 fixed bytes: 0xff, 0x44, 0x43, 0x42.

SHA_256_Hash:
: 32 Bytes. SHA-256 hash digest of the dictionary {{SHA-256}}.
```

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -326,14 +306,27 @@ Link: <https://example.org/dict.dat>; rel="compression-dictionary"
...
+The "dcb" content encoding identifies a resource that is a
 "Dictionary-Compressed Brotli" stream.
 
-A "Dictionary-Compressed Brotli" stream is a Brotli {{RFC7932}} stream for a
-response that has been compressed with an external dictionary using a
-compression window not larger than 16 MB.
+A "Dictionary-Compressed Brotli" stream is a binary stream that starts with a
+36-byte header and is followed by a Brotli {{RFC7932}} stream of the response
+that has been compressed with an external dictionary using a compression window
+not larger than 16 MB.
```
martinthomson, line 315: I think that it's about this point that you might acknowledge that describing a thing completely in one sentence is hurting readability a bit.

```suggestion
A "Dictionary-Compressed Brotli" stream has a fixed header that is followed by
a Brotli {{RFC7932}} stream. The header consists of a fixed 4 byte sequence
and a 32 byte hash of the external dictionary that was used.  The Brotli stream
is created using the referenced external dictionary and a compression window
that is at most 16 MB in size.
```

martinthomson (approved):
Overall, I like this.  I want to flag for others that it doesn't use the skippable frame concept.

pmeenan:
FWIW, it doesn't look like it would be particularly difficult to use a skippable frame instead of a custom header in the Zstandard case if that's the way we want to go.  It's 4-bytes bigger (for the frame size) so it will essentially have an 8-byte "magic number" (since the size will always be the same).

It won't help with the creation of the files but if there's benefit to have the existing tooling be able to decode them (without verifying the hash because it will just skip it) we can go that route.

On the client side, it feels like it will be cleaner on the implementation to use the same flow for verifying the dictionary hash and magic number for both encodings without having to deal with different sized magic numbers (and always buffer and remove the first 36 bytes) but it's not a significant difference.

yoavweiss (approved):
LGTM

yoavweiss:
Still LGTM

pmeenan:
@felixhandte could you please take a look and make sure I got the byte sequence for the skippable frame correct?

I tested with the zstd cli and it was able to decode the file with the header without issue when created like this:

```sh
echo -en '\x5e\x2a\x4d\x18\x20\x00\x00\x00' > data.txt.dcz && \
openssl dgst -sha256 -binary dictionary.txt >> data.txt.dcz && \
zstd -D dictionary.txt -f -o tmp.zstd data.txt && \
cat tmp.zstd >> data.txt.dcz
```

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -340,19 +340,23 @@ The "dcz" content encoding identifies a resource that is a
...
 of the response that has been compressed with an external dictionary.
 
-The 36-byte stream header consists of a 4-byte signature followed by the
+The 40-byte header consists of a fixed 8-byte sequence followed by the
 32-byte SHA-256 hash of the external dictionary that was used to compress the
 resource:
 
 Magic_Number:
-: 4 fixed bytes: 0xff, 0x44, 0x45, 0x5a.
+: 8 fixed bytes: 0x5e, 0x2a, 0x4d, 0x18, 0x20, 0x00, 0x00, 0x00.
```
felixhandte, line 351: LGTM!

felixhandte (approved):
Looks like you just need to update the PR summary to reflect this change.

pmeenan:
> Looks like you just need to update the PR summary to reflect this change.

Thanks. Updated and added an example with the zstd cli.

MNocenyelli01 (approved):

PR #2787, opened 2024-05-22
[compression-dictionary] Restrict Accept-Encoding

pmeenan:
Adds `MUST NOT` language to the `Accept-Encoding` to not send the dictionary-aware encodings when a dictionary is not available (or when one is available but not being used).

The dictionary-based encodings don't work without a dictionary (they contain a header that has the dictionary hash) and should not be advertised when they are not really available.


royfielding (approved):
Thanks

PR #2788, opened 2024-05-23
[compression-dictionary] Tightened up the CORS language

### LINKED ISSUE #2634: CORS is inappropriate - Raised by Roy T. Fielding
For the existing draft, there is a lot of unnecessary confusion regarding features of fetch, like CORS, that don't make any sense from a security perspective. That's not what CORS is capable of covering, nor how it is implemented in practice, so reusing it doesn't make any sense.

[Original listserv response](https://lists.w3.org/Archives/Public/ietf-http-wg/2023JulSep/0130.html)
cc @royfielding 

Moved from private ID [issue #7](https://github.com/pmeenan/i-d-compression-dictionary/issues/7)

pmeenan:
This simplifies the CORS processing language in the doc, leaving all of the client-side readability processing to fetch and clearly spelling out the server responsibilities.

Fix #2634

### DIFF
--- a/draft-ietf-httpbis-compression-dictionary.md
+++ b/draft-ietf-httpbis-compression-dictionary.md
@@ -530,53 +530,30 @@ In browser terms, that means that both are either same-origin to the context
 they are being fetched from or that the response is cross-origin and passes
 the CORS check (https://fetch.spec.whatwg.org/#cors-check).
 
-#### Same-Origin
+### Server Responsibility
 
-On the client-side, same-origin determination is defined in the fetch spec (https://html.spec.whatwg.org/multipage/browsers.html#origin).
+A timing attack may still be possible on non-readable responses to extract
+information from them if they were compressed using a dictionary. To mitigate
+this risk, a server MUST NOT use dictionary compression on a response when the
+HTTP headers indicate that the request is a "cors" request and that the
+response would not be readable by the client.
 
-On the server-side, a request with a "Sec-Fetch-Site:" request header with a value of "same-origin" is to be considered a same-origin request.
+The following algorithm will return TRUE if a response MAY use a
+dictionary-based content-encoding and FALSE if a response MUST NOT use a
+dictionary-based content-encoding:
 
-* For any request that is same-origin:
-    * Response MAY be used as a dictionary.
-    * Response MAY be compressed by an available dictionary.
-
-#### Cross-Origin
-
-For requests that are not same-origin ({{same-origin}}), the "mode" of the request can be used to determine the readability of the response.
-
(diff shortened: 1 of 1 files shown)


royfielding:
The new language is much better, but I think it still misses the point. We don't need to be shy about why these requirements are in the spec and under what circumstances they are necessary. I suggest starting with an explanation:

> As with any usage of compressed content in a secure context, a potential timing attack exists if the attacker can control any part of the dictionary, or if it can read the dictionary and control any part of the content being compressed, while performing multiple requests that vary the dictionary or injected content. Under such an attack, the changing size of the response reveals information about the content, which might be sufficient to read the supposedly secure response.
> 
> In general, a server can mitigate such attacks by preventing variations per request, as in preventing active use of multiple dictionaries for the same content, disabling compression when any portion of the content comes from uncontrolled sources, and securing access and control over the dictionary content in the same way as the response content. In addition, the following requirements on a server are intended to disable dictionary-aware compression when the client provides CORS request header fields that indicate a cross-origin request context.

pmeenan:
Thanks.  I pulled in your explanation and tweaked the language to make it clear that precautions should be considered, one of which may be to not use dictionary-based content encoding. I like the new version a lot better as it allows for cacheable, public data to still use dictionary compression even if it is fetched in a cross-origin context and isn't client-readable.

I'll have to be careful with the fetch language as, currently, the client will fail a cross-origin non-readable response in case the server hasn't applied relevant protections. It may be reasonable to relax that and allow for the client to process it and put the burden on the server but it might also let a weak configuration go unnoticed (though since attacker-controlled dictionary content isn't particularly practical in this model, maybe it's not really a concern).

draft-ietf-httpbis-compression-dictionary.md
```diff
@@ -530,53 +530,41 @@ In browser terms, that means that both are either same-origin to the context
...
-1. If the mode is any other value (including "no-cors"):
-    * Response MUST NOT be used as a dictionary.
-    * Response MUST NOT be compressed by an available dictionary.
+### Server Responsibility
+
+As with any usage of compressed content in a secure context, a potential
+timing attack exists if the attacker can control any part of the dictionary,
+or if it can read the dictionary and control any part of the content being
+compressed, while performing multiple requests that vary the dictionary or
+injected content. Under such an attack, the changing size of the response
```
royfielding, line 539:
```suggestion
injected content. Under such an attack, the changing size or processing time of the response
```

royfielding: I'm not sure if we need to mention processing time as well, since we mention timing attack, but this looks good either way.

royfielding (approved):

Issue #2789, opened 2024-05-28
[RFC6265bis] Why Max-Age limitation is specified for User Agent ?


Jxck:
in [Section 4.1.2.2](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-14#section-4.1.2.2)

> The user agent MUST limit the maximum value of the Max-Age attribute. The limit SHOULD NOT be greater than 400 days (34560000 seconds) in duration. The RECOMMENDED limit is 400 days in duration, but the user agent MAY adjust the limit (see [Section 7.2](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-14#cookie-policy)). Max-Age attributes that are greater than the limit MUST be reduced to the limit.

This section mentions Max-Age limitation.
But Max-Age is an attribute for `Set-Cookie` not for `Cookie`, and `Set-Cookie` is sent by Server not Client.
So I'm wondering why this section is written for "User-Agent" not "Implementation".

It can be considered as "Server could send any max-age, but Client will trim it to < 400day", but in that case, that recommendation is not for "Max-Age" attributes itself, but for lifetime management for implementation.

> The RECOMMENDED limit is 400 days in duration

if this recommendation is for `Max-Age` attributes, it'll apply not (only?) for User Agent but for Server. So it should be use not "User Agent" but "Implementation".

miketaylr:
I'm not sure I agree that removing any mention of the user-agent in the Set-Cookie section is a better outcome. IMHO,  it's useful to consider the client even in the context of server requirements. 

> This is an issue for server side framework "Hono". It restricts max-age shorter than 400day according to RFC recommendation.

That seems like a valid choice for the framework. I'm not sure I would make the same choice, but frameworks like to be opinionated, and I can see the logic behind the thinking. I don't see how this is a bug in the draft.

Jxck:
> IMHO, it's useful to consider the client even in the context of server requirements.

I'm not against that.
but please think it simply,  this section is for "Max-Age attributes", and who sends "Max-Age" is SERVER.
So RECOMMENDATION should given for server, isn't it ?

sbingler:
Hi,

Both the `Max-Age`  and `Expires` are ways for the Server to specify the maximum amount of time the Server would like the cookie to exist. It's an upper bound. UAs can, and frequently do, remove cookies before the server specified time and this is allowed (and sometime encouraged) by the spec.

[...]

PR #2790, opened 2024-05-28
Clarify Max-Age limit and recommendation on RFC6265bis

Jxck:
In Max-Age Attributes section, there are unclear point about limitation to the duration.

- Max-Age attributes is for server (sender) but limitations are for User Agent
- Recommended duration is for limitation by user agent, not for the Max-Age value itself.
- RECOMMENDATION is not for server, and also this spec doesn't give recommendations for server too.

Clarify them will make it easy to understand that
- Server can set any duration value for Max-Age
- Client will limit the value to the convenient duration, recommended upper limit is < 400days.

closes: #2789


sbingler:
This is obsoleted by #2792

PR #2792, opened 2024-05-31
Refactor Cookie Expiration Limits

### LINKED ISSUE #2789: [RFC6265bis] Why Max-Age limitation is specified for User Agent ?

sbingler:
Closes #2789

Moves the user agent instructions out of the, non-normative, server sections.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -665,28 +665,18 @@ ignore unrecognized cookie attributes (but not the entire cookie).
 #### The Expires Attribute {#attribute-expires}
 
 The Expires attribute indicates the maximum lifetime of the cookie,
-represented as the date and time at which the cookie expires. The user agent is
-not required to retain the cookie until the specified date has passed. In fact,
-user agents often evict cookies due to memory pressure or privacy concerns.
-
-The user agent MUST limit the maximum value of the Expires attribute.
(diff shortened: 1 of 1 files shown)


mikewest (approved):
LGTM. This doesn't create any normative changes, so I think it's unnecessary to add it to the changelog.

PR #2804, opened 2024-06-12
Add multiple cookie header tolerance suggestions & length limit warnings

### LINKED ISSUE #2796: 6265bis: what is tolerance to multiple cookies?
RFC 6265 had a requirement that a client MUST generate a single cookie header. HTTP/2 and HTTP/3 allow "cookie crumbling" that permits multiple Cookie headers. 6265bis says in https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#section-4.2: 

> Servers MUST be tolerant of multiple cookie headers. For example, an HTTP/2 [[RFC9113](https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#RFC9113)] or HTTP/3 [[RFC9114](https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#RFC9114)] connection might split a cookie header to improve compression.

What is tolerance really? Why are we stating MUST requirement that are immediately invalidated by two of the versions of HTTP in wide usage? Section 5 is huge, what requirement is the text in section 4.2 forward referencing to?

I think there's some room for improvement in this area. An HTTP/2 or HTTP/3 is being instructed as a MUST requirement in the relevant RFCs to combine multiple cookie headers into one e.g. https://www.rfc-editor.org/rfc/rfc9113.html#section-8.2.3-2

> To allow for better compression efficiency, the Cookie header field MAY be split into separate header fields, each with one or more cookie-pairs. If there are multiple Cookie header fields after decompression, these MUST be concatenated into a single octet string using the two-octet delimiter of 0x3b, 0x20 (the ASCII string "; ") before being passed into a non-HTTP/2 context, such as an HTTP/1.1 connection, or a generic HTTP server application.

but if the server MUST be tolerant to multiple Cookies, maybe we can relax the HTTP/2 or HTTP/3 requirements?

### LINKED ISSUE #2797: 6265bis: suggested `Cookie:` header maximum/minimum length?
It is not unusual for servers to reject HTTP requests that contain individual headers longer than a certain limit. If a client sends a `Cookie:` header longer than around 8-9K bytes, the risk is quite large that it will get a 40x response back which  effectively blocks the client from continuing what it is set out to do.

Because of the rather complicated results of such a "too long" cookie header, I think it could be worth mentioning this in the spec. For newcomers in cookie land, this is perhaps not completely obvious.

The spec currently mentions size-limits per cookie, but a server can easily send a few large cookies and have the combined size of them outgrow 8K.

I would perhaps suggest wording that says a client SHOULD keep its lines shorter than NNN bytes and servers SHOULD accept cookie lines up to at least NNN bytes. (in curl we have the limit set to 8190 bytes)

HTTP related specs rarely speak of this kind of size limits, but I know my implementation could have avoided some issues if this real-world limit had been mentioned.

sbingler:
Closes #2796
Closes #2797 

Adds suggestions for how servers could fulfill the `MUST be tolerant of multiple cookie headers` requirement

Adds suggestion that servers be mindful of the number and size of the cookies they set to avoid hitting their own header length limits. Also mention that UA should consider splitting longer fields.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -854,7 +854,7 @@ cookie        = cookie-string
 cookie-string = cookie-pair *( ";" SP cookie-pair )
 ~~~
 
-While Section 5.4 of {{RFC9110}} does not define a length limit for header
+While {{Section 5.4 of RFC9110}} does not define a length limit for header
 fields it is likely that the web server's implementation does impose a limit;
 many popular implementations have default limits of 8K. Server SHOULD avoid
 setting a large number of large cookies such that the final cookie-string
@@ -1845,6 +1845,12 @@ the cookie-string following the algorithm defined in {{retrieval-algorithm}},
 where the retrieval's URI is the request-uri, the retrieval's same-site status
 is computed for the HTTP request as defined in {{same-site-requests}}, and the
 retrieval's type is "HTTP".
+
+Note: Previous versions of this specification required that only one Cookie
+header field be sent in requests. This is no longer a requirement. While this
+specification requires that a single cookie-string be produced, some user agents
+may split that string across multiple cookie header fields. For examples, see
+{{Section 8.2.3 of RFC9113}} and {{Section 4.2.1 of RFC9114}}.
 
 ### Non-HTTP APIs {#non-http}
 
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -2740,6 +2740,9 @@ The "Cookie Attribute Registry" should be created with the registrations below:
 * Update HTTPSEM to RFC9110
   <https://github.com/httpwg/http-extensions/pull/2795>
 
+* Add multiple cookie header tolerance suggestions & length limit warnings
+  <https://github.com/httpwg/http-extensions/pull/2804>
+
 # Acknowledgements
 {:numbered="false"}
 RFC 6265 was written by Adam Barth. This document is an update of RFC 6265,
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -871,8 +871,8 @@ any header field limits when deciding which approach to take.
 
 Note: Since intermediaries can modify cookie headers they should also be
 mindful of common server header field limits in order to avoid sending servers
-headers that they cannot process. For example, by concatenating multiple cookie
-headers into a single header larger than a server's limit.
+headers that they cannot process. For example, concatenating multiple cookie
+headers into a single header might exceed a server's size limit.
 
 
 ### Semantics
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -856,18 +856,24 @@ cookie-string = cookie-pair *( ";" SP cookie-pair )
 
 While {{Section 5.4 of RFC9110}} does not define a length limit for header
 fields it is likely that the web server's implementation does impose a limit;
-many popular implementations have default limits of 8K. Server SHOULD avoid
+many popular implementations have default limits of 8K. Servers SHOULD avoid
 setting a large number of large cookies such that the final cookie-string
 would exceed their header field limit. Not doing so could result in requests
 to the server failing.
 
 Servers MUST be tolerant of multiple cookie headers. For example, an HTTP/2
-{{RFC9113}} or HTTP/3 {{RFC9114}} connection might split a cookie header to
-improve compression. Servers are free to determine what form this tolerance
-takes. For example, the server could process each cookie header individually
-or the server could concatenate all the cookie headers into one and then
-process that final, single, header. The server should be mindful of any
-header field limits when deciding which approach to take.
+{{RFC9113}} or HTTP/3 {{RFC9114}} client or intermediary might split a cookie
+header to improve compression. Servers are free to determine what form this
+tolerance takes. For example, the server could process each cookie header
+individually or the server could concatenate all the cookie headers into one
+and then process that final, single, header. The server should be mindful of
+any header field limits when deciding which approach to take.
+
+Note: Since intermediaries can modify cookie headers they should also be
+mindful of common server header field limits in order to avoid sending servers
+headers that they cannot process. For example, by concatenating multiple cookie
+headers into a single header larger than a server's limit.
+
 
 ### Semantics
 
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1846,12 +1846,6 @@ where the retrieval's URI is the request-uri, the retrieval's same-site status
 is computed for the HTTP request as defined in {{same-site-requests}}, and the
 retrieval's type is "HTTP".
 
-Note: While this spec requires that a single cookie-string be produced, a user
-agent may consider splitting large request headers within their {{RFC9110}}
-implementation. This is due to potential server header length limits (See
-{{server-syntax}}) being exceeded due to servers setting too many large
-cookies.
-
 ### Non-HTTP APIs {#non-http}
 
 The user agent MAY implement "non-HTTP" APIs that can be used to access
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1835,6 +1846,12 @@ where the retrieval's URI is the request-uri, the retrieval's same-site status
 is computed for the HTTP request as defined in {{same-site-requests}}, and the
 retrieval's type is "HTTP".
 
+Note: While this spec requires that a single cookie-string be produced, a user
+agent may consider splitting large request headers within their {{RFC9110}}
```
sbingler, line 1850: @reschke Not sure how you feel that I suggest implementers deviate from the http spec.

This problem isn't specific to cookies (even if cookies are much likely to encounter it) so I don't think this spec should directly suggest that the `cookie:` header should be split.

mikewest: Is this a thing any implementer does at the moment? It seems different in kind from the H2/H3 quirks discussed above... It's not clear to me that making this suggestion is helpful, and that the developer-facing admonition to reduce the number and size of cookies above would be more appropriate to focus on.

If implementers take this action today, it might be better to define how exactly such a split should be made, and likewise to define exactly how servers should handle multiple headers. If implementers don't take this action today, let's not tacitly approve of a choice to do so in the future?

miketaylr:
>  If implementers don't take this action today...

If that's the case, I'd recommend punting on getting this in 6265bis - which has had a goal of documenting reality.

draft-ietf-httpbis-rfc6265bis.md
MikeBishop: "While we require this, consider not following the requirement" feels very RFC 6919.

sbingler: Safari https://github.com/httpwg/http-extensions/issues/2541#issuecomment-1552101655 has been splitting the cookie header for a few years

Chrome does not and I don't believe Firefox does either (from my own testing)

guoye-zhang: Yes, Safari started splitting Cookie in iOS 14 and macOS 11 (2020). It has become a load-bearing optimization since several internal websites won't load otherwise.

I was not aware that neither Chrome nor Firefox does it.

sbingler: After consideration I think this note should be removed.

martinthomson: It might be worth a note to the opposite effect:

> Previous versions of this specification required that only one Cookie header fields be sent in requests.
> This is no longer a requirement. Some user agents split cookies across multiple header fields.
> See also {{Section 8.2.3 of ?RFC9113}} and {{Section 4.2.1 of ?RFC9114}}.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -854,9 +854,20 @@ cookie        = cookie-string
 cookie-string = cookie-pair *( ";" SP cookie-pair )
 ~~~
 
+While Section 5.4 of {{RFC9110}} does not define a length limit for header
```
martinthomson, line 857:
```suggestion
While {{Section 5.4 of RFC9110}} does not define a length limit for header
```

draft-ietf-httpbis-rfc6265bis.md
sbingler: I think that's a good way of phrasing it.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -854,9 +854,20 @@ cookie        = cookie-string
...
+fields it is likely that the web server's implementation does impose a limit;
+many popular implementations have default limits of 8K. Server SHOULD avoid
+setting a large number of large cookies such that the final cookie-string
+would exceed their header field limit. Not doing so could result in requests
+to the server failing.
+
 Servers MUST be tolerant of multiple cookie headers. For example, an HTTP/2
 {{RFC9113}} or HTTP/3 {{RFC9114}} connection might split a cookie header to
-improve compression.
+improve compression. Servers are free to determine what form this tolerance
```
LPardue, line 866: Commenting on the line above, the `connection` doesn't do the splitting per se. Instead, the protocol allows the user agent to split the cookie. As a server you have to accept the split cookie when processing, and then move on to figuring out your tolerance policy (as you're covering in the new text)

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -854,9 +854,20 @@ cookie        = cookie-string
...
+takes. For example, the server could process each cookie header individually
+or the server could concatenate all the cookie headers into one and then
+process that final, single, header. The server should be mindful of any
+header field limits when deciding which approach to take.
```
LPardue, line 870: It might be besides the point right at this point in the text, but something common in my experience is for a reverse proxy to do H2 or H3 termination. It might tend to pick some strategy such as merging, in order to build a compliant single cookie that is forwards. 

Such a proxy might not validate the cookie in any way - not length nor content. Instead, it passes on this jumbo cookie that then fails at the origin.

This can be surprising to the origin admin. They might have followed the set-cookie advise and might control the UA to ensure that cookie splitting is happening. But somewhere along the chain the cookies are getting merged, leading to some size limit check somewhere failing.

draft-ietf-httpbis-rfc6265bis.md
sbingler: Are you asking that the phrase `...connection might split a cookie header...` be changed?

mikewest (approved):
Thanks for iterating on this, and for correcting my understanding of the status quo behavior of cookie headers. I think this patch is good to land as-is at this point. LGTM.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -854,9 +854,20 @@ cookie        = cookie-string
 cookie-string = cookie-pair *( ";" SP cookie-pair )
 ~~~
 
+While {{Section 5.4 of RFC9110}} does not define a length limit for header
+fields it is likely that the web server's implementation does impose a limit;
+many popular implementations have default limits of 8K. Server SHOULD avoid
```
MikeBishop, line 859:
```suggestion
many popular implementations have default limits of 8K. Servers SHOULD avoid
```

draft-ietf-httpbis-rfc6265bis.md
MikeBishop: Perhaps s/connection/client or intermediary/?

draft-ietf-httpbis-rfc6265bis.md
MikeBishop: If we want to address it here, perhaps something like:

> For example, the server could process each cookie header individually. Alternatively, the server or an intermediary could concatenate all the cookie headers into one and then process/forward that final, single, header. The server and any intermediaries should be mindful of any header field limits when deciding which approach to take.

MikeBishop (approved):

draft-ietf-httpbis-rfc6265bis.md
sbingler: Sounds good to me

draft-ietf-httpbis-rfc6265bis.md
sbingler: Thanks for that catch

draft-ietf-httpbis-rfc6265bis.md
sbingler: I added a note that intermediaries should make large headers. While it'd be great if intermediary implementation followed that advice I think there is a benefit in giving a server's admin a hint as to why their requests might be failing and where to trying troubleshooting.

mikewest (approved):
Still LGTM, thanks!

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -854,9 +854,26 @@ cookie        = cookie-string
...
+header to improve compression. Servers are free to determine what form this
+tolerance takes. For example, the server could process each cookie header
+individually or the server could concatenate all the cookie headers into one
+and then process that final, single, header. The server should be mindful of
+any header field limits when deciding which approach to take.
+
+Note: Since intermediaries can modify cookie headers they should also be
+mindful of common server header field limits in order to avoid sending servers
+headers that they cannot process. For example, by concatenating multiple cookie
+headers into a single header larger than a server's limit.
```
MikeBishop, line 875:
```suggestion
headers that they cannot process. For example, concatenating multiple cookie
headers into a single header might exceed a server's size limit.
```

Let's fix this sentence fragment, but I'm fine with the message.

Issue #2791, opened 2024-05-30
[RFC6265bis WGLC] Lax-allowing-unsafe narrative  


bc-pi:
I don't believe this example given in [8.8.6. ](https://www.ietf.org/archive/id/draft-ietf-httpbis-rfc6265bis-14.html#section-8.8.6)[Top-level requests with "unsafe" methods](https://www.ietf.org/archive/id/draft-ietf-httpbis-rfc6265bis-14.html#name-top-level-requests-with-uns) particularity well reflects the situation that precipitated the behavior described in [5.5.7.2. ](https://www.ietf.org/archive/id/draft-ietf-httpbis-rfc6265bis-14.html#section-5.5.7.2)["Lax-Allowing-Unsafe" enforcement](https://www.ietf.org/archive/id/draft-ietf-httpbis-rfc6265bis-14.html#name-lax-allowing-unsafe-enforce):

> For example, a login flow may involve a cross-site top-level POST request to an endpoint which expects a cookie with login information. For such a cookie, "Lax" enforcement is not appropriate, as it would cause the cookie to be excluded due to the unsafe HTTP request method. On the other hand, "None" enforcement would allow the cookie to be sent with all cross-site requests, which may not be desirable due to the cookie's sensitive contents.

It's only an explanatory example so arguably doesn't even really matter.  But RFCs are _forever_ and they are often cited so it'd be nice to have this bit of the historical narrative better reflect the realities at the time (I say this mostly from perspective of having experienced some of the difficulties caused by the introduction of SameSite cookies and their default enforcement mode).

As best I can recall, login type flows were the main driver for the "Lax-allowing-unsafe" default mode but the need manifested itself more in the later part of the flow where the authenticating site (IDP) caused the browser to make a POST request callback to the site requesting authentication (RP). At the start of the flow, prior to sending the user to the IDP, the RP would typically set a cookie with or referencing some transaction specific state and use that info during processing of the callback for validation of the token provided by the IDP. Cookies with Lax as the default enforcement mode would not be sent with the callback POST request and the whole login flow would fail.  The "Lax-allowing-unsafe" default mode (aka "Lax + POST mitigation") allowed those flows to not break for a while and gave RPs (and their vendors, frameworks, etc.) a little grace period to update to explicitly designate those cookies as SameSite=None. 

[...]

PR #2812, opened 2024-06-27
Tweak the example in the explanation of "Lax-allowing-unsafe"

b---c:
minimalist attempt at a fix for #2791

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -2421,8 +2421,8 @@ resulting in an unrecoverable failure of the whole login flow.
 
 The "Lax-allowing-unsafe" enforcement mode described in {{lax-allowing-unsafe}}
 retains some of the protections of "Lax" enforcement (as compared to "None")
-while still allowing cookies to be sent cross-site with unsafe top-level
-requests.
+while still allowing recently created cookies to be sent cross-site with unsafe
+top-level requests.
 
 As a more permissive variant of "Lax" mode, "Lax-allowing-unsafe" mode
 necessarily provides fewer protections against CSRF. Ultimately, the provision
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -2413,11 +2413,11 @@ cross-site requests with "unsafe" HTTP methods (as was the case prior to the
 introduction of the `SameSite` attribute).
 
 For example, the concluding step of a login flow may involve a cross-site top-level
-`POST` request to an endpoint which expects a cookie with transactional state
-information necessary to securely complete the login. For such a cookie, "Lax"
-enforcement is not appropriate, as it would cause the cookie to be excluded due
-to the unsafe HTTP request method, resulting in an unrecoverable failure of the
-whole login flow.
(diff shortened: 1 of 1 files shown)


mikewest (approved):
LGTM.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -2412,12 +2412,12 @@ not explicitly specifying a `SameSite` attribute being included on top-level
...
-"Lax" enforcement is not appropriate, as it would cause the cookie to be
-excluded due to the unsafe HTTP request method. On the other hand, "None"
-enforcement would allow the cookie to be sent with all cross-site requests,
-which may not be desirable due to the cookie's sensitive contents.
+For example, the concluding step of a login flow may involve a cross-site top-level
+`POST` request to an endpoint which expects a cookie with transactional state
+information necessary to securely complete the login. For such a cookie, "Lax"
+enforcement is not appropriate, as it would cause the cookie to be excluded due
+to the unsafe HTTP request method, resulting in an unrecoverable failure of the
+whole login flow.
```
sbingler, line 2420:
```suggestion
For example, the concluding step of a login flow may involve a cross-site top-level
`POST` request to an endpoint; this endpoint expects a recently created cookie
containing transactional state information, necessary to securely complete the
login. For such a cookie, "Lax" enforcement is not appropriate, as it would
cause the cookie to be excluded due to the unsafe HTTP request method,
resulting in an unrecoverable failure of the whole login flow.
```

I think it's a good idea to also motivate the reason for the 2 min limit.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -2412,12 +2412,12 @@ not explicitly specifying a `SameSite` attribute being included on top-level
...
 
 The "Lax-allowing-unsafe" enforcement mode described in {{lax-allowing-unsafe}}
```
sbingler, line 2422: Seems github won't let me add a suggestion for this text, but I also suggest a change of

```
The "Lax-allowing-unsafe" enforcement mode described in {{lax-allowing-unsafe}}
retains some of the protections of "Lax" enforcement (as compared to "None")
while still allowing recently created cookies to be sent cross-site with unsafe
top-level requests.
```

sbingler (commented):
Thanks!

I made some suggestions around the 2 min rule (plus some minor grammatical ones)

draft-ietf-httpbis-rfc6265bis.md
b---c: Yeah, github doesn't let you suggest changes too far from the context of the original changes in the PR. But I'll apply this locally and push and the whole dance. Thanks for the suggestion.

b---c: done in 4590b6c0b949791f907eb240c59c3fe0a5a6bace

b---c:
Thanks for the review @sbingler! I've incorporated all your suggestions.

sbingler (approved):
LGTM, thanks! I'll wait for @mikewest to take another look before merging

mikewest (approved):
Catching up on things after being out sick: this still LGTM. :)

bc-pi:
Thanks @mikewest and @sbingler! I dare say it's okay merge?

PR #2818, opened 2024-07-04
[optimistic-upgrade] Deprecate the "HTTP/*.*" upgrade token

### LINKED ISSUE #2737: Determine the status of "Upgrade: HTTP/2.0"
The [IANA Upgrade Token Registry](https://www.iana.org/assignments/http-upgrade-tokens/http-upgrade-tokens.xhtml) lists an entry for HTTP/\*.\*, and specifically identifies "2.0" as a possible version value, as instructed by [RFC 9110 Section 18.10](https://datatracker.ietf.org/doc/html/rfc9110#name-upgrade-token-registration).  However, this flow seems to have been superseded by "Upgrade: h2c", which was introduced in RFC 7540 and deprecated in RFC 9113.  I am not aware of any implementations of "Upgrade: HTTP/2.0" in clients or servers.

I think we have two main options:

1. Status Quo: Regard "HTTP/\*.\*" as a mechanism that is formally defined but not implemented anywhere.  Provide formal analysis of its security for completeness.
2. Deprecation: Update RFC 9110 in this draft to remove the definition of this upgrade token.  Mark "HTTP" as "obsolete" in the Upgrade Tokens Registry.

bemasc:
Fixes #2737

### DIFF
--- a/draft-ietf-httpbis-optimistic-upgrade.md
+++ b/draft-ietf-httpbis-optimistic-upgrade.md
@@ -14,7 +14,7 @@ workgroup: "HTTPBIS"
 keyword:
 venue:
   github: "httpwg/http-extensions"
-updates: 9298
+updates: 9298, 2817
 
 author:
  -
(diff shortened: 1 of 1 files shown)


michaelrsweet:
Note that the Internet Printing Protocol (STD 92) REQUIRES support for TLS upgrade when the printer supports TLS. Since TLS has been a standard features of most network printers since 2012 (and was in most enterprise printers prior to that), you can expect somewhere north of 2 billion network printers support it...

I understand wanting to deprecate TLS upgrade for ordinary web browser traffic where proxies are commonly used that don't know how to deal with the Upgrade: header at all, but TLS upgrade *is* used for printing.

royfielding:
I think this suggestion is nuts.

We have registries so that the same name is not being used for two different things. We don't need to deprecate things that are not used. If they are used where they shouldn't be, that would be a note placed on their specification by full IETF review, not on their name registry.

If the folks in the room just don't understand HTTP and the variety of its uses over the Internet, they will foolishly suggest that names should be deprecated just because their own software doesn't use them. That's not how we are supposed to work. We don't have a room large enough for all of the implementations.

bemasc:
Thanks, I didn't know about IPP using "Upgrade: TLS".  I've removed that deprecation suggestion and retitled the PR.  I've moved the technical observation about the safety of "Upgrade: TLS" into #2828.

bemasc:
From the discussion during the session yesterday, it seems like there isn't consensus to declare this token deprecated, and it's not clear what to say about it since it's not actually in use.  I've opened #2845 to ignore it instead.

Issue #2825, opened 2024-07-09
Address Compression Dictionary AD review issues


pmeenan:
(tried using ietf-comments but ran into issues so just tracking manually in one combined issue)...

# AD Review of draft-ietf-httpbis-compression-dictionary-06

 

cc @fpalombini

 

Thank you for the work on this document.

 

Almost all my comments are about references. I think a new version is necessary before starting IETF Last Call, to avoid process issues along the way.

 

Francesca

 

### Duplicated BCP 14 boilerplate

 

The boilerplate is duplicate, please remove the second occurrence.

 

### Structured fields

 

Can you please update the reference to 8941 to draft-ietf-httpbis-sfbis ? That doc is with the RFC Editor so should not be holding this document up.

 

Also, I believe the reference to draft-ietf-httpbis-sfbis should be normative, not informative, since terminology from that doc is used. Alternatively, if you want to keep the ref informative, you can import the part of the terminology that is necessary for this doc. I think that's a uglier solution, so I highly prefer sfbis to be made normative, but won't block on it.

 

### whatwg reference

 

[URLPattern]

"URL Pattern Standard", March 2024, https://urlpattern.spec.whatwg.org/.

 

needs to be indicated as Living standard (see RFC 9110 or 9421 for eample of whatwg specs references).

 

### Fetch missing reference

 

> The "match-dest" value of the Use-As-Dictionary header is an Inner List of String values that provides a list of request destinations for the dictionary to match (https://fetch.spec.whatwg.org/#concept-request-destination).

 

> and passes the CORS check (https://fetch.spec.whatwg.org/#cors-check).

 

Please fix this so that the Fetch spec is properly referenced (normatively is needed, I believe).

 

### Missing reference

 

> NOTE: '\' line wrapping per RFC 8792

 

RFC 8792 should be (informatively) referenced.

 

### RFC 5861

 

I agree with Mark's write up, 5861 should really be informative.

 

### Section 2.2.2.

 

There is several occurrences of {Origin}, please fix.

 

### CRIME Ref

 

> The CRIME attack shows that it's a bad idea to compress data from mixed (e.g. public and private) sources

 

Please add a reference.

 

### Cookies

 

> To mitigate any additional tracking concerns, clients MUST treat dictionaries in the same way that they treat cookies.

 

It would be good to have an informative reference to 6265 (or even 6265bis).

 

## Notes

 

This review is in the ["IETF Comments" Markdown format][ICMF], You can use the

[...]

PR #2826, opened 2024-07-09
[compression-dictionary] Fixed references and comments from AD review

pmeenan:
This addresses the reference issues, comments and nits from the AD review (see #2825)


pmeenan:
@fpalombini would you mind taking a quick look before I publish a new draft?

PR #2827, opened 2024-07-11
[optimistic-upgrade] Recommend GET for future Upgrade Tokens

### LINKED ISSUE #2738: Conside restrictions on bodies in Upgrade requests
Formally, Upgrade can be attached to any HTTP request, in which case the request is processed in HTTP/1.1, and the server has the option to Upgrade before returning the response.  This includes requests with and without nonempty bodies.  Handling of Upgrade with a nonempty body is well-defined, but the Upgrade paths that are actually implemented today (WebSockets and MASQUE) only use GET requests with empty bodies.

@gstrauss raised concerns that this is something of a "sharp edge" for security.  For example, a server implementation might assume that there is no body in an Upgrade request, leading to a mismatch between the client and server when a body is present.

We could make this problem go away by mandating that Upgrade only be used with "Content-Length: 0", and perhaps only with "GET".  However, this requirement would appear to contradict RFC 2817, which says "A client MAY offer to switch to secured operation during any clear HTTP request", so that document would have to be Updated or marked Historic.

bemasc:
Fixes #2738

### DIFF
--- a/draft-ietf-httpbis-optimistic-upgrade.md
+++ b/draft-ietf-httpbis-optimistic-upgrade.md
@@ -100,7 +100,11 @@ A related category of attacks use protocol disagreement to exploit vulnerabiliti
 
 If the server rejects the transition request, the connection can continue to be used for HTTP/1.1.  There is no requirement to close the connection in response to a rejected transition, and keeping the connection open has performance advantages if additional HTTP requests to this server are likely.  Thus, it is normally inappropriate to close the connection in response to a rejected transition.
 
+<<<<<<< HEAD
 # Impact on HTTP Upgrade with Existing Upgrade Tokens
+=======
+# Impact on Existing Upgrade Tokens {#existing}
+>>>>>>> ac690a6f (Remove normative SHOULD and clarify rationale)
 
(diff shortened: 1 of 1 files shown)


royfielding:
I do not agree with this suggestion.

Upgrade is intended to support a not-yet-defined future. Limits placed upon it must be evaluated in terms of what will be excluded from future use, not what is currently being used in practice.

If you want to limit what can be sent in a body, do that instead. The entire premise of this document is that a client will send something incredibly stupid and uncontrolled in a request body while asking for an upgrade. Just forbid that and be done.

In any case, Upgrade with GET, HEAD, and OPTIONS has no such issues. Upgrade should also be possible to use with QUERY with any compliant server. If you can find a server that fails to process a valid request body and then proceeds to treat it as a pipelined request, that is a vulnerability in their code.

bemasc:
This PR originally assumed the deprecation of "TLS" in #2818.  I've removed that deprecation there, so other HTTP methods are necessarily allowed.

Due to that change, I've rewritten the recommendation here, removed the PR stacking, and removed the "draft PR" flag.

PR #2845, opened 2024-07-26
[optimistic upgrade] Remove mention of HTTP/*.*

### LINKED ISSUE #2737: Determine the status of "Upgrade: HTTP/2.0"

bemasc:
This avoids having to determine whether `HTTP/*.*` is an actual existing standard or not.

This is an alternative to #2818

Fixes #2737

### DIFF
--- a/draft-ietf-httpbis-optimistic-upgrade.md
+++ b/draft-ietf-httpbis-optimistic-upgrade.md
@@ -85,15 +85,7 @@ If the server rejects the upgrade, the connection can continue to be used for HT
 
 # Impact on Existing Upgrade Tokens
 
-At the time of writing, there are four distinct Upgrade Tokens that are registered, associated with published documents, and not marked obsolete.  This section considers the impact of this document's considerations on each registered Upgrade Token.
-
-## "HTTP"
-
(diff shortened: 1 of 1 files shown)


bemasc:
Merging for draft cutoff.

PR #2851, opened 2024-08-02
(Content-)Location in QUERY

MikeBishop:
#1746 reborn.


draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -201,6 +201,26 @@ q=foo&amp;limit=10&amp;sort=-published
       QUERY responses are illustrated in <xref target="examples" />.
     </t>
 
+    <t>
+      A successful response to a QUERY request can include a Content-Location
+      header field (see <xref target="HTTP" section="8.7"/>) containing an
+      identifier for a specific resource corresponding to the results of the
```
martinthomson, line 207:
```suggestion
      identifier for a resource corresponding to the results of the
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -201,6 +201,26 @@ q=foo&amp;limit=10&amp;sort=-published
...
+    </t>
+
+    <t>
+      A server &MAY; create or locate a resource that identifies the query
+      operation for future use. If the server does so, the URI of the resource
+      can be included in the Location header field of the response (see <xref
+      target="HTTP" section="10.2.2"/>). This represents a claim that a client can
+      send a GET request to the indicated URI to repeat the query operation just
+      performed without resending the query parameters. This resource might be
+      temporary; if a future request fails, the client &SHOULD; retry using the
```
martinthomson, line 220:
```suggestion
      temporary; if a future request fails, the client can retry using the
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -346,6 +367,15 @@ Dubois, Camille, camille.dubois@example.net
       considerations as all HTTP methods as described in
       <xref target="HTTP"/>.
     </t>
+
+    <t>
+      The QUERY method can be used as an alternative to passing query
+      information in the query portion of a URI. This is preferred in some
+      cases, as the URI is more likely to be logged than the request content. If
+      a server creates a temporary resource to represent the results of a QUERY
+      request (e.g. for use in the Location or Content-Location field), the URI of this
```
martinthomson, line 376:
```suggestion
      request (e.g., for use in the Location or Content-Location field), the URI of this
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -346,6 +367,15 @@ Dubois, Camille, camille.dubois@example.net
...
+      resource &SHOULD-NOT; expose the original request content in plaintext.
```
martinthomson, line 377: I'm not sure that this is a consideration that warrants a "SHOULD NOT".  Yes, there is a risk that logging a URL will cause stuff to be logged that you shouldn't log, but that is only if the QUERY contains stuff that shouldn't be logged.  This reads as being unconditional.

reschke: I agree.

reschke:
@martinthomson - thanks for the feedback; I have applied the trivial changes and will look at he last comment later on.

draft-ietf-httpbis-safe-method-w-body.xml
reschke: maybe... "... the URI SHOULD be chosen in a way that potentially sensitive information is not exposed."?

reschke:
(added an example with L and C-L, and how to use them; this certainly can be enhanced, for instance with Last-Modified fields).

draft-ietf-httpbis-safe-method-w-body.xml
martinthomson: I'm looking for an additional qualification on the statement, something like:

> If a server creates a temporary resource to represent the results of a QUERY request (e.g., for use in the Location or Content-Location field), the URI of this resource &SHOULD-NOT; expose **any sensitive information from** the original request content in plaintext.

PR #2852, opened 2024-08-04
Revert "(Content-)Location in QUERY"

reschke:
Reverts httpwg/http-extensions#2851

### DIFF
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -306,7 +306,6 @@ select surname, givenname, email limit 10
 <artwork type="http-message">
 HTTP/1.1 200 OK
 Content-Type: text/csv
-Content-Location: http://example.org/contacts/query123
 
 surname, givenname, email
 Smith, John, john.smith@example.org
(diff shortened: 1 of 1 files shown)


reschke:
(sorry for accidentally merging, now reverted)

reschke:
Q; in the Security Considerations, should we mention that using L or C-L needs to be done carefully when the original QUERY method's behavior was subject to some kind of access control?

MikeBishop:
I'm confused. #2851 appears to be in `main`, while this PR proposes reverting it. Was the reverse intended?

reschke:
I probably screwed up things when I accidentally pressed "merge" when I wanted "rebase".

Let me try to fix it today.

reschke:
Ok, I have restored the original branch, but I'm not sure how to remove the commits from main.

I think something like (on main)
---
git reset d26f7c3b9f38badc3598e742824e82667df2f092
---

ought to work, but I'm not totally convinced.

Maybe @mnot can help?

MikeBishop:
Three options I can see:

- If there's nothing else merged we need to keep, `git reset [last good commit] && git push -f` will work, but it's potentially disruptive if anything has been merged from other documents or people have cloned/pulled in the meantime.
- Merge the reversion PR to main; it's a lot of unneeded commits, but it doesn't lose anything and it's consistent.
- `git show [last good commit]:draft-ietf-httpbis-safe-method-w-body.xml > draft-ietf-httpbis-safe-method-w-body.xml && git add draft-ietf-httpbis-safe-method-w-body.xml && git commit -m "Revert PR#2851"` would also do it as a single commit to revert the draft without touching anything else.

reschke:
In the meantime, I did the simplest thing:

- recreated the branch without reverts
- copied the previous start of the draft over to main, so we don't have the contents of the PR in there. That left the version history uglx, but there we go...

PR #2857, opened 2024-08-06
Restore cookie-name = token

markt-asf:
This is a partial revert of 5469d51 from PR #1145


mikewest:
Talked with @sbingler a moment ago. I think there are some additional changes that will be necessary, but I don't object to the idea of this revert. Seems like a mistake in the original PR.

markt-asf:
Happy to amend the PR if you can give me an idea of what additional changes are required.

sbingler:
I wrote up a change that should cover all the points we want to make.

Namely, we still want UAs to accept that larger character set while encouraging server to use the more limited character set.

mikewest (approved):
This looks reasonable to me, thanks to you both.

Issue #2875, opened 2024-08-28
Integrate definitions in Conventions


ioggstream:
## I expect

- representation metadata, content, ... in Conventions and Definitions.

PR #2876, opened 2024-08-28
Fix: #2875. More HTTP definitions.

ioggstream:
## This PR

- adds more used definitions from HTTP.


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -85,7 +85,7 @@ This document defines an optional mechanism for HTTP that enables resumable uplo
 The terms Byte Sequence, Item, String, Token, Integer, and Boolean are imported from
 {{!STRUCTURED-FIELDS=RFC8941}}.
 
-The terms client and server are from {{HTTP}}.
+The terms "representation", "representation data", "representation metadata", "content", "client" and "server" are from {{HTTP}}.
```
Acconut, line 88: Is there a preferred way of written the referenced terms? Here you quote them, but in the sentence above we capitalize them without quotes. Does it depend on how the referenced document writes the terms?

Issue #2883, opened 2024-08-28
Effective request uri is not RFC9110.


ioggstream:
## I expect

- target URI

## Instead

- Effective request URI

PR #2884, opened 2024-08-28
Fix: #2883. Use target URI.

ioggstream:
## This PR

- use Target URI

Fixes: #2883


MikeBishop:
Superceded by #2891. Please reopen if that's not correct.

PR #2902, opened 2024-09-18
Reword guidance around URI generation

### LINKED ISSUE #2853: Sensitive information in QUERY URLs
I'm looking for an additional qualification on the statement, something like:

> If a server creates a temporary resource to represent the results of a QUERY request (e.g., for use in the Location or Content-Location field), the URI of this resource &SHOULD-NOT; expose **any sensitive information from** the original request content in plaintext.

_Originally posted by @martinthomson in https://github.com/httpwg/http-extensions/pull/2851#discussion_r1703436067_

MikeBishop:
Mostly editorial, but touches normative text. Fixes #2853.

### DIFF
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -432,8 +432,9 @@ Dubois, Camille, camille.dubois@example.net
       information in the query portion of a URI. This is preferred in some
       cases, as the URI is more likely to be logged than the request content. If
       a server creates a temporary resource to represent the results of a QUERY
-      request (e.g., for use in the Location or Content-Location field), the URI of this
-      resource &SHOULD; be chosen such that it does not expose any sensitive
+      request (e.g., for use in the Location or Content-Location field) and the request
+      contains sensitive information that cannot be logged, then the URI of this
+      resource &SHOULD; be chosen such that it does not include any sensitive
       portions of the original request content.
     </t>
   </section>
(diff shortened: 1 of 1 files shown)


reschke (approved):

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -433,7 +433,8 @@ Dubois, Camille, camille.dubois@example.net
       cases, as the URI is more likely to be logged than the request content. If
       a server creates a temporary resource to represent the results of a QUERY
       request (e.g., for use in the Location or Content-Location field), the URI of this
```
martinthomson, line 435:
```suggestion
      request (e.g., for use in the Location or Content-Location field) and the request
      contains sensitive information that cannot be logged, then the URI of this
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -433,7 +433,8 @@ Dubois, Camille, camille.dubois@example.net
...
-      resource &SHOULD-NOT; expose the original request content in plaintext.
+      resource &SHOULD; be chosen such that it does not expose any sensitive
```
martinthomson, line 436:
```suggestion
      resource &SHOULD; be chosen such that it does not include any sensitive
```

martinthomson (approved):
This is all so conditional that I wonder whether it even needs a SHOULD (NOT).  Would it be enough to say instead.

> A server can avoid this risk by constructing resource URIs
> (for use in Location or Content-Location) 
> that do not include sensitive information.

The other options are fine.

PR #2905, opened 2024-09-27
Add draft-ietf-httpbis-no-vary-search

jeremyroman:
This has been adopted by the working group as of Sept 25, 2024, after previously existing at:
  https://github.com/jeremyroman/http-no-vary-search
  https://github.com/WICG/nav-speculation


jeremyroman:
@mnot @tfpauly PTAL?

PR #2910, opened 2024-10-07
Resumable Upload: Add guidance on upload creation strategy

Acconut:
The definition of upload creation requests allows the client to include the entire file, some file data or no data at all. While we provide this flexibility, we don't give guidance on when which approach is best suited.

This PR adds this missing piece and explains how an optimistic client can attempt to upload the file in one request and how a careful client uses a separate request for creating the upload. In addition, it also mentions how to transparently upgrade to resumable uploads, which wasn't explained in detail before.

Let me know if you have any feedback on this.

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -520,15 +520,15 @@ The following subsections describe two typical upload strategies that are suited
 
 ## Optimistic Upload Creation
 
-An "optimistic upload creation" can be used if the client assumes that the server supports resumable uploads under the target URI and that the client is capable of handling and processing interim responses. The upload creation request then includes the full file because the client anticipates that the file will be transferred without interruptions or resumed if an interruption occurs.
+An "optimistic upload creation" can be used independent of the client's knowledge about the server's support for resumable uploads. However, the client must be capable of handling and processing interim responses. An upload creation request then includes the full file because the client anticipates that the file will be transferred without interruptions or resumed if an interruption occurs.
 
 The benefit of this method is that if the upload creation request succeeds, the file was transferred in a single request without additional round trips.
 
-A possible drawback is that the client might be unable to resume an upload. If an upload is interrupted before the client received a `104 (Upload Resumption Supported)` intermediate response with the upload URL, the client cannot resume that upload due to the missing upload URL. The intermediate response might not be received if the interruption happens too early in the message exchange, the server does not support sending the `104 (Upload Resumption Supported)` intermediate response, or an intermediary dropped the intermediate response. Without a 104 response, the client needs to either treat the upload as failed or retry the entire upload creation request if this is allowed by the application.
+A possible drawback is that the client might be unable to resume an upload. If an upload is interrupted before the client received a `104 (Upload Resumption Supported)` intermediate response with the upload URL, the client cannot resume that upload due to the missing upload URL. The intermediate response might not be received if the interruption happens too early in the message exchange, the server does not support resumable uploads at all, the server does not support sending the `104 (Upload Resumption Supported)` intermediate response, or an intermediary dropped the intermediate response. Without a 104 response, the client needs to either treat the upload as failed or retry the entire upload creation request if this is allowed by the application.
 
 ### Upgrading To Resumable Uploads
 
-Optimistic upload creation allows clients and servers to automatically upgrade non-resumable uploads to resumable ones. In a non-resumable upload, the file is transferred in a single request, usually `POST` or `PUT`, without any ability to resume from interruptions. The client can offer the server to upgrade such a request to a resumable upload by adding the `Upload-Complete: ?1` header field to the original request. The request is not changed otherwise.
+Optimistic upload creation allows clients and servers to automatically upgrade non-resumable uploads to resumable ones. In a non-resumable upload, the file is transferred in a single request, usually `POST` or `PUT`, without any ability to resume from interruptions. The client can offer the server to upgrade such a request to a resumable upload (see {{feature-detection}}) by adding the `Upload-Complete: ?1` header field to the original request. The request is not changed otherwise.
 
 A server that supports resumable uploads at the target URI can create a resumable upload resource and send its upload URL in a `104 (Upload Resumption Supported)` intermediate response for the client to resume the upload after interruptions. A server that does not support resumable uploads or does not want to upgrade to a resumable upload for this request ignores the `Upload-Complete: ?1` header. The transfer then falls back to a non-resumable upload without additional cost.
 
@@ -536,7 +536,7 @@ This upgrade can also be performed transparently by the client without the user 
 
 ## Careful Upload Creation
 
-With a "careful upload creation" the client sends an empty upload creation request without including any file data. Upon successful response reception, the client can use the included upload URL to transmit the file data ({{upload-appending}}) and resume the upload at any stage if an interruption occurs. The client should inspect the response for the `Upload-Limit` header field, which would indicate limits applying to the remaining upload procedure.
+For a "careful upload creation" the client knows that the server supports resumable uploads and sends an empty upload creation request without including any file data. Upon successful response reception, the client can use the included upload URL to transmit the file data ({{upload-appending}}) and resume the upload at any stage if an interruption occurs. The client should inspect the response for the `Upload-Limit` header field, which would indicate limits applying to the remaining upload procedure.
 
 The retransmission of file data or the ultimate upload failure that can happen with an "optimistic upload creation" is therefore avoided at the expense of an additional request that does not carry file data.
 
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -512,6 +512,34 @@ If the client knows the integrity digest of the content from an upload creation 
 
 The server might process the uploaded data and make its results available in another resource during or after the upload. This subsequent resource is different from the upload resource created by the upload creation request ({{upload-creation}}). The subsequent resource does not handle the upload process itself, but instead facilitates further interaction with the uploaded data. The server MAY indicate the location of this subsequent resource by including the `Content-Location` header field in the informational or final responses generated while creating ({{upload-creation}}), appending to ({{upload-appending}}), or retrieving the offset ({{offset-retrieving}}) of an upload. For example, a subsequent resource could allow the client to fetch information extracted from the uploaded data.
 
+# Upload Strategies
+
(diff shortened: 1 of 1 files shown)


guoye-zhang:
Are "Optimistic Upload Creation" and "Upgrading To Resumable Uploads" the exact same way?

Maybe the 3rd way could be checking `Upload-Limit` first before uploading?

Acconut:
> Are "Optimistic Upload Creation" and "Upgrading To Resumable Uploads" the exact same way?

The way they are described right now they are similar, yes. In theory, an optimistic upload creation can also be used where the client doesn't upload the entire file, but just a large part of it (think 500 MB of a 10 GB file). When upgrading a non-resumable request to a resumable upload, this act of division is not allowed. But I am not sure if we should go into that much detail here. If not, then we can merge these two sections.

> Maybe the 3rd way could be checking `Upload-Limit` first before uploading?

Upload limits are a good point! With the careful upload, the client also knows the limits before sending the first PATCH request. The optimistic client must be prepared to receive a 104 response with limits that it has to adapt to.

We can mention that a careful client may also send an OPTIONS request before hand if it deems that necessary.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -512,6 +512,34 @@ If the client knows the integrity digest of the content from an upload creation
...
 
+# Upload Strategies
+
+The definition of the upload creation request ({{upload-creation}}) provides the client with flexibility to choose whether the file is fully or partially transferred in the first request, or if no file data is included at all. The decision which behavior is best largely depends on the client's capabilities, its intention to avoid data transmission, and its knowledge about the server's support for resumable uploads.
+
+The following subsections describe three different, typical upload strategies that are suited for common environments. Note that these modes are never explicitly communicated to the server and clients are not required to stick to one strategy, but can mix and adapt them to their needs.
+
+## Optimistic Upload Creation
+
+An "optimistic upload creation" can be used if the client knows or assumes with a high degree of confidence that the server supports resumable uploads under the target URI and that the client is capable of handling and processing interim responses. The upload creation request then includes the full file because the client assumes that the file will be transferred without interruptions.
```
smatsson, line 523: >... because the client assumes that the file will be transferred without interruptions.

> ... because the client assumes that the file will be transferred without interruptions or resumed if an interruption occurs.

smatsson (commented):
Just a minor nit. Otherwise I think this is a very good addition and agree with @guoye-zhang   that we should add some text with regards to the `Upload-Limit`.

Proposal for optimistic: The client needs to adapt to the `Upload-Limit` header provided in the `104 (Upload Resumption Supported)` intermediate response and fail the request if the file is larger than the maximum allowed size.

Proposal for Careful: The client can use the  `Upload-Limit` header provided in the `104 (Upload Resumption Supported)` to verify that the file is allowed to be uploaded before sending any data.

guoye-zhang:
I think upload-limit can just be enforced on the server side. If it sends a 104 and a 5xx response, the client would automatically retry. We have deployed a client-side implementation that is not aware of the upload-limit header field.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -512,6 +512,34 @@ If the client knows the integrity digest of the content from an upload creation
...
 
+# Upload Strategies
+
+The definition of the upload creation request ({{upload-creation}}) provides the client with flexibility to choose whether the file is fully or partially transferred in the first request, or if no file data is included at all. The decision which behavior is best largely depends on the client's capabilities, its intention to avoid data transmission, and its knowledge about the server's support for resumable uploads.
```
martinthomson, line 517:
```suggestion
The definition of the upload creation request ({{upload-creation}}) provides the client with flexibility to choose whether the file is fully or partially transferred in the first request, or if no file data is included at all. Which behavior is best largely depends on the client's capabilities, its intention to avoid data transmission, and its knowledge about the server's support for resumable uploads.
```

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -512,6 +512,34 @@ If the client knows the integrity digest of the content from an upload creation
...
+
+The benefit of this method is that if the upload creation request succeeds, the file was transferred in a single request avoiding additional round trips. A drawback is that the client may be restricted in its ability to resume that upload. If the upload was interrupted before the client received a `104 (Upload Resumption Supported)` intermediate response with the upload URL, the client cannot resume that upload due to the missing upload URL. The intermediate response might not be received if the interruption happens too early in the message exchange, the server does not support sending the `104 (Upload Resumption Supported)` intermediate response, or an intermediary dropped the intermediate response. The client must either ultimately fail the entire upload or retry the upload creation request if this is allowed by the application at the expense of transferring file data again.
```
martinthomson, line 525:
```suggestion
The benefit of this method is that if the upload creation request succeeds, the file was transferred in a single request without additional round trips.

A possible drawback is that the client might be unable to resume an upload. If an upload is interrupted before the client received a `104 (Upload Resumption Supported)` intermediate response with the upload URL, the client cannot resume that upload due to the missing upload URL. The intermediate response might not be received if the interruption happens too early in the message exchange, the server does not support sending the 104 (Upload Resumption Supported) intermediate response, or an intermediary dropped the intermediate response. Without a 104 response, the client needs to either treat the upload as failed or retry the entire upload creation request if this is allowed by the application.
```

I'll note that if the 104 doesn't have time to arrive, then the client isn't really losing much progress.  As long as they are able to start over, this seems like a very reasonable approach.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -512,6 +512,34 @@ If the client knows the integrity digest of the content from an upload creation
...
+
+The "optimistic upload creation" attempts to upload the file in one request at the potential expense of either not finishing the upload at all or re-transmitting file data again.
+
```
martinthomson, line 528:
```suggestion
```

This is redundant.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -512,6 +512,34 @@ If the client knows the integrity digest of the content from an upload creation
...
+To perform an upgrade, the client adds the `Upload-Complete: ?1` header field to the original request, which attempts to transmit the entire file in one request. A server that supports resumable uploads at the target URI, can create a resumable upload resource and send its upload URL in a `104 (Upload Resumption Supported)` intermediate response for the client to resume the upload after interruptions. On the other hand, if the server does not support resumable uploads or does not want to upgrade to a resumable upload for this request, it can ignore the `Upload-Complete: ?1` header and treat the request as a non-resumable upload, where the client cannot recover from interruptions.
```
martinthomson, line 539:
```suggestion
To perform an upgrade, the client adds the `Upload-Complete: ?1` header field to the original request, which attempts to transmit the entire file in one request. A server that supports resumable uploads at the target URI, can create a resumable upload resource and send its upload URL in a 104 (Upload Resumption Supported) intermediate response for the client to resume the upload after interruptions.

A server that does not support resumable uploads or does not want to upgrade to a resumable upload for this request can ignore the `Upload-Complete: ?1` header and treat the request as a non-resumable upload.
```

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -512,6 +512,34 @@ If the client knows the integrity digest of the content from an upload creation
...
+
+The "optimistic upload creation" attempts to upload the file in one request at the potential expense of either not finishing the upload at all or re-transmitting file data again.
+
+## Careful Upload Creation
+
+With a "careful upload creation" the client sends an empty upload creation request without including any file data. Upon successful response reception, the client can use the included upload URL to transmit the file data ({{upload-appending}}) and resume the upload at any stage if an interruption occurs. The retransmission of file data or the ultimate upload failure that can happen with an "optimistic upload creation" is therefore avoided at the expense of an additional request that does not carry file data.
+
+This approach best suited if the client cannot receive intermediate responses, e.g. due to a limitation in the provided HTTP interface, or if large files are transferred where the cost of the additional request is miniscule compared to the effort of transferring the large file itself.
+
+## Upgrading To Resumable Uploads
```
martinthomson, line 535: This seems indistinguishable from the optimistic approach to me.  The use of Upload-Complete might distinguish it, but I don't see how these are functionally different.

martinthomson (commented):
Some editorial suggestions, but one big question:

What is the material difference between the first and third strategy?  Could these be the same strategy?

Acconut:
Thanks for all of your comments! I updated the PR to put the section about upgrading to resumable uploads under the optimistic upload creation since they are similar. I also added a brief mention of Upload-Limit and shortened the text a bit.

Let me know if you think the text needs more adjustments.

draft-ietf-httpbis-resumable-upload.md
Acconut: You're correct. I merged them.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -512,6 +512,36 @@ If the client knows the integrity digest of the content from an upload creation
...
 
+# Upload Strategies
+
+The definition of the upload creation request ({{upload-creation}}) provides the client with flexibility to choose whether the file is fully or partially transferred in the first request, or if no file data is included at all. Which behavior is best largely depends on the client's capabilities, its intention to avoid data re-transmission, and its knowledge about the server's support for resumable uploads.
+
+The following subsections describe two typical upload strategies that are suited for common environments. Note that these modes are never explicitly communicated to the server and clients are not required to stick to one strategy, but can mix and adapt them to their needs.
+
+## Optimistic Upload Creation
+
+An "optimistic upload creation" can be used if the client assumes that the server supports resumable uploads under the target URI and that the client is capable of handling and processing interim responses. The upload creation request then includes the full file because the client anticipates that the file will be transferred without interruptions or resumed if an interruption occurs.
```
guoye-zhang, line 523: > assumes that the server supports resumable uploads under the target URI

This seems to be a bit in conflict with the "Upgrading To Resumable Uploads" subsection now that they are merged. Maybe we can say "assumes that the server supports resumable uploads under the target URI, or has no knowledge of the server support", but then that sentence doesn't really say anything.

"Careful Upload Creation" is where the server must support resumable uploads. "Optimistic Upload Creation" is OK if the server doesn't.

Acconut: Yes, the client's knowledge about the server support was a difference of the two sections at first. Now that they are merged, the text gets a bit more unwieldy, but a added https://github.com/httpwg/http-extensions/commit/de5992146b5957168dc2c7e0af2a2dbe64086ab5 to correct the statements about the client's knowledge. What do you think? Is that sufficient or do we need more significant changes?

guoye-zhang: Looks good, thanks

guoye-zhang (approved):

Issue #2833, opened 2024-07-18
Resumable Upload: Discovering upload limits upfront


Acconut:
## Problem

#2747 introduced `Upload-Limit` for the server to indicate limits on the upload resource (e.g. maximum size). These limits are announced to the client during upload creation in the informational or final responses. However, the client cannot discover these limits without attempting to create an upload resource. A client may want to avoid creating an upload resource if it knows the the server limits cannot be satisfied.

## Possible solution

A possible solution is to allow clients to send an OPTIONS request upfront to the URL from upload creation. The server might then include the Upload-Limit header in its response, so that the client knows them before trying to create an upload resource. However, a client is not required to send this OPTIONS request upfront.

This would also be handy in a possible future where browsers natively implement resumable uploads. The response for a CORS preflight request might then already include information whether the endpoints supports resumable uploads and, if so, under which constraints. The browser can use this data to optimize the uploads.

smatsson:
Early thought: Would it be beneficial to include some kind of marker that this endpoints supports resumable upload in the OPTIONS response? For client's that does not support upgrading requests (e.g. due to lack of support for 1xx responses) who whishes to know upfront if the upload can be split (Upload-Complete: ?0) or if a regular upload (without resumability) should be performed.

Acconut:
Interesting idea! The presence of the Upload-Limit header could be used as an indicator for support of resumable uploads. In theory, its value can be an empty dictionary as well if the server does not impose any limits.
If the response does not include the Upload-Limit header, the client can either attempt an upgradable upload or directly fall back to regular uploads.

LPardue:
https://httpwg.org/specs/rfc8941.html#rfc.section.3.2

> As with Lists, an empty Dictionary is represented by omitting the entire field. This implies that fields defined as Dictionaries have a default empty value.

smatsson:
Too bad we can't use the empty dictionary header as an indicator. It's probably an edge case anyway that a client, not knowing the support of the server, would issue an OPTIONS request to find out? Or maybe not, given CORS? What do others think?

Acconut:
Thanks for the pointer, Lucas! If we cannot use an empty value, a server could also fill in a dummy value, e.g. `Upload-Limit: min-size=0`. This is only necessary if the server does not apply any limit at all (not even `max-size`), which I suppose is very rare in reality. 

[...]

PR #2911, opened 2024-10-09
Resumable Upload: Discover upload limits through OPTIONS request

Acconut:
This change allows client to discover support for resumable uploads and potential limits via an OPTIONS request.

Closes https://github.com/httpwg/http-extensions/issues/2833.

A brief comment on this sentence:

> The limits announced in an `OPTIONS` response SHOULD NOT be less restrictive than the limits applied to an upload once the upload resource has been created.

Once the server has created an upload resource it might have more information available (user details, file metadata etc) and apply different limits than it would have known upfront for an OPTIONS request where this information was not available. We allow this difference but recommend that the upfront announced limits are not looser than the limits after the upload resource has been created. The client can then assume that it can create an upload resource if it obeys to the upfront announced limits even if looser limits will then apply later. Does this make sense? Or is this addition rather unnecessary?

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -437,6 +437,8 @@ The `Upload-Limit` response header field indicates limits applying the upload re
 
 When parsing this header field, unrecognized keys MUST be ignored and MUST NOT fail the parsing to facilitate the addition of new limits in the future.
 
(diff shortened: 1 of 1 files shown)


guoye-zhang:
Is OPTIONS a required feature on the server side? What about `OPTIONS *` requests?

Acconut:
> Is OPTIONS a required feature on the server side?

Yes, I think a server should be required to support this. If it only wants to provide resumable uploads for specific users, it does not have to announce this for unauthorized users.

> What about `OPTIONS *` requests?

If a server supports resumable uploads for all target URIs, then it should respond with `Upload-Limit` to an `OPTIONS *` request. But I am unsure how often this actually happens.

Issue #2834, opened 2024-07-18
Resumable Upload: Requesting digests from server


Acconut:
## Problem

https://github.com/httpwg/http-extensions/pull/2748 added some guidelines on how a client may include integrity fields to upload requests for protecting the integrity of a single request or the entire upload. However, this either requires the client to compute the digest upfront (before sending a request or starting the upload), or the use of request trailers, which is not very widespread and not always possible (e.g. the [Fetch API does not support it currently](https://github.com/whatwg/fetch/issues/981)).

## Possible solution

Instead of requiring the client to compute the digest upfront, the server and client could calculate the digests while the upload is ongoing. Once the upload is completed, the server can send its digest to the client in the response to the request that completed the upload. The client can then compare the server's digest to its computed value. If they don't match, it can raise an error to the end user and decide to not continue using the uploaded data.

The client should indicate its interest in the server's digest when an upload is created, so the server is not required to compute the digest if the client has no interest. I am wondering if the `Want-Repr-Digest` from [RFC 9530](https://www.rfc-editor.org/rfc/rfc9530.html#name-integrity-preference-fields) can be used for this. Can a client include `Want-Repr-Digest` in the first upload creation request, but the server delivers the digest potentially later in a response for a subsequent request? An upload may be spread across multiple requests.

smatsson:
Should `Want-Content-Digest` also be supported? For multi request uploads, there's really no use to continue if request 3 of 100 got corrupted.

Acconut:
As @LPardue mentioned in https://github.com/httpwg/http-extensions/pull/2748#discussion_r1668595785, `Want-Content-Digest` is not directly applicable here because it asks the server to send the digest corresponding to the response content, not the received request content.

> For multi request uploads, there's really no use to continue if request 3 of 100 got corrupted.

Good point that a client may validate digests while a multi-request upload is progressing. A server could respond with the latest digest of the entire uploaded data (not just the data from the last request) using `Repr-Digest` after POST and PATCH requests. The client can then validate and decide to abort if the digests don't match.

This usage would be in line with the PATCH example in [section B.9. of RFC 9530](https://www.rfc-editor.org/rfc/rfc9530.html#name-digest-with-patch). The server responds with the digest of the representation after the PATCH has been applied. The example also notes that the server is not required to respond with the representation itself and could also have returned a 204 (as is done for resumable uploads):

> Note that a 204 No Content response without content, but with the same Repr-Digest field-value, would have been legitimate too.

[...]

PR #2917, opened 2024-10-15
Resumable Upload: Requesting digests from server using `Want-Repr-Digest`

Acconut:
Closes https://github.com/httpwg/http-extensions/issues/2834.

A client can express its interest in digests using `Want-Repr-Digest` during upload creation. The server can then respond with `Repr-Digest` after POST and PATCH requests to share its digest with the client. The client computes its digest as well during the upload and compares the digest to spot issues. The representation digest of an upload resource covers the entire bytes that have been uploaded so far (i.e. right until the upload offset).

@ioggstream @LPardue Is this in line with the semantics of RFC 9530?


guoye-zhang (approved):

ioggstream:
@Acconut Need a week for the review :)

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -510,10 +511,12 @@ Unlike `Content-Encoding` (see {{Section 8.4.1 of HTTP}}), `Transfer-Encoding` (
...
 
+In the other case, that the client does not know the integrity digest of the entire data upfront, it can ask the server to compute the integrity digests during the upload and send them back to the client. The client expresses its interest using the `Want-Repr-Digest` header field with its preferred algorithms when creating an upload ({{upload-creation}}). The server SHOULD compute the representation digests using the preferred algorithms once the upload is complete and include the corresponding `Repr-Digest` header field in the response. Alternatively, the server MAY compute the digest continuously during the upload and include the `Repr-Digest` header field in responses to upload creation ({{upload-creation}}) and upload appending requests ({{upload-appending}}) even when the upload is not completed yet. This allows the client to simultaneously compute the digest of the transmitted upload data, compare its digest to the server's digest, and spot data integrity issues. If an upload is spread across multiple requests, data integrity issues can be found even before the upload is fully completed.
```
ioggstream, line 518:
```suggestion
When creating an upload (see {{upload-creation}}),
the client MAY ask the server to compute and return the integrity digests using a `Want-Repr-Digest` field conveying the preferred algorithms.
The response SHOULD include at least one of the requested digests,
but MAY not include it. <!-- If the server is overloaded -->

The server SHOULD compute the representation digests using the preferred algorithms once the upload is complete and include the corresponding `Repr-Digest` header field in the response. Alternatively, the server MAY compute the digest continuously during the upload and include the `Repr-Digest` header field in responses to upload creation ({{upload-creation}}) and upload appending requests ({{upload-appending}}) even when the upload is not completed yet. This allows the client to simultaneously compute the digest of the transmitted upload data, compare its digest to the server's digest, and spot data integrity issues. If an upload is spread across multiple requests, data integrity issues can be found even before the upload is fully completed.
```

ioggstream:
Preliminary feedback:

1. I'd simplify the language/split text on multiple lines;
2. Describe processing in separate subsections, e.g.

```
## Reply on finish

...


## Reply in transit

...

```

Acconut:
Thank you for the feedback, Roberto. I added subsections for representation and content digests in https://github.com/httpwg/http-extensions/pull/2917/commits/664679150c9ece5f5f5cb3b89b3b398fb225841b. I think this division helps the reader and also makes sense since representation digests can cover the entire upload, while content digests only cover one request. Let me know if you think differently.

PR #2920, opened 2024-10-17
query: tune language about redirects, add context about Location; add examples (fixes #1916)

### LINKED ISSUE #1916: Method rewriting and Redirects
>  In some cases, the server may choose to respond indirectly to the QUERY request by returning a 3xx Redirection with a Location header field specifying an alternate Request URI from which the results can be retrieved using an HTTP GET request.

Some 3xx will rewrite the method, but others won't.

reschke:


mnot (approved):

reschke:
@MikeBishop - are you ok with this as well? We can always tune later...

reschke:
fixed via 066061d2b3ba6bede757c837cd754279e1560d6b

Issue #2855, opened 2024-08-05
connect-tcp: require capsules


DavidSchinazi:
As of draft-ietf-httpbis-connect-tcp-05, connect-tcp now optionally supports the capsule protocol. However, this optionality adds non-trivial complexity to implementations. As currently written, if a client wishes to use capsules, the server can reject the use of capsules, and now the client needs to cache that this server doesn't support them. However it's undefined for how long the client should cache this information, and within what scope. We could potentially figure out recommendations for all these questions, but fundamentally it might be best to instead make capsules mandatory for connect-tcp, like they are for connect-udp and connect-ip. That would simplify implementations and improve consistency.

tfpauly:
Individually, I think it would be interesting to try out this direction — probably based on implementation experience for how easy it is to add.

On one hand, this makes connect-tcp less of a trivial drop-in. On the other, it opens up more possible use cases.

bemasc:
I heard a good deal of support for this direction in the HTTPBIS meeting.

PR #2921, opened 2024-10-17
Make Capsule Protocol support mandatory for servers

bemasc:
This increases the complexity of a minimal server implementation and decreases the complexity of a rich client implementation.

Addresses #2855


tfpauly (approved):

DavidSchinazi (approved):
Thanks!

DavidSchinazi:
Another option would be to say that connect-tcp always uses capsules, that would reduce complexity even further

bemasc:
> Another option would be to say that connect-tcp always uses capsules, that would reduce complexity even further

That would reduce the complexity of a rich server implementation, but it would increase the complexity of a minimal client implementation.  I don't think that is a wise trade.

DavidSchinazi:
Do we think minimal client implementations will exist? I was imagining that they'd use regular CONNECT

LPardue:
The spec probably wants more language to describe what a client should do when a server does not enable the capsule protocol, since thst a failure mode for this optional behaviour.

bemasc:
@DavidSchinazi Template-based configuration is valuable with or without the capsule protocol.  Also, if you're not actually using capsules, it adds bandwidth overhead without any benefit.

@LPardue We don't usually say "the server MUST $X, but if it doesn't the client SHOULD $Y".  Is there a reason why that is necessary in this case?

LPardue:
It seems weird to me that a server that the server echos something back that the client doesn't check.  

Since the server has a MUST requirement to process whatever format the client tells it, we could just drop the response header and state that a 2xx is the signal that server accepted connect-tcp with or without capsules.

bemasc:
The response header is for the benefit of intermediaries.  Clients don't rely on it.  This is also true for CONNECT-UDP.

LPardue:
That's a good point and might be worth saying explicitly in the text.

However, we do have a pattern in RFC9298 that levies requirements on request and response message and says "if any of these requirements are not met, the client MUST treat the procying attempt as failed and abort the request". Since connect-tcp levies a MUST about capsule-protocol on the response message, using similar text seems consistent to me.

Issue #2874, opened 2024-08-28
Move `upload resource` and `target resource` to Conventions and Definitions


ioggstream:
## Question

I suggest to move "upload resource" and "target resource" to "Conventions and Definitions".

While it is OK to refer to the final resource as the "target resource", it  may have a slightly different meaning wrt HTTP, thing that's probably worth mentioning in some way.

## Note

In other specs we have "Notational Conventions", is there a specific reason for having a different section name?

ioggstream:
e kept the name "Conventions and Definitions" from the I-D template (https://github.com/martinthomson/internet-draft-template/blob/main/draft-todo-yourname-protocol.md#conventions-and-definitions)
 
Being on Martin's template is a good reason to leave it as it is :)

PR #2929, opened 2024-10-28
Resumable Upload: Editorial pass

Acconut:
The document has grown organically over time as features and clarifications have been added. Unfortunately, this makes the draft harder to read and less structured.

This draft attempts to improve the situation by
- placing a definition of "upload resource" and "target resource" at the beginning of the document (closes https://github.com/httpwg/http-extensions/issues/2874).
- adding a definition of a "file" and using it consistently throughout the document (we avoided using the term before because resumable uploads can also upload from streaming resources and not only files on disk, but the new definition explicitly mentions that a file can also be a streaming resource).
- introducing a section dedicated to the upload resource, which is the main way through which client interact with resumable uploads (the section houses the description for upload offset, length, completeness, as well as the HEAD, PATCH, DELETE requests).
- splitting the requirements for request handling into client-side and server-side requirements to make their sections more structured.

This PR is not yet completed, but any feedback is welcome.


Acconut:
Closing this in favor of smaller PRs that are easier to review. See https://github.com/httpwg/http-extensions/pull/2993 and https://github.com/httpwg/http-extensions/pull/2992

PR #2930, opened 2024-10-30
no-vary-search: Allow more efficient caches by assuming consistency

### LINKED ISSUE #2908: No-Vary-Search: Handling multiple matches and efficient cache implementation
Migrated from https://github.com/WICG/nav-speculation/issues/210.

The most active discussion is at the bottom of that issue -- permitting cache implementations to assume a single non-empty `No-Vary-Search` field value per path in order to allow simpler & more efficient implementations, and missing if this isn't the case.

jeremyroman:
For most known use cases, a particular resource is consistently invariant to the same query parameters (i.e., either only certain parameters are ever meaningful, or certain parameters are always meaningless).

In such cases, it is possible to avoid searching several stored responses and take a more efficient approach. This shortcut might be desirable in, for example, browser disk cache implementations, where the number of accesses is known to affect cache latency significantly.

Implementations are not required (neither servers nor caches) to take this shortcut; the worst consequence is an avoidable cache miss. Incorrect stored responses are never used.

Resolves #2908.

### DIFF
--- a/draft-ietf-httpbis-no-vary-search.md
+++ b/draft-ietf-httpbis-no-vary-search.md
@@ -35,6 +35,8 @@ author:
     email: jbroman@chromium.org
 
 normative:
+  HTTP: RFC9110
+  HTTP-CACHING: RFC9111
   FETCH:
    target: https://fetch.spec.whatwg.org/
    title: Fetch Living Standard
@@ -389,6 +391,36 @@ So, for example, given any non-default value for `No-Vary-Search`, such as `No-V
   </dt>
   <dd><tt>+</tt> and <tt>%20</tt> are both parsed as U+0020 SPACE</dd>
 </dl>
+
+# Caching
+
+If a cache {{HTTP-CACHING}} implements this specification, the presented target URI requirement in {{Section 4 of HTTP-CACHING}} is replaced with:
+
+* one of the following:
+  * the presented target URI ({{Section 7.1 of HTTP}}) and that of the stored response match, or
+  * the presented target URI and that of the stored response are equivalent modulo search variance ({{comparing}}), given the variance obtained ({{obtain-a-url-search-variance}}) from the stored response.
+
+Servers SHOULD send no more than one distinct non-empty value for the `No-Vary-Search` field in response to requests for a given pathname.
+
+Cache implementations MAY fail to reuse a stored response whose target URI matches _only_ modulo URL search variance, if the cache has more recently stored a response which:
+
+* has a target URI which is equal to the presented target URI, excluding the query, and
+* has a non-empty value for the `No-Vary-Search` field, and
+* has a `No-Vary-Search` field value different from the stored response being considered for reuse.
+
+{:aside}
+> Caches aren't required to reuse stored responses, generally. However, the above expressly empowers caches to, if it is advantageous for performance or other reasons, search a smaller number of stored responses. Such a cache might take steps like the following to identify a stored response (before checking the other conditions in {{Section 4 of HTTP-CACHING}}):
+>
+> 1. Let exactMatch be cache\[presentedTargetURI\]. If it is a stored response that can be reused, return it.
+> 1. Let targetPath be presentedTargetURI, with query parameters removed.
+> 1. Let lastNVS be mostRecentNVS\[targetPath\]. If it does not exist, return null.
+> 1. Let simplifiedURL be the result of simplifying presentedTargetURI according to lastNVS (by removing query parameters which are not significant, and stable sorting parameters by key, if key order should be ignored).
+> 1. Let nvsMatch be cache\[simplifiedURL\]. If it does not exist, return null. (It is assumed that this was written when storing in the cache, in addition to the exact URL.)
+> 1. Let searchVariance be obtained ({{obtain-a-url-search-variance}}) from nvsMatch.
+> 1. If nvsMatch's target URI and presentedTargetURI are not equivalent modulo search variance ({{comparing}}) given searchVariance, then return null.
+> 1. If nvsMatch is a stored response that can be reused, return it. Otherwise, return null.
+>
+> Such implementations might "miss" some stored responses that could otherwise have been reused. It is therefore useful for servers to avoid sending different values for the `No-Vary-Search` field when possible.
 
 # Security Considerations
 


draft-ietf-httpbis-no-vary-search.md
```diff
@@ -390,6 +392,15 @@ So, for example, given any non-default value for `No-Vary-Search`, such as `No-V
   <dd><tt>+</tt> and <tt>%20</tt> are both parsed as U+0020 SPACE</dd>
 </dl>
 
+# Caching
+
+{:aside}
+> A more thorough discussion of how the above algorithms are used in HTTP caches {{HTTP-CACHING}} generally is likely merited.
+
+Servers should send no more than one distinct non-empty value for the `No-Vary-Search` field in response to requests for a given pathname. Cache implementations may fail to reuse a stored response whose target URI is not equal to the presented target URI ({{Section 7.1 of HTTP}}) even though the two are equivalent modulo URL search variance for the variance obtained ({{obtain-a-url-search-variance}}) from the stored response, if the cache has more recently stored a response with a target URI which is equal, excluding the query, and a different non-empty value for the `No-Vary-Search` field.
```
domenic, line 400: The final sentence here is so long. It is very hard to follow. And as noted, it's kind of modifying a caching integration that doesn't exist.

My attempt at clearing this up would be something like the following:

This specification modifies the "Constructing Responses from Caches" section of RFC 9911 by replacing the statement that a cache must not reuse a stored response unless "the presented target URI and that of the stored response match". Instead, cache implementations use the condition that **the cache URIs match** given the presented target URI and the stored response.

The **cache URIs match**, given a URL |url| and a stored response |response|, if the following algorithm returns true:

1. An implementation may perform the following step:
   1. If all of the following are true:
      - The cache has more recently stored a response |moreRecentResponse| with a target URI which is equal, excluding the query, to |url|;
      - |moreRecentResponse|'s `No-Vary-Search` field is non-empty; and
      - |moreRecentResponse|'s `No-Vary-Search` field does not equal |response|'s `No-Vary-Search` field,
     
      then return true if |response|'s target URI equals |url|. Otherwise return false.
1. Let |urlSearchVariance| be the result of obtaining a URL search variance given |response|.
1. Return whether |url| is equivalent modulo URL search variance to |response|'s target URI, given |urlSearchVariance|.

This only resolves my confusion with the long sentence, and doesn't resolve my confusion with how this actually helps the implementation. (See below.)

draft-ietf-httpbis-no-vary-search.md
```diff
@@ -390,6 +392,15 @@ So, for example, given any non-default value for `No-Vary-Search`, such as `No-V
...
+
+This permits cache implementations to reduce the number of accesses required to search for a response (to a single search for an exact match, plus one more if the cache has seen `No-Vary-Search` used for that path, using a simplified query string). Since such implementations might "miss" some stored responses that could otherwise have been reused, it is useful for servers to avoid sending different values where possible.
```
domenic, line 402: It's still not obvious to me how this works. Maybe you could spell it out with pseudo-code? And annotate the steps which depend on the optional step above? I tried to write some myself but got confused.

jeremyroman: I previously described it here: https://github.com/WICG/nav-speculation/issues/210#issuecomment-2364516756

Is that helpful enough? I'm trying not to mandate anything in particular, just make algorithms along those lines permissible.

draft-ietf-httpbis-no-vary-search.md
jeremyroman: Alright, explicitly integrated with RFC 9111. Not as bad as I'd feared, mostly because it just lists cases where a cache MUST NOT reuse a stored response (and therefore implies that everything else MAY be reused) -- at least, from my quick read.

I've made this a bit more descriptive than algorithmic but hopefully still respecting the spirit of your comment. PTAL.

draft-ietf-httpbis-no-vary-search.md
```diff
@@ -390,6 +392,36 @@ So, for example, given any non-default value for `No-Vary-Search`, such as `No-V
...
+>
+> 1. Let exactMatch be cache\[presentedTargetURI\]. If it is a stored response that can be reused, return it.
+> 1. Let targetPath be presentedTargetURI, with query parameters removed.
+> 1. Let lastNVS be mostRecentNVS\[targetPath\]. If it does not exist, return null.
+> 1. Let simplifiedURL be the result of simplifying presentedTargetURI according to lastNVS (by removing query parameters which are not significant, and stable sorting parameters by key, if key order should be ignored).
+> 1. Let nvsMatch be cache\[simplifiedURL\]. (It is assumed that this was written when storing in the cache, in addition to the exact URL.)
+> 1. Let searchVariance be obtained ({{obtain-a-url-search-variance}}) from nvsMatch.
+> 1. If nvsMatch's target URI and presentedTargetURI are not equivalent modulo URL search variance given searchVariance, then return null.
+> 1. If nvsMatch is a stored response that can be reused, return it. Otherwise, return null.
```
domenic, line 421: I'm a bit unclear on what nvsMatch could be, besides either "a stored response that can be reused, or null". The null case seems a bit broken (you perform various operations on nvsMatch before null-checking it).

draft-ietf-httpbis-no-vary-search.md
```diff
@@ -390,6 +392,36 @@ So, for example, given any non-default value for `No-Vary-Search`, such as `No-V
   <dd><tt>+</tt> and <tt>%20</tt> are both parsed as U+0020 SPACE</dd>
 </dl>
 
+# Caching
+
+If a cache {{HTTP-CACHING}} implements this specification, the presented target URI requirement in {{Section 4 of HTTP-CACHING}} should be replaced with:
```
domenic, line 397: should -> MUST? It feels a bit strong but if you're a cache implementing this spec then I think that this requirement is the whole point, so I guess it makes sense.

draft-ietf-httpbis-no-vary-search.md
```diff
@@ -390,6 +392,36 @@ So, for example, given any non-default value for `No-Vary-Search`, such as `No-V
...
+Servers SHOULD send no more than one distinct non-empty value for the `No-Vary-Search` field in response to requests for a given pathname.
+
+Cache implementations MAY fail to reuse a stored response whose target URI matches _only_ modulo URL search variance, if the cache has more recently stored a response which:
+
+* has a target URI which is equal to the presented target URI, excluding the query, and
+* has a non-empty value for the `No-Vary-Search` field, and
+* has a `No-Vary-Search` field value different from the stored response being considered for reuse.
+
+{:aside}
+> Caches aren't required to reuse stored responses, generally. However, the above expressly empowers caches to, if it is advantageous for performance or other reasons, search a smaller number of stored responses. Such a cache might take steps like the following pseudocode to identify a stored response (before checking the other conditions in {{Section 4 of HTTP-CACHING}}):
```
domenic, line 412: This doesn't really look like pseudocode so probably just "steps like the following".

draft-ietf-httpbis-no-vary-search.md
```diff
@@ -390,6 +392,36 @@ So, for example, given any non-default value for `No-Vary-Search`, such as `No-V
...
+>
+> 1. Let exactMatch be cache\[presentedTargetURI\]. If it is a stored response that can be reused, return it.
```
domenic, line 414: Should we italicize all the variables?

draft-ietf-httpbis-no-vary-search.md
```diff
@@ -390,6 +392,36 @@ So, for example, given any non-default value for `No-Vary-Search`, such as `No-V
   <dd><tt>+</tt> and <tt>%20</tt> are both parsed as U+0020 SPACE</dd>
 </dl>
 
+# Caching
+
+If a cache {{HTTP-CACHING}} implements this specification, the presented target URI requirement in {{Section 4 of HTTP-CACHING}} should be replaced with:
+
+* one of the following:
+  * the presented target URI ({{Section 7.1 of HTTP}}) and that of the stored response match, or
+  * the presented target URI and that of the stored response are equivalent modulo URL search variance, given the variance obtained ({{obtain-a-url-search-variance}}) from the stored response.
```
domenic, line 401: I know the cross-links are cumbersome rendered but I feel like at least one part of this section should cross-reference "equivalent modulo URL search variance".

draft-ietf-httpbis-no-vary-search.md
```diff
@@ -390,6 +392,36 @@ So, for example, given any non-default value for `No-Vary-Search`, such as `No-V
...
+> 1. Let exactMatch be cache\[presentedTargetURI\]. If it is a stored response that can be reused, return it.
+> 1. Let targetPath be presentedTargetURI, with query parameters removed.
+> 1. Let lastNVS be mostRecentNVS\[targetPath\]. If it does not exist, return null.
+> 1. Let simplifiedURL be the result of simplifying presentedTargetURI according to lastNVS (by removing query parameters which are not significant, and stable sorting parameters by key, if key order should be ignored).
+> 1. Let nvsMatch be cache\[simplifiedURL\]. (It is assumed that this was written when storing in the cache, in addition to the exact URL.)
+> 1. Let searchVariance be obtained ({{obtain-a-url-search-variance}}) from nvsMatch.
+> 1. If nvsMatch's target URI and presentedTargetURI are not equivalent modulo URL search variance given searchVariance, then return null.
+> 1. If nvsMatch is a stored response that can be reused, return it. Otherwise, return null.
+>
+> Such implementations might "miss" some stored responses that could otherwise have been reused. It is therefore useful for servers to avoid sending different values for the `No-Vary-Search` values when possible.
```
domenic, line 423:
```suggestion
> Such implementations might "miss" some stored responses that could otherwise have been reused. It is therefore useful for servers to avoid sending different values for the `No-Vary-Search` field when possible.
```

domenic (approved):
This is beautiful. I think it slots in nicely with the existing ecosystem and specification style.

I have various editorial suggestions, and we could do another round of review to double-check if you want, but I'm giving you the green checkmark since this is already quite nice as-is.

draft-ietf-httpbis-no-vary-search.md
jeremyroman: Sure, done. I'm just trying not to imply it's strict WHATWG steps, either. But meh, it's just an aside.

draft-ietf-httpbis-no-vary-search.md
jeremyroman: Added the missing "return if null". How unclear is it after that?

I'm assuming `cache` is a map of some kind from URI to stored response. In a real cache that could in theory be a multimap (but in browsers, it isn't). Whether a stored response can be reused is defined by RFC 9111 (and depends on Vary, freshness, and other such things).

I could add more detail but this I'm loathe to make this section too wordy as it's just a handwave about why the previous text exists at all.

draft-ietf-httpbis-no-vary-search.md
jeremyroman: Done.

draft-ietf-httpbis-no-vary-search.md
jeremyroman: Asides are already presented entirely in italics in the HTML rendering, so I think it would only add noise to the plaintext rendering.

draft-ietf-httpbis-no-vary-search.md
jeremyroman: Strictly speaking I don't think any uppercase (BCP 14) word makes sense because it's not a requirement on the implementation to replace the wording? Just changed it to "is replaced with" to make it sound more descriptive?

draft-ietf-httpbis-no-vary-search.md
jeremyroman: Done.

draft-ietf-httpbis-no-vary-search.md
domenic: I think this is good now, but I'm still not sure what nvsMatch could be in this step besides "a stored response that can be reused". I guess maybe it could be a stored response that _can't_ be reused? Maybe because it was stored but now has expired? That makes sense.

Anyway, I think your current phrasing strikes a good balance, and this was mostly for my own curiosity at what you were getting at with that exact sub-phrase.

domenic (approved):

draft-ietf-httpbis-no-vary-search.md
jeremyroman: Right, it might have expired, a request header included in `Vary` might not match, the request method might not allow it to be reused, it might have `Cache-Control: no-cache` and not validate successfully (notably different from `no-store`), etc.

PR #2931, opened 2024-10-30
no-vary-search: Fill in IANA Considerations section

jeremyroman:
The header field should be registered.


jeremyroman:
This is purely editorial.

PR #2932, opened 2024-11-01
Be explicit about the name of a shared group

hawkinsw:
Make the calculation of group responses more explicit with respect to the name of a shared group.


hawkinsw:
@mnot I really, _really_ appreciate the work that you have done on this draft. I am sorry for only chiming in now. Please feel free to ignore my small addition if you think that it is not helpful.

mnot:
Hi Will, thanks for the kind words. I'm going to leave this one out for now.

Issue #2860, opened 2024-08-07
query: Interaction between Accept-Query and Allow


hawkinsw:
Hello. I hope that these comments/questions are helpful. If not, please feel free to ignore them.

The QUERY extension says that a server may respond with an Accept-Query header whose value is a list of _one or more_ content types it supports for encoding queries. 

1. How does a server advertise its support for QUERY if it places no restriction on the content type? The grammar for the header value says that at least one content type must be specified. 
2. Should we specify (for the sake of explicitness _not_ correctness) how the Accept-Query header interacts/overlaps with Allow? In particular, does a 405 response that includes an Accept-Query header mean that QUERY does not need to appear in Allow?

Again, I hope that these questions are helpful. If they are, and there is a consensus on how to proceed, I would be more than happy to help draft some changed language. If these are not helpful, please just tell me to go away.

Thanks for your work on this interesting extension!

martinthomson:
@hawkinsw (hi Will!) asks a good question here.  

That made me ask something else:

1. Can we have an example of Accept-Query?
2. Can we have an explanation how Accept-Query compares to Accept in a response, and why there is a need for a new field for this purpose.

Of course, the real question is what Will asks: if the resource responds with 405 and lists a number of methods, how do you interpret the Accept/Accept-Query/Accept-Encoding fields of that response?  It is entirely reasonable for a resource to support different media types for different methods.

One approach would be to say that managing the matrix of combinations is too difficult.  These fields therefore only apply to the specific request that was made.  That leads to Accept-* having no real meaning for a 405 response.  Under that interpretation, I can't see any reason to define Accept-Query over a straight Accept.  [This](https://datatracker.ietf.org/doc/html/rfc9110#section-12.5.1-2) is the most that RFC 9110 says, which isn't much.

hawkinsw:
> @hawkinsw (hi Will!) asks a good question here.
> 

Hello back!!

> That made me ask something else:
> 
>     1. Can we have an example of Accept-Query?
> 
>     2. Can we have an explanation how Accept-Query compares to Accept in a response, and why there is a need for a new field for this purpose.

I was wondering this very thing, too. Read what comes next with the caveat that I am not an expert at HTTP semantics but have started added `QUERY` support to freenginx:

[...]

PR #2935, opened 2024-11-05
query: switch Accept-Query to SF (closes #2934)

### LINKED ISSUE #2934: query: format for Accept-Query field
1) "#mediatype (for consistency, for example, with PATCH: https://greenbytes.de/tech/webdav/rfc5789.html#accept-patch)
2) Structured Fields, because that's how we define new fields (and one could argue that we shouldn't make up reasons not to use that format)

My proposal is 2).

reschke:
Summary:

- use list of Tokens holding "type/subtype" media types (not Strings @mnot) - not that if we'd use HTTP's definition of type/subtype as tokes, this would allow "*" in both places (is this a problem in HTTP spec???)
- allows parameters

TODO:

- clarify that the Tokens do not hold full media types, which would include parameters
- clarity that these are not media ranges, thus no wildcards (but see above)

### DIFF
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -340,8 +340,8 @@ q=foo&amp;limit=10&amp;sort=-published
       <li>Media types do not exactly map to Tokens, for instance they
       allow a leading digit. In cases like these, the String format needs to
       be used.</li>
-      <li>The use of wildcards for anything except "*/*" or "type/*" is invalid
-      and will never match an actual media type.</li>
+      <li>The only supported uses of wildcards are "*/*", which matches any type, 
+      or "xxxx/*", which matches any subtype of the indicated type.</li>
       <li>The order of types listed in the field value is not significant.</li>
       <li>The only allowed format for parameters is String.</li>
     </ul>
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -357,7 +357,7 @@ q=foo&amp;limit=10&amp;sort=-published
     <artwork type="example">
 Accept-Query: "application/jsonpath", application/sql;charset="UTF-8"</artwork>
     <t>
-      Note that although the syntax appears to be similar to other
+      Although the syntax for this field appears to be similar to other
       fields, such as "Accept" (<xref target="HTTP" section="12.5.1"/>),
       it is a Structured Field and thus &MUST; be processed as specified in
       <xref target="STRUCTURED-FIELDS" section="4"/>.
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -337,7 +337,7 @@ q=foo&amp;limit=10&amp;sort=-published
     </t>
     <ul>
       <li>The choice of Token vs. String is semantically insignificant.</li>
-      <li>Media types do not exactly map to Tokens, for instance they do
+      <li>Media types do not exactly map to Tokens, for instance they
       allow a leading digit. In cases like these, the String format needs to
       be used.</li>
       <li>The use of wildcards for anything except "*/*" or "type/*" is invalid
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -336,7 +336,9 @@ q=foo&amp;limit=10&amp;sort=-published
       Note:
     </t>
     <ul>
-      <li>The choice of Token vs. String is semantically insignificant.</li>
+      <li>The choice of Token vs. String is semantically insignificant. That is,
+      recipients &MAY; convert Tokens to Strings, but &MUST-NOT; process them
+      differently based on the received type.</li>
       <li>Media types do not exactly map to Tokens, for instance they
       allow a leading digit. In cases like these, the String format needs to
       be used.</li>
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -329,24 +329,36 @@ q=foo&amp;limit=10&amp;sort=-published
       the specific query format media type(s) that may be used.
     </t>
     <t>
-      "Accept-Query" contains a list of media types (<xref target="HTTP" section="8.3.1"/>),
-      represented by a List Structured Header Field, containing
-      Token Items, optionally parameterized (<xref target="STRUCTURED-FIELDS" section="3"/>).
+      "Accept-Query" contains a list of media ranges (<xref target="HTTP" section="12.5.1"/>)
+      using "Structured Fields" syntax (<xref target="STRUCTURED-FIELDS"/>).
+      Media ranges are represented by a List Structured Header Field of either Tokens or
+      Strings, containing the media range value without parameters. Parameters,
+      if any, are mapped to Parameters of type String.
     </t>
     <t>
-      The order of types listed in the field value is not significant.
+      Note:
     </t>
+    <ul>
+      <li>The choice of Token vs. String is semantically insignificant.</li>
+      <li>Media types do not exactly map to Tokens, for instance they do
+      allow a leading digit. In cases like these, the String format needs to
+      be used.</li>
+      <li>The use of wildcards for anything except "*/*" or "type/*" is invalid
+      and will never match an actual media type.</li>
+      <li>The order of types listed in the field value is not significant.</li>
+      <li>The only allowed format for parameters is String.</li>
+    </ul>
     <t>
       Accept-Query's value applies to every URI on the server that shares the same path; in
       other words, the query component is ignored. If requests to the same resource return
-      different Accept-Query values, the most recently received fresh (per
-      <xref target="HTTP-CACHING" section="4.2"/>) value is used.
+      different Accept-Query values, the most recently received fresh value (per
+      <xref target="HTTP-CACHING" section="4.2"/>) is used.
     </t>
     <t>
       Example:
     </t>
     <artwork type="example">
-Accept-Query: application/jsonpath, application/sql;charset="UTF-8"</artwork>
+Accept-Query: "application/jsonpath", application/sql;charset="UTF-8"</artwork>
     <t>
       Note that although the syntax appears to be similar to other
       fields, such as "Accept" (<xref target="HTTP" section="12.5.1"/>),
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -340,7 +340,7 @@ q=foo&amp;limit=10&amp;sort=-published
       <li>Media types do not exactly map to Tokens, for instance they
       allow a leading digit. In cases like these, the String format needs to
       be used.</li>
-      <li>The only supported uses of wildcards are "*/*", which matches any type, 
+      <li>The only supported uses of wildcards are "*/*", which matches any type,
       or "xxxx/*", which matches any subtype of the indicated type.</li>
       <li>The order of types listed in the field value is not significant.</li>
       <li>The only allowed format for parameters is String.</li>
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -333,20 +333,25 @@ q=foo&amp;limit=10&amp;sort=-published
       if any, are mapped to Parameters of type String.
     </t>
     <t>
-      Note:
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -285,27 +285,35 @@ q=foo&amp;limit=10&amp;sort=-published
...
-<sourcecode type="abnf">
-Accept-Query = 1#media-type
-</sourcecode>
     <t>
-      The Accept-Query header field specifies a comma-separated listing of media
-      types (with optional parameters) as defined by
-      <xref target="HTTP" section="8.3.1"/>. <cref>field syntax currently discussed in <eref target="https://github.com/httpwg/http-extensions/issues/2860"/></cref>
+      "Accept-Query" contains a list of media types (<xref target="HTTP" section="8.3.1"/>),
+      represented by a List Structured Header Field, containing
+      Token Items, optionally parametrized (<xref target="STRUCTURED-FIELDS" section="3"/>).
```
zenomt, line 295: "parameterized" is misspelled here as "parametrized".

zenomt (approved):
LGTM other than the typo i mentioned.

draft-ietf-httpbis-safe-method-w-body.xml
reschke: ack

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -285,27 +285,35 @@ q=foo&amp;limit=10&amp;sort=-published
...
       the specific query format media type(s) that may be used.
     </t>
-<sourcecode type="abnf">
-Accept-Query = 1#media-type
-</sourcecode>
     <t>
-      The Accept-Query header field specifies a comma-separated listing of media
-      types (with optional parameters) as defined by
-      <xref target="HTTP" section="8.3.1"/>. <cref>field syntax currently discussed in <eref target="https://github.com/httpwg/http-extensions/issues/2860"/></cref>
+      "Accept-Query" contains a list of media types (<xref target="HTTP" section="8.3.1"/>),
```
zenomt, line 293: sorry, i missed this on first review. here you're saying "a list of media types" with reference to the "Media Types" section of RFC 9110. however, "Media Types" doesn't allow wildcards like `image/*`. for that you need "Media Ranges".

zenomt (changes requested):
doing a "request changes" to undo my approval, to address the "media types" vs "media-ranges" concern.

draft-ietf-httpbis-safe-method-w-body.xml
reschke: That's intentionally.

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -285,27 +285,35 @@ q=foo&amp;limit=10&amp;sort=-published
...
+      represented by a List Structured Header Field, containing
+      Token Items, optionally parameterized (<xref target="STRUCTURED-FIELDS" section="3"/>).
```
mnot, line 295: What is/are the type(s) of the parameters?

draft-ietf-httpbis-safe-method-w-body.xml
zenomt: a justification for the intentional "explicit media types, no wildcards" (either in the Issue or here) would be appreciated.

reschke: We discussed this in the other ticket: what's the use case? Can you give an example?

zenomt: sorry, i'm not finding where a decision to not allow wildcards was made. i saw you mention `*/*` in https://github.com/httpwg/http-extensions/issues/2860#issuecomment-2335436391 , and in https://github.com/httpwg/http-extensions/issues/2860#issuecomment-2452378251 you list as an open question
> how does a server say that it supports any format? Does this make any sense at all? We currently require one value, and IMHO we can stick with that

which i took to mean "at least one value".

a use case that comes immediately to mind is a fancy AI™-powered image search that can accept any image type (at least as "any" as a web browser that tells a server it accepts any image type):
```
Accept-Query: image/*
```

a possible use case for `*/*` could be a query-by-value of a collection/container resource: given any kind of serialization of any kind of object, answer my contained URI of a resource (and maybe metadata about it) if i happen to contain a resource matching that serialization and media type, or `204` if i don't.

draft-ietf-httpbis-safe-method-w-body.xml
zenomt: the `parameter-value` of a `media-type` or `media-range` can be a `token` or a `quoted-string`. in order to express media types and their parameters in a SF as commonly given, both token and quoted string need to be supported. quoted string is definitely needed because some media types' parameters (can) have spaces in them that mean something (for example, the `profile` parameter of `application/ld+json`).

draft-ietf-httpbis-safe-method-w-body.xml
reschke: I'm not convinced by these use cases, for "image/*" I would expect that more than the image would be needed, thus some kind of multipart would be used. For "*/*", the situation seems similar.

With that said, I'd propose the simplest possible way to get this resolved would be to indeed allow these values.

draft-ietf-httpbis-safe-method-w-body.xml
martinthomson: Though not likely useful, `q=0.9` would be sf-decimal .

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -324,26 +324,46 @@ q=foo&amp;limit=10&amp;sort=-published
...
-      The order of types listed by the Accept-Query header field is not significant.
+      Note:
     </t>
+    <ul>
+      <li>The choice of Token vs. String is semantically insignificant.</li>
+      <li>Media types do not exactly map to Tokens, for instance they do
+      allow a leading digit. In cases like these, the String format needs to
+      be used.</li>
+      <li>The use of wildcards for anything except "*/*" or "type/*" is invalid
+      and will never match an actual media type.</li>
```
martinthomson, line 347:
```suggestion
      <li>The only supported uses of wildcards are "*/*", which matches any type, 
      or "xxxx/*", which matches any subtype of the indicated type.</li>
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -324,26 +324,46 @@ q=foo&amp;limit=10&amp;sort=-published
...
-      <xref target="HTTP" section="8.3.1"/>. <cref>field syntax currently discussed in <eref target="https://github.com/httpwg/http-extensions/issues/2860"/></cref>
+      "Accept-Query" contains a list of media ranges (<xref target="HTTP" section="12.5.1"/>)
+      using "Structured Fields" syntax (<xref target="STRUCTURED-FIELDS"/>).
+      Media ranges are represented by a List Structured Header Field of either Tokens or
+      Strings, containing the media range value without parameters. Parameters,
+      if any, are mapped to Parameters of type String.
     </t>
     <t>
-      The order of types listed by the Accept-Query header field is not significant.
+      Note:
```
martinthomson, line 339: Is this really a note, or just a list of important considerations.  Does it need to be a list?

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -324,26 +324,46 @@ q=foo&amp;limit=10&amp;sort=-published
...
+      different Accept-Query values, the most recently received fresh value (per
+      <xref target="HTTP-CACHING" section="4.2"/>) is used.
+    </t>
+    <t>
+      Example:
+    </t>
+    <artwork type="example">
+Accept-Query: "application/jsonpath", application/sql;charset="UTF-8"</artwork>
+    <t>
+      Note that although the syntax appears to be similar to other
```
martinthomson, line 363:
```suggestion
      Although the syntax for this field appears to be similar to other
```

martinthomson (approved):

draft-ietf-httpbis-safe-method-w-body.xml
reschke: I thought a list would be more readable. Do you suggest to make it a plain paragraph?

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -324,26 +324,46 @@ q=foo&amp;limit=10&amp;sort=-published
...
+      Media ranges are represented by a List Structured Header Field of either Tokens or
+      Strings, containing the media range value without parameters. Parameters,
+      if any, are mapped to Parameters of type String.
     </t>
     <t>
-      The order of types listed by the Accept-Query header field is not significant.
+      Note:
     </t>
+    <ul>
+      <li>The choice of Token vs. String is semantically insignificant.</li>
```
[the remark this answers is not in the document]
reschke: Ok. Not sure about MAY/SHOULD/MUST though.

reschke:
@mnot , @martinthomson , @MikeBishop : "This branch cannot be rebased due to conflicts" - what's going on here?

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -321,26 +321,46 @@ q=foo&amp;limit=10&amp;sort=-published
...
-Accept-Query = 1#media-type
-</sourcecode>
     <t>
-      The Accept-Query header field specifies a comma-separated listing of media
-      types (with optional parameters) as defined by
-      <xref target="HTTP" section="8.3.1"/>. <cref>field syntax currently discussed in <eref target="https://github.com/httpwg/http-extensions/issues/2860"/></cref>
+      "Accept-Query" contains a list of media ranges (<xref target="HTTP" section="12.5.1"/>)
+      using "Structured Fields" syntax (<xref target="STRUCTURED-FIELDS"/>).
+      Media ranges are represented by a List Structured Header Field of either Tokens or
+      Strings, containing the media range value without parameters. Parameters,
```
zenomt, line 332: saying "Parameters, if any, are mapped to Parameters of type String" is both unnecessarily prescriptive and confusing. if i say `application/foo+json; x=3.1`, where parameter `x` is (or could be interpreted as) a SF decimal, i should (1) be allowed to say it that way (that is, i shouldn't be **forced** to say `application/foo+json; x="3.1"`), and (2) it's nobody else's business whether i, when receiving this field, treat the `3.1` as a string or a number. that should be between the definition & semantics of the media-type/range and my implementation.

how about words like
> "Parameters of the media-range, if any, are mapped to Structured Field Parameters. The media-range and its parameters \<bcp14>MUST\</bcp14> be parsable as a valid Structured Field."

perhaps using "expressed" or "serialized" in place of "mapped" in the first sentence.

reschke: The restriction is intentional. For instance, when not quoting values, surprising things may happen (integer vs floats etc). 

On-the-wire verbosity really is no issue here.

zenomt: media types & ranges have an established syntactic form & semantics, and an established ecosystem beyond HTTP. we want to use this field to tell clients what media types (or ranges of them) are acceptable for this method.

> The restriction is intentional. For instance, when not quoting values, surprising things may happen (integer vs floats etc).

i don't think this restriction is a _necessary_ one. i'm not convinced that guarding against a hypothetical insufficient implementation warrants restricting otherwise valid expressions of media types/ranges in a structured field. i think the only restrictions should be the minimum actually necessary so that an expression of a media type/range is **syntactically** valid as both a `media-range` and a structured field, thereby maintaining as much compatibility as possible with the media-type ecosystem (and all of the different normative and common forms of different media types) while conforming to Structured Field syntax.

draft-ietf-httpbis-safe-method-w-body.xml
reschke: @martinthomson - I believe @mnot prefers to explicitly restrict the allowed types to the bare minimum.

draft-ietf-httpbis-safe-method-w-body.xml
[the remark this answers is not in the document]
reschke: @martinthomson - please check now.

draft-ietf-httpbis-safe-method-w-body.xml
reschke: Turned into plain paragraphs(s).

reschke:
@mnot , @martinthomson , @MikeBishop - absent further comments I plain to "merge" this soonish.

"merge" in double quotes, as git does not let me. So I'll first create a new PR, link to that from here, and then close this one.

reschke:
Superseded by https://github.com/httpwg/http-extensions/pull/2966

PR #2945, opened 2024-11-11
Correct HTTP reference

mnot:


mikewest (approved):
Thanks! Looks like this snuck back in via https://github.com/httpwg/http-extensions/commit/455a360129ce47c5cebe15a99fb003a1e324f6a2. Skimming the rest of the doc, I don't see any additional references to the older RFCs.

Issue #2938, opened 2024-11-06
RFC 6265bis: clarify that cookie names are case sensitive


bagder:
I think the spec would benefit from getting this fact explicitly specified.  `FooBar` and `foobar` are two different cookies.

sbingler:
If you had asked me if this already existed I would have guessed yes. Seems like an obvious improvement

PR #2946, opened 2024-11-11
Clarify that cookie names are case-sensitive and may be empty

sbingler:
Closes: #2938
Closes: #2963

Adds an example to the Examples section to clarify and illustrate that cookie names are case-sensitive.
Additionally adds a clarification that cookie names could be empty to improve readability.


mikewest (approved):
LGTM. It'd be very nice if we could remove the notion of empty cookie names, but since we can't we should document them clearly. :/

sbingler:
Agreed

PR #2965, opened 2024-12-02
fix(resumable-upload): Fix Upload-Complete response header value for incomplete append

provokateurin:
If the client gives Upload-Complete=false then the server must respond with Upload-Complete=false and not Upload-Complete=true.


guoye-zhang (approved):

Acconut (approved):
Thank you for this correction!

PR #2967, opened 2024-12-08
query: Reference HTTP spec for terminology (closes #2953)

### LINKED ISSUE #2953: query: should reference RFC9110 terminology

reschke:


martinthomson (approved):
It's customary to put the terminology and BCP 14 block in the same section, but this works.

PR #2976, opened 2025-01-03
Standardize on "case-sensitive"

### LINKED ISSUE #2974: rfc6265bis: "case sensitive" vs "case-sensitive" (nit)
In draft-18, the first version is used once in the document, the second five times. Shouldn't it be written the same way in all six instances?

sbingler:
Closes #2974
Make every use of "case-sensitive" hyphenated.


mikewest (approved):
LGTM. Skimming through, I do think this is the only "case (in)sensitive" that wasn't hyphenated. Good catch!

Deanious04 (commented):
Not sure if this will work just yet

Deanious04 (approved):
Just hoping this works right now

Issue #2983, opened 2025-01-28
Make Lax-by-default optional


edgul:
RFC6265ter already aims to make [lax-by-default optional](https://datatracker.ietf.org/doc/html/draft-annevk-johannhof-httpbis-cookies-00#section-5.5.2-5.5.1) because in the absence of a `SameSite` attribute Firefox defaults to None and I'm told that Safari also defaults to None as well. It might be nice to include this as optional in 6265bis as well.

(Sorry for the lateness of this issue)

edgul:
https://github.com/httpwg/http-extensions/pull/2984

miketaylr:
I think this decision would benefit from multi-vendor discussion (some of which started at TPAC last year), and not changed at the last minute. Given the plans to fast-pursue 6265ter, what's the benefit of making the change now?

edgul:
I was unaware of plans to fast-pursue 6265ter, but I suppose my original concern was that if ter takes longer than expected to go beyond draft.  It just seemed like an easy change to make the soon-to-be-current spec more accurately reflecting the state of browser implementations in the meantime.

sbingler:
I don't believe it's a good idea to introduce a material change to 6265bis this late. As Mike said, we intend to quickly follow 6265bis's publication with a new draft.

It's worth exploring the deprecation of lax-by-default but that should be done on the next document.

LPardue:
Any material change now is a thin end of a wedge to other changes. 

As a member of the WG not deeply invested in a cookie implementation, I would appreciate the bis just getting published ASAP after the many years of work on it. Rather than extend the process further and further.

PR #2984, opened 2025-01-28
Change Lax-by-default to be optional

edgul:


sbingler:
As stated in #2983 I don't believe now is the time for this change. We can revisit it in the next draft.

PR #2987, opened 2025-01-29
Add TLS as an informative reference

### LINKED ISSUE #2982: TLS reference
_Comment by @fpalombini_

[Section 4.1.2.5](https://www.ietf.org/archive/id/draft-ietf-httpbis-rfc6265bis-19.html#section-4.1.2.5)

> (typically HTTP over Transport Layer Security (TLS) [HTTP])

I believe an informative reference to TLS should be added. Note that TLS is mentioned a couple of times throughout the doc (always as informative).

sbingler:
Closes #2982

Adds RFC8446 (TLS) as an informative reference and links to it where references are made

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -115,6 +115,7 @@ informative:
   RFC6265:
   RFC7034:
   RFC8446:
+     display: TLS13
   RFC9113:
   RFC9114:
   CSRF:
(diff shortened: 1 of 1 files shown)


mikewest:
HTTP references it as `TLS13`: https://datatracker.ietf.org/doc/html/rfc9110#:~:text=rfc793%3E.-,%5BTLS13%5D,%2C%20%3Chttps%3A//www.rfc%2Deditor.org/info/rfc8446%3E.,-%5BURI%5D. So, let's do that?

sbingler:
Sounds good

Issue #2980, opened 2025-01-28
SHOULD NOT


sbingler:
_Comment by @fpalombini_

[Section 3](https://www.ietf.org/archive/id/draft-ietf-httpbis-rfc6265bis-19.html#section-3):

> Origin servers SHOULD NOT fold multiple Set-Cookie header fields into a single header field.

I wonder why this is a SHOULD NOT and not a MUST NOT. I do see this was a SHOULD NOT in the original RFC 6265. If it's the case that it has been kept as a SHOULD NOT because existing implementations do it despite this recommendation, I suggest explicitly motivate it here (for example by adding a note, as it has been done in other sections of the document).

sbingler:
@mikewest 

Do you have any more context as to why this is a "SHOULD NOT" instead of a "MUST NOT"?

Looking at the mail list for RFC6265 I found https://mailarchive.ietf.org/arch/msg/http-state/FwQWLsmRynEwtCKxOgUOWjFa7gE/

>>>> [S3 P4] It seems like there should be an all-caps requirement here, e.g., that "Origin servers MUST NOT fold multiple Set-Cookie header fields into a single header field".
>>> 
>>> Origin servers certainly can fold multiple Set-Cookie header fields
>>> into a single header field if they desire.  However, doing so changes
>>> the semantics.  Given the bytes on the wire, there's no experiment we
>>> can run to see if the origin server did or did not perform this
>>> folding.  Therefore, a MUST-level requirement is inappropriate.
>> 
>> As I understand it, folding would cause multiple Set-Cooking header values to be considered instead as setting one cookie with many attributes.  Is there a situation where this is "acceptable or even useful", to quote RFC 2119?  It seems to me like folding is virtually guaranteed to be a bug.
> 
> Ok.  I've changed this to a SHOULD.  I suspect, actually, that this
> requirement already exists in the document (in an obtuse way) because
> the recommend grammar for servers won't produce a "," in the right
> syntactic surrounds to be a folded Set-Cookie header.

Sounds like it's this way because it's a bad idea but we can't check/enforce that it isn't.

mnot:
For what it's worth -- SHOULD vs MUST does not depend on whether it can be tested/enforced; it's whether there are exceptions to the rule.

mikewest:
[...]

PR #2988, opened 2025-01-30
[RFC6265bis] Harden requirements for the `Set-Cookie` profile.

mikewest:
This patch shifts several recommendations in Section 4's definition of a well-behaved `Set-Cookie` profile from SHOULD to MUST. This change makes server requirements less ambiguous, but does not change Section 5's definition of clients' obligations to gracefully handle servers' errors.

Closes httpwg/http-extensions#2980.

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -349,7 +349,7 @@ origin server can include multiple Set-Cookie header fields in a single response
 The presence of a Cookie or a Set-Cookie header field does not preclude HTTP
 caches from storing and reusing a response.
 
-Origin servers SHOULD NOT fold multiple Set-Cookie header fields into a single
+Origin servers MUST NOT fold multiple Set-Cookie header fields into a single
 header field. The usual mechanism for folding HTTP headers fields (i.e., as
 defined in {{Section 5.3 of RFC9110}}) might change the semantics of the Set-Cookie header
 field because the %x2C (",") character is used by Set-Cookie in a way that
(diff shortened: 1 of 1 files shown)


mikewest:
WDYT, @sbingler and/or @mnot? :)

mnot:
LGTM

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -607,15 +607,15 @@ as enhanced by {{RFC1123}}, Section 2.1. Thus, domain-value is a string of
 {{Section 2.3.2.1 of RFC5890}}.
 
 The portions of the set-cookie-string produced by the cookie-av term are
-known as attributes. To maximize compatibility with user agents, servers SHOULD
+known as attributes. To maximize compatibility with user agents, servers MUST
 NOT produce two attributes with the same name in the same set-cookie-string.
```
martinthomson, line 611: If there is a prohibition on folding multiple Set-Cookie fields into one, does this need to be reformulated?  (set-cookie-string, as I understand it, is the formulation for a single field value instance)

mikewest: This requirement covers attributes (the repeated `cookie-av` grammar) within a single field value instance. That is, servers ought not produce something like `Set-Cookie: name=value;SameSite=Strict;SameSite=Lax`. We could reword this as "... servers MUST NOT produce a `set-cookie-string` with more than one `cookie-av` of a given type", but it's not clear to me that that's any more helpful. 🤷 Do you have a suggestion about how we could word this more explicitly?

sbingler:
Overall looks good to me. Waiting for a resolution to @martinthomson 's comment before signing off on the PR

draft-ietf-httpbis-rfc6265bis.md
martinthomson: Ah, that was probably more of a misread on my part.  This seems good once you corrected my misconception.

sbingler (approved):
LGTM

PR #2991, opened 2025-02-03
Errata fix

sbingler:
Fixes erratum [6093](https://www.rfc-editor.org/errata/eid6093)


draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -349,11 +349,11 @@ origin server can include multiple Set-Cookie header fields in a single response
...
-Origin servers SHOULD NOT fold multiple Set-Cookie header fields into a single
-header field. The usual mechanism for folding HTTP headers fields (i.e., as
-defined in {{Section 5.3 of RFC9110}}) might change the semantics of the Set-Cookie header
-field because the %x2C (",") character is used by Set-Cookie in a way that
-conflicts with such folding.
+Origin servers SHOULD NOT combine multiple Set-Cookie header fields into a
+single header field. The usual mechanism for combining HTTP headers fields
+(i.e., as defined in {{Section 5.3 of RFC9110}}) might change the semantics of
+the Set-Cookie header field because the %x2C (",") character is used by
+Set-Cookie in a way that conflicts with such combining.
```
martinthomson, line 356: I know that the change is editorial, but it seems to me if you are concerned about combining by intermediaries, this is the opposite of the advice that you should be giving.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -552,7 +552,7 @@ set-cookie        = set-cookie-string
 set-cookie-string = BWS cookie-pair *( BWS ";" OWS cookie-av )
 cookie-pair       = cookie-name BWS "=" BWS cookie-value
 cookie-name       = token
-cookie-value      = *cookie-octet / ( DQUOTE *cookie-octet DQUOTE )
+cookie-value      = ( DQUOTE *cookie-octet DQUOTE ) / *cookie-octet
```
martinthomson, line 555: This change is pointless.  It is exactly the same as the original.  (Yes, I've read the erratum: people writing parsers that depend on order can rewrite the ABNF for themselves.)

martinthomson (approved):

draft-ietf-httpbis-rfc6265bis.md
mikewest: 1. We're changing this to a MUST in https://github.com/httpwg/http-extensions/pull/2988. Perhaps changing the wording in that PR would be simpler than rebasing this on top of that?

2. I'm not sure I follow your suggestion, @martinthomson. What advice should we be giving here?

draft-ietf-httpbis-rfc6265bis.md
mikewest: I do suspect that anyone writing a parser in the last ~15 years has probably either not had this problem, or solved it. Changing the spec doesn't seem terribly helpful at this point, and I agree with @martinthomson that this is the same from a grammatical standpoint so the churn it introduces seems confusing without reading the errata. 🤷  Optimizing for parsers' potential ordering issues doesn't seem necessary.

sbingler: Works for me, I'll revert the change.

draft-ietf-httpbis-rfc6265bis.md
sbingler: > Perhaps changing the wording in that PR would be simpler than rebasing this on top of that?

Now that it's just a single change in this PR I think combining them is just as easy as not (maybe easier to avoid merge issues)

> I'm not sure I follow your suggestion, @martinthomson. What advice should we be giving here?

I'm also not clear on why this isn't the correct advice.

martinthomson: The statement says that you should not produce `set-cookie: a=b;c=d` from `set-cookie: a=b` and `set-cookie: c=d`.  It justifies that by saying that the usual combining rules for fields don't work.  However, the advice you want is: if you combine `set-cookie`, use a semicolon rather than a comma.

The thing that got me to react here was the implication that leaving the cookie fields alone carries a risk that an intermediary might combine fields without knowing that `set-cookie` is special.  In THAT case, combining (using ";") at the origin server would be an excellent idea because it would avoid that happening.

(There are other concerns, like the effect of combining on HTTP/2 and HTTP/3 header compression efficiency.  But maybe we don't need to get into that level of detail.)

It's possible that this is just poor wording.  Maybe if you say that implementations (origin servers, any servers, intermediaries, clients, anyone) should not combine `set-cookie` using ",".  That this specifically names origin servers invokes all of the above questions.  (If you agree, we should open an issue to clarify, rather than continuing here, where it will definitely be lost.)

sbingler: > However, the advice you want is: if you combine set-cookie, use a semicolon rather than a comma.

We don't want to give that advice either.
Within a set-cookie string any values separated by a semicolon will be treated as attributes for the first `name=value` pair. 

From Section 5.6

> 2. If the set-cookie-string contains a %x3B (";") character:
> &nbsp;&nbsp;1. The name-value-pair string consists of the characters up to, but not including, the first %x3B (";"), and the unparsed-attributes consist of the remainder of the set-cookie-string (including the %x3B (";") in question).

So `set-cookie: a=b;c=d ` would set a cookie `a=b` with some attribute `c` that has a value of `d`. Since `c` is an unrecognized attribute it'll be ignored and a default `a=b` cookie will be set.

> The thing that got me to react here was the implication that leaving the cookie fields alone carries a risk that an intermediary might combine fields without knowing that set-cookie is special. 

Presumably an intermediary would have implemented the [HTTP spec](https://www.rfc-editor.org/rfc/rfc9110#name-field-order) which specifically calls out `Set-Cookie` as special. Granted that's specified with an "ought".

> That this specifically names origin servers invokes all of the above questions

Fair.

I think we can resolve this ambiguity by calling out origin servers and intermediaries. Something like 

`Origin servers and intermediaries MUST NOT each combine multiple Set-Cookie header fields into a...`

I'll upload a patch

sbingler:
@mikewest Thoughts on the recent change? If it LGTY we can merge it with your PR.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -349,11 +349,11 @@ origin server can include multiple Set-Cookie header fields in a single response
...
-Origin servers SHOULD NOT fold multiple Set-Cookie header fields into a single
-header field. The usual mechanism for folding HTTP headers fields (i.e., as
-defined in {{Section 5.3 of RFC9110}}) might change the semantics of the Set-Cookie header
-field because the %x2C (",") character is used by Set-Cookie in a way that
-conflicts with such folding.
+Origin servers and intermediaries MUST NOT each combine multiple Set-Cookie
+header fields into a single header field. The usual mechanism for combining
+HTTP headers fields (i.e., as defined in {{Section 5.3 of RFC9110}}) might
+change the semantics of the Set-Cookie header field because the %x2C (",")
+character is used by Set-Cookie in a way that conflicts with such combining.
```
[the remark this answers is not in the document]
sbingler: Good suggestion

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -349,11 +349,11 @@ origin server can include multiple Set-Cookie header fields in a single response
 The presence of a Cookie or a Set-Cookie header field does not preclude HTTP
 caches from storing and reusing a response.
 
-Origin servers SHOULD NOT fold multiple Set-Cookie header fields into a single
-header field. The usual mechanism for folding HTTP headers fields (i.e., as
-defined in {{Section 5.3 of RFC9110}}) might change the semantics of the Set-Cookie header
-field because the %x2C (",") character is used by Set-Cookie in a way that
-conflicts with such folding.
+Origin servers and intermediaries MUST NOT each combine multiple Set-Cookie
```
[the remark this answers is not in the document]
sbingler: On second thought I agree. Removed.

PR #2992, opened 2025-02-07
Resumable Upload: Correct terminology around representations

Acconut:
The current draft mixes the use of _representation (data)_, _file_, or just _data_ when referring to the bytes being uploaded. This inconsistency is confusing and the use of _file_ is not entirely correct because resumable uploads can transfer non-file representations as well, as was rightfully [mentioned by Roy Fielding on the mailing list](https://mailarchive.ietf.org/arch/msg/httpbisa/m-iXTJQGdb8XpxGjZWxXGj6Siic/).

This PR attempts to unify the terminology and embrace the HTTP terms _representation_ and _representation data_ where possible.

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -66,19 +66,19 @@ informative:
 
 --- abstract
 
-HTTP clients often encounter interrupted data transfers as a result of canceled requests or dropped connections. Prior to interruption, part of a representation may have been exchanged. To complete the data transfer of the entire representation, it is often desirable to issue subsequent requests that transfer only the remainder of the representation. HTTP range requests support this concept of resumable downloads from server to client. This document describes a mechanism that supports resumable uploads from client to server using HTTP.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -89,17 +89,19 @@ The terms Byte Sequence, Item, String, Token, Integer, and Boolean are imported
 
 The terms "representation", "representation data", "representation metadata", "content", "client" and "server" are from {{HTTP}}.
 
+The term "patch document" is taken from {{RFC5789}}.
```
Acconut, line 92: While not strictly related to the topic of the PR, I noticed that we were missing a reference to where _patch document_ comes from.

guoye-zhang (approved):
Looks good. Thanks! I think a few instances of "representation data" is missing "the" at the front.

Acconut:
Thank you, Guoye! Just went over the changes again and touched some of my mistakes up.

smatsson (approved):

PR #2995, opened 2025-02-11
s/intermediate response/interim response

LPardue:
RFC 9110 uses "interim response" throughout and intermediate is confusingly similar
looking to intermediary, which is something entirely different.

I didn't touch the appendix instances, since those are getting axed anyway.


guoye-zhang (approved):

Acconut (approved):
Good call, thank you!

PR #2997, opened 2025-02-11
[optimistic-upgrade] Adjust text related to HTTP/1.1 CONNECT

### LINKED ISSUE #2942: Consider impact of proxy-initiated close on HTTP 407
@ekinnear noted today that the performance impact of proxies closing the connection after returning HTTP 407 are not merely hypothetical, and are significant in practice.  This draft currently says that proxies MAY adopt this behavior as a mitigation for potential misbehaving clients.

Some things we could do:
* Remove this MAY.  After a closer look at some earlier anecdotes, we have not actually found a client that exhibits the misbehavior.
* Conduct some kind of study or outreach to see if we can find any misbehaving clients (i.e. CONNECT clients that send first-flight payload data without waiting for the 200 response).
* Define a request header like "Vulnerable-To-Request-Smuggling: ?0", by which the client promises that it is proxying TCP connections for a trusted TCP client or delaying payload data until after the success response, so there is no need for defensive measures.

bemasc:
Closes #2942

### DIFF
--- a/draft-ietf-httpbis-optimistic-upgrade.md
+++ b/draft-ietf-httpbis-optimistic-upgrade.md
@@ -158,9 +158,9 @@ Future specifications for Upgrade Tokens should restrict their use to "GET" requ
 
 # Guidance for HTTP CONNECT
 
-In HTTP/1.1, clients that send CONNECT requests on behalf of untrusted TCP clients MUST wait for a 2xx (Successful) response before sending any TCP payload data.
+In HTTP/1.1, proxy clients that send CONNECT requests on behalf of untrusted TCP clients MUST wait for a 2xx (Successful) response before forwarding any TCP payload data.  Proxy clients that start forwarding before confirming the response status code are vulnerable to a trivial request smuggling attack ({{request-smuggling}}).
 
-To mitigate vulnerabilities from any clients that do not conform to this requirement, proxy servers MAY close the underlying connection when rejecting an HTTP/1.1 CONNECT request, without processing any further data sent to the proxy server on that connection.  Note that this behavior may impair performance, especially when returning a "407 (Proxy Authentication Required)" response.
+To mitigate the impact of such vulnerable clients, proxy servers MAY close the underlying connection when rejecting an HTTP/1.1 CONNECT request, without processing any further data on that connection.  Note that this behavior will frequently impair the performance of correctly implemented clients, especially when returning a "407 (Proxy Authentication Required)" response.  This performance loss can be be avoided by using HTTP/2 or HTTP/3.
 
 # IANA Considerations
 


draft-ietf-httpbis-optimistic-upgrade.md
```diff
@@ -158,9 +158,9 @@ Future specifications for Upgrade Tokens should restrict their use to "GET" requ
...
+In HTTP/1.1, proxy clients that send CONNECT requests on behalf of untrusted TCP clients MUST wait for a 2xx (Successful) response before forwarding any TCP payload data.  Proxy clients that start forwarding before confirming the response status code are vulnerable to a trivial request smuggling attack ({{request-smuggling}}).
 
-To mitigate vulnerabilities from any clients that do not conform to this requirement, proxy servers MAY close the underlying connection when rejecting an HTTP/1.1 CONNECT request, without processing any further data sent to the proxy server on that connection.  Note that this behavior may impair performance, especially when returning a "407 (Proxy Authentication Required)" response.
+To mitigate the impact of such vulnerable clients, proxy servers MAY close the underlying connection when rejecting an HTTP/1.1 CONNECT request, without processing any further data on that connection.  Note that this behavior will frequently impair the performance of correctly implemented clients, especially when returning a "407 (Proxy Authentication Required)" response.  This performance loss can be be avoided by using HTTP/2 or HTTP/3.
```
LPardue, line 163: nit: the reason it doesn't affect performance is because the stream is closed without affecting connection lifecycle by default. That's implicit but maybe it makes sense to state it explicitly? I.e. you could add a qualifier like ", which decouple request and connection lifecycles and allow the connection to be safely used for subsequent requests".

Just an idea, it doesn't need to block the PR

LPardue (approved):

draft-ietf-httpbis-optimistic-upgrade.md
bemasc: I don't think the full explanation is needed, but I added a few more words here to avoid confusion.

LPardue: Your text is much better. LgTm

draft-ietf-httpbis-optimistic-upgrade.md
```diff
@@ -158,9 +158,9 @@ Future specifications for Upgrade Tokens should restrict their use to "GET" requ
...
 
-To mitigate vulnerabilities from any clients that do not conform to this requirement, proxy servers MAY close the underlying connection when rejecting an HTTP/1.1 CONNECT request, without processing any further data sent to the proxy server on that connection.  Note that this behavior may impair performance, especially when returning a "407 (Proxy Authentication Required)" response.
+To mitigate the impact of such vulnerable clients, proxy servers MAY close the underlying connection when rejecting an HTTP/1.1 CONNECT request, without processing any further data on that connection.  Note that this behavior will frequently impair the performance of correctly implemented clients, especially when returning a "407 (Proxy Authentication Required)" response.  This performance loss can be be avoided by using HTTP/2 or HTTP/3, which are not vulnerable to this attack.
```
martinthomson, line 163:
```suggestion
To mitigate the impact of such vulnerable clients, proxy servers MUST close the underlying connection when rejecting an HTTP/1.1 CONNECT request, without processing any further data on that connection.  Note that this behavior will frequently impair the performance of correctly implemented clients, especially when returning a "407 (Proxy Authentication Required)" response.  This performance loss can be be avoided by using HTTP/2 or HTTP/3, which are not vulnerable to this attack.
```

Happy to create an issue to track the change, but it seems like this is the only level of requirement that will work for this.

bemasc: OK, let's keep that question separate from this editorial PR.

PR #3003, opened 2025-02-17
Incremental draft

martinthomson:
This is a direct copy from https://github.com/kazuho/draft-kazuho-httpbis-incremental-http/blob/db3b05696db34eb18cafc4947f5c21cdaff7e032/draft-kazuho-httpbis-incremental-http.md with just the `docname` changed.

I removed the extra "-http" from the name and ran `make update-codeowners` which seems to have also updated the `.gitignore` (that's emergent behaviour that surprised me a little, even though I think that it is correct to do so).


martinthomson:
Requesting review, but I'll merge this soon if I don't get something.

PR #3011, opened 2025-02-25
Resumable Upload: Require client to adhere to limits in `Upload-Limit`

Acconut:
Currently, the client is not required to adhere to limits announced in the `Upload-Limit` header field in responses. As discussed in https://github.com/httpwg/http-extensions/pull/3006#discussion_r1959738722, it doesn't make a lot of sense to give the client the option to willingly ignore limits that the server announces. Hence, this PR changes the normative language to require the client to adhere to any limits that the client received and recognizes.

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -311,7 +311,7 @@ The `Upload-Complete` header field is set to true if the request content include
 
 If the client knows the representation data's length, it SHOULD include the `Upload-Length` header field ({{upload-length}}) in the request to help the server allocate necessary resources for the upload and provide early feedback if the representation violates a limit ({{upload-limit}}).
 
-The client SHOULD respect any limits ({{upload-limit}}) announced in the `Upload-Limit` header field in interim or final responses. In particular, if the allowed maximum size is less than the amount of representation data the client intends to upload, the client SHOULD stop the current request immediately and cancel the upload ({{upload-cancellation}}).
+The client MUST respect recognized limits announced by the `Upload-Limit` header field in a received interim or final response. In particular, if the allowed maximum size is less than the amount of representation data the client intends to upload, the client MUST stop the current request immediately and cancel the upload ({{upload-cancellation}}).
 
 The request content MAY be empty. If the `Upload-Complete` header field is then set to true, the client intends to upload an empty representation. An `Upload-Complete` header field is set to false is also valid. This can be used to create an upload resource URL before transferring any representation data. Since interim responses are optional, this technique provides another mechanism to learn the URL, at the cost of an additional round-trip before data upload can commence.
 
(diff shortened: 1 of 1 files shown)


guoye-zhang:
I'm hesitant about making this a MUST. Currently, our client does not parse `Upload-Limit` at all, as we only operate in the mode where we upgrade a traditional upload to a resumable upload. Let's suppose that the upload itself is larger than `max-size`, it might not be resumable but it might still be fine to continue as a traditional upload. And respecting `max-append-size` requires a lot of additional work on the client side since we currently only do `Upload-Complete: ?1`.

The server can always enforce the limits by sending a `413 Content Too Large` response, or a 5xx response so the client would try to resume. I'm not sure if it's worth the additional complexity to adhere to the limits on the client side since the server has to enforce it anyway.

LPardue:
I tend to agree with Guoye. I'd treat this as an advisory header, in the vein of Retry-After https://www.rfc-editor.org/rfc/rfc9110.html#section-10.2.3. 

In fact, removing the SHOULD might be closer to reality. Instead, we can explain why clients benefit from consuming the hint and taking some action. Then we can leave it up to them to make their own call.

Acconut:
I understand that implementing the client behavior can be challenging, especially `max-append-size`. My main concern is that making it optional for the client to follow limits could result in interoperability issues. Our experience with tus has been that a sizeable number of server deployments for resumable uploads enforce a maximum request content size. Usually to accommodate for restrictions in proxies or other parts of the infrastructure. I would assume that similar restrictions will exist if people deploy resumable uploads as defined in this draft.

A client that ignores Upload-Limit can then run into issues uploading files that are larger than the server's maximum request size. Proxies usually reject requests that are too large with a 413, preventing the client from resuming, causing the upload to fail. As Guoye mentioned, the upload would work if the proxy consumes request content until reaching the size limit and then responding with a 5xx. The client will resume and the upload can continue, but I'm unsure if such a proxy behavior is feasible in reality.

If a client respects the limits announced by the server, it can directly adjust to the server's restrictions and increase interoperability.

LPardue:
I agree with these points. However, isn't that pretty much the situation we're already in today? 

I know that I operate proxy servers that have variable upload limits, and variable behaviour when it comes to buffering of upload before responding vs. Responding immediately and allowing uploads to continue unbuffered (aka streamed).

The language in the Upload Limits section is currently that a server `MAY` not do something if the action exceeds a limit it advertised. I don't think it's acceptable to levy a MUST requirement on a client if a server might be more flexible in how it applies limits it advertised. 

I know it feels like it might help to be strict but HTTP is generally quite loose. The market will decide what it wants to do. And if people are already saying "it's easier for me to deal with rejection staus codes like I do with all HTTP upload services than use Upload-Limit" the a MUST probably won't change their mind.

Acconut:
Closing this for now as I don't see strong support for this.

PR #3012, opened 2025-02-26
Resumable Upload: Rename `expires` limit to `max-age`

### LINKED ISSUE #2940: Better name for the "expires" attribute in Upload-Limit
Expires typically is an absolute timestamp, but it is a number of seconds in Upload-Limit. It might be better named as `timeout` or something similar.

Acconut:
Closes #2940.

The Upload-Limit header can provide the remain lifetime of the upload resource in number of seconds under the `expires` key. However, this name is misleading as it implies that the value is an absolute timestamp. A better name could be `max-age`, which is already used in the [Cache-Control](https://www.rfc-editor.org/rfc/rfc9111.html#name-max-age), [Set-Cookie](https://datatracker.ietf.org/doc/html/rfc6265#autoid-13), and [Alt-Svc](https://datatracker.ietf.org/doc/html/rfc7838#section-3.1) header fields to denominate the lifetime of cookies and cache entries.


LPardue (approved):
Lgtm

guoye-zhang (approved):

PR #3016, opened 2025-02-28
32 x 32

mnot:


martinthomson (approved):
This seems good.  Both dimensions seem large enough that most uses will fit well within.  I can see how someone might want a 10-20 character label, but not 10-20 of them, so maybe the total number of groups could be 16 instead, to reduce this to a <<1k max, but we can afford a little generosity.  After all, the 1k this requires is not that significant relative to other metadata necessary to operate a good cache.

PR #3018, opened 2025-02-28
Require less response headers for upload creation and appending

Acconut:
Currently, the draft requires responses for upload creation and appending requests to include the `Upload-Complete` and `Upload-Offset` header fields. The original intention was to make the resource signal the new upload state back to the client. However, this isn't necessary for resumable uploads to work. If the client received a successful response, it can assume that the upload state was updated accordingly. If it didn't get a response, it retrieves the state using a HEAD request.

So make this document less restrictive, this PR removes the requirement to send `Upload-Offset` ~~and `Upload-Complete`~~ in responses to upload creation and append. The append response also doesn't have to include the `Upload-Limit` anymore. If one exists, the client has already received it in the upload creation or offset retrieval response.

EDIT: `Upload-Complete` is still necessary to identify whether a response is related to the resumable upload or the processing of the representation.


guoye-zhang:
Probably need to bump the interop version for this change since I think our client is checking the presence of these header fields

Acconut:
@guoye-zhang We can definitely bump the interop version. How do you feel about this PR in general?

jbaehr:
While the `Upload-Complete` response header may not be required in the case of a successful response or in case of no response, it is currently needed in the case of an unsuccessful response: Specifically the `409 (Conflict)` response becomes ambiguous if concerning the last/final request. There are two ways of interpreting it:
- The `Upload-Offset` provided by the client *does not match* the server's value, so the chunk uploaded with this request cannot be used. In turn the upload stays incomplete. A client may retry with the `Upload-Offset` provided in the response, which may results in a success.
- The  `Upload-Offset` provided by the client *does match* the server's value and the upload gets finalized. Now quoting [the current draft version](https://github.com/httpwg/http-extensions/blob/021878764a05a901c3b2958105995c4447edddd9/draft-ietf-httpbis-resumable-upload.md?plain=1#L491):
  > the upload is marked as complete and the upload resource SHOULD generate the response that matches what the resource, that was targeted by the initial upload creation ({{upload-creation}}), would have generated if it had received the entire representation in the initial request.

  In this step, while processing the entire resource, the server may want to reply with a 409 as well. This time however meaning that the new target resource conflicts. Any retry attempt made by the client is likely to fail as well, either because appending to an already finalized upload resource results in a subsequent `400 Bad Request`, a `404 File Not Found` because the upload resource is already gone, or when redoing the entire upload the `409 Conflict` will show up again. Because the upload itself was not the problem.

My preference would be to have the server reply with a `400 Bad Request` in the first case (just like when appending to an already completed upload). The current way of using the `Upload-Complete` response header field to figure out the meaning of the HTTP status code is not ideal. And with this PR here even becomes impossible.

Has this already been discusses somewhere? I haven't found anything in the issues labeled "resumable-upload". Shall I create a dedicated issue for this discussion? Although triggered by this PR, it may be a bit beyond the scope of the proposed changes here.

Acconut:
@jbaehr The draft currently requires a 409 Conflict response to be accompanied with the correct offset in the Upload-Offset header field:

> If the Upload-Offset request header field value does not match the current offset ({{upload-offset}}), the upload resource MUST reject the request with a 409 (Conflict) status code. The response MUST include the correct offset in the Upload-Offset header field.

This would allow a client to distinguish whether a 409 is related to the resumable upload itself or the processing of the uploaded representation after the upload is complete.

That being said, the document also currently says that a client should not try to resume an upload if it received a 4xx, including a 409. Just like any other 4xx response, a 409 is treated as a terminal state regarding the upload. The reasoning behind this is that a well-behaving client and server should never trigger a 409. A 409 indicates that the offset counted by the client has diverged from the correct offset of the upload resource, which can only be the result of a bug (in the client or server) or an intermediary fiddling with the messages. When a client resumes an upload after an interruption, it fetches the current offset via a HEAD request instead of guessing the correct offset in a PATCH request.

However, with this argumentation, there is no reason to use 409 for mismatching offsets instead of a 400, I agree with you.

I am open to discussing
- whether 409 should be replaced with a 400, while continuing to say that clients shouldn't resume in these cases, or
- whether 409 + Upload-Offset is kept, and clients are allows to resume in those cases (diverging from the current wording in the spec).

guoye-zhang:
Using `Upload-Complete: ?1` to distinguish the final response is the only way today. I don’t think changing 409 to 400 is sufficient since it can also be the final response. We will have to register a new 4xx status code exclusively for resumable upload failures for that to work.

guoye-zhang:
Technically, we don’t disallow the final response to include Upload-Offset. Our current server implementation in NIO puts Upload-Offset on all responses.

Acconut:
You're right, we don't disallow it. If we want to allow clients to resume an upload after they attempted to append at a wrong offset, we could either use a new 4xx status code (as @guoye-zhang suggested) or make the server return a 5xx, so that the client retrieves the new offset. 5xx feels a bit odd at first, but if we frame the mismatching offsets as "server cannot write data at that position", it could technically be interpreted as a server error.

But this requires that we want clients to resume after mismatching offsets (which the spec currently doesn't allow). Do we want to allow that?

guoye-zhang:
I don't think we should. Upload-Offset in the 409 is merely for debugging purposes. Not something that people write code to handle.

Acconut:
I'm fine with keeping it that way but would also be open to changing that. At the end of the day, I have to admit that I wrote client code that explicitly retries on 409, mostly to make uploads work with weird setups that I don't control.

> @jbaehr Has this already been discusses somewhere? I haven't found anything in the issues labeled "resumable-upload". Shall I create a dedicated issue for this discussion? Although triggered by this PR, it may be a bit beyond the scope of the proposed changes here.

Yes, please! Feel free to open a separate issue to discuss this.

guoye-zhang (approved):

Acconut:
@guoye-zhang @jbaehr Thank you for your feedback. I updated the PR to ensure that responses for append requests always include the `Upload-Complete` header so the client can distinguish whether error responses belonging the resumable upload versus the representation processing. Please let me know what you think.

I will bump the interop version in a separate PR to keep this diff readable.

guoye-zhang (approved):

jbaehr:
> However, with this argumentation, there is no reason to use 409 for mismatching offsets instead of a 400, I agree with you.
> 
> I am open to discussing
> -  whether 409 should be replaced with a 400, while continuing to say that clients shouldn't resume in these cases, or
> -  whether 409 + Upload-Offset is kept, and clients are allows to resume in those cases (diverging from the current wording in the spec).

I guess the tradeoff here is between one additional request (HEAD, as in "resume after interruption") and the ease to retrofit existing protocols with resumeability (where the 409 status is already well-established).

To my liking there are several places in the current draft that seem to have this retrofitting use case in mind. The section I quoted above or §17.1.1 "Upgrading To Resumable Uploads". Let's make this a first-class concept and remove the last remaining roadblock of conflicting error codes.

---

> Using `Upload-Complete: ?1` to distinguish the final response is the only way today.

Yes. And reinterpreting the very same status code based on the (non-)existence of some other headers is far from straight forward.

> I don’t think changing 409 to 400 is sufficient since it can also be the final response. We will have to register a new 4xx status code exclusively for resumable upload failures for that to work.

That's also true. One could argue that only one potentially clashing status is better than having two of them. Or that the likelihood of encountering a 409 further down the pipeline (when the entire upload is done and the payload processing starts) is higher than a 400 (which by definition is only scoped to a request, not a resource)...
At the end of the day it's just me who happens to encounter the problem with 409, not 400 -- so a subjective choice. The objective way is to register a new 4xx status. Similar to `411 Length Required`.

---

> I updated the PR to ensure that responses for append requests always include the Upload-Complete header so the client can distinguish whether error responses belonging the resumable upload versus the representation processing. Please let me know what you think.

We do have now (again, but this time explicitly) a way to distinguish between errors from the upload itself and from processing the uploaded resource (as it would be processed without resume capabilities). This is good.

I still find it odd to have to look into other headers to determine the definite meaning of a status code, though. However, this is better discussed separately. I'll open a dedicated issue for that, but it may take a few days.

Acconut:
Thank you for the feedback, @jbaehr. Let's discuss this in a separate issue (or on the mailing list)!

PR #3023, opened 2025-03-03
Resumable Upload: Problem type for inconsistent length values

Acconut:
This PR adds a problem type for servers to use when responding to requests that include inconsistent upload length values as explained in https://httpwg.org/http-extensions/draft-ietf-httpbis-resumable-upload.html#name-length.

With this PR, the draft defines a problem type for all (three) upload-specific error cases that are mentioned in the document. I think this might be useful for implementation.

Let me know if you think this isn't a good idea.

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -283,7 +283,7 @@ Even the client might not know the total length of the representation data when 
 - If the request includes the `Upload-Complete` field value set to true and a valid `Content-Length` header field, the request content is the remaining representation data. The length is then the sum of the current offset ({{upload-offset}}) and the `Content-Length` header field value.
 - The request can include the `Upload-Length` header field, whose value is the number of bytes of the entire representation data as an Integer.
 
-If both indicators are present in the same request, their indicated lengths MUST match. If multiple requests include indicators, their indicated values MUST match.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -777,6 +792,17 @@ Recommended HTTP status code:
...
 : This document
 
+IANA is asked to register the following entry in the "HTTP Problem Types" registry:
+
+Type URI:
+: https://iana.org/assignments/http-problem-types#inconsistent-upload-length
+Title:
+: Inconsistent Upload Length Values
+Recommended HTTP status code:
+: 400
```
guoye-zhang, line 802: 409?

Acconut: I used 400 for now to save the use of 409 to mismatching offsets in this document. Depending on how the discussion following from https://github.com/httpwg/http-extensions/pull/3018 turns out, we might get a different status code for mismatching offsets.

guoye-zhang: The example above shows 400 though

Acconut: Oh my bad! Thank you, I'll fix that!

PR #3031, opened 2025-03-07
Make two examples clearer

smatsson:
First example had the Upload-Limit in the wrong 104 response.

In the second example I added a 104 to Upload-Complete: ?0 to make it clear that this interim response is sent here as well. No Upload-Offset interim though to show that it's optional.


Acconut (approved):
Thanks for this PR! I wonder if we should emphasize more that the 104 responses are optional. The server doesn't have to send them and the client does not always need them.

guoye-zhang (approved):

smatsson:
@Acconut IMO the "Careful Upload Creation" section is enough for that.

Issue #2739, opened 2024-02-15
Consider closing the connection on Upgrade failure


bemasc:
@martinthomson writes

> The server treating Upgrade as implying Connection: close might be a good start, counter to what Section 4 currently says.  We should not be recommending mitigations that only one affected party can deploy.

In other words, we would instruct servers to close the connection after responding, without reading any more requests, when they reject an Upgrade.

This behavior would be compatible and compliant, and would foreclose the security issue in question.  However, it would have a notable performance cost when returning a response that will trigger a retry (e.g. 307, 401, 407).

bemasc:
I've added this to #2821 in the context of HTTP CONNECT.

I don't think this mitigation makes sense for Upgrade, because there is no widely deployed vulnerable client that would justify a costly server-side intervention.  However, for HTTP CONNECT, it does seem like there may be some clients that are already doing the wrong thing, so a server-side mitigation may be justified.

bemasc:
The text now reads:

> To mitigate the impact of such vulnerable clients, proxy servers MAY close the underlying connection when rejecting an HTTP/1.1 CONNECT request

@ekinnear voiced a concern that similar "aggressive close" behaviors have had a noticeable negative impact on performance in real deployments (#2942).  This suggests that we might not want servers to apply this mitigation.  To help decide what to do about this, I went looking for clients exhibiting this "vulnerable behavior" and found a few (https://github.com/httpwg/http-extensions/issues/2942#issuecomment-2652165660).  This strikes me as sufficient to justify the "MAY" here.

@martinthomson suggests converting this MAY to a MUST (https://github.com/httpwg/http-extensions/pull/2997#discussion_r1955446429), in order to foreclose the vulnerability entirely.  This would presumably Update RFC 9112.

We have a choice to make:

* MAY close
* MUST close
* SHOULD/conditional MUST, and under what conditions.

"MUST" seems impractical to me, given the decades of deployed CONNECT proxy servers.  "SHOULD" is possible but I think we would have to say something like:

> If some portion of a proxy server's clients are likely to be vulnerable in this way and can carry malicious traffic, the proxy server SHOULD ...

We could also discuss User-Agent sniffing...

bemasc:
Settled on a MUST with some conditions, per advice from the WG session at IETF 122.

PR #3048, opened 2025-03-18
[connect-tcp] Require close after CONNECT fails

bemasc:
This changes makes it REQUIRED for proxy servers to close the connection after a failed HTTP/1.1 CONNECT ... with a legalistic escape hatch: this requirement only applies "when communicating with potentially vulnerable clients".

Presumably any client is potentially vulnerable unless you know something about it specifically.

See #2739

### DIFF
--- a/draft-ietf-httpbis-optimistic-upgrade.md
+++ b/draft-ietf-httpbis-optimistic-upgrade.md
@@ -14,7 +14,7 @@ workgroup: "HTTPBIS"
 keyword:
 venue:
   github: "httpwg/http-extensions"
-updates: 9298
+updates: 9112, 9298
 
 author:
  -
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-optimistic-upgrade.md
```diff
@@ -160,7 +160,7 @@ Future specifications for Upgrade Tokens should restrict their use to "GET" requ
...
 
-To mitigate the impact of such vulnerable clients, proxy servers MAY close the underlying connection when rejecting an HTTP/1.1 CONNECT request, without processing any further data on that connection.  Note that this behavior will frequently impair the performance of correctly implemented clients, especially when returning a "407 (Proxy Authentication Required)" response.  This performance loss can be be avoided by using HTTP/2 or HTTP/3, which are not vulnerable to this attack.
+At the time of writing, some proxy clients are believed to be vulnerable as described.  When communicating with potentially vulnerable clients, proxy servers MUST close the underlying connection when rejecting an HTTP/1.1 CONNECT request, without processing any further data on that connection.  Note that this mitigation will frequently impair the performance of correctly implemented clients, especially when returning a "407 (Proxy Authentication Required)" response.  This performance loss can be be avoided by using HTTP/2 or HTTP/3, which are not vulnerable to this attack.
```
martinthomson, line 163:
```suggestion
At the time of writing, some proxy clients are believed to be vulnerable as described.  When communicating with potentially vulnerable clients, proxy servers MUST close the underlying connection when rejecting an HTTP/1.1 CONNECT request, without processing any further data on that connection.  Note that this mitigation will frequently impair the performance of correctly implemented clients, especially when returning a "407 (Proxy Authentication Required)" response.  This performance loss can be avoided by using HTTP/2 or HTTP/3, which are not vulnerable to this attack.
```

martinthomson (commented):
Works for me.  I might prefer if your loophole was phrased more like "unless the server can guarantee that the client is not vulnerable".  The current phrasing is a little indirect.  But that's stylistic more than anything; this has the substance right.

Issue #3060, opened 2025-04-01
query: shorten HTTP method name


reschke:
It has been pointed out to me that the name "QUERY"

1. may be confusing (vs 'query parameters'), but also because of "SEARCH" and "REPORT".
2. wastes a few bytes on the wire (even in H2+)

My proposal is to change "QUERY" to "?" - given the fact that "query" does not even appear in the Internet Draft name, this is a really simple change.

mnot:
-1. Four bytes is not meaningful in a protocol like HTTP, and a non-alphanumeric method is more likely to break assumptions in software as well as confuse people. I think the linkage to 'query parameters' is actually a plus.

reschke:
@LPardue - interesting idea, but let's not bikeshed here.

reschke:
FWIW, aliasing GET/PUT with `<<` and  `>>` might be good as well. Unfortunately, changing existing method names theoretically could break things.

reschke:
@mnot - would using `QR` address your concerns?

reschke:
Closing this due to lack of enthusiasm. Maybe in a bis spec.

PR #3061, opened 2025-04-01
qr: smplf mthd nm (clss #3060)

reschke:


reschke:
For a simple change like that, there's no problem merging the PR just with @LPardue 's approval.

mnot:
/me looks at date

leave me alone

PR #3063, opened 2025-04-04
Resumable Upload: Retries, rejections, and security

Acconut:
This PR addresses feedback received on the latest resumable upload draft:

- It clarifies what the server has to do when an upload append request provides too much data in the request content, exceeding the upload length. It has to reject the request, invalidate the upload resource, and reject further interaction. Such behavior is indicative of a client bug and has to be fixed on the client's end, so preventing resumption is not a problem.
- Clients are currently recommended to not resume the upload after 4xx responses. This PR relaxes this rule to allow retries for some 4xx responses, like 429 or 421.
- The security considerations are expanded to include additional recommendations based on experience from current and past implementations:
  - Upload resource URIs should not leak confidential information, e.g. the storage server where the representation is saved at.
  - Representation metadata should be properly sanitized before consulting it to determine how representations are processed (I witnessed people using non-sanitized filenames as the location of where to save uploads in the past)
  - Reiterate that uploaded data is untrusted input and has to be handled carefully.


guoye-zhang:
Can we rephrase it so we don't explicitly mention 429? There is also 421 that clients can retry on a new connection.

In general, resumable upload implementations use 4xx to signal fatal errors, and 5xx to signal recoverable errors. I guess that's not true in general for HTTP, and different status codes can be handled by different layers of the stack.

Acconut:
> Can we rephrase it so we don't explicitly mention 429? There is also 421 that clients can retry on a new connection.

Good point, I can try. How about something like this:

> [if the response has a] `4xx (Client Error)` status code, the client SHOULD NOT attempt to retry or resume the upload, unless the schemantics of the response allow or recommend the client to retry the request.

Does this cover 429, 421 and potential future status codes sufficiently, yet precisely, enough?

guoye-zhang:
Yeah, that sounds fine

Acconut:
Thanks for the feedback, @guoye-zhang. I updated the PR with the new wording.

smatsson (approved):

guoye-zhang (approved):

guoye-zhang:
Looks good, thanks

PR #3064, opened 2025-04-09
resumable: add keywords

reschke:
RFC Editor will ask anyway


Acconut:
That's a good idea, thanks for bringing it up, but this PR doesn't contain changes. Is this intentional?

reschke:
Oops :-)

reschke:
Apparently this duplicates an earlier change.

PR #3065, opened 2025-04-09
[connect-tcp] Proposal: END_DATA and STOP_SENDING_DATA

### LINKED ISSUE #3000: Resolve how to signal TCP RST vs. FIN
In the current Editor's Copy of `connect-tcp`, TCP FIN and RST are distinguished by HTTP Stream Errors, or (especially in HTTP/1.1, where there is no such thing) by injecting a deliberate syntax error in the Capsule Protocol or an unclean termination of the HTTP connection itself.  This approach is likely simplest for recipients, as it avoids the need for new codepaths (beyond the error handling already required for broken connections), but it might be difficult to implement for senders, who would need detailed access to the HTTP stream, Capsule Protocol, TLS, or TCP connection (in order to inject an error).

PR #2949 draws up an alternative: a FINAL_DATA capsule that represents the FIN bit.  Streams that end without FINAL_DATA are presumed to be unclean (timeout, RST, etc.) and are converted to RST.    This approach matches the abstraction used by WebTransport and avoids the need for error injection, at the cost of some additional complexity.

Some reviewers have asked for an explicit RST signal as well.  This would allow a sender to terminate a stream explicitly, whether by FIN or RST, and then send additional capsules to communicate some metadata about the connection (e.g. summary statistics).  This could be encoded by an additional capsule type (e.g. "ERROR") or by a more flexible version of FINAL_DATA with a boolean flag for clean/unclean shutdown ("END_DATA").  One could even imagine a TCP_FLAGS_DATA that carries the entire TCP flags field, allowing senders to convey FIN, RST, PSH, and URG.

These are the 5 options that I am aware of: implicit (current editor's copy), FINAL_DATA (#2949), FINAL_DATA+ERROR, END_DATA, TCP_FLAGS_DATA.  We need to make a choice to unblock progress on this draft.

bemasc:
This is an alternative to #2949

Fixes #3000


bemasc:
Reviewers, please express preferences for this approach vs. #2949.

draft-ietf-httpbis-connect-tcp.md
```diff
@@ -72,6 +72,12 @@ An intermediary MAY merge and split successive DATA capsules, subject to the fol
 * There are no intervening capsules of other types.
 * The order of payload content is preserved.
 
+An END_DATA capsule indicates that sender has finished sending data for this stream.  Its payload is a QUIC varint representing an error code.  A value of 0 means "completed", semantically equivalent to TCP FIN.  Any other value indicates truncated data, corresponding to a TCP RST, timeout, or other error.  In this specification, the only other defined value is 1, meaning "truncated".
+
+A STOP_SENDING_DATA capsule indicates that the sender has begun discarding DATA capsules that it receives on this stream.  Its payload is empty.
```
martinthomson, line 77: This is going to end up being confusing because it is close to the semantic of QUIC's STOP_SENDING, but not quite.

I understand that you might want to have this capability for HTTP/2 and HTTP/1.1, where our model does not properly decouple send and receive sides of a bidirectional exchange.  This is a function that is already provided by the reset function in each protocol (RST_STREAM in HTTP/2, TCP RST in HTTP/1.1).  

While you might think that it is necessary to cover for those shortcomings, I don't see how that is necessary.  You are  only providing a redundant mechanism/signal.

A positive signal that a stream ended cleanly is useful, but the absence of that signal is sufficient to indicate truncation.

martinthomson (commented):
Given what we know about HTTP/1.1, and the need for this mechanism to work over that protocol, I'm comfortable with the inclusion of a positive "stream complete" signal, but the other mechanisms duplicate existing mechanisms in confusing ways.

draft-ietf-httpbis-connect-tcp.md
DavidSchinazi: I agree with MT, I don't think we need STOP_SENDING_DATA. For h2 and h3, sending RST_STREAM/RESET_STREAM without sending FINAL_DATA is enough to indicate an abortive close. For h1, we have a mechanism (TCP RST / TLS error alert) that's good enough even if it's not possible to implement for everyone.

bemasc:
Closing in favor of revised #2949.

PR #3080, opened 2025-05-01
query: some editorial nits

martinthomson:
Overall, this is very good, so these are very minor.

Feel free to cherry-pick or reject as you see fit.

### DIFF
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -419,7 +419,7 @@ Accept-Query: "application/jsonpath", application/sql;charset="UTF-8"</artwork>
     <t>
       Caches that normalize QUERY content incorrectly or in ways that are
       significantly different from how the resource processes the content
-      can return the incorrect response if normalization results in a false positive.
+      can return an incorrect response if normalization results in a false positive.
     </t>
     <t>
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -218,8 +219,8 @@ q=foo&amp;limit=10&amp;sort=-published
       representation of the resource identified by the target URI
       (as defined by <xref target="HTTP" section="7.1"/>), the QUERY
       method is used to ask the server to perform a query operation
-      (described by the request content) over some set of data scoped to the
-      target URI. The content returned in response to a QUERY
+      (described by the request content) over some set of data at the
+      resource. The content returned in response to a QUERY
```
reschke, line 223: That's the only change ("scoped" vs "at") where I'm not completely sure it's an improvement. For "scoped", did we intend to have this an hierarchical quality?

martinthomson: I have no idea what this is supposed to imply with "scoped", honestly.  My conception of this feature is that you query a resource and it uses it state ("data at the resource") to answer the question.  The URI only identifies the resource.

reschke: yes, "URI -> resource" is obvious.

PR #3084, opened 2025-05-10
no-vary-search: make document title more descriptive

reschke:
Just a proposal; but the title absolutely needs to be clarified.

### DIFF
--- a/draft-ietf-httpbis-no-vary-search.md
+++ b/draft-ietf-httpbis-no-vary-search.md
@@ -1,5 +1,5 @@
 ---
-title: No-Vary-Search
+title: The HTTP No-Vary-Search Field
 abbrev:
 category: std
 


domenic:
Happy to make any changes here, but I'm unsure about this particular pattern. Looking at the sidebar of https://httpwg.org/specs/ I see:

- 3 bare-header names or name-value pairs (`Cache-Control: stale-while-revalidate and stale-if-error`; `Cache-Control: immutable`; `Content-Disposition`)
- 3 "X Header Field" (`Cache-Status Header Field`, `Prefer Header Field`, `ALPN Header Field`)
- 1 "X Field" (`Proxy-Status Field`)
- 1 "X HTTP Header Field" (`Client-Cert HTTP Header Field`)

This change would move us from joining the first group (which is tied for plurality right now) to creating a fifth novel group, "The HTTP X Field".

Of course my data is just from surveying that sidebar and there are probably many more IETF specs one could check...

reschke:
"Use of the Content-Disposition Header Field in the Hypertext Transfer Protocol (HTTP)"

And that's of course _too_ verbose (blame me).

domenic:
Alright, well, there's no consensus pattern, but the one I like the best is `The X HTTP Response Header Field` (used twice: [Proxy-Status](https://httpwg.org/specs/rfc9209.html), [Cache-Status](https://httpwg.org/specs/rfc9211.html)). That gives clarity on request vs. response and header vs. header+trailer.

However, now I'm wondering if there's more spec work that needs to be done to talk about trailer vs. header... Proxy-Status [discusses](https://httpwg.org/specs/rfc9209.html#rfc.section.2) the possibility, but most specs do not seem to do so.

Any thoughts or advice on that? Or should I just go ahead and pull the trigger on "The No-Vary-Search HTTP Response Header Field"?

PR #3092, opened 2025-05-15
Resumable Upload: Correct usage of BCP14 keywords

Acconut:
In his [review](https://mailarchive.ietf.org/arch/msg/httpbisa/9dsNnPwl3j3TCuv1o7jL5Em5380/), Julian rightfully called out some incorrect uses of BCP14 keywords in the document.

This PR attempts to address these point by

- turning MAYs into "can"s where they just express a possibility,
- turning SHOULDs into MUSTs where appropriate,
- explaining SHOULD to aid reader in the decision process of whether to follow the advice or not.

Some of these changes are normative:

- Handling `OPTIONS *` is now optional
- Server must announce limits using `Upload-Limit`
- Clients must adhere to limits that they are aware of
- Server must prevent issues from concurrent requests, but are not required to achieve this by terminating previous requests. This gives server the ability to implement the approach that works best while focusing on the outcome.
- 204 response is no longer mandated for DELETE responses 

Some additional paragraphs or sentences have been added to provide more context where helpful.


guoye-zhang (approved):
Thanks for doing this cleanup!

PR #3093, opened 2025-05-15
Separate Garbage Collect Cookies from Store a Cookie algorithm

DCtheTall:


draft-ietf-httpbis-layered-cookies.md
```diff
@@ -804,30 +804,59 @@ agent MUST remove from the cookie store all cookies whose expiry-time is null.
...
+    1. Otherwise:
+
+        1. Let _cookie_ be the first item of _secureCookies_.
+
+        1. Remove _cookie_ from _secureCookies_.
 
-    1. Otherwise, remove the first item of _secureCookies_ and delete the corresponding cookie from the user agent's cookie store.
+        1. Remove _cookie_ from the user agent's cookie store.
+
+        1. Append _cookie_ to _excessHostCookies_.
```
annevk, line 839: Shouldn't we return _excessHostCookies_ at this point? Do we know for a fact that cookie-related events are fired for excess-cookie removals?

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -1245,11 +1277,37 @@ boolean _httpOnlyAllowed_, boolean _allowNonHostOnlyCookieForPublicSuffix_, and
...
+To **Garbage Collect Cookies** given a host _host_:
 
-1. Run Remove Global Excess Cookies.
+1. Let _expiredCookies_ be the result of running Remove Expired Cookies.
+
+1. Let _excessHostCookies_ be the result of running Remove Excess Cookies for Host given _host_.
+
+1. Let _excessGlobalCookies_ be the result of running Remove Global Excess Cookies.
+
+1. Let _removedCookies_ be an empty list.
```
annevk, line 1304:
```suggestion
1. Let _removedCookies_ be « ».
```

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -1361,6 +1419,10 @@ SHOULD be processed as follows:
    _isSecure_, _host_, _path_, _httpOnlyAllowed_, _allowNonHostOnlyCookieForPublicSuffix_, and
    _sameSiteStrictOrLaxAllowed_.
 
+1. If _cookie_ is null then return.
```
annevk, line 1422:
```suggestion
1. If _cookie_ is null, then return.
```

annevk (commented):
I see how this separates out expired cookies as per https://github.com/johannhof/draft-annevk-johannhof-httpbis-cookies/issues/14 but I don't see the corresponding adjustments to the other sections.

In particular https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-layered-cookies-00#section-5.2 needs to be removed now and instead be done from [Requirements Specific to Non-Browser User Agents](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-layered-cookies-00#name-requirements-specific-to-no) through this new algorithm.

draft-ietf-httpbis-layered-cookies.md
DCtheTall: Fixed.

draft-ietf-httpbis-layered-cookies.md
DCtheTall: Added comma.

DCtheTall:
> I see how this separates out expired cookies as per https://github.com/johannhof/draft-annevk-johannhof-httpbis-cookies/issues/14 but I don't see the corresponding adjustments to the other sections.

Ah I needed to address the point about the Cookie Store Eviction section. I made some changes to this PR to address that:

- I moved Remove Expired Cookies to Cookie Store Eviction section.
- I moved the SHOULD/MUST from Cookie Store Eviction's intro to a new section called Cookie Store Eviction for Non-Browser User Agents which calls Remove Expired Cookies.

Let me know if this is what you had in mind!

johannhof (approved):
LGTM, will hold off on merging for a bit to give @annevk the chance to weigh in though

PR #3095, opened 2025-05-16
Resumable Upload: Replace section 'Subsequent Resources'

Acconut:
Resumable uploads have the problem that if the client misses the response to the last PATCH request, it has no direct way of recovering this specific response. For that reason, the draft currently includes a section about subsequent resources that tries to explain how an additional resource can be used to make the final response available for later retrieval. However, the paragraph doesn't explain the actual root problem well and provides little context in general.

Therefore, this PR replaces this section with one that attempts to describe the issue properly, as well as different approaches using existing HTTP mechanism to work around the problem. It also gets rid of the odd term of _subsequent resources_.


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -724,9 +724,13 @@ Content digests help verify the integrity of the content in an individual reques
...
+
+To address this issue, the server can choose to send the corresponding information in header fields, which are then included in the final response, once all representation data has been transferred, as well as responses when the client fetches the upload state via a HEAD request ({{offset-retrieving}}). This way, the client can retrieve the information from the header even if it failed to receive the final response. In addition, if the server responds with a representation and wants to make it available for clients for later retrieval as well, it can include the `Content-Location` header field in the final response and responses when fetching the upload state. The client can then retrieve the representation from the provided URI, even if it failed to receive the final response.
```
smatsson, line 733: I found this section a bit hard to read. Could we simplify it? 

Suggestion:
> To fix this issue, the server can send the needed information in headers that are included in both the final response 
 and also in responses when the client queries the upload state with a HEAD request. Additionally, if the server sends a representation and wants to make it available for later, it can include the `Content-Location` header field in both the final response and in responses when checking the upload state. This allows the client to get the information even if it didn't receive the final response.

guoye-zhang, line 733: Responses to HEAD are not allowed to contain bodies, so it does not make sense to use offset retrieving to retrieve the final response if it has a body. I would just recommend that the server accepting a 0-length upload as a way to retrieve the final response again. Our client already does that.

smatsson: >Responses to HEAD are not allowed to contain bodies, so it does not make sense to use offset retrieving to retrieve the final response if it has a body.

The HEAD response cannot but the GET response from the URL in the Content-Location header can. I think it does make sense as this is a common pattern for servers when they want to cache a response for later use. 

> I would just recommend that the server accepting a 0-length upload as a way to retrieve the final response again. Our client already does that.

While this is an interesting approach it seems very custom for this protocol as most servers already uses the Content-Location header for other setups. I'm not against your approach here but from experience I haven't seen any servers that actually implement it this way (for this protocol or others). It will need to be documented in the specification if we want wide adoption of it.

guoye-zhang: The reason we used 0-length append is that it's the least implementation effort on the client side. We already handle connection interruptions by doing an offset retrieving and an append. And we don't currently look at the Content-Location header in the offset retrieving response.

The server can always use a 303 redirect as the append response, and we will follow the Location header with a new GET request.

Acconut: Thanks, I incorporated this into the PR

Issue #3107, opened 2025-06-09
query: review caching requirements


reschke:
(follow up to https://github.com/httpwg/http-extensions/issues/2896)

Currently:

>[2.4. ](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-11.html#section-2.4)[Caching](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-11.html#name-caching)
>
>The response to a QUERY method is cacheable; a cache MAY use it to satisfy subsequent QUERY requests as per [Section 4](https://rfc-editor.org/rfc/rfc9111#section-4) of [[HTTP-CACHING](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-11.html#HTTP-CACHING)]).
>The cache key for a QUERY request (see [Section 2](https://rfc-editor.org/rfc/rfc9111#section-2) of [[HTTP-CACHING](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-11.html#HTTP-CACHING)]) MUST incorporate the request content. When doing so, caches SHOULD first normalize request content to remove semantically insignificant differences, thereby improving cache efficiency, by:
>
>- Removing content encoding(s) ([Section 8.4](https://rfc-editor.org/rfc/rfc9110#section-8.4) of [[HTTP](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-11.html#HTTP)]).
>- Normalizing based upon knowledge of format conventions, as indicated by any media subtype suffix in the request's Content-Type field (e.g., "+json", see [Section 4.2.8](https://rfc-editor.org/rfc/rfc6838#section-4.2.8) of [[RFC6838](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-11.html#RFC6838)]).
>- Normalizing based upon knowledge of the semantics of the content itself, as indicated by the request's Content-Type field.
>
> Note that any such normalization is performed solely for the purpose of generating a cache key; it does not change the request itself.

Issues:

1. "... MUST incorporate the request content" - that's misleading, because any kind of stable indentifier (such as a SHA-256 of (normalized) content) should be good enough
2. "When doing so..." means "When responses are cached...", right?
3. "...SHOULD first normalize..." is misleading, as all the steps that follow are more or less optional ... "...can normalize..." or "...are encouraged to normalize..." maybe?
[...]

PR #3108, opened 2025-06-10
QUERY: improve description of CACHING, clarifying what is required (addresses #3107)

reschke:


reschke:
In the context of normalizing JSON, do we need to talk about known issues like repeating members, ordering, etc? Also, that any attempt of normalization that leads to an error condition needs to be handled properly in that no normalization can happen?

mnot (approved):

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -317,20 +317,25 @@ q=foo&amp;limit=10&amp;sort=-published
       </t>
       <t>
         The cache key for a QUERY request (see <xref target="HTTP-CACHING" section="2"/>) &MUST;
-        incorporate the request content. When doing so, caches &SHOULD; first normalize request
-        content to remove semantically insignificant differences, thereby improving cache
-        efficiency, by:
+        incorporate the request content.
+      </t>
+      <t>
+        Caches are encouraged to remove semantically insignificant differences first, thereby
```
nicowilliams: s/are encouraged to/may/ (lower-case may at that)

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -317,20 +317,25 @@ q=foo&amp;limit=10&amp;sort=-published
...
+        <li>normalizing based upon knowledge of format conventions, as indicated by any
           media subtype suffix in the request's Content-Type field (e.g., "+json", see
           <xref target="RFC6838" section="4.2.8"/>).</li>
-        <li>Normalizing based upon knowledge of the semantics of the content itself, as indicated by
+        <li>normalizing based upon knowledge of the semantics of the content itself, as indicated by
           the request's Content-Type field.</li>
       </ul>
       <t>
-        Note that any such normalization is performed solely for the purpose of generating a cache key; it
+        Note that any such transformation is performed solely for the purpose of generating a cache key; it
```
nicowilliams, line 352: s/is/should be/

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -317,20 +317,25 @@ q=foo&amp;limit=10&amp;sort=-published
...
         does not change the request itself.
```
nicowilliams, line 353: s/the request/the semantics of/

draft-ietf-httpbis-safe-method-w-body.xml
nicowilliams: See @vidraj's recent commentary.  @vidraj believes (and I suspect they're right) that request-side `no-transform` has long been in HTTP and RFC 9110's failure to mention it is accidental.

mnot: See RFC9111.

nicowilliams: @mnot ah, I was searching for `cache-request-directive`.

nicowilliams: Still, I encourage you to say "may" (lower case) rather than "are encouraged to".  I don't see the need to "encourage".

reschke: "encouraged to" is supposed to be between "may" and "should". Saying "may" wouldn't de facto mean anything, because a cache doesn't need permission for this in any case.

draft-ietf-httpbis-safe-method-w-body.xml
reschke: No, it does not change the request at all.

draft-ietf-httpbis-safe-method-w-body.xml
reschke: "is" is a statement of fact.

If we would say "should", we'd have to consider under which circumstances it would be ok to ignore the requirement.

draft-ietf-httpbis-safe-method-w-body.xml
nicowilliams: Let's be specific and exact.  A change to the request body's byte sequence can be not-a-change of the request's logical contents and semantics, but it _is_ a change of the request.  To say that the permissible normalizations are those that do not change the request's semantics is quite correct, precise, sufficient, and necessary.  To try to further distinguish what sort of change is allowed seems pointless given that not changing the _semantics_ of the request is enough.

For example, a proxy could change the encoding of the request if the point of the proxy is to do just that because some clients have an encoding impedance mismatch with the server.  But it would be quite wrong to change the semantics of the request query.

draft-ietf-httpbis-safe-method-w-body.xml
nicowilliams: "Econraged to" is not closer to "should" as far as RFC 2119 is concerned -- neither "encouraged to" nor "should" nor "may" are recognized by RFC 2119.  So to a point we're splitting hairs.

But!  I don't see a need to encourage these transformations.  Especially given that I have serious doubts that any cache will ever implement them because it's between `ETOOHARD` and `ENOTWORTHWHILE` (because users won't be ~~posting~~ sending the same queries anyways).  You don't have running code to demonstrate it either.

Incidentally, I work with a [proprietary] RESTful directory service that has caching for non-query end-points (because these have an ORM N+1 problem and there are commonly-requested entities) and no caching for query end-points precisely because there is no point to caching random queries that are very fast (because those do not have an ORM N+1 problem) (also it is difficult to invalidate _bulk_ query cached results, so any caching of those must be very short-lived).  Thus my skepticism.

If you have experience contrary to mine it'd be good to hear about it.

draft-ietf-httpbis-safe-method-w-body.xml
nicowilliams: "should" not being normative, I disagree, but we can split this by using "would".  "Note that any such transformation _would be_ performed solely for ...".

draft-ietf-httpbis-safe-method-w-body.xml
reschke: The cache does not change the request. The whole paragraph is about generating the cache key.

draft-ietf-httpbis-safe-method-w-body.xml
reschke: I don't think this changes the meaning of the statement in any way.

draft-ietf-httpbis-safe-method-w-body.xml
reschke: You are now again questioning normalization in general it seems.

We had this discussion before, and the conclusion was to keep it but to clarify the description. This is what this PR is about.

draft-ietf-httpbis-safe-method-w-body.xml
vidraj: It does change the request from the point of view of the client who came second and got a cached response originally generated for a different – even if semantically identical – request.

What's the rationale for requiring the first request to be passed on unchanged? I think it would be better to require caches to send the transformed (normalized / canonicalized) request if they perform such changes.

This would ensure that every client gets the same response no matter who came first, that the transformation is something sane (as opposed to whatever suboptimal or borderline broken thing the first client happened to generate), and that the server operators get to see the transformation in action (good for inspecting buggy / misconfigured proxies).

draft-ietf-httpbis-safe-method-w-body.xml
nicowilliams: > You are now again questioning normalization in general it seems.

> We had this discussion before, and the conclusion was to keep it but to clarify the description. This is what this PR is about.

That was _your_ conclusion.  You yourself indicated that you had no idea how to normalize things like SQL.  You've also not indicated whether you have running code, which I assume means you don't have any.  I think at this point this matter has to be brought to the [ietf-http-wg@w3.org](mailto:ietf-http-wg@w3.org) list for a consensus call.

draft-ietf-httpbis-safe-method-w-body.xml
nicowilliams: I don't want to give the impression that this is something that caches should do.  You want exactly that near as I can tell.  Clearly we differ on this.  I think we've reached the end of productive discussion on GitHub issues commentary.

reschke:
@nicowilliams - pleas check whether the latest changes help (https://github.com/httpwg/http-extensions/pull/3108/files#diff-fbb7f16de95d1030dd8fd422ce02cdfe9a7aec318a23a81db89b0aa924a6584f) - and ignore the changes to the other files; I screwed up the commit log.

reschke:
closed in favor of https://github.com/httpwg/http-extensions/pull/3131

PR #3110, opened 2025-06-16
Cookie layering - HttpOnly and HostHttpOnly prefixes

yoavweiss:
As discussed at [IETF 122](https://httpwg.org/wg-materials/ietf122/minutes.html#delete-cookie-and-_httponly-prefix), this is a stab at integrating https://yoavweiss.github.io/httponly_prefix/draft-httponlyprefix-weiss-http.html with the Cookie layering processing model.


draft-ietf-httpbis-layered-cookies.md
```diff
@@ -1260,11 +1289,35 @@ boolean _httpOnlyAllowed_, boolean _allowNonHostOnlyCookieForPublicSuffix_, and
 
    then return null.
 
+1. If _cookie_'s name, byte-lowercased, starts with `__httponly-` and not all of the following are true:
+
+    1. _cookie_'s secure-only is true; and
+
+    1. _cookie_'s http-only is true;
```
annevk:
```suggestion
    1. _cookie_'s http-only is true,
```

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -1260,11 +1289,35 @@ boolean _httpOnlyAllowed_, boolean _allowNonHostOnlyCookieForPublicSuffix_, and
 
    then return null.
 
+1. If _cookie_'s name, byte-lowercased, starts with `__httponly-` and not all of the following are true:
```
annevk: I think this would be clearer as (same for the one directly below):

1. If _cookie_'s name, ... and:
   1. _cookie's secure-only is false; or
   2. _cookie's http-only is false,

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -669,6 +669,35 @@ Set-Cookie: __Host-SID=12345; Secure; Path=/
...
+This combination yields a cookie that hews as closely as a cookie can to
+treating the origin as a security boundary, while at the same time ensuring developers
+and server operators that its scope is limited to HTTP requests. The lack of a `Domain`
+attribute ensures that cookie's host-only is true, locking the cookie to a
+particular host, rather than allowing it to span subdomains. Setting the `Path`
+to `/` means that the cookie is effective for the entire host, and won't be
+overridden for specific paths. The `Secure` attribute ensures that the cookie
+is unaltered by non-secure origins, and won't span protocols. The `HttpOnly` attribute
+ensures that the cookie is not exposed by the user agent to non-HTTP APIs.
+
```
annevk, line 700: You've eaten a newline here.

annevk:
I like the idea of the prefix being just `Http` for brevity's sake. I also think we should have an issue for the multiple discussion as it'll come back.

draft-ietf-httpbis-layered-cookies.md
yoavweiss: done

draft-ietf-httpbis-layered-cookies.md
yoavweiss: done

draft-ietf-httpbis-layered-cookies.md
yoavweiss: Added back

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -1178,6 +1208,22 @@ cookie attribute "wins".
 
 ### Store a Cookie {#store-a-cookie}
 
+A cookie _cookie_ is **Host prefix compatible** if:
```
annevk:
```suggestion
A cookie _cookie_ is **Host-prefix compatible** if:
```

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -1178,6 +1208,22 @@ cookie attribute "wins".
...
+
+1. _cookie_'s secure-only is true;
+
+1. _cookie_'s host-only is true; and
+
+1. _cookie_'s path's size is 1 and _cookie_'s path[0] is the empty string,
+
+
+A cookie _cookie_ is **Http prefix compatible** if:
```
annevk:
```suggestion
A cookie _cookie_ is **Http-prefix compatible** if:
```

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -1178,6 +1208,22 @@ cookie attribute "wins".
...
+
+1. _cookie_'s secure-only is true; and
+
+1. _cookie_'s http-only is false,
```
annevk: Both of these should go to Cookie Struct Miscellaneous.

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -1250,21 +1296,21 @@ boolean _httpOnlyAllowed_, boolean _allowNonHostOnlyCookieForPublicSuffix_, and
 
    Note: The check here and those below are with a byte-lowercased value in order to protect servers that process these values in a case-insensitive manner.
 
-1. If _cookie_'s name, byte-lowercased, starts with `__host-` and not all of the following are true:
+1. If _cookie_'s name, byte-lowercased, starts with `__host-` and it is not **Host prefix compatible**, then return null.
```
annevk:
```suggestion
1. If _cookie_'s name, byte-lowercased, starts with `__host-` and _cookie_ is not **Host-prefix compatible**, then return null.
```

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -1250,21 +1296,21 @@ boolean _httpOnlyAllowed_, boolean _allowNonHostOnlyCookieForPublicSuffix_, and
...
 
-    1. _cookie_'s secure-only is true;
+1. If _cookie_'s name, byte-lowercased, starts with `__http-` and it is not **Http prefix compatible**, then return null.
```
annevk:
```suggestion
1. If _cookie_'s name, byte-lowercased, starts with `__http-` and _cookie_ is not **Http-prefix compatible**, then return null.
```

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -1250,21 +1296,21 @@ boolean _httpOnlyAllowed_, boolean _allowNonHostOnlyCookieForPublicSuffix_, and
...
 
-    1. _cookie_'s host-only is true; and
+1. If _cookie_'s name, byte-lowercased, starts with `__hosthttp-` and it is not both **Host prefix compatible** and **Http prefix compatible*, then return null.
```
annevk:
```suggestion
1. If _cookie_'s name, byte-lowercased, starts with `__hosthttp-` and _cookie_ is not both **Host-prefix compatible** and **Http-prefix compatible*, then return null.
```

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -1250,21 +1296,21 @@ boolean _httpOnlyAllowed_, boolean _allowNonHostOnlyCookieForPublicSuffix_, and
...
+1. If _cookie_'s name is the empty byte sequence and one of the following is true:
 
-   then return null.
+    * _cookie_'s value, byte-lowercased, starts with `__secure-`,
 
-1. If _cookie_'s name is the empty byte sequence and one of the following is true:
+    * _cookie_'s value, byte-lowercased, starts with `__host-`,
 
-    * _cookie_'s value, byte-lowercased, starts with `__secure-`, or
+    * _cookie_'s value, byte-lowercased, starts with `__http-`
```
annevk:
```suggestion
    * _cookie_'s value, byte-lowercased, starts with `__http-`, or
```

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -1250,21 +1296,21 @@ boolean _httpOnlyAllowed_, boolean _allowNonHostOnlyCookieForPublicSuffix_, and
...
 
-    * _cookie_'s value, byte-lowercased, starts with `__host-`
+    * _cookie_'s value, byte-lowercased, starts with `__hosthttp-`
```
annevk:
```suggestion
    * _cookie_'s value, byte-lowercased, starts with `__hosthttp-`,
```

draft-ietf-httpbis-layered-cookies.md
yoavweiss: done!

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -669,6 +669,36 @@ Set-Cookie: __Host-SID=12345; Secure; Path=/
 ~~~
 
 
+#### The "__Http-" Prefix
+
+If a cookie's name begins with a case sensitive match for the string
+`__Http`, then the cookie will have been set with a `Secure` attribute, and an
```
miketaylr:
```suggestion
`__Http-`, then the cookie will have been set with a `Secure` attribute, and an
```

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -669,6 +669,36 @@ Set-Cookie: __Host-SID=12345; Secure; Path=/
...
+`HttpOnly` attribute.
+
+This helps developers and server operators to know that the cookie was set using
+a `Set-Cookie` header, and is limited in scope to HTTP requests.
+
+
+#### The "__HostHttp-" prefix
+
+If a cookie's name begins with a case sensitive match for the string
+`__HostHttp`, then the cookie will have been set with a `Secure` attribute, an
```
miketaylr:
```suggestion
`__HostHttp-`, then the cookie will have been set with a `Secure` attribute, an
```

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -669,6 +669,36 @@ Set-Cookie: __Host-SID=12345; Secure; Path=/
...
+`HttpOnly` attribute, a `Path` attribute with a value of `/`, and no `Domain` attribute.
+
+This helps developers and server operators to know that the cookie was set using
+a `Set-Cookie` header, and is limited in scope to HTTP requests.
+
+This combination yields a cookie that hews as closely as a cookie can to
+treating the origin as a security boundary, while at the same time ensuring developers
+and server operators that its scope is limited to HTTP requests. The lack of a `Domain`
```
miketaylr: server operators **know**?

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -790,6 +820,21 @@ A cookie's **last-access-time** is a time. It is initially the current time.
 
 A cookie is **expired** if its expiry-time is non-null and its expiry-time is in the past.
 
+A cookie _cookie_ is **Host-prefix compatible** if:
+
+1. _cookie_'s secure-only is true;
+
+1. _cookie_'s host-only is true; and
+
+1. _cookie_'s path's size is 1 and _cookie_'s path[0] is the empty string,
```
miketaylr, line 833: I realize you're inheriting this from the cookie layering draft, but is the empty string correct? 

Above (in prose) this is described as:

> `Path` attribute with a value of `/`

(that's also what 6265bis has in 5.7.21.3)

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -1250,21 +1295,21 @@ boolean _httpOnlyAllowed_, boolean _allowNonHostOnlyCookieForPublicSuffix_, and
 
    Note: The check here and those below are with a byte-lowercased value in order to protect servers that process these values in a case-insensitive manner.
 
-1. If _cookie_'s name, byte-lowercased, starts with `__host-` and not all of the following are true:
+1. If _cookie_'s name, byte-lowercased, starts with `__host-` and _cookie_ is not **Host-prefix compatible**, then return null.
 
-    1. _cookie_'s secure-only is true;
+1. If _cookie_'s name, byte-lowercased, starts with `__http-` and _cookie_ is not **Http-prefix compatible**, then return null.
```
miketaylr: Why do we lowercase to `__http-` if it has to be a case-sensitive match with `__Http-` to begin with (above, in prose)?

draft-ietf-httpbis-layered-cookies.md
yoavweiss: In case there's a Path attribute, I see the following algorithm:
```
1. If _attributeName_ is a byte-case-insensitive match for `Path`:
    1. If _attributeValue_ is not empty and if the first byte of  _attributeValue_ is 0x2F (/), then set _cookie_'s path to _attributeValue_  split on 0x2F (/).
```
So I think a list of size 1 with an empty string is correct here.
  
In case there's no `Path` attribute, we fall back to the **Cookie Default Path** algorithm which I think intends to do the same thing and set the path as a list of size 1 with an empty string if the path contains only a single `/`.
  
 At the same time, that algorithm operates on a list (and currently receives a URL). I'm also not sure if we should accept a `__Host` prefix if the cookie doesn't include an explicit `Path=/` attribute.

draft-ietf-httpbis-layered-cookies.md
yoavweiss: Everyone else were doing it? :shrug:

Should we change the prose?

draft-ietf-httpbis-layered-cookies.md
yoavweiss: done, thanks

draft-ietf-httpbis-layered-cookies.md
annevk: If you want to change the conformance requirements, please file a separate issue. We shouldn't change that as part of this PR.

draft-ietf-httpbis-layered-cookies.md
annevk: We should not, according to bis: https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#section-5.7

However, that means we have to keep additional data around that we're currently not. We need to store some information about the attributes on the parsed cookie in some manner. That's unfortunate. https://github.com/httpwg/http-extensions/issues/3112

draft-ietf-httpbis-layered-cookies.md
yoavweiss: Oh, so there's a difference between the conformance requirements and the processing (which is more lenient). Got it!

draft-ietf-httpbis-layered-cookies.md
yoavweiss: > I'm also not sure if we should accept a `__Host` prefix if the cookie doesn't include an explicit `Path=/` attribute.

RFC 6265bis suggests we shouldn't, but I'm not sure there's a way to make that distinction ATM. I can add a `default_path` flag that would enable it

annevk: I think it would be more natural to add a field to cookie along the lines of "had explicit path" that we then check here in addition to checking that it had the correct value.

Alternatively we could maybe compute ahead of time whether it's a host-prefix cookie and store that as a bit during parsing.

yoavweiss: Added a `path-attribute` boolean. Let me know if this makes sense (and if this should be split to a separate PR, as it impacts `__Host`)

yoavweiss: (I'm on train wifi, so I made the above comments oblivious to your comments. Apologies for any incoherence)

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -669,6 +669,36 @@ Set-Cookie: __Host-SID=12345; Secure; Path=/
 ~~~
 
 
+#### The "__Http-" Prefix
+
+If a cookie's name begins with a case sensitive match for the string
```
annevk:
```suggestion
If a cookie's name begins with a case-sensitive match for the string
```

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -669,6 +669,36 @@ Set-Cookie: __Host-SID=12345; Secure; Path=/
...
+`__Http-`, then the cookie will have been set with a `Secure` attribute, and an
+`HttpOnly` attribute.
+
+This helps developers and server operators to know that the cookie was set using
+a `Set-Cookie` header, and is limited in scope to HTTP requests.
+
+
+#### The "__HostHttp-" prefix
+
+If a cookie's name begins with a case sensitive match for the string
```
annevk:
```suggestion
If a cookie's name begins with a case-sensitive match for the string
```

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -980,13 +1029,15 @@ A host _host_ **Domain-Matches** a string _domainAttributeValue_ if at least one
...
+2.  Let _splitPath_ be _path_ split on 0x2F (/).
 
-2.  If _path_'s size is greater than 1, then remove _path_'s last item.
+3.  If _splitPath_'s size is greater than 1, then remove _splitPath_'s last item.
 
-3.  Otherwise, set _path_[0] to the empty string.
+4.  Otherwise, set _splitPath_[0] to the empty string.
 
-4.  Return _path_.
+5.  Return _splitPath_.
```
annevk: This is all wrong. A URL path is defined as a list.

draft-ietf-httpbis-layered-cookies.md
```diff
@@ -770,6 +800,8 @@ A cookie's **host-only** is a boolean. It is initially false.
 
 A cookie's **path** is a URL path.
 
+A cookie's **path-attribute** is a boolean. It is initially false.
```
annevk: I went through the document to check whether this is needed for the Secure attribute and it's not. It also doesn't appear to be needed for the Domain attribute.

However, I do think we should rename this to "has path attribute" for clarity.

draft-ietf-httpbis-layered-cookies.md
yoavweiss: I did not know that. I'll revert

draft-ietf-httpbis-layered-cookies.md
yoavweiss: done

draft-ietf-httpbis-layered-cookies.md
yoavweiss: Resolving as I believe this is now done

draft-ietf-httpbis-layered-cookies.md
miketaylr: I can file an issue.

miketaylr (approved):

Issue #3121, opened 2025-06-29
QUERY: HTTPDIR/RF review - Abstract


reschke:
> Abstract
>
>    This specification defines a new HTTP method, QUERY, as a safe,
>    idempotent request method that can carry request content.

The above is too generic. It should say why it is being defined:

   This specification defines the QUERY method for HTTP.
   A QUERY requests that the request target process the enclosed
   content in a safe/idempotent manner and then respond with the
   result of that processing. This is similar to POST requests but
   can be automatically repeated or restarted without concern for
   partial state changes. The kind of processing desired to perform
   the QUERY is indicated by the Content-Type header field value
   enclosed with the request.

reschke:
It's a bit verbose for my taste, but let's go ahead with the proposal.

PR #3124, opened 2025-06-30
QUERY: address HTTPDIR/RF review for Abstract (addresses #3121)

reschke:


royfielding (approved):

royfielding:
I think a single paragraph is better for an abstract. The last sentence is just a clarification that could be moved into section 2 to really explain how Content-Type is being used, not just implied by a couple examples.

reschke:
Abandoned in favor of https://github.com/httpwg/http-extensions/pull/3125

PR #3125, opened 2025-07-01
QUERY: address HTTPDIR/RF review for Abstract

reschke:


draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -61,8 +61,12 @@
 
     <abstract>
       <t>
-        This specification defines a new HTTP method, QUERY, as a safe, idempotent
-        request method that can carry request content.
+        This specification defines the QUERY method for HTTP.
```
CxRes, line 64: Proposal for slightly crisper language: 

"This specification defines the QUERY method for HTTP.  QUERY is a safe and idempotent method that requests a target to process the enclosed body and respond with the result. ..."

PR #3132, opened 2025-07-14
[Unencoded-Digest] Allow parameterized values.

mikewest:
The spec is currently silent on the subject of parameters, and it seems reasonable to consistently make room for future extension by explicitly defining them as allowed but ignored. This has little practical impact, but hopefully ensures that implementers would treat `sha-256=:...:;options=plz` and `sha-256=:...:` as equivalent declarations.

### DIFF
--- a/draft-ietf-httpbis-unencoded-digest.md
+++ b/draft-ietf-httpbis-unencoded-digest.md
@@ -137,7 +137,11 @@ where each:
   compute the digest;
 * value is a `Byte Sequence` ({{Section 3.3.5 of STRUCTURED-FIELDS}}), that
   conveys an encoded version of the byte output produced by the digest
-  calculation.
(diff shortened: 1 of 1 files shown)


LPardue:
This looks good. Although it makes me a little sad to realize it's something we probably should also have said in RFC 9530 (but too late now). Therefore, we should add the same text to the Want-Unencoded-Digest field, just in case some future extension needs it.

mikewest:
Yeah. The immutable nature of RFCs is a double-edged sword. 🤷 I guess this isn't enough of a reason to supersede that document, though.

In any event, updated the PR. The linter is unhappy with this iteration (and the last one) due to trailing whitespace in both places we make this change. It seems like that check is going to be difficult to satisfy while also having paragraphs in bulleted lists. :/

PR #3135, opened 2025-07-21
QUERY: HTTPDIR/RF review - Redirection (closes #3119)

### LINKED ISSUE #3119: QUERY: HTTPDIR/RF review - Redirection
> 2.2.  Redirection
>
>    In some cases, the server may choose to respond indirectly to the
>    QUERY request by redirecting the user agent to a different URI (see
>    Section 15.4 of [HTTP]).  The semantics of the redirect response do
>    not differ from other methods.
>
>    For instance, a 303 (See Other, Section 15.4.4 of [HTTP]) response
>    would indicate that the Location field identifies an alternate URI
>    from which the results can be retrieved using a GET request (this use
>    case is also covered by the use of the Location response field in a
>    2xx response).

No, that is incorrect. In a 303 response, the Location field identifies
a replacement target resource that will perform the same query when it
receives a GET request. To complete processing of the original QUERY, the
user agent will need to perform a GET request on the resource referenced
by Location. This allows the original query to be identified for reuse as
a normal resource and for the results to be cached.

reschke:
This changes the definition of 303 -> Location to identify the "stored query" for GET, representing _current_ results.

Interestingly enough, this is what the example showed already.

### DIFF
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -293,23 +293,28 @@ q=foo&amp;limit=10&amp;sort=-published
       <t>
         In some cases, the server may choose to respond indirectly to the QUERY
         request by redirecting the user agent to a different URI (see
-        <xref target="HTTP" section="15.4"/>). The semantics of the redirect
-        response do not differ from other methods.
+        <xref target="HTTP" section="15.4"/>). The semantics of the various redirect
+        status codes are the same as those defined in in that specification,
       </t>
       <t>
-        For the status code 303 (See Other, <xref target="HTTP" section="15.4.4"/>) in a
-        response to a QUERY request, this specification defines the semantics
+        In particular, the status codes 307 (Temporary Redirect, <xref target="HTTP" format="comma" section="15.4.8"/>) and 308
+        (Permanent Redirect, <xref target="HTTP" format="comma" section="15.4.9"/>) indicate that
+        the target resource resides at a different URI, either temporarily or permanently.
+      </t>
+      <t>
+        Note that in the context of the QUERY method, status codes 301 (Moved Permanently, <xref target="HTTP" format="comma" section="15.4.2"/>)
+        and 302 (Found, <xref target="HTTP" format="comma" section="15.4.3"/>) have exactly the
+        same meaning as status codes 308 and 307, as their only difference in behaviour is restricted
+        to responses to POST requests.
+      </t>
+      <t>
+        For the status code 303 (See Other, <xref target="HTTP" section="15.4.4"/>), this specification defines the semantics
         of the Location response field (<xref target="HTTP" sectionFormat="comma" section="10.2.2"/>) to
         identify an alternate URI that represents the QUERY request, and from which current results
         can be retrieved using a GET request.
       </t>
       <t>
-        On the other hand, response codes such as 307 (Temporary Redirect, <xref target="HTTP" format="comma" section="15.4.8"/>) and 308
-        (Permanent Redirect, <xref target="HTTP" format="comma" section="15.4.9"/>) can be used to request the user agent to redo
-        the QUERY request on the URI specified by the Location field.
-      </t>
-      <t>
-        See <xref target="example.indirect"/> for an example.
+        See <xref target="example.indirect"/> for an example use of the status code 303.
       </t>
     </section>
 
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -309,10 +309,8 @@ q=foo&amp;limit=10&amp;sort=-published
         similar QUERY request to the new target URI referenced by Location.
       </t>
       <t>
-        Note that in the context of the QUERY method, status codes 301 (Moved Permanently, <xref target="HTTP" format="comma" section="15.4.2"/>)
-        and 302 (Found, <xref target="HTTP" format="comma" section="15.4.3"/>) have exactly the
-        same meaning as status codes 308 and 307, as their only difference in behaviour is restricted
-        to responses to POST requests.
+        Note that the exceptions for redirecting a POST as a GET request after a 301 or 302 response
+        do not apply to QUERY requests.
       </t>
       <t>
         For the status code 303 (See Other, <xref target="HTTP" section="15.4.4"/>), this specification defines the semantics
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -297,20 +297,19 @@ q=foo&amp;limit=10&amp;sort=-published
         response do not differ from other methods.
       </t>
       <t>
-        For instance, a 303 (See Other, <xref target="HTTP" section="15.4.4"/>)
-        response would indicate that the Location field identifies an
-        alternate URI from which the results can be retrieved
-        using a GET request (this use case is also covered by the
-        use of the Location response field in a 2xx response).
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -297,20 +297,19 @@ q=foo&amp;limit=10&amp;sort=-published
         response do not differ from other methods.
```
```suggestion
        response are defined by the particular status code.
```
I think this needs elaboration because of 301/302 (what that history means for QUERY given that the historical notes are only about POST) and the reasons for choosing particular codes. It isn't useful to make statements ambiguous with "In some cases" and "such as". 

For example, I suggest:

The redirect status code semantics are the same as for GET except for the following distinctions:
  1) 301 indicates the resource motivating the query has permanently moved (i.e., the form or context that guides construction of a query for this resource has moved to a different Location); the user agent is expected to perform a GET request on the identified Location resource for further instruction.
  2) 302 indicates that the results for this QUERY are not enclosed, but can be obtained or described by performing a retrieval request (a GET request if using HTTP) targeting the URI reference supplied in the [Location](https://www.rfc-editor.org/rfc/rfc9110.html#field.location) header field. Note that this redirection URI is not considered equivalent to the original QUERY target URI because it might only reflect the results of that specific query, not the results of similar future queries.
  3) 303 indicates that the results for this QUERY are not enclosed, but the same query can be performed via a retrieval request (a GET request if using HTTP) targeting the URI reference supplied in the [Location](https://www.rfc-editor.org/rfc/rfc9110.html#field.location) header field. Note that this redirection URI is intended to be equivalent to the original QUERY request as a whole, much like a stored procedure for a database, and accessed via GET to improve cacheability and subsequent reuse as a linkable resource.
  4) 307 indicates a temporary redirect to the resource identified in Location; the user agent MAY continue the request by sending the same QUERY to the resource identified in Location but should not assume that the new location is a permanent replacement for the original request URI.
  5) 308 indicates a permanent redirect to the resource identified in Location; the user agent MAY continue the request by sending the same QUERY to the resource identified in Location and can use this new identifier as a permanent replacement for the original request URI.

royfielding (changes requested):

draft-ietf-httpbis-safe-method-w-body.xml
reschke: I hear you, but that interpretation of 301/302 is inconsistent with RFC 9110. We should stick with what we say there ~; we don't _expect_ a client to do a GET request, but we allow that. On all other aspects, 301/302 are the same as 308/307~.

reschke: In particular, we carefully said in the current RFCs, that for 301 and 302, _only_ POST can be rewritten to GET - and that actually was tested with user agents.

Thus, trying to special-case these for POST seems to be entirely unproductive.

What you suggest for 307 and 308 appears to be exactly what RFC 9110 says.

We can certainly tune the text for 303/307/308 when needed. We could also say that for the case of QUERY, 301/302 mean exacly the same thing as 308/307.

reschke: I have update the PR slightly - but I still believe that adding detailed definitions of 301/302/307/308 does not make any sense here, in particular if these details conflict with that RFC 9110 says.

Next step would be to add one more example wrt 301/302/307/308.

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -293,24 +293,28 @@ q=foo&amp;limit=10&amp;sort=-published
       <t>
         In some cases, the server may choose to respond indirectly to the QUERY
         request by redirecting the user agent to a different URI (see
-        <xref target="HTTP" section="15.4"/>). The semantics of the redirect
-        response do not differ from other methods.
+        <xref target="HTTP" section="15.4"/>). The semantics of the various redirect
+        status codes are the same as those defined in in that specification,
```
```suggestion
        <xref target="HTTP" section="15.4"/>).
```
The sentence doesn't say anything useful.

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -293,24 +293,28 @@ q=foo&amp;limit=10&amp;sort=-published
...
       </t>
       <t>
-        For instance, a 303 (See Other, <xref target="HTTP" section="15.4.4"/>)
-        response would indicate that the Location field identifies an
-        alternate URI from which the results can be retrieved
-        using a GET request (this use case is also covered by the
-        use of the Location response field in a 2xx response).
+        In particular, the status codes 307 (Temporary Redirect, <xref target="HTTP" format="comma" section="15.4.8"/>) and 308
+        (Permanent Redirect, <xref target="HTTP" format="comma" section="15.4.9"/>) indicate that
+        the target resource resides at a different URI, either temporarily or permanently.
```
royfielding: Since we aren't changing the definition, these can be simplified 
```suggestion
        A response with either of the status codes
        301 (Moved Permanently, <xref target="HTTP" format="comma" section="15.4.2"/>) or
        308 (Permanent Redirect, <xref target="HTTP" format="comma" section="15.4.9"/>)
        indicates that the target resource has permanently moved to a different URI referenced by the
        Location response field (<xref target="HTTP" sectionFormat="comma" section="10.2.2"/>).
        Likewise, a response with either
        302 (Found, <xref target="HTTP" format="comma" section="15.4.3"/> or
        307 (Temporary Redirect, <xref target="HTTP" format="comma" section="15.4.8"/>)
        indicates that the target resource has temporarily moved. In all four cases, the server is
        suggesting that the user agent can accomplish its original QUERY request by sending a
        similar QUERY request to the new target URI referenced by Location.
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -293,24 +293,28 @@ q=foo&amp;limit=10&amp;sort=-published
...
       </t>
       <t>
-        On the other hand, response codes 307 (Temporary Redirect, <xref target="HTTP" section="15.4.8"/>) and 308
-        (Permanent Redirect, <xref target="HTTP" section="15.4.9"/>) can be used to request the user agent to redo
-        the QUERY request on the URI specified by the Location field.
+        Note that in the context of the QUERY method, status codes 301 (Moved Permanently, <xref target="HTTP" format="comma" section="15.4.2"/>)
+        and 302 (Found, <xref target="HTTP" format="comma" section="15.4.3"/>) have exactly the
+        same meaning as status codes 308 and 307, as their only difference in behaviour is restricted
+        to responses to POST requests.
```
```suggestion
        Note that the exceptions for redirecting a POST as a GET request after a 301 or 302 response
        do not apply to QUERY requests.
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -293,24 +293,28 @@ q=foo&amp;limit=10&amp;sort=-published
...
       <t>
-        Various non-normative examples of successful
-        QUERY responses are illustrated in <xref target="examples"/>.
+        For the status code 303 (See Other, <xref target="HTTP" section="15.4.4"/>), this specification defines the semantics
+        of the Location response field (<xref target="HTTP" sectionFormat="comma" section="10.2.2"/>) to
+        identify an alternate URI that represents the QUERY request, and from which current results
+        can be retrieved using a GET request.
+      </t>
+      <t>
+        See <xref target="example.indirect"/> for an example use of the status code 303.
```
```suggestion
        A response to QUERY with the status code 303 (See Other, <xref target="HTTP" section="15.4.4"/>)
        indicates that the original query can be accomplished via a normal retrieval request on the URI referenced
        by the Location response field (<xref target="HTTP" sectionFormat="comma" section="10.2.2"/>).
        For HTTP, this means sending a GET request to the new target URI, as illustrated by the example
        in <xref target="example.indirect"/>.
```

royfielding (changes requested):

reschke:
@royfielding - thanks a lot for the feedback.

PR #3138, opened 2025-07-23
Disable buffering in endpoints too

martinthomson:
Thanks to @bemasc who suggested that this field might be a signal to endpoint software.

### DIFF
--- a/draft-ietf-httpbis-incremental.md
+++ b/draft-ietf-httpbis-incremental.md
@@ -115,7 +115,7 @@ to be forwarded, intermediaries can still buffer the entire header and trailer
 sections of the message before forwarding them downstream.
 
 The request to use incremental forwarding also applies to HTTP implementations.
-Though most HTTP APIs provide the ability to incrementally write message content,
+Though most HTTP APIs provide the ability to incrementally transfer message content,
 those that do not for any reason, SHOULD use the presence of the Incremental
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-incremental.md
```diff
@@ -114,6 +114,11 @@ arrive. As the Incremental header field indicates only how the message content i
 to be forwarded, intermediaries can still buffer the entire header and trailer
 sections of the message before forwarding them downstream.
 
+The request to use incremental forwarding also applies to HTTP implementations.
+Though most HTTP APIs provide the ability to incrementally write message content,
```
bemasc: Specifically, my understanding is that iOS NSURLRequest streams requests and responses, but only streams the response body after the request body is complete.  For "incremental in both directions" to work as intended, that behavior would need to change.  I'm not sure this text (focused on "writes") quite captures that.

martinthomson: I would not regard iOS as meeting this requirement, because it involves unbounded buffering.  Though maybe there is a simple change to make...

```suggestion
Though most HTTP APIs provide the ability to incrementally transfer message content,
```

tfpauly: NSURL allows doing incremental reading or waiting for a complete read, etc. That's a fairly common model I believe.

PR #3148, opened 2025-08-03
QUERY: tune method definition (addresses #3213)

reschke:
This should address the points raised in #3213, except for a discussion of other status codes.

### DIFF
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -238,42 +238,45 @@ q=foo&amp;limit=10&amp;sort=-published
   <section title="QUERY" anchor="query">
     <t>
       The QUERY method is used to initiate a server-side query. Unlike
-      the HTTP GET method, which requests that a server return a
+      the GET method, which requests a
       representation of the resource identified by the target URI
       (as defined by <xref target="HTTP" section="7.1"/>), the QUERY
-      method is used to ask the server to perform a query operation
-      (described by the request content) over some set of data at the
-      resource. The content returned in response to a QUERY
-      cannot be assumed to be a representation of the resource identified by
-      the target URI.
+      method is used to ask the target resource to perform a query operation
+      (described by the request content) within the scope of that target
+      resource (as defined by its origin).
     </t>
     <t>
       The content of the request and it's media type define the query.
-      Implementations &MAY; use a request content of any media type with
-      the QUERY method, provided that it has appropriate query semantics.
+      Servers <bcp14>MUST</bcp14> fail the request if the Content-Type
+      request field (<xref target="HTTP" sectionFormat="comma" section="8.3"/>)
+      is missing or is inconsistent with the request content.
     </t>
     <t>
       As for all HTTP methods in general, the target URI's query part
-      takes part in identifying the resource being queried and therefore is
-      not part of the actual query. Whether and how the URI's query part
-      directly affects the result of the query is implementation specific
+      takes part in identifying the resource being queried. Whether and how it
+      directly affects the result of the query is specific to the resource
       and out of scope for this specification.
     </t>
     <t>
-      QUERY requests are both safe and idempotent with regard to the
-      resource identified by the request URI. That is, QUERY requests do not
-      alter the state of the identified resource. However, while processing a
-      QUERY request, a server can be expected to allocate computing and memory
-      resources or even create additional HTTP resources through which the
-      response can be retrieved.
+      QUERY requests are safe with regard to the target resource
+      (<xref target="HTTP" sectionFormat="comma" section="9.2.1"/>) --
+      that is, the client does not request or expect any change to the
+      state of the target resource. This does not prevent the server
+      from creating additional HTTP resources through which additional
+      information can be retrieved (see sections below).
     </t>
     <t>
-      A successful response to a QUERY request is expected to provide some
-      indication as to the final disposition of the operation. For
-      instance, a successful query that yields no results can be represented
-      by a 204 (No Content, <xref target="HTTP" section="15.3.5"/>) response. If the response includes content,
-      it is expected to describe the results of the operation.
+      Furthermore, QUERY requests are idempotent
+      (<xref target="HTTP" sectionFormat="comma" section="9.2.2"/>) --
+      they can be retried or repeated when needed, for instance after
+      a connection failure.
     </t>
+    <t>
+      A 200 (OK) response to a QUERY request indicates that the query
+      was successfully processed and the results of that processing are
+      enclosed as the response content.
+    </t>
+    <cref>describe other status codes</cref>
 
     <section title="Content-Location Response Field" anchor="content-location">
       <t>


draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -238,42 +238,45 @@ q=foo&amp;limit=10&amp;sort=-published
...
     <t>
       The content of the request and it's media type define the query.
-      Implementations &MAY; use a request content of any media type with
-      the QUERY method, provided that it has appropriate query semantics.
+      Servers <bcp14>MUST</bcp14> fail the request if the Content-Type
+      request field (<xref target="HTTP" sectionFormat="comma" section="8.3"/>)
+      is missing or is inconsistent with the request content.
     </t>
     <t>
       As for all HTTP methods in general, the target URI's query part
```
mnot:
```suggestion
      As for all HTTP methods, the target URI's query part
```

reschke:
Ok, I believe this is ready. Please review - @royfielding @mnot @martinthomson @MikeBishop ...

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -238,42 +238,88 @@ q=foo&amp;limit=10&amp;sort=-published
...
       representation of the resource identified by the target URI
       (as defined by <xref target="HTTP" section="7.1"/>), the QUERY
-      method is used to ask the server to perform a query operation
-      (described by the request content) over some set of data at the
-      resource. The content returned in response to a QUERY
-      cannot be assumed to be a representation of the resource identified by
-      the target URI.
+      method is used to ask the target resource to perform a query operation
+      (described by the request content) within the scope of that target
+      resource (as defined by its origin).
```
martinthomson: There are a lot of parentheticals here.  Would this be clearer?


```suggestion
      within the scope of that target resource.
      The query operation is described by the request content.
      The origin server determines the scope of the operation based on the target resource.
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -238,42 +238,88 @@ q=foo&amp;limit=10&amp;sort=-published
...
-      alter the state of the identified resource. However, while processing a
-      QUERY request, a server can be expected to allocate computing and memory
-      resources or even create additional HTTP resources through which the
-      response can be retrieved.
+      QUERY requests are safe with regard to the target resource
+      (<xref target="HTTP" sectionFormat="comma" section="9.2.1"/>) --
+      that is, the client does not request or expect any change to the
+      state of the target resource. This does not prevent the server
+      from creating additional HTTP resources through which additional
+      information can be retrieved (see sections below).
```
martinthomson: Can you cite a specific section or sections rather than "below"?

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -238,42 +238,88 @@ q=foo&amp;limit=10&amp;sort=-published
...
+      </t>
+      <ul>
+        <li>
+          A request lacking media type information by definition is incorrect and needs to fail
+          with a 4xx status code such as 400 (Client Error).
+        </li>
+        <li>
+          If a media type is specified, but not supported by the resource, a 415 (Unsupported Media Type)
+          is appropriate. This specifically includes the case where the media type is
+          known in principle, but lacks semantics specific to QUERY (for example "text/xml"). In both cases,
```
martinthomson:
```suggestion
          known in principle, but lacks semantics specific to a QUERY to the target resource. In both cases,
```

I understand what you are trying to say here, but I think that "text/xml" is an entirely valid thing to use, if the resource knows what to do with it.  The reason to use 415 is that *this particular resource* cannot process *this specific media type* as a QUERY.

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -238,42 +238,88 @@ q=foo&amp;limit=10&amp;sort=-published
...
+        </li>
+        <li>
+          If a media type is specified, but is inconsistent with the actual request content,
+          a 400 (Bad Request) can be returned. Thus, "content sniffing" is not allowed.
+        </li>
+        <li>
+          If the media type is specified, is understood, and the content is indeed consistent with the
+          type, but the query can not be processed due to the actual contents of the query,
+          the status 422 (Unprocessable Content) can be used. An example would be a syntactically
+          correct SQL query querying a non-existing table.
```
martinthomson:
```suggestion
          correct SQL query that identifies a non-existent table.
```

martinthomson (approved):
Some suggestions that you can take or leave as you prefer.

draft-ietf-httpbis-safe-method-w-body.xml
reschke: yes -> https://github.com/httpwg/http-extensions/commit/e302b3a593b6a1cf64024260191368dd5ceaeed0

draft-ietf-httpbis-safe-method-w-body.xml
reschke: I'm not 100% that we can't be more rigid - but that would be a different discussion.

reschke:
Thanks, @martinthomson - feeback incorporated.

@mnot , @royfielding , @MikeBishop - any more feedback?

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -238,42 +238,90 @@ q=foo&amp;limit=10&amp;sort=-published
...
-      resource. The content returned in response to a QUERY
-      cannot be assumed to be a representation of the resource identified by
-      the target URI.
+      method is used to ask the target resource to perform a query operation
+      within the scope of that target resource.
+      The query operation is described by the request content.
+      The origin server determines the scope of the operation based on the target resource.
     </t>
     <t>
       The content of the request and it's media type define the query.
```
MikeBishop:
```suggestion
      The content of the request and its media type define the query.
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -238,42 +238,90 @@ q=foo&amp;limit=10&amp;sort=-published
...
+      <ul>
+        <li>
+          A request lacking media type information by definition is incorrect and needs to fail
+          with a 4xx status code such as 400 (Client Error).
+        </li>
+        <li>
+          If a media type is specified, but not supported by the resource, a 415 (Unsupported Media Type)
+          is appropriate. This specifically includes the case where the media type is
+          known in principle, but lacks semantics specific to a QUERY to the target resource. In both cases,
+          the Accept-Query response field (<xref target="field.accept-query"/>) can be used for inform
```
MikeBishop:
```suggestion
          the Accept-Query response field (<xref target="field.accept-query"/>) can be used to inform
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -238,42 +238,90 @@ q=foo&amp;limit=10&amp;sort=-published
...
+          the client of media types which are supported.
+        </li>
+        <li>
+          If a media type is specified, but is inconsistent with the actual request content,
+          a 400 (Bad Request) can be returned. Thus, "content sniffing" is not allowed.
```
MikeBishop: Fine with the error, but I'm unclear on the "content sniffing is not allowed." Isn't it content-sniffing to inspect the content and determine that it's not the claimed content type? Or are you meaning that a server isn't allowed to infer the correct content-type from the payload and override a missing/erroneous content-type header?

Assuming it's the latter, I'd change it to something like "The server MUST NOT attempt to infer the correct media type."

reschke: The latter. That said, this is in the list of recommendations for status codes; the normative requirements are at the start of the subsection.

reschke: @MikeBishop - please see https://github.com/httpwg/http-extensions/pull/3148/commits/d6ccf95ef59130707b8ff0f3da6eb5888c5cd48b

PR #3149, opened 2025-08-04
Gen-ART review - Address technical issues

sbingler:
This PR address the remaining technical issues from the [Gen-ART review](https://lists.w3.org/Archives/Public/ietf-http-wg/2025JanMar/0124.html).

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1145,16 +1145,13 @@ there isn't a clear link between a top-level traversable and a worker.
 This is especially true for Service Workers {{SERVICE-WORKERS}}, which may
 execute code in the background, without any document visible at all.
 
-Note: The descriptions below assume that workers must be same-origin with
-the documents that instantiate them. How to handle non-same-origin workers is
-out of scope of this document.
-
 #### Dedicated and Shared Workers {#dedicated-and-shared-requests}
 
-Dedicated workers are simple, as each dedicated worker is bound to one and only
-one document. Requests generated from a dedicated worker (via `importScripts`,
-`XMLHttpRequest`, `fetch()`, etc) define their "site for cookies" as that
-document's "site for cookies".
+Dedicated workers are simple, as each dedicated worker is bound to one and
+only one document. The worker's "site for cookies" is the document's "site for
+cookies" if the worker's origin is same-site with the document's "site for
+cookies", otherwise its "site for cookies" is an origin set to an opaque
+origin.
 
 Shared workers may be bound to multiple documents at once. As it is quite
 possible for those documents to have distinct "site for cookies" values, the
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -368,6 +368,11 @@ cookie with a path of /c,d=e.
 User agents MAY ignore Set-Cookie header fields based on response status codes or
 the user agent's cookie policy (see {{ignoring-cookies}}).
 
+Note: A cookie's octets MUST be processed as {{USASCII}} characters. While
+it's possible a non-HTTP API could pass a set-cookie-string with one or more
+non-{{USASCII}} characters, no attempt should be made to interpret these octets
+as anything other than {{USASCII}} characters.
+
 ## Examples
 
 Using the Set-Cookie header field, a server can send the user agent a short string
@@ -610,6 +615,9 @@ Per the grammar above, the cookie-value MAY be wrapped in DQUOTE characters.
 Note that in this case, the initial and trailing DQUOTE characters are not
 stripped. They are part of the cookie-value, and will be included in Cookie
 header fields sent to the server.
+
+Per the grammar above, cookie-avs MUST NOT contain leading or trailing WSP
+characters as they will be interpreted as BWS and removed.
 
 The domain-value is a subdomain as defined by {{Section 3.5 of RFC1034}}, and
 as enhanced by {{Section 2.1 of RFC1123}}. Thus, domain-value is a string of
@@ -1138,8 +1146,8 @@ This is especially true for Service Workers {{SERVICE-WORKERS}}, which may
 execute code in the background, without any document visible at all.
 
 Note: The descriptions below assume that workers must be same-origin with
-the documents that instantiate them. If this invariant changes, we'll need to
-take the worker's script's URI into account when determining their status.
+the documents that instantiate them. How to handle non-same-origin workers is
+out of scope of this document.
 
 #### Dedicated and Shared Workers {#dedicated-and-shared-requests}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -368,6 +368,11 @@ cookie with a path of /c,d=e.
 User agents MAY ignore Set-Cookie header fields based on response status codes or
 the user agent's cookie policy (see {{ignoring-cookies}}).
 
+Note: A cookie's octets MUST be processed as {{USASCII}} characters. While
+it's possible a non-HTTP API could pass a set-cookie-string with one or more
+non-{{USASCII}} characters, no attempt should be made to interpret these octets
+as anything other than {{USASCII}} characters.
```
sbingler, line 374: This is meant to address

>RFC 9110 prescribes that HTTP headers be in US-ASCII.  However the
existence of non-HTTP APIs for the cookie store raises the possibility
that a larger set of characters might be supported.  As far as I can
tell, this document does not specify the universe of characters.

I'm trying to communicate that when we do "look at" the octets that we process them as ascii, and to basically avoid trying to interpret anything like
`document.cookie = "foo=🍪"

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -611,6 +616,9 @@ Note that in this case, the initial and trailing DQUOTE characters are not
 stripped. They are part of the cookie-value, and will be included in Cookie
 header fields sent to the server.
 
+Per the grammar above, cookie-avs MUST NOT contain leading or trailing WSP
+characters as they will be interpreted as BWS and removed.
+
```
sbingler, line 621: This is to address the first issue in the review regarding whitespace around cookie attributes.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1138,8 +1146,8 @@ This is especially true for Service Workers {{SERVICE-WORKERS}}, which may
 execute code in the background, without any document visible at all.
 
 Note: The descriptions below assume that workers must be same-origin with
-the documents that instantiate them. If this invariant changes, we'll need to
-take the worker's script's URI into account when determining their status.
+the documents that instantiate them. How to handle non-same-origin workers is
+out of scope of this document.
```
sbingler: Addresses

>5.2.2.  Worker-based requests

>   Note: The descriptions below assume that workers must be same-origin
   with the documents that instantiate them.  If this invariant changes,
   we'll need to take the worker's script's URI into account when
   determining their status.

>The import of this is not clear to me.  One possibility is that
"Currently, all implementations require that workers must be
same-origin with the documents that instantiate them."  Another
possibility is "We only discuss the case when a worker is same-origin
with the document that instantiates it."  In either case, this should
include/point to a discussion of what to do when this is *not* the
case, or a statement that it is out of scope for this document.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1642,8 +1650,8 @@ user agent MUST process the cookie as follows:
 
     1.  Let the domain-attribute be the empty string.
 
-8.  If the domain-attribute contains a character that is not in the range of {{USASCII}}
-    characters, abort this algorithm and ignore the cookie entirely.
+8.  If the domain-attribute contains a character that is not in CHAR, abort
+    this algorithm and ignore the cookie entirely.
```
sbingler, line 1651: Addresses

> "[USASCII]" doesn't seem to be strictly defined.  E.g. the term is
used descriptively in sec 2.2 but a definition is not explicitly
imported from RFC 5234, and 5234 doesn't use the term.  Probably what
is meant is "is not in CHAR", which is explicitly defined.

draft-ietf-httpbis-rfc6265bis.md
mikewest: I think this is consistent with the RFC we're updating, and that it's perfectly reasonable to leave clearer integration with HTML and DOM APIs to the `tre` layering document in progress.

draft-ietf-httpbis-rfc6265bis.md
mikewest: We do now support `new Worker("data:application/javascript,...")`, which creates a dedicated Worker with an opaque origin. I'd suggest that we instead drop this note entirely, and change the dedicated worker text on lines 1154-1157 to support that case explicitly:

> Dedicated workers are simple, as each dedicated worker is bound to one and only one document. The worker's "site for cookies" is the document's "site for cookies" if the worker's origin is same-site with the document's "site for cookies", and an opaque origin otherwise.

WDYT?

mikewest (changes requested):
Looks good, but I think we can take another pass at the worker note. WDYT about the suggestion inline?

draft-ietf-httpbis-rfc6265bis.md
martinthomson: This seems right, and is consistent with the use of opaque origin in HTML.  It's a shame that the use of opaque origin in HTML is just not very good.  It says that opaque origins can be compared for equality, then uses "an opaque origin" often, without specifying which other opaque origins qualify as equal.

In this case, the only opaque origin that would be equal to the opaque origin for this dedicated worker would be the result of any attempt to get the "site for cookies" for exactly the same worker.

That is, another call to `new Worker ("data:...")` would produce a different opaque origin.  I assume.  Otherwise, you might imagine that each attempt to get an origin produces a different opaque origin, which would never compare as equal.

mikewest: > This seems right, and is consistent with the use of opaque origin in HTML. It's a shame that the use of opaque origin in HTML is just not very good. It says that opaque origins can be compared for equality, then uses "an opaque origin" often, without specifying which other opaque origins qualify as equal.

The [same-origin](https://html.spec.whatwg.org/multipage/browsers.html#same-origin) definition clarifies this a bit, though somewhat unfortunately reuses the word "same" with a slightly different meaning (identity vs equality). That could possibly be improved if you have a suggestion for a better framing? 

> In this case, the only opaque origin that would be equal to the opaque origin for this dedicated worker would be the result of any attempt to get the "site for cookies" for exactly the same worker.

Yup. My understanding here matches yours.

draft-ietf-httpbis-rfc6265bis.md
sbingler: Ok. Just to be clear, this text LGTY?

draft-ietf-httpbis-rfc6265bis.md
sbingler: The change sgtm, I went with a slightly different phrasing to stay consistent with nearby text. ptal.

sbingler:
Merging. I'll assume you're okay with https://github.com/httpwg/http-extensions/pull/3149#discussion_r2250651368, but if not I can edit it in a later PR.

PR #3153, opened 2025-08-11
Rename __HostHttp- to __Host-Http- for improved backwards compat

yoavweiss:
Following discussions on Matrix, this renames the `__HostHttp-` to be `__Host-Http-`.

While a bit uglier, this makes it easier to deploy this to non-supporting browsers while not losing the characteristics that `__Host-` provides.

Tested in https://github.com/web-platform-tests/wpt/pull/54226 and https://github.com/web-platform-tests/wpt/pull/54260. Relevant cookie store changes in https://github.com/whatwg/cookiestore/pull/286

### DIFF
--- a/draft-ietf-httpbis-layered-cookies.md
+++ b/draft-ietf-httpbis-layered-cookies.md
@@ -1305,7 +1305,7 @@ boolean _httpOnlyAllowed_, boolean _allowNonHostOnlyCookieForPublicSuffix_, and 
 
 1. If _cookie_'s name, byte-lowercased, starts with `__http-` and _cookie_ is not Http-prefix compatible, then return null.
 
-1. If _cookie_'s name, byte-lowercased, starts with `__hosthttp-` and _cookie_ is not both Host-prefix compatible and Http-prefix compatible, then return null.
+1. If _cookie_'s name, byte-lowercased, starts with `__host-http-` and _cookie_ is not both Host-prefix compatible and Http-prefix compatible, then return null.
 
 1. If _cookie_'s name is the empty byte sequence and one of the following is true:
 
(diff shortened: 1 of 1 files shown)


yoavweiss:
(build failure seems unrelated)

annevk:
Filed https://github.com/httpwg/http-extensions/issues/3155 on the build failure.

This seems okay to me, but we'll need test coverage changes and a corresponding change to the Cookie Store API. And ideally they're all linked from OP for clarity.

annevk:
I don't think a name change is editorial. This has normative impact.

simon-friedberger:
> Following discussions on Matrix, this renames the `__HostHttp-` to be `__Host-Http-`.

Which channel? Could you summarize here?

yoavweiss:
The #cookies channel. 

<details><summary>Discussion</summary>
<ul>
<li>Yoav - Hey folks! Thinking about the deployment of __HttpHost- prefixes (with Shopify's cookie folks) I think we should've spelled it __Host-__Http- (or something similar) to enable deployment in non-supporting browsers and have __Host as a fallback. Otherwise, folks that rely on __Host- today will not be able to deploy __HttpHost- until all browsers support it (or with UA sniffing server-side logic that adds a bunch of complexity)</li>
<li>bvandersloot - Yeah, trying to get the progressive enhancement angle of __Host-__Http- seems nice. Especially in the long tail of users that have out-of-date browsers.</li>
<li>Yoav -Dylan Cutler, thoughts on modifying Chromium's implementation to match that? I'm assuming usage is still tiny/non-existent.. I guess we just need to add a dash and rename from __HostHttp- to __Host-Http-
(to be clear: I can take on the implementation change/intent if you're cool with that)</li>
<li>bvandersloot - valentin from our networking team would know about Gecko's status here that I would</li>
<li>valentin - landed [1974979 - Add `__Http-` and `__HostHttp-` cookie prefixes](https://bugzilla.mozilla.org/show_bug.cgi?id=1974979) recently, but I think we're open to changing</li>
</details>

PR #3156, opened 2025-08-11
Ensure that http and host-http prefix rules are enforced on read

yoavweiss:
To ease deployments of `__Http-` and `__Host-Http-` prefixes, servers need to be sure that changing cookie names today won't enable attacking browsers that don't yet support these semantics once they would start supporting them.

Enforcing the prefix semantics when reading the cookies can help provide these guarantees. (and can avoid complex server-side logic when deploying these prefixes)

### DIFF
--- a/draft-ietf-httpbis-layered-cookies.md
+++ b/draft-ietf-httpbis-layered-cookies.md
@@ -1405,6 +1405,18 @@ boolean _httpOnlyAllowed_, and string _sameSite_:
 
        * cookie's same-site is "`none`".
 
+   * One of the following is true:
+
+       * cookie's name, byte-lowercased, does not start with `__http-`.
+
+       * cookie is Http-prefix compatible.
+
+   * One of the following is true:
+
+       * cookie's name, byte-lowercased, does not start with `__host-http-`.
+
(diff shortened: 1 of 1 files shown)


annevk:
It seems weird to only enforce this for these prefixes. And since this can't be tested I'm a bit worried about it not getting implemented.

johannhof:
I'd be okay with this.

FWIW we do have a privileged way to set cookies in webdriver, though I don't know off hand if it enforces prefix rules for the context it's being called in...

johannhof:
> > FWIW we do have a privileged way to set cookies in webdriver, though I don't know off hand if it enforces prefix rules for the context it's being called in...
> 
> I didn't see anything in https://web-platform-tests.org/writing-tests/testdriver.html#cookies

Hmm WebDriver the protocol has an [Add Cookie](https://w3c.github.io/webdriver/#add-cookie) function but it may not be implemented in testdriver?

yoavweiss:
@annevk - would adding a test driver interface to add cookies (and make this WPT testable) change things from your perspective?

yoavweiss:
Closing as there's no consensus around this.

PR #3157, opened 2025-08-11
QUERY: recommendations for media type related errors

reschke:
I intended this to become a set of examples, but then realized that I was only recommending status codes for the various cases. With that, it feels misplaced in an appendix. Move closer to the QUERY method definition?


reschke:
overtaken by - https://github.com/httpwg/http-extensions/pull/3148

PR #3200, opened 2025-08-22
Tweak defintiion of Upload-Length, and forward ref to its use in responses

### LINKED ISSUE #3170: RESUMABLE: definition of Upload-Length for both requests and responses is unclear
@gstrauss provided the following feedback on https://lists.w3.org/Archives/Public/ietf-http-wg/2025JulSep/0094.html wrt Section 4.1.3

> "The request can include the Upload-Length request and response header field."  What is "and response" header field supposed to mean in the context of the request?
> With HTTP/1.1 Transfer-Encoding: chunked and HTTP/2 and HTTP/3 framing, the lenght might never be specified.  That is okay, though its absense means that it can not be cross-checked with the upload resource size.

At the point the header is defined, it is not clear on how it is used in requests or responses. Text later in the document fills in some gaps but this is not sufficent. The definition at the time of declaration can be can be improved, most likely through editorial changes.

LPardue:
Fixes #3170


Acconut (approved):
That helps readers, thank you!

PR #3201, opened 2025-08-22
Tweak definition of Upload-Limit KVs

### LINKED ISSUE #3173: RESUMABLE: text about defined/undefined Upload-Limit dict entries is unclear
@gstrauss provided the following feedback on https://lists.w3.org/Archives/Public/ietf-http-wg/2025JulSep/0094.html wrt Section 4.1.4

> "Keys with values other than defined here MUST be ignored."  Do you mean only for the keys defined here?  Or is this implying there will never be other keys?  "Unrecognized keys MUST be ignored.  Keys define here but with values other than defined here MUST be ignored."

Changes in this area are mosts likely to be an editorial change to improve the wording while maintaining the intent.

LPardue:
First, use a list rather than bullets. This helps highlight the key name, and its value type.

Second, borrow text from https://www.rfc-editor.org/rfc/rfc9218#section-4-7 about how we
deal with unknown or known but unexpected things. Unlike RFC 9218, there are no additional
constrains on the allowed values carried in Upload-Limit, so I dropped that clause here. If
someone defines a new one later on, they can declare any additional special behaviour that might be
needed IMO

Fixes #3173

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -347,7 +347,7 @@ The following key-value pairs are defined:
 
 Except for the `max-age` limit, the existence of a limit or its value MUST NOT change throughout the lifetime of the upload resource.
 
-Receivers of `Upload-Limit` parse the Dictionary as described in {{Section 4.2 of STRUCTURED-FIELDS}}. Where the Dictionary is successfully parsed, this document places the additional requirement that Dictionary members with unknown keys, or members using known keys with values of unexpected types, MUST be ignored.
+Receivers of `Upload-Limit` parse the Dictionary as described in {{Section 4.2 of STRUCTURED-FIELDS}}. Where the Dictionary is successfully parsed, this document places two additional requirements on Dictionary members. First, a member with an unknown key MUST be ignored. Second, a member with a known key but a value of unexpected type MUST cause the entire `Upload-Limit` header field to be ignored, or alternatively the complete HTTP message MUST be treated as malformed.
 
 A server that supports the creation of a resumable upload resource ({{upload-creation}}) for a target URI MUST include the `Upload-Limit` header field with the corresponding limits in a response to an `OPTIONS` request sent to this target URI. If a server supports the creation of upload resources for any target URI, it SHOULD include the `Upload-Limit` header field with the corresponding limits in a response to an `OPTIONS` request with the `*` target unless the server is not capable of handling `OPTIONS *` requests. If the server does not apply any limits, it MUST use `min-size=0` instead of an empty header value.
 


LPardue:
@Acconut PTAL when you can

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -314,15 +314,24 @@ An upload resource MAY enforce one or multiple limits, which are communicated to
...
+
+`max-age`:
+: Specifies the remaining lifetime of the upload resource in seconds counted from the generation of the response. After the resource's lifetime is reached, the server might make the upload resource inaccessible and a client SHOULD NOT attempt to access the upload resource as these requests will likely fail. The lifetime MAY be extended but SHOULD NOT be reduced unless the server makes the upload resource immediately inaccessible. The value is an Integer.
 
 Except for the `max-age` limit, the existence of a limit or its value MUST NOT change throughout the lifetime of the upload resource.
 
-When parsing the `Upload-Limit` header field, unrecognized keys MUST be ignored and MUST NOT fail the parsing to facilitate the addition of new limits in the future. Keys with values other than defined here MUST be ignored.
+Receivers of `Upload-Limit` parse the Dictionary as described in {{Section 4.2 of STRUCTURED-FIELDS}}. Where the Dictionary is successfully parsed, this document places the additional requirement that Dictionary members with unknown keys, or members using known keys with values of unexpected types, MUST be ignored.
```
Acconut: Previously, receivers could error out if unknown keys held values with unexpected types. This now requires them to ignore these. I'm fine with that, but just something to be aware of.

LPardue: Thanks for the callout. To explain a bit more my though process,

1. If the field is a valid `Dictionary` then it will pass https://www.rfc-editor.org/rfc/rfc9651#section-4.2
2. If the field in an invalid `Dictionary` then the requirements in https://www.rfc-editor.org/rfc/rfc9651#section-4.2-10 apply 

> If parsing fails, either the entire field value MUST be ignored (i.e., treated as if the field were not present in the section), or alternatively the complete HTTP message MUST be treated as malformed. This is intentionally strict to improve interoperability and safety, and field specifications that use Structured Fields are not allowed to loosen this requirement.

3. If the field is a valid `Dictionary`, then apply the additional requirements of this document on the member `value`. If you know the key, you can parse the value and apply this documents rules. If you don't know the key, then you can't know the rules and can't do validation.

When I first authored the commit, I took the opinion that we should ignore the individual members. However, revisiting now I wonder if we should state stricter text more akin to RFC 9651 e.g., 

```suggestion
Receivers of `Upload-Limit` parse the Dictionary as described in {{Section 4.2 of STRUCTURED-FIELDS}}. Where the Dictionary is successfully parsed, this document places the two additional requirements on Dictionary members. First, a member with an unknown key MUST be ignored. Second, a member with a known key but a value of unexpected type MUST cause the entire `Upload-Limit` header field to be ignored, or alternatively the complete HTTP message MUST be treated as malformed.
```

The stricter approach may help to avoid unintended consequences caused by ignoring individual dictionary members.

LPardue: @Acconut I pushed the above text suggestion in a new commit, happy with it?

Acconut: Your new suggestion looks good to me. This also seems to address @MikeBishop's comment on the WGLC that advocates for errors on unexpected value types for known keys:

> In 4.1.4, I'm surprised that keys with unexpected values are ignored rather than an error; does that mean I can include max-age="Really Long, I Swear" and the limit will still work? I commend the future-proofing of ignoring unknown keys, but I'd consider being stricter on keys you know the semantics of.

Acconut (approved):

draft-ietf-httpbis-resumable-upload.md
LPardue: cool, thanks!

PR #3202, opened 2025-08-22
RFC6265bis: Address Secdir nits

sbingler:
Addresses issues brought up during the Secdir review https://lists.w3.org/Archives/Public/ietf-http-wg/2025JanMar/0125.html

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -2062,7 +2062,7 @@ site being browsed. Some block cookies based upon user agent cookie policy and/o
 While this document does not endorse or require a specific approach, it is RECOMMENDED that user
 agents adopt a policy for third-party cookies that is as restrictive as compatibility constraints
 permit. Consequently, resources cannot rely upon third-party cookies being treated consistently by
-user agents for the foreseeable future.
+user agents for the foreseeable future. TODO
 
 
 ## Cookie Policy {#cookie-policy}
@@ -2172,7 +2172,7 @@ and Set-Cookie header fields is transmitted in the clear.
     direction, with unpredictable results.
 
 3.  A malicious client could alter the Cookie header fields before transmission,
-    with unpredictable results.
+    with unpredictable results. TODO
 
 Servers SHOULD encrypt and sign the contents of cookies (using whatever format
 the server desires) when transmitting them to the user agent (even when sending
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -2167,7 +2172,7 @@ and Set-Cookie header fields is transmitted in the clear.
     direction, with unpredictable results.
 
 3.  A malicious client could alter the Cookie header fields before transmission,
-    with unpredictable results.
+    with unpredictable results. TODO
```
sbingler: > 3. Section 8.3 lists security risks when cookies are sent in clear and clause 3
states: "A malicious client could alter the Cookie header before transmission,
with unpredictable results". I failed to understand how this is concerned with
the use of TLS. In my understanding, malicious clients can do this in any case,
regardless on whether TLS is used or not. Am I missing something?

This text has existed since draft-ietf-httpstate-cookie-00 but isn't present in rfc2965, so Adam Barth seems to have added it.

Perhaps Adam was considering a malicious client on the user's system? In that case, would TLS necessarily protect the cookies? If it was a malicious user agent then all bets are off.

I'm currently in agreement with the reviewer but I'm hoping someone else have a better understanding of why this is listed.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -2058,7 +2062,7 @@ site being browsed. Some block cookies based upon user agent cookie policy and/o
 While this document does not endorse or require a specific approach, it is RECOMMENDED that user
 agents adopt a policy for third-party cookies that is as restrictive as compatibility constraints
 permit. Consequently, resources cannot rely upon third-party cookies being treated consistently by
-user agents for the foreseeable future.
+user agents for the foreseeable future. TODO
```
sbingler: > 4. Section 7.1, last para states that "it is RECOMMENDED that user agents adopt
a policy for third-party cookies that is as restrictive as compatibility
constraints permit". This gives no concrete recommendations, and no examples,
thus making this requirement vague and subjective. I think this requirement
should be provided with more details how to deal with it.

My draft response for this is:

>This is difficult to do given the range of compatibility constraints among user agents. Some agents could strive for maximum compatibility while others for maximum user privacy. The third-party cookie policies will differ greatly between those two types of UAs.

Curious if anyone has other thoughts on how to handle this.

draft-ietf-httpbis-rfc6265bis.md
mikewest: I agree that malicious client behavior has little to do with whether the client is sending information in an encrypted or unencrypted fashion. Dropping the third item here seems fine, as the risks listed in the first and second bullets seem exhaustive.

draft-ietf-httpbis-rfc6265bis.md
mikewest: For this document, I agree with your approach. The primitives that user agents seem to be aligning around (CHIPS, etc) didn't make it into this document, so the advice we could provide here would be at best incomplete.

It might be reasonable to file this as feedback to https://www.ietf.org/archive/id/draft-ietf-httpbis-layered-cookies-00.html#name-third-party-cookies, however.

PR #3203, opened 2025-08-22
Update requirements on use of Location during upload creation

### LINKED ISSUE #3183: RESUMABLE: proposal to change the requirements related to 104 Upload Resumption Supported and Location header
@gstrauss provided the following (slightly edited) feedback on https://lists.w3.org/Archives/Public/ietf-http-wg/2025JulSep/0094.html wrt Section 4.2.2

> "These interim responses MUST NOT include the Location header field"
Over-specified.  MUST NOT should be SHOULD NOT, and/or the doc should say that Location should not be included in any 104 response other than in the initial 104 response when the upload resource is created, and that a client SHOULD ignore Location in any 104 response besides the initial 104 response for the upload resource.  Icky.  It would also be okay to say the Location of the upload resource MUST NOT change if Location is sent more than once. A more robust approach would be a request for creation of a temporary resource with a 0-length request body, which is separate from requests sending the request body to the upload resource.

This sounds like it would be design change, albeit a minor one.

LPardue:
Previously, there were "mixed" requirements on including a Location
in interim or final response to an upload creation request. This
created some annoying dealing with 104 responses depending on
if they were first or not.
   
The PR proposes more consistency. If you're going to generate a
Location, it must always be sent and must always be the same value
in all responses. Only a SHOULD requirement is levied on the client
to check these - making it MUST seems a bit too onerous.
    
Fixes #3183

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -365,7 +365,7 @@ The server MUST record the length according to {{upload-length}} if the necessar
 
 While the request content is being received, the server MAY send multiple interim responses with a `104 (Upload Resumption Supported)` status code and the `Upload-Offset` header field set to the current offset to inform the client about the upload progress.
 
-Where a response requires a `Location` header field to be included, all interim and final response messages for the same request MUST contain an identical `Location` value. Clients SHOULD verify the `Location` value does not change between response messages.
+Where a response requires a `Location` header field to be included, all interim and final response messages for the same request MUST contain an identical `Location` value. However, final responses including the `Upload-Complete: ?1` header field are exempt from this requirement because they are the result of processing the transferred representation and the `Location` value does not necessarily represent the upload location. Where the `Location` value is expected to be identical across multiple messages, clients SHOULD verify this. If verification fails, clients SHOULD abort the current request and cancel the upload ({{upload-cancellation}}).
 
 If the server does not receive the entire request content, for example because of canceled requests or dropped connections, it SHOULD append as much of the request content as possible to the upload resource. The upload resource MUST NOT be considered complete then.
 


LPardue:
@Acconut PTAL when you can

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -353,17 +353,19 @@ If the client received a final response with a
...
 
 The server MUST record the length according to {{upload-length}} if the necessary header fields are included in the request.
 
-While the request content is being received, the server MAY send additional interim responses with a `104 (Upload Resumption Supported)` status code and the `Upload-Offset` header field set to the current offset to inform the client about the upload progress. These interim responses MUST NOT include the `Location` header field.
+While the request content is being received, the server MAY send multiple interim responses with a `104 (Upload Resumption Supported)` status code and the `Upload-Offset` header field set to the current offset to inform the client about the upload progress.
+
+Where a response requires a `Location` header field to be included, all interim and final response messages for the same request MUST contain an identical `Location` value. Clients SHOULD verify the `Location` value does not change between response messages.
```
Acconut: Should there be a recommendation for the client to abort the upload if it detects a changed Location value? How should a client react in this situation?

In addition, I'm wondering if this is too strict for the final response:

> Clients SHOULD verify the `Location` value does not change between response messages.

For example, the client uploads a large video in a single request. Since the file is large, the server offers a resumable upload via a 104 response and Location pointing to the upload resource. However, the upload completes fine and the server now wants to indicate the video's URI on the CDN to the client in the Location header. That value would then be different from the Location value in the 104 response. But that's OK because the upload is completed. Should we consider that as well as an exemption?

LPardue: It feels to me we might want a little more input from the WG on this?

Acconut: > It feels to me we might want a little more input from the WG on this?

We can prompt the WG for their opinions, but I don't think this is necessary to address my concern. That can also be done by stating explicitly that Location header field in responses with Upload-Complete: ?1 are exempt from this rule (these responses are the final ones after the upload is done and don't necessarily point to the upload resource).

How about something like this:


```suggestion
Where a response requires a `Location` header field to be included, all interim and final response messages for the same request MUST contain an identical `Location` value. Final responses including the `Upload-Complete: ?1` header field are exempt from this requirement as they are the result of processing the transferred representation. Clients SHOULD verify the `Location` value does not change between response messages. They SHOULD abort the current request and cancel the upload ({{upload-cancellation}}) if a change is detected.
```

LPardue: Thanks, I'll take this but tweak it a bit. Give me a mo

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -353,17 +353,19 @@ If the client received a final response with a
...
+While the request content is being received, the server MAY send multiple interim responses with a `104 (Upload Resumption Supported)` status code and the `Upload-Offset` header field set to the current offset to inform the client about the upload progress.
+
+Where a response requires a `Location` header field to be included, all interim and final response messages for the same request MUST contain an identical `Location` value. However, final responses including the `Upload-Complete: ?1` header field are exempt from this requirement because they are the result of processing the transferred representation and the `Location` value does not necessarily represent the upload location. Where the `Location` value is expected to be identical across multiple messages, clients SHOULD verify this. If verification fails, clients SHOULD abort the current request and cancel the upload ({{upload-cancellation}}).
```
LPardue, line 368: @Acconut PTAL

PR #3204, opened 2025-08-22
Clarify when a server MUST supply an upload-length

### LINKED ISSUE #3187: RESUMABLE: "MUST if" is a poor normative construction
@gstrauss provided the following feedback on https://lists.w3.org/Archives/Public/ietf-http-wg/2025JulSep/0094.html wrt 4.3.2

> "MUST include the length in the Upload-Length header field if known"
Over-specified.  If it is okay to exclude it if not known, then this should be SHOULD, not MUST.

Stating how the client deals with a missing field probably would not go amiss either. Borderline editorial/minor design change.

LPardue:
We have the requirement "if the client knows the representation data's length,
it SHOULD include the `Upload-Length`". Which means the server can be a position
to either know, or not know the value.

This change makes a minor tweak to the text to highlight that when the server
was given a length it really MUST respond with it but when it didn't get given
one it can't.

Fixes #3187

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -476,7 +476,7 @@ A successful response to a `HEAD` request against an upload resource
 
 - MUST include the offset in the `Upload-Offset` header field ({{upload-offset}}),
 - MUST include the completeless state in the `Upload-Complete` header field ({{upload-complete}}),
-- MUST include the length in the `Upload-Length` header field if known ({{upload-length}}),
+- MUST include the length in the `Upload-Length` header field unless the client has not supplied one ({{upload-length}}),
 - MUST indicate the limits in the `Upload-Limit` header field ({{upload-limit}}), and
 - SHOULD include the `Cache-Control` header field with the value `no-store` to prevent HTTP caching ({{CACHING}}).
 


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -476,7 +476,7 @@ A successful response to a `HEAD` request against an upload resource
 
 - MUST include the offset in the `Upload-Offset` header field ({{upload-offset}}),
 - MUST include the completeless state in the `Upload-Complete` header field ({{upload-complete}}),
-- MUST include the length in the `Upload-Length` header field if known ({{upload-length}}),
+- MUST include the length in the `Upload-Length` header field unless the client has not supplied one ({{upload-length}}),
```
guoye-zhang: The client can also specify the complete upload length implicitly through the combination of `Content-Length` and `Upload-Complete: ?1`. Wondering if the phrasing should be "unless the total length is unknown".

LPardue: or `"unless the client has not supplied one (either by supplying its own upload-length or providing an Upload-Complete:?1 with a Content-Length)"`

guoye-zhang: Yeah, that works

guoye-zhang (approved):

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -476,7 +476,7 @@ A successful response to a `HEAD` request against an upload resource
 
 - MUST include the offset in the `Upload-Offset` header field ({{upload-offset}}),
 - MUST include the completeless state in the `Upload-Complete` header field ({{upload-complete}}),
-- MUST include the length in the `Upload-Length` header field if known ({{upload-length}}),
+- MUST include the length in the `Upload-Length` header field unless the client has not supplied one, either by providing its own `Upload-Length` ({{upload-length}}) or providing an Upload-Complete:?1 with a Content-Length ({{upload-complete}}),
```
Acconut: The text could also just reference Section 4.1.3, which explains how the length of an upload can be discovered: 

> There are two different ways for the client to indicate and the upload resource to discover the length from requests for creating the upload resource ([Section 4.2](https://httpwg.org/http-extensions/draft-ietf-httpbis-resumable-upload.html#upload-creation)) or appending to it ([Section 4.4](https://httpwg.org/http-extensions/draft-ietf-httpbis-resumable-upload.html#upload-appending)):
> - If the request includes the Upload-Complete field value set to true and a valid Content-Length header field, the request content is the remaining representation data. The length is then the sum of the current offset ([Section 4.1.1](https://httpwg.org/http-extensions/draft-ietf-httpbis-resumable-upload.html#upload-offset)) and the Content-Length header field value.
> - The request can include the Upload-Length request and response header field. Upload-Length is an Item Structured Header Field ([[STRUCTURED-FIELDS](https://httpwg.org/http-extensions/draft-ietf-httpbis-resumable-upload.html#STRUCTURED-FIELDS)]). Its value is a non-negative Integer ([Section 3.3.1](https://rfc-editor.org/rfc/rfc9651#section-3.3.1) of [[STRUCTURED-FIELDS](https://httpwg.org/http-extensions/draft-ietf-httpbis-resumable-upload.html#STRUCTURED-FIELDS)]) and indicates the number of bytes of the entire representation data. Other values MUST cause the entire header field to be ignored.

(from https://httpwg.org/http-extensions/draft-ietf-httpbis-resumable-upload.html#section-4.1.3-2)

Maybe something along these lines:


```suggestion
- MUST include the length in the `Upload-Length` header field unless the client has not supplied one, by providing the corresponding headers as described in Section 4.1.3,
```

That might be a bit more watertight.

LPardue: I'm happy to avoid restating requirements in multiple places in the document but I'd probably want to land https://github.com/httpwg/http-extensions/pull/3200 first

PR #3205, opened 2025-08-22
RFC6265bis: Address Deb Cooley's feedback

sbingler:
Addresses Deb Cooley's feedback https://lists.w3.org/Archives/Public/ietf-http-wg/2025JanMar/0129.html

### DIFF
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -1609,27 +1609,27 @@ user agent MUST process the cookie as follows:
     creation-time and the last-access-time to the current date and time.
 
 6.  If the cookie-attribute-list contains an attribute with an attribute-name
-    of "Max-Age" or an attribute with an attribute-name of "Expires":
+    of "Max-Age":
 
     1.  Set the cookie's persistent-flag to true.
 
-    2.  If the cookie-attribute-list contains an attribute with an attribute-name
-    of "Max-Age":
+    2.  Set the cookie's expiry-time to attribute-value of the last
+        attribute in the cookie-attribute-list with an attribute-name of
+        "Max-Age".
 
-        1. Set the cookie's expiry-time to attribute-value of the last
-           attribute in the cookie-attribute-list with an attribute-name of
-           "Max-Age".
+    Otherwise, if the cookie-attribute-list contains an attribute with an
+    attribute-name of "Expires" (and does not contain an attribute with an
+    attribute-name of "Max-Age"):
 
-        Otherwise:
+    1.  Set the cookie's persistent-flag to true.
 
-        1. Set the cookie's expiry-time to attribute-value of the last
-           attribute in the cookie-attribute-list with an attribute-name of
-           "Expires".
+    2.  Set the cookie's expiry-time to attribute-value of the last
+        attribute in the cookie-attribute-list with an attribute-name of
+        "Expires".
 
     Otherwise:
 
     1.  Set the cookie's persistent-flag to false.
-
     2.  Set the cookie's expiry-time to the latest representable date.
 
 7.  If the cookie-attribute-list contains an attribute with an
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -2114,8 +2114,6 @@ set to expire in two weeks.
 
 ## Overview
 
-TODO
-
 Cookies have a number of security pitfalls. This section overviews a few of the
 more salient issues.
 
--- a/draft-ietf-httpbis-rfc6265bis.md
+++ b/draft-ietf-httpbis-rfc6265bis.md
@@ -70,7 +70,7 @@ normative:
       org: WHATWG
   SAMESITE:
     target: https://html.spec.whatwg.org/#same-site
-    title: HTML - Living Standard
+    title: HTML - Living Standard TODO
     date: 2021-01-26
     author:
     -
@@ -1609,23 +1609,22 @@ user agent MUST process the cookie as follows:
     creation-time and the last-access-time to the current date and time.
 
 6.  If the cookie-attribute-list contains an attribute with an attribute-name
+    of "Max-Age" or an attribute with an attribute-name of "Expires":
+
+    1.  Set the cookie's persistent-flag to true.
+
+    2.  If the cookie-attribute-list contains an attribute with an attribute-name
     of "Max-Age":
 
-    1.  Set the cookie's persistent-flag to true.
+        1. Set the cookie's expiry-time to attribute-value of the last
+           attribute in the cookie-attribute-list with an attribute-name of
+           "Max-Age".
 
-    2.  Set the cookie's expiry-time to attribute-value of the last
-        attribute in the cookie-attribute-list with an attribute-name of
-        "Max-Age".
+        Otherwise:
 
-    Otherwise, if the cookie-attribute-list contains an attribute with an
-    attribute-name of "Expires" (and does not contain an attribute with an
-    attribute-name of "Max-Age"):
-
-    1.  Set the cookie's persistent-flag to true.
-
-    2.  Set the cookie's expiry-time to attribute-value of the last
-        attribute in the cookie-attribute-list with an attribute-name of
-        "Expires".
+        1. Set the cookie's expiry-time to attribute-value of the last
+           attribute in the cookie-attribute-list with an attribute-name of
+           "Expires".
 
     Otherwise:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -70,7 +70,7 @@ normative:
       org: WHATWG
   SAMESITE:
     target: https://html.spec.whatwg.org/#same-site
-    title: HTML - Living Standard
+    title: HTML - Living Standard TODO
     date: 2021-01-26
     author:
```
sbingler: > Section 10.1:  Most (all?) of the WHATWG documents can be referenced as a
snapshot to make them immutable.  There might be other ways to do this, but
this is the one I've seen used.

I recall there being pushback against linking to snapshot some time ago, but I can't seem to find it.
I'm inclined to leave this as-is but open to other opinions.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -2115,6 +2114,8 @@ set to expire in two weeks.
 
 ## Overview
 
+TODO
+
 Cookies have a number of security pitfalls. This section overviews a few of the
 more salient issues.
```
sbingler: > Section 8:  I agree with Valery that this section picks and chooses some
example issues ('more salient issues').  I wonder if it isn't possible to give
a 1-2 sentence overview of the general security issues associated with cookies.
 Something to set the stage, where what follows are examples of issues that
have been seen over time (with or without mitigations).  Sadly, I do not have
proposed text, and indeed, it may not be possible/feasible

Perhaps I'm missing something but I think the overview section gives a decent overview of the major security issues. I, at least, can't think of a way to sum up all of them in just a few sentences.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1609,23 +1609,22 @@ user agent MUST process the cookie as follows:
...
 
-    1.  Set the cookie's persistent-flag to true.
+        Otherwise:
 
-    2.  Set the cookie's expiry-time to attribute-value of the last
-        attribute in the cookie-attribute-list with an attribute-name of
-        "Expires".
+        1. Set the cookie's expiry-time to attribute-value of the last
+           attribute in the cookie-attribute-list with an attribute-name of
+           "Expires".
```
sbingler: > Section 5:  (Recognizing that this is from the original RFC) The nested
numbered lists are difficult to parse.  For example Section 5.7, #6 has 3 sets
of sub numbered lists that appear to be distinct.  If these sub numbered lists
are necessary (and when there is merely a #1 without a #2, one might argue it
isn't 'necessary') then perhaps characters other than numbers might be clearer.

I changed the conditionals around in a way that, imo, is easier to read. Unfortunately it still has 3 levels of nesting. I'm fine with keeping, reverting, or doing something else to this, open to feedback.

draft-ietf-httpbis-rfc6265bis.md
mikewest: @mnot probably has a more canonically IETFish opinion here. WDYT?

draft-ietf-httpbis-rfc6265bis.md
mikewest: I agree that the overview does a reasonable job of this. I don't think adding a sentence or two to those two paragraphs would usefully increase reader's understanding.

If there was a concrete suggestion to evaluate, I'd happily do so, but as-is, I'd suggest leaving it as written.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1609,23 +1609,22 @@ user agent MUST process the cookie as follows:
...
+           attribute in the cookie-attribute-list with an attribute-name of
+           "Max-Age".
 
-    1.  Set the cookie's persistent-flag to true.
+        Otherwise:
 
-    2.  Set the cookie's expiry-time to attribute-value of the last
-        attribute in the cookie-attribute-list with an attribute-name of
-        "Expires".
+        1. Set the cookie's expiry-time to attribute-value of the last
```
mikewest: I think this item could drop the nesting without much impact to readability. That is:

> 2.  If the cookie-attribute-list contains an attribute with an attribute-name
    of "Max-Age", set the cookie's expiry-time to attribute-value of the last attribute in the cookie-attribute-list with an attribute-name of "Max-Age".
>
>     Otherwise, set the cookie's expiry-time to the attribute-value of the last attribute in the cookie-attribute-list with an attribute-name of "Expires".
>

mikewest: (The downside of this is that it's inconsistent with the rest of this section (and probably the rest of the spec), which separates out single step "then" clauses. So, I guess we should make a stylistic choice about which to prefer and run with it. I'm fine with either direction, actually: as below, nesting doesn't bother me at all.)

draft-ietf-httpbis-rfc6265bis.md
mikewest: I actually found the original approach easier to parse (or, at least easier to recognize the priority placed on `max-age` over `expires`). I'm also perfectly happy with long numbered lists, possibly because I'm more used to algorithmic specs from places like WHATWG?

For my personal aesthetics:

1.  The lists in 5.7.6 seem necessary, as they're describing the steps you take for each option in a disjunction. They can be combined and moved around, as in this PR, but I'm not sure they can fundamentally be removed.
2.  Switching between numbers and letters would make the algorithm more difficult for me to understand than sticking with a numbered list of steps throughout.
3.  I don't mind nesting, but, as above, we could drop the nesting for single-step lists pretty easily without compromising readability (IMO).

In any event: I think any change here must be non-normative, and I can live with any of the editorial directions proposed. I would be least happy switching between letters and numbers, but it's not the end of the world.

sbingler: > Switching between numbers and letters would make the algorithm more difficult for me to understand than sticking with a numbered list of steps throughout.

Agreed

>  I think any change here must be non-normative

Agreed

> I don't mind nesting, but, **as above**, we could drop the nesting for single-step lists pretty easily

Maybe I'm missing it, but what are you referring to with "as above"?

draft-ietf-httpbis-rfc6265bis.md
sbingler: SGTM

draft-ietf-httpbis-rfc6265bis.md
mikewest: > "as above"

I was referring to the comment on line 1625 (https://github.com/httpwg/http-extensions/pull/3205/files#r2295487115) about removing nesting.

draft-ietf-httpbis-rfc6265bis.md
sbingler: > The downside of this is that it's inconsistent with the rest of this section (and probably the rest of the spec)

My thoughts as well, I dislike having just the one part inlined. Given that I think I'll revert this back to the way it was.

draft-ietf-httpbis-rfc6265bis.md
mikewest: cc @tfpauly, who might also wish to weigh in from HTTPWG's perspective.

mnot: Practice is to link to the living document and make sure that the anchor (if used) is [permanent](https://whatwg.org/working-mode#anchors).

See eg
* HTMLURL in rfc9421
* FETCH in rfc9163, rfc9205

sbingler: Thanks for the recommendation. 

> See eg
>  * HTMLURL in rfc9421

Just to confirm, you're referring to the usage within https://datatracker.ietf.org/doc/html/rfc9421#section-2.2.8
>... in the format defined in Section [5](https://url.spec.whatwg.org/#application/x-www-form-urlencoded) (["application/x-www-form-urlencoded"](https://url.spec.whatwg.org/#application/x-www-form-urlencoded)) of [[HTMLURL](https://url.spec.whatwg.org/)]...

?

sbingler: With https://github.com/whatwg/html/issues/11646 and https://github.com/whatwg/html/issues/6329 both the anchor are permanent.

I tried to match the format that RFC9421 uses but ran into some difficultly since dom-document-cookie isn't actually its own section.  I feel keeping the separate [SAMESITE] and [DOM_DOCUMENT_COOKIE] references is fine, but do you have any thoughts @mikewest ?

sbingler: Ping @mikewest

mikewest: I think leaving those links untouched is appropriate, given that both are permanent links. Closing this out seems fine to me.

sbingler:
Closing, there are no changes to the file but whatwg references are now permanent.

PR #3208, opened 2025-08-23
Increase upload and append sizes to the 100 MB range

### LINKED ISSUE #3207: RESUMABLE: examples use small byte ranges that seem a bit contrived
During WGLC review, @MikeBishop raise the folloing issue on https://lists.w3.org/Archives/Public/ietf-http-wg/2025JulSep/0123.html

> Given that this mechanism really only makes sense for large uploads, the examples of 50-200 byte resources seem a bit *too* contrived. No one would use this for such tiny transfers, would they? Since you're not actually including the payload in this document, why not just use more realistic numbers thoughout?

LPardue:
Fixes #3207

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -369,16 +369,16 @@ If the server does not receive the entire request content, for example because o
 
 ### Examples {#upload-creation-example}
 
-A) The following example shows an upload creation, where the entire 100 bytes are transferred in the initial request. The server sends multiple interim responses and one final response from processing the uploaded representation.
+A) The following example shows an upload creation, where the entire 100000000 bytes are transferred in the initial request. The server sends multiple interim responses and one final response from processing the uploaded representation.
 
 ~~~ http-message
 POST /project/123/files HTTP/1.1
 Host: example.com
 Upload-Complete: ?1
-Content-Length: 100
-Upload-Length: 100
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -369,16 +369,16 @@ If the server does not receive the entire request content, for example because o
 
 ### Examples {#upload-creation-example}
 
-A) The following example shows an upload creation, where the entire 100 bytes are transferred in the initial request. The server sends multiple interim responses and one final response from processing the uploaded representation.
+A) The following example shows an upload creation, where the entire 100000000 bytes are transferred in the initial request. The server sends multiple interim responses and one final response from processing the uploaded representation.
```
LPardue, line 372: Note that here and other prose sections, we could attempt to be kinder and use megabytes rather than bytes. However, on balance I felt it easier and clearer to stick to bytes ; it also also avoids mega vs mibi byte nonsense

wtarreau: A trick when dealing with large numbers in examples that you fear will be hard to read or compare such as 100000000 is to replace most/all zeroes with recognizable numbers, such as 123456789. This way you can easily show that after sending 56789 bytes, 123400000 remain to be sent.

LPardue: Nice suggestion!

PR #3232, opened 2025-08-25
Add some light example details about upload cancellation and dropped connections

### LINKED ISSUE #3216: RESUMABLE: what is meant by "canceled request"

> A pointer to what is meant by "canceled request" might be useful. An HTTP-naive reader might interpret this as the client deciding it doesn't wish to do the transfer any more, in which case why would it be resumed? Or just skip cancellation here -- the broken connection example suffices. You could even mention why connections get broken (e.g. client mobility, NAT reboot, etc.).

LPardue:
Fixes #3216

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -91,7 +91,7 @@ explicit server cancellation:
 implicit server cancellation:
 : e.g., DoS mitigation or internal error
 
-Connection can be dropped due to a variety of network or transport layer reasons triggered by endpoints or on-path elements.
+Connections can be dropped due to a variety of network or transport layer reasons triggered by endpoints or on-path elements.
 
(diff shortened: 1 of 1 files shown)


guoye-zhang (approved):

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -69,13 +69,29 @@ informative:
...
+implicit client cancellation:
+: e.g., terminating a tab, garbage collecting a process or internal error
+
+explicit server cancellation:
+: e.g., scheduled maintenance
+
+implicit server cancellation:
+: e.g., DoS mitigation or internal error
+
+Connection can be dropped due to a variety of network or transport layer reasons triggered by endpoints or on-path elements.
```
```suggestion
Connections can be dropped due to a variety of network or transport layer reasons triggered by endpoints or on-path elements.
```

Issue #3231, opened 2025-08-25
Large numbers in examples


MikeBishop:
A trick when dealing with large numbers in examples that you fear will be hard to read or compare such as 100000000 is to replace most/all zeroes with recognizable numbers, such as 123456789. This way you can easily show that after sending 56789 bytes, 123400000 remain to be sent.

_Originally posted by @wtarreau in https://github.com/httpwg/http-extensions/pull/3208#discussion_r2296494416_

Preserving this suggestion so it doesn't get lost; the PR might have been merged before it came in?

PR #3234, opened 2025-09-01
Apply Willy's suggestion to large number examples

LPardue:
Taking the suggestion from https://github.com/httpwg/http-extensions/issues/3231

Closes https://github.com/httpwg/http-extensions/issues/3231

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -628,17 +628,17 @@ HTTP/1.1 204 No Content
 Upload-Complete: ?0
 ~~~
 
-B) The next example shows an upload append, where the client transfers the remaining 23456789 bytes and completes the upload. The server processes the uploaded representation and generates the responding response, in this example containing extracted meta data:
+B) The next example shows an upload append, where the client transfers the remaining 4567890 bytes and completes the upload with a length of 1234567890 bytes. The server processes the uploaded representation and generates the responding response, in this example containing extracted meta data:
 
 ~~~ http-message
 PATCH /upload/d38d6ffe8 HTTP/1.1
 Host: example.com
 Upload-Complete: ?1
-Upload-Offset: 46913578
-Content-Length: 23456789
+Upload-Offset: 1230000000
+Content-Length: 4567890
 Content-Type: application/partial-upload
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -577,38 +577,38 @@ While the request content is being received, the server SHOULD send interim resp
...
-B) The next example shows an upload append, where the client transfers the remaining 25000000 bytes and completes the upload. The server processes the uploaded representation and generates the responding response, in this example containing extracted meta data:
+B) The next example shows an upload append, where the client transfers the remaining 23456789 bytes and completes the upload. The server processes the uploaded representation and generates the responding response, in this example containing extracted meta data:
 
 ~~~ http-message
 PATCH /upload/b530ce8ff HTTP/1.1
 Host: example.com
 Upload-Complete: ?1
-Upload-Offset: 25000000
-Content-Length: 25000000
+Upload-Offset: 46913578
```
guoye-zhang: Should we make sure that the final completed length is 1234567890? So each chunk can be 1200000000, 34000000, 560000, 7800, 90?

LPardue: My understanding of the examples was that they are fairly isolated from each other. Even prior to the change to use larger numbers, in https://www.ietf.org/archive/id/draft-ietf-httpbis-resumable-upload-09.html#section-4.4.3 the appends result in an object that has a total length of 300 bytes, which doesn't correlate to any upload length in other examples.

Furthermore, I'd prefer to keep examples concise. Adding multiple append operations that don't do much beyond using a different size increases the length of example and might add more source of confusion. I created https://github.com/httpwg/http-extensions/issues/3250 that highlights this via reuse of the same URL

guoye-zhang: Oh I wasn't suggesting adding more appends to the example, I was just thinking 46913578 kind of loses the benefit of the easily recognizable number. If offset is 1230000000 and length is 4567890, it's obvious that they add up together as 1234567890

LPardue: Ah I undestand now, I can do that

PR #3252, opened 2025-09-19
Add some possible response types for rejected completed uploads

### LINKED ISSUE #3225: RESUMABLE: In 4.4.2, with what status code does the server reject appending to an already-completed upload?

> In 4.4.2, with what status code does the server reject appending to an already-completed upload? 409 might be reasonable here as well, since it's a conflict with the state of the upload; 410 (Gone) might also be reasonable since the upload facility is no longer available at that URI.

LPardue:
Closes #3225


wtarreau:
LGTM!

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -563,7 +563,12 @@ If the upload resource does not receive the entire patch document, for example b
...
+
+- If the client attempted to append a non-zero length document, the server MUST treat this as an inconsistent length failure. The server can use the problem type {{PROBLEM}} of "https://iana.org/assignments/http-problem-types#inconsistent-upload-length" ({{inconsistent-length}}) in the response.
+- If the client attempted to append a zero-length document:
+  - If the server retains knowledge of the upload resource but it is not available, the `410 (Gone)` status code can be used; the response could also contain a `Location` header field to indicate the location of the completed upload. The server can use the problem type {{PROBLEM}} of "https://iana.org/assignments/http-problem-types#completed-upload" ({{completed-upload}}) in the response.
+  - If the server has not retained knowledge of the upload resource, a `404 (Not Found)` status code can be used. The server can use the problem type {{PROBLEM}} of "https://iana.org/assignments/http-problem-types#completed-upload" ({{completed-upload}}) in the response.
```
smatsson: This reads odd to me. If the server has no knowledge of the upload resource a 404 seems fine- But if it does not have knowledge of the upload, how does it know to use the completed-upload problem detail?

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -563,7 +563,12 @@ If the upload resource does not receive the entire patch document, for example b
...
+If the upload is already complete ({{upload-complete}}), the server MUST NOT modify the upload resource and MUST reject the request. The choice of response depends on the nature of the upload request and server state, including but not limited to:
+
+- If the client attempted to append a non-zero length document, the server MUST treat this as an inconsistent length failure. The server can use the problem type {{PROBLEM}} of "https://iana.org/assignments/http-problem-types#inconsistent-upload-length" ({{inconsistent-length}}) in the response.
+- If the client attempted to append a zero-length document:
+  - If the server retains knowledge of the upload resource but it is not available, the `410 (Gone)` status code can be used; the response could also contain a `Location` header field to indicate the location of the completed upload. The server can use the problem type {{PROBLEM}} of "https://iana.org/assignments/http-problem-types#completed-upload" ({{completed-upload}}) in the response.
```
guoye-zhang, line 570: I think this is a weird state that no implementation should get into. Maybe `completed-upload` shouldn't even have been an error in the first place.

draft-ietf-httpbis-resumable-upload.md
LPardue: Oops copy-paste mistake, youre right

draft-ietf-httpbis-resumable-upload.md
LPardue: It does feel a bit like an "error! success" style result. 410 Gone plus Location seems like it does the same thing without needing the custom problem type

guoye-zhang: I was mainly thinking that 410 and 404 make almost no difference to the client, and it doesn't make much sense for a server to forget the final response but remember an upload happened to a particular temporary resource.

If we decide to use a 0-length append to retrieve the final response, every other error case is covered by `mismatching-offset` or `inconsistent-length` so we don't need `completed-upload` problem type.

Acconut (approved):

draft-ietf-httpbis-resumable-upload.md
LPardue: I'd like to land this change and open a new issue to discuss whether we keep completed-upload error type

LPardue: See https://github.com/httpwg/http-extensions/issues/3306

PR #3254, opened 2025-09-19
Forward ref to security considerations about upload resource URI

### LINKED ISSUE #3217: RESUMABLE: section 4 should add forward ref to section 13

> Section 4:

>  * I see that in Section 13, you talk about authentication/authorization of the upload resource. The mechanisms the server uses to perform this are probably out of scope, but the server's obligation needs to be mentioned. You also talk about making the URI of the upload resource difficult to guess, such that accessing it in the first place is some indication of authenticity. It's worth a forward reference to these requirements from Section 4, since they discuss the resource and how to implement it.

LPardue:
Closes #3217


LPardue:
Ah yes, sorry. Closing this then.

Issue #3137, opened 2025-07-22
QUERY: define what the selected representation is for QUERY


reschke:
See https://www.rfc-editor.org/rfc/rfc9110.html#representations

> A [target resource](https://www.rfc-editor.org/rfc/rfc9110.html#target.resource) might be provided with, or be capable of generating, multiple representations that are each intended to reflect the resource's current state. An algorithm, usually based on [content negotiation](https://www.rfc-editor.org/rfc/rfc9110.html#content.negotiation) ([Section 12](https://www.rfc-editor.org/rfc/rfc9110.html#content.negotiation)), would be used to select one of those representations as being most applicable to a given request. This "selected representation" provides the data and metadata for evaluating conditional requests ([Section 13](https://www.rfc-editor.org/rfc/rfc9110.html#conditional.requests)) and constructing the content for [200 (OK)](https://www.rfc-editor.org/rfc/rfc9110.html#status.200), [206 (Partial Content)](https://www.rfc-editor.org/rfc/rfc9110.html#status.206), and [304 (Not Modified)](https://www.rfc-editor.org/rfc/rfc9110.html#status.304) responses to GET ([Section 9.3.1](https://www.rfc-editor.org/rfc/rfc9110.html#GET)).

It seems that this affects other issues such as #3117.

Thoughts:

- should be the same for all "new" methods, in particular, but not limited to all safe and idempotent methods including request content.
- what precisely is it for POST, and can the definition for QUERY be derived from that?

MikeBishop:
The definition of Content-Location provides additional context on the selected representation for POST:
> - For a response to a GET or HEAD request, this is an indication that the target URI refers to a resource that is subject to content negotiation and the Content-Location field value is a more specific identifier for the [selected representation](https://www.rfc-editor.org/rfc/rfc9110.html#selected.representation).
> - Otherwise, such a Content-Location indicates that this content is a representation reporting on the requested action's status and that the same report is available (for future access with GET) at the given URI. For example, a purchase transaction made via a POST request might include a receipt document as the content of the [200 (OK)](https://www.rfc-editor.org/rfc/rfc9110.html#status.200) response; the Content-Location field value provides an identifier for retrieving a copy of that same receipt in the future.

reschke:
@martinthomson - if we made that "the same as GET except the server needs to take the content into account", we'd still have the problem that GET and QUERY would share the same set of selected representations.

*Could* we state that QUERY-without-content actually is the *same thing* as a GET request?

[...]

PR #3255, opened 2025-09-19
QUERY: selected representation (addresses  #3137)

reschke:
First step: make concept of "GET-table" resource independant of whether it has a URI. Will require word-smithing. Actual uses of this terminology comes later.

EDIT 2025-09-21: use term "equivalent resource" with shorter introduction; add example for use of If-Not-Modified. Q: Etags? More conneg information? Sequence diagram?

cc @mnot @martinthomson @MikeBishop @royfielding

### DIFF
--- a/draft-ietf-httpbis-safe-method-w-body.xml
+++ b/draft-ietf-httpbis-safe-method-w-body.xml
@@ -1479,11 +1479,11 @@ Last-Modified: Thu, 20 Feb 2025 06:10:01 GMT
   Ashok Malhotra participated in early discussions leading to this
   specification:
 </t>
-<author initials="A." surname="Malhotra" fullname="Ashok Malhotra">
+<contact initials="A." surname="Malhotra" fullname="Ashok Malhotra">
   <address>
     <email>malhotrasahib@gmail.com</email>
   </address>
-</author>
+</contact>
 </section>
 </back>
 </rfc>
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -93,7 +93,7 @@ implicit server cancellation:
 
 Connections can be dropped due to a variety of network or transport layer reasons triggered by endpoints or on-path elements.
 
-This specification defines a new mechanism for resumable uploads from client to server in a way that is backwards-compatible with conventional HTTP uploads. When an upload is interrupted, clients can send subsequent requests to query the server state and use this information to send the remaining representation data. Alternatively, they can cancel the upload entirely. Unlike ranged downloads, this protocol does not support transferring an upload as multiple requests in parallel.
+This specification defines a new mechanism for resumable uploads from client to server that can seamlessly fall back to conventional uploads. When an upload is interrupted, clients can send subsequent requests to query the server state and use this information to send the remaining representation data. Alternatively, they can cancel the upload entirely. Unlike ranged downloads, this protocol does not support transferring an upload as multiple requests in parallel.
 
 Utilizing resumable uploads, applications can recover from unintended interruptions, but also interrupt an upload on purpose to later resume it, for example, when a user wants to pause an upload, the device's network connectivity changes, or bandwidth should be saved for higher priority tasks.
 
@@ -285,7 +285,7 @@ An upload resource is specific to the upload of one representation. For uploadin
 
 The server can clean up an upload resource and make it inaccessible immediately after the upload is complete. However, if a client didn't receive the last response acknowledging the upload's completion and the upload resource is not available anymore, the client cannot verify the upload's state with the server. Therefore, the server SHOULD keep the upload resource available for a reasonable amount of time after the upload is complete.
 
-An upload resource SHOULD NOT reuse the URI from a previous upload resource, unless reasonable time has passed to ensure that no client will attempt to access the previous upload resource. Otherwise, a client might access the upload resource corresponding to a different representation than it intends to transfer.
+An upload resource SHOULD be unique. Reuse of a URI for a different upload resource SHOULD be avoided in order to reduce the chance of misdirected or corrupted upload resources, as well as the potential security issues described in {{security-considerations}}.
 
 ## State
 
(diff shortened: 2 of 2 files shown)


royfielding (commented):
This can be simplified quite a bit by just calling it "an equivalent resource". Being a resource with safe semantics implies that it responds to GET, and being semantically equivalent to the QUERY request (including all semantically significant bits) is sufficient to define the resource as GET-able.

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -342,8 +361,8 @@ q=foo&amp;limit=10&amp;sort=-published
 
     <section title="Location Response Field" anchor="location">
       <t>
-        A server can create or locate a resource that identifies the query
-        operation for future use. If the server does so, the URI of the resource
+        A server can assign a URI to the GET-equivalent resource (<xref target="get-equivalent"/>)
+        of a QUERY request. If the server does so, the URI of that resource
```
```suggestion
        A server can define a resource that is semantically equivalent to a given QUERY request. The URI of an equivalent resource
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -325,6 +325,25 @@ q=foo&amp;limit=10&amp;sort=-published
...
+        In other words, the GET-equivalent resource is derived from the resource
+        implementing QUERY by incorporating the request content.
+      </t>
+      <t>
+        The definition of the GET-equivalent resource is used as a means to define
+        behavior for other HTTP aspects, such as selected representations. Servers can
+        assign URIs to these resources, making them accessible for GET requests.
+      </t>
+    </section>
+
```
```suggestion
```

draft-ietf-httpbis-safe-method-w-body.xml
royfielding: IOW, this is all that is needed because GET is implied and the description of how to use such a resource follows.

reschke:
Thanks for the feedback, Roy.

I still think a separate section (albeit smaller) would be good, as we want to re-use that definition/explanation in more places, and you explanation may not be sufficient for all readers.

Will work on that soonish.

royfielding (approved):

reschke:
@martinthomson - SVG in https://github.com/httpwg/http-extensions/pull/3255/commits/766cdf8482ee73eae996e4246683570dc47f2665

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -908,13 +948,292 @@ See stored query at "/contacts/stored-queries/42".
...
+|                                  +--------------+
+|                                         |
+| QUERY with content                      |
+| (conditional on 'foo')                  |
++---------------------------------------->|
+|                                         |
+|                                         |
+|                                  200 OK |
+|                          Validator: bar |
+|<----------------------------------------+
```
martinthomson, line 1090:
```suggestion
|<----------------------------------------+
|                                         |
```

The tails on the lines are necessary to avoid making the diagram look a bit strange.

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -908,13 +948,292 @@ See stored query at "/contacts/stored-queries/42".
...
+      <t>
+        The diagrams below illustrate the use of conditional requests and how they can differ
+        when a URI is assigned to the equivalent resource (and when the client is taking advantage
+        of it). The fictitious field name "Validator" is used for demonstration purposes.
+      </t>
+      <figure title="Data Flow with QUERY only">
+        <artset>
+          <artwork type="ascii-art" name="flow-query" align="center"><![CDATA[
+Client                             Resource
+|                                         |
```
martinthomson, line 1064: You could make this slightly more compact:

```suggestion
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -908,13 +948,292 @@ See stored query at "/contacts/stored-queries/42".
...
+|                                         |
+| QUERY with content                      |
++---------------------------------------->|
+|                                         |
```
martinthomson, line 1067:
```suggestion
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -908,13 +948,292 @@ See stored query at "/contacts/stored-queries/42".
...
+|                                         |
+|                                  200 OK |
+|                          Validator: foo |
+|<----------------------------------------+
+|                                         |
```
martinthomson, line 1071:
```suggestion
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -908,13 +948,292 @@ See stored query at "/contacts/stored-queries/42".
...
+|                                         |
+| QUERY with content                      |
+| (conditional on 'foo')                  |
++---------------------------------------->|
+|                                         |
```
martinthomson, line 1075:
```suggestion
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -908,13 +948,292 @@ See stored query at "/contacts/stored-queries/42".
...
+|                                         |
+|                                  +--------------+
+|                                  | State Change |
+|                                  +--------------+
+|                                         |
+| QUERY with content                      |
+| (conditional on 'foo')                  |
++---------------------------------------->|
+|                                         |
```
martinthomson, line 1087:
```suggestion
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -908,13 +948,292 @@ See stored query at "/contacts/stored-queries/42".
...
+|                                                                +--------------+
+|                                                                       |
+| GET                                                                   |
+| (conditional on 'foo')                                                |
++---------------------------------------------------------------------->|
+|                                                                       |
+|                                                                       |
+|                                                                200 OK |
+|                                                        Validator: bar |
+|<----------------------------------------------------------------------+]]></artwork>
```
martinthomson: Similar comments to the above diagram, I trust you to use good judgment.

martinthomson (approved):
The SVG looks good (I continue to be amazed by aasvg, because even though I maintain the wrapper, I never wrote the original).  Some nits only (this will probably mean regenerating the SVG).

martinthomson:
Just looking again at [the second diagram](https://httpwg.org/http-extensions/reschke-3137/draft-ietf-httpbis-safe-method-w-body.html#appendix-A.5-17.1.2) I see that it's a bit smaller, probably because it's a wider source image.  You might be able to trim a few columns out to avoid that.

Also, the "Equivalent Resource" label in that diagram could be more directly placed over the corresponding line.  Maybe by moving down the line above "/xyz"

reschke:
> Just looking again at [the second diagram](https://httpwg.org/http-extensions/reschke-3137/draft-ietf-httpbis-safe-method-w-body.html#appendix-A.5-17.1.2) I see that it's a bit smaller, probably because it's a wider source image. You might be able to trim a few columns out to avoid that.

Actually it isn't with xml2rfc's default CSS, not with my XSLT's.

grafik

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -908,13 +948,280 @@ See stored query at "/contacts/stored-queries/42".
...
+|                                         |
+| QUERY with content                      |           Equivalent Resource
++---------------------------------------->|              (generates /xyz)
+|                                         +-----------------------------o
+|                                         |                             |
+|                                  200 OK |                             |
+|                          Validator: foo |                             |
+|                          Location: /xyz |                             |
+|<----------------------------------------+                             |
+|                                                                       |
```
martinthomson:
```suggestion
|<----------------------------------------+                             |
|                                         |                             |
```

draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -908,13 +948,280 @@ See stored query at "/contacts/stored-queries/42".
...
+| GET                                                                   |
+| (conditional on 'foo')                                                |
++---------------------------------------------------------------------->|
+|                                                                       |
+|                                                      304 Not Modified |
+|                                                        Validator: foo |
+|<----------------------------------------------------------------------+
+|                                                                       |
+|                                                                +--------------+
```
martinthomson: I make this line out at 81 columns, which is why things end up looking a little squished.

martinthomson (commented):
> Actually it isn't with xml2rfc's default CSS, not with my XSLT's.

Ah, I see what is going on.  My CSS is much more carefully tuned to a "reasonable" page width (I know that we might differ in terms of what that means...).  So when you have an 81 column figure like this, it gets squashed a little.  Not a lot, but enough.

You might need to adjust that anyway, just to ensure that you fit the text output size constraints.  I generally recommend 69 columns for figures, just to maintain the left margin.  That might be a little tricky in this case, but it might still be feasible.

Issue #3120, opened 2025-06-29
QUERY: HTTPDIR/RF review - Caching


reschke:
**_QUOTED FROM SPEC TEXT, COMMENTS ARE ROY'S (@royfielding)_**

> 2.4.  Caching
>
>    The response to a QUERY method is cacheable; a cache MAY use it to
>    satisfy subsequent QUERY requests as per Section 4 of
>    [HTTP-CACHING]).

No, just no. A cache does not have access to the request content when
making a hit/miss decision. Use the 303 response, as designed.

The reason why this is not allowed in HTTP is because routing decisions
are based on the connection context, host, and entire target URI.
A cache cannot know what parts may apply. The origin doesn't know either.

The actual server recipient of a request containing query parameters
might have been passed along a completely different internal routing
path, with its own security filtering, from the same request with
those parameters hidden within the request content.

Allowing a cache to change the key by moving identifiers from the
content would allow a generic resource to poison the cache for other,
more specific resources.

reschke:
See also https://github.com/httpwg/http-extensions/issues/3122#issuecomment-3025288750

royfielding:
As much as I disagree with the idea, the way forward here is to clearly document how QUERY caching differs from (or is the same as) existing GET caches, in all respects. If the specification accurately describes the tradeoffs for implementation, I won't have a problem with it.

That means define QUERY in the same level of detail as we define GET for HTTP. In addition to defining how to construct the key, It means going through all of the conditional request header fields and defining how they are to be interpreted (mostly like GET does for testing when to send the response, not like PUT and POST does as a precondition for performing the method). It means demonstrating how a client would make a conditional QUERY request, why it would do so, and what to do for each significant status code returned (i.e., 304). Likewise for Range requests.

Note that all of the above will have to be defined as updates to 9110, since HTTP semantics assumes (for a reason) that there will never be another method like GET. This spec needs to redefine that, for its own reasons, and seriously discuss the negative tradeoffs for existing deployments of the Web and caching.

reschke:
If an origin server is willing to compute a digest, why not instead use it in a URI of a GETtable resource, which will work without any other changes for caching, conditionals and potentially range requests?

reschke:
That's

https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-11.html#section-2.1-2

no?

Re:

[...]

PR #3259, opened 2025-09-26
QUERY: caching - mention GET alternative wrt caching (addresses #3120)

reschke:
(note that this is on top of #3137)


draft-ietf-httpbis-safe-method-w-body.xml
```diff
@@ -460,6 +461,13 @@ q=foo&amp;limit=10&amp;sort=-published
         that they wish that no such transformation happens (but note that this directive is just
         advisory).
       </t>
+      <t>
+        Note that caching QUERY method responses is inherently more complex than caching responses
+        to GET, as complete reading of the request's content is needed in order to determine
+        the cache key. When a resource uses the Location response field (<xref target="location"/>)
+        to indicate a URI for the equivalent resource (<xref target="equivalent-resource"/>), clients
+        can switch to GET for subsequent requests, thereby simplifying processing. Furthermoe
```
```suggestion
        the cache key. If a QUERY response supplies a Location response field (<xref target="location"/>)
        to indicate a URI for an equivalent resource (<xref target="equivalent-resource"/>), clients
        can switch to GET for subsequent requests, thereby simplifying processing.
```

royfielding: minor tweaks

royfielding (changes requested):

PR #3267, opened 2025-10-01
no vary: terminology

reschke:


domenic (approved):
I guess "header" is no longer acceptable shorthand for "header field"? Is there a style guide for this somewhere?

PR #3284, opened 2025-10-14
no-vary: URI/URL/Caching terminology (fixes #3283)

### LINKED ISSUE #3283: no-vary: URI/URL terminology
The introduction mixes "URI" and "URL"; that might be confusing to 
readers. For URI, please cite RFC 3986. (maybe also explain why both 
terms are needed)

"resource has multiple URLs" -> "when different URLs that only differ in 
certain query components identify the same resource"

"some or all parts of the query" -> "...query component"

query parameter key needs a reference

"do not semantically affect the served resource" -> "do not affect what 
resource is identified"

"raw output" -> "output" (maybe)

(see https://lists.w3.org/Archives/Public/ietf-http-wg/2025OctDec/0000.html)

reschke:
This is an attempt (probably incomplete) to clean up terminology of URIs vs URLs vs resources vs responses etc.

IMHO the goal should be:

- use URL when referencing algorithms and formats in the WHATWG spec,
- use URI when talking about HTTP, Caching and related topics.

...and also explain what's going on wrt this.

### DIFF
--- a/draft-ietf-httpbis-no-vary-search.md
+++ b/draft-ietf-httpbis-no-vary-search.md
@@ -112,7 +112,7 @@ This specification defines a proposed HTTP response header field for changing ho
 
 # Introduction
 
-HTTP caching {{HTTP-CACHING}} is based on reusing resources which match across a number of cache keys. One of the most prominent is the presented target URI ({{Section 7.1 of HTTP}}). However, sometimes multiple URIs can represent the same resource. This leads to caches not always being as helpful as they could be: if the cache contains a response under one URI, but the resource is then requested under another, the cached version will be ignored.
+HTTP caching {{HTTP-CACHING}} is based on reusing resources which match across a number of cache keys. One of the most prominent is the presented target URI ({{Section 7.1 of HTTP}}). However, sometimes multiple URIs can represent the same resource. This leads to caches not always being as helpful as they could be: if the cache contains a response under one URI, but the response is then requested under another, the cached version will be ignored.
 
 The `No-Vary-Search` HTTP header field tackles a specific subset of this general problem, for when different URIs that only differ in
 certain query parameters identify the same resource. It allows resources to declare that some or all parts of the query component do not semantically affect the served resource, and thus can be ignored for cache matching purposes. For example, if the order of the query parameters do not affect which resource is identified, this is indicated using


draft-ietf-httpbis-no-vary-search.md
```diff
@@ -105,15 +106,16 @@ informative:
...
 
-This specification defines a proposed HTTP response header field for changing how URL search parameters impact caching.
+This specification defines a proposed HTTP response header field for changing how URI query parameters impact caching.
 
 --- middle
 
 # Introduction
 
-HTTP caching {{HTTP-CACHING}} is based on reusing resources which match across a number of cache keys. One of the most prominent is the presented target URI ({{Section 7.1 of HTTP}}). However, sometimes multiple URLs can represent the same resource. This leads to caches not always being as helpful as they could be: if the cache contains the resource under one URI, but the resource is then requested under another, the cached version will be ignored.
+HTTP caching {{HTTP-CACHING}} is based on reusing resources which match across a number of cache keys. One of the most prominent is the presented target URI ({{Section 7.1 of HTTP}}). However, sometimes multiple URIs can represent the same resource. This leads to caches not always being as helpful as they could be: if the cache contains a response under one URI, but the resource is then requested under another, the cached version will be ignored.
```
nidhijaju: It's a bit odd to call it "response" and then "resource" immediately after, as it's not clear which resource we're talking about. I recognize it also doesn't make sense to simply replace "resource" with "response", so can we say "the same resource" or something else to tie it back to the response somehow?

Should we also say "for one URI" instead of "under one URI"?

draft-ietf-httpbis-no-vary-search.md
```diff
@@ -131,12 +133,15 @@ And if the resource instead wants to take an allowlist-based approach, where onl
...
+{{header-definition}} defines the header field, using the {{STRUCTURED-FIELDS}} framework. {{data-model}} and {{parsing}} illustrate the data model for how the field value can be represented in specifications, and the process for parsing the raw output from the structured field parser into that data model. {{comparing}} gives the key algorithm for comparing if two URLs are equivalent under the influence of the header field; notably, it leans on the decomposition of the query component into keys and values given by the [application/x-www-form-urlencoded](https://url.spec.whatwg.org/#concept-urlencoded) format specified in {{WHATWG-URL}}. (As such, this header field is not useful for URLs whose query component does not follow that format.) Finally, {{caching}} explains how to extend {{Section 4 of HTTP-CACHING}} to take this new equivalence into account.
 
 # Conventions and Definitions
 
 {::boilerplate bcp14-tagged}
 
+In this document, the terms "URI" and "URL" are used mostly interchangeably, depending on context ({{URI}} and {{HTTP}} vs the algorithms specified in {{WHATWG-URL}}).
```
nidhijaju: It's not totally clear which is used when, so suggesting some text to make that clearer hopefully:

```suggestion
In this document, the terms "URI" and "URL" are used interchangeably, depending on context. "URI" is used in the context of {{URI}}, {{HTTP}}, and {{HTTP-CACHING}}, whereas "URL" is used in the context of the algorithms specified in {{WHATWG-URL}}.
```

draft-ietf-httpbis-no-vary-search.md
```diff
@@ -293,13 +298,13 @@ To _parse a key_ given an ASCII string _keyString_:
 
 ### Examples
 
-The parse a key algorithm allows encoding non-ASCII key strings in the ASCII structured header field format, similar to how the [application/x-www-form-urlencoded](https://url.spec.whatwg.org/#concept-urlencoded) format {{WHATWG-URL}} allows encoding an entire entry list of keys and values in ASCII URL format. For example,
+The parse a key algorithm allows encoding non-ASCII key strings in the ASCII structured header field format, similar to how the [application/x-www-form-urlencoded](https://url.spec.whatwg.org/#concept-urlencoded) format {{WHATWG-URL}} allows encoding an entire entry list of keys and values in URI (which is constricted to ASCII characters). For example,
```
nidhijaju:
```suggestion
The parse a key algorithm allows encoding non-ASCII key strings in the ASCII structured header field format, similar to how the [application/x-www-form-urlencoded](https://url.spec.whatwg.org/#concept-urlencoded) format {{WHATWG-URL}} allows encoding an entire entry list of keys and values in a URI (which is restricted to ASCII characters). For example,
```

draft-ietf-httpbis-no-vary-search.md
reschke: Should both be "response".

PR #3286, opened 2025-10-14
no-vary: updating CACHING (fixes #3285)

### LINKED ISSUE #3285: no-vary: updating CACHING
I found the first part of 
https://www.ietf.org/archive/id/draft-ietf-httpbis-no-vary-search-03.html#name-caching 
a bit hard to understand with respect to what is replaced by what.

See (text version):

> 7.  Caching
> 
>    If a cache [HTTP-CACHING] implements this specification, the
>    presented target URI requirement in Section 4 of [HTTP-CACHING] is
>    replaced with:
> 
>    *  one of the following:
> 
>       -  the presented target URI (Section 7.1 of [HTTP]) and that of
>          the stored response match, or
> 
>       -  the presented target URI and that of the stored response are
>          equivalent modulo search variance (Section 6), given the
>          variance obtained (Section 5.2) from the stored response.

Proposal: make both list items quotes; simplify the replacement text so 
that it does not require more nesting, and do not lose the ", and" at 
the end.

(see https://lists.w3.org/Archives/Public/ietf-http-wg/2025OctDec/0050.html)

reschke:


draft-ietf-httpbis-no-vary-search.md
```diff
@@ -419,11 +419,15 @@ So, for example, given any non-default value for `No-Vary-Search`, such as `No-V
...
 
-* one of the following:
-  * the presented target URI ({{Section 7.1 of HTTP}}) and that of the stored response match, or
-  * the presented target URI and that of the stored response are equivalent modulo search variance ({{comparing}}), given the variance obtained ({{obtain-a-url-search-variance}}) from the stored response.
+> the presented target URI (Section 7.1 of [HTTP]) and that of the stored response match, and
+{:quote}
+
+is replaced with:
+
+> the presented target URI (Section 7.1 of [HTTP]) and that of the stored response match or are equivalent modulo search variance, and
```
nidhijaju, line 429: I feel like preserving the structure of the original text in HTTP-CACHING is more clear than saying it's replaced with "one of the following" and sub-bullets.

If we think it's valuable, we could still include the references to the other sections:
```suggestion
> the presented target URI (Section 7.1 of [HTTP]) and that of the stored response match, or are equivalent modulo search variance ({{comparing}}), given the variance obtained ({{obtain-a-url-search-variance}}) from the stored response, and
```

nidhijaju (approved):

draft-ietf-httpbis-no-vary-search.md
nidhijaju: @reschke Friendly ping, what do you think about the suggestion?

reschke: I still prefer "my suggestion, but at the end of the day, it's mostly an editorial choice.

 Maybe ask @mnot ?

mnot: I don't have strong feelings here.

Issue #3165, opened 2025-08-11
RESUMABLE: the original target cannot be updated


LPardue:
@gstrauss provided the following feedback on https://lists.w3.org/Archives/Public/ietf-http-wg/2025JulSep/0094.html wrt Section 4

> Related, the server is required to maintain server-side state including the original target.  The target is then unable to be later changed, e.g. a date in the title of the doc or blog post.  That does not matter in all case, but might matter in some.  In any case, it would be better if the target was required with Upload-Complete: ?1.  The target could still be provided in the initial request, in order for the server to do some early checks, but those checks should be repeated immediately before processing after Upload-Complete: ?1

gstrauss:
The comment also questions how much state the server is required to store on the initial request, and suggests that it might be better for the client to be required to provide some/all of that state when the temporary resource is complete.  My comment is not so much a request for a feature to change the target, but more an overall questioning of the design.

There can be a large amount of time between the initial request to create the temporary resource and the completion of upload of the temporary resource.  The server should revalidate the request against all appropriate policies when the temporary resource is submitted to the target resource for processing.  (The draft also mentions this briefly in Security Considerations, but might also mention that the entire request, including headers, is untrusted.)

Acconut:
> There can be a large amount of time between the initial request to create the temporary resource and the completion of upload of the temporary resource. The server should revalidate the request against all appropriate policies when the temporary resource is submitted to the target resource for processing.

That's a valuable suggestion we should add to the draft. Thank you for bringing this up!

PR #3294, opened 2025-10-16
RESUMABLE: Security consideration regarding time-of-check to time-of-use

Acconut:
This PR adds a security consideration addressing issues that can arise from the delay between the start and end of an upload. It was brought up in https://github.com/httpwg/http-extensions/issues/3165. Not all implementations are vulnerable to this, but one has to be careful when a process is split up across multiple, long-running requests.

Closes https://github.com/httpwg/http-extensions/issues/3165.


Acconut:
That's an interesting perspective. Token expiration is indeed a problem with resumable uploads, especially if they are running for extended periods. In the past, I've had a few inquiries from people seeking help with handling expired tokens in my client implementations.

That being said, resumable uploads shouldn't be able to circumvent access restrictions. For example, if an administrator removes my write privileges while my upload is ongoing, the file shouldn't be written to the destination once the upload is complete. Even if I had the privilege when the upload began. That was the motivation behind this change.

However, I don't think our perspectives are exclusive. Token validation is best performed when the upload begins, but privileges have to be checked at the end again. Let me update the text to account for that.

Issue #3172, opened 2025-08-11
RESUMABLE: proposal to relax the requirements on changs to limits


LPardue:
@gstrauss provided the following feedback on https://lists.w3.org/Archives/Public/ietf-http-wg/2025JulSep/0094.html wrt Section 4.1.4

> "... the existence of a limit or its value MUST NOT change ..."
Over-specified.  Maybe SHOULD NOT.  A more intelligent server might want to supply hints or make adjustments to adapt to changing conditions.  Many of these "limits" are over-specified and might be better as hints.

To add:

* The current text does not specify how to deal with a change in limits, which might be an indicator that this MUST NOT requirement is not being adhered to
* There is no rationale provided about why changes are prohibited in this way. Explanation could help, whether or not the MUST NOT is changed to a SHOULD not. Text could cover things such as whether limit decreases are worse than limit increases.

Changes in this topic would be a (marginal) design change

PR #3295, opened 2025-10-16
RESUMABLE: Relax rules on limit changes

Acconut:
Closes https://github.com/httpwg/http-extensions/issues/3172.

Limits changes are now allowed, but not recommended if they can cause ongoing uploads to fail. This was the main reason behind the previous use of MUST NOT, but that can also be expressed without such a strict requirement.


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -347,9 +347,9 @@ The following key-value pairs are defined:
...
+: Specifies the remaining lifetime of the upload resource in seconds counted from the generation of the response. After the resource's lifetime is reached, the server might make the upload resource inaccessible and a client SHOULD NOT attempt to access the upload resource as these requests will likely fail. The value is an Integer.
 
-Except for the `max-age` limit, the existence of a limit or its value MUST NOT change throughout the lifetime of the upload resource.
+Clients usually discover limits through the `Upload-Limit` header field when the upload resource is created ({{upload-creation}}). Throughout the lifetime of the upload resource, these limits SHOULD NOT change in a way that causes failures for clients adhering to the initially discovered limits. In particular, whereas the values for `max-size`, `max-append-size`, and `max-age` can increase without harm, they SHOULD NOT decrease as they can cause ongoing uploads to fail. Similarly, the value for `min-append-size` SHOULD NOT increase.
```
LPardue: double-checking, is it clearly stated elsewhere how client might fail for all limits?

Acconut: The section on upload creation includes the following:

> The client SHOULD respect any limits ({{upload-limit}}) announced in the Upload-Limit header field in interim or final responses. In particular, if the allowed maximum size is less than the amount of representation data the client intends to upload, the client SHOULD stop the current request immediately and cancel the upload ({{upload-cancellation}}).

Basically, when an upload is started, but with the first response from the server the client then realizes that it cannot meet the limits, it should stop and cancel the upload.

Do you think we should expand on this?

LPardue: That text seems fine. Maybe we can just cross ref to it here as a helpful pointer?

LPardue (approved):

Issue #3184, opened 2025-08-11
RESUMABLE: upload creation server behaviour text is confusing


LPardue:
@gstrauss provided the following feedback on https://lists.w3.org/Archives/Public/ietf-http-wg/2025JulSep/0094.html wrt Section 4.2.2

> "If the server does not receive the entire request content, for example because of canceled requests or dropped connections, it SHOULD append as much of the requst content as possible to the upload resource."
> This is confusing unless the doc has been read carefully multiple times.  This also conflicts with wording elsewhere in the doc which suggests that the request should be aborted immediately when there is a new request for state of the upload resource (4.6 Concurrency).  Perhaps you meant that pending data should be read without blocking and appended to the upload resource, and that the server should then abort the request instead of waiting for additional data.

PR #3297, opened 2025-10-16
RESUMABLE: Clarify handling of partial content

Acconut:
Closes https://github.com/httpwg/http-extensions/issues/3184.

The text was a bit confusing. The mentioned interruptions would occur naturally because connections get interrupted. I tried to express this a bit better and then also added that saving partial content messages allows the client to resume from the last byte. This should also help making these paragraphs more understandable.

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -396,7 +396,7 @@ The server MUST record the length according to {{upload-length}} if the `Upload-
 
 While the request content is being received, the server MAY send additional interim responses with a `104 (Upload Resumption Supported)` status code and the `Upload-Offset` header field set to the current offset to inform the client about the upload progress. These interim responses MUST NOT include the `Location` header field.
 
-The server might not receive the entire request content when the upload is interrupted, for example because of dropped connections or canceled requests. In this case, the server SHOULD append as much of the request content as possible to the upload resource, allowing the client to resume the upload from where it was interrupted. In addition, the upload resource MUST NOT be considered complete then.
+The server might not receive the entire request content when the upload is interrupted, for example because of dropped connection or canceled request. In this case, the server SHOULD append as much of the request content as possible to the upload resource, allowing the client to resume the upload from where it was interrupted. In addition, the upload resource MUST NOT be considered complete then.
 
 ### Examples {#upload-creation-example}
 


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -396,7 +396,7 @@ The server MUST record the length according to {{upload-length}} if the `Upload-
 
 While the request content is being received, the server MAY send additional interim responses with a `104 (Upload Resumption Supported)` status code and the `Upload-Offset` header field set to the current offset to inform the client about the upload progress. These interim responses MUST NOT include the `Location` header field.
 
-If the server does not receive the entire request content, for example because of canceled requests or dropped connections, it SHOULD append as much of the request content as possible to the upload resource. The upload resource MUST NOT be considered complete then.
+The server might not receive the entire request content when the upload is interrupted, for example because of dropped connections or canceled requests. In this case, the server SHOULD append as much of the request content as possible to the upload resource, allowing the client to resume the upload from where it was interrupted. In addition, the upload resource MUST NOT be considered complete then.
```
```suggestion
The server might not receive the entire request content when the upload is interrupted, for example because of dropped connection or canceled request. In this case, the server SHOULD append as much of the request content as possible to the upload resource, allowing the client to resume the upload from where it was interrupted. In addition, the upload resource MUST NOT be considered complete then.
```

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -571,7 +571,7 @@ If the client received a final response with a
...
 
-If the upload resource does not receive the entire patch document, for example because of canceled requests or dropped connections, it SHOULD append as much of the patch document as possible, starting at its beginning and without discontinuities. Appending a continuous section starting at the patch document's beginning constitutes a successful PATCH as defined in {{Section 2 of PATCH}}.
+The server might not receive the entire patch document when the upload is interrupted, for example because of dropped connections or canceled requests. In this case, the server SHOULD append as much of the patch document as possible to the upload resource, starting at its beginning and without discontinuities. Appending a continuous section starting at the patch document's beginning constitutes a successful PATCH as defined in {{Section 2 of PATCH}}. Saving the received data allows the client to resume the upload from where it was interrupted. In addition, the upload resource MUST NOT be considered complete then.
```
```suggestion
The server might not receive the entire patch document when the upload is interrupted, for example because of a dropped connection or canceled request. In this case, the server SHOULD append as much of the patch document as possible to the upload resource, starting at its beginning and without discontinuities. Appending a continuous section starting at the patch document's beginning constitutes a successful PATCH as defined in {{Section 2 of PATCH}}. Saving the received data allows the client to resume the upload from where it was interrupted. In addition, the upload resource MUST NOT be considered complete then.
```

Issue #3263, opened 2025-09-29
RESUMABLE: diagram shows client waiting on 104


Acconut:
During WGLC review, @wtarreau raised the comment in https://mailarchive.ietf.org/arch/msg/httpbisa/JjZSzo1siwNKqsilcw8Gz8WE6tc/:

> 
> I was initially confused by the first diagram in 3.1:
>   ```
>    Client                                  Server
>    |                                            |
>    | POST /project/123/files                    |
>    | Upload-Complete: ?1                        |
>    |------------------------------------------->|
>    |                                            |
>    |                                            | Reserve resources
>    |                                            | for upload
>    |                                            |-----------------.
>    |                                            |                 |
>    |                                            |<----------------'
>    |                                            |
>    |            104 Upload Resumption Supported |
>    |            Location: /uploads/abc          |
>    |<-------------------------------------------|
>    |                                            |
>    X--------------Flow Interrupted--------------X
>    ```
> Which I've read as:
> 
> - client sends POST with headers
> - server returns 104
> - client the possibly uploads data
> - transfer is interrupted
> 
> As an intermediary implemented it worried me that we'd have a new header
> that would cause uploads to wait for an interim response. When reading
> the rest I figured it does not make sense and is just my incorrect
> interpretation of the diagram. But it also means that others could
> possibly implement it like this (especially clients which could then
> wait for the 104 before sending data).
>
> I don't know how we could fix the diagram. Maybe adding an arrow from the
> client to the server in the middle, like this:
> 
> ```
>    Client                                  Server
>    |                                            |
>    | POST /project/123/files                    |
>    | Upload-Complete: ?1                        |
>    |------------------------------------------->|
>    | body                                       |
>    |------------------------------------------->| Reserve resources
>    |                                            | for upload
>    |                                            |-----------------.
>    |                                            |                 |
>    |                                            |<----------------'
>    |                                            |
>    |            104 Upload Resumption Supported |
>    |            Location: /uploads/abc          |
>    |<-------------------------------------------|
>    |                                            |
> ```

While a client could in theory be waiting on a 104 response to check for Upload-Limit, I don't expect most client to behave that way. They will likely just start transferring the data as soon as the request is opened. Therefore, we should update the diagram to make this clearer.

PR #3303, opened 2025-10-20
RESUMABLE: include request content in examples

Acconut:
Closes https://github.com/httpwg/http-extensions/issues/3263

@wtarreau noted that the examples indicate that clients should wait for the 104 response before starting to send representation data. They can do this but it's neither mandatory nor usual. This PR updates the non-normative description of the example. It also adds the request content to the example to better illustrate which requests carry what representation data.

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -125,7 +125,7 @@ The remainder of this section uses examples to illustrate different interactions
 
 In this example, the client first attempts to upload representation data with a known size in a single HTTP request to the resource at `/project/123/files`. An interruption occurs and the client then attempts to resume the upload using subsequent HTTP requests to the upload resource at `/uploads/abc`.
 
-1) The client notifies the server that it wants to begin an upload ({{upload-creation}}). The server reserves the required resources to accept the upload from the client and then sends an interim response to the client, which signals the server's support of resumable upload as well as the upload resource's URI via the Location header field ({{Section 10.2.2 of HTTP}}). The client can start sending the representation data in the request content without delay right after the request header. Alternatively, it could also await the acknoledgement in form of the interim response.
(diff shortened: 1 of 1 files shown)


wtarreau:
Looks good to me, thanks!

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -125,15 +125,15 @@ The remainder of this section uses examples to illustrate different interactions
...
 
-1) The client notifies the server that it wants to begin an upload ({{upload-creation}}). The server reserves the required resources to accept the upload from the client, and the client begins transferring the entire representation data in the request content.
-
-An interim response can be sent to the client, which signals the server's support of resumable upload as well as the upload resource's URI via the Location header field ({{Section 10.2.2 of HTTP}}).
+1) The client notifies the server that it wants to begin an upload ({{upload-creation}}). The server reserves the required resources to accept the upload from the client and then sends an interim response to the client, which signals the server's support of resumable upload as well as the upload resource's URI via the Location header field ({{Section 10.2.2 of HTTP}}). The client can start sending the representation data in the request content without delay right after the request header. Alternatively, it could also await the acknoledgement in form of the interim response.
```
```suggestion
1) The client notifies the server that it wants to begin an upload ({{upload-creation}}). The server reserves the required resources to accept the upload from the client and then sends an interim response to the client, which signals the server's support of resumable upload as well as the upload resource's URI via the Location header field ({{Section 10.2.2 of HTTP}}). The client can start sending the representation data in the request content immediately after the request header. Alternatively, it could also await the acknowledgement in form of the interim response.
```

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -269,6 +277,8 @@ Client                                       Server
 | Upload-Offset: Z                                |
 | Upload-Complete: ?1                             |
```
LPardue, line 277: nit: is there any reason that `upload-offset` appears before `upload-complete` here? The diff makes it clear that the ordering is inconsistent across the examples. I know ordering doesn't matter in general but it seems our document examples could be consistent

LPardue (approved):
lgtm modulo nits

draft-ietf-httpbis-resumable-upload.md
Acconut: `acknowledgement` is my arch enemy. It happens so often...

draft-ietf-httpbis-resumable-upload.md
Acconut: Good call, I'll address that.

PR #3309, opened 2025-10-22
Shepherd feedback on Incremental

martinthomson:
This will conflict.  I'll rebase next.

### DIFF
--- a/draft-ietf-httpbis-incremental.md
+++ b/draft-ietf-httpbis-incremental.md
@@ -1,5 +1,5 @@
 ---
-title: "Requesting Incremental Forwarding of HTTP Messages"
+title: "Incremental Forwarding of HTTP Messages"
 docname: draft-ietf-httpbis-incremental-latest
 category: std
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-incremental.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Incremental HTTP Messages"
+title: "Requesting Incremental Forwarding of HTTP Messages"
```
mnot: I'm a tiny bit uncomfortable with 'Requesting', as it might lead readers to believe this is only a request-side mechanic. YMMV.

martinthomson:
```suggestion
title: "Incremental Forwarding of HTTP Messages"
```

Better ?

mnot: Yep, thanks.

kazuho (approved):
Thank you for doing this!

PR #3311, opened 2025-10-23
[resumable] Fix missed value in creation section

smatsson:
Fixes a minor issue in the text for server behavior in the creation part. 

If the client sends an empty creation request, i.e. zero byte body and `Upload-Complete: ?0`, it does not make sense for the server to record the length in `Content-Length` given that it will be zero bytes.


guoye-zhang (approved):

Acconut (approved):
Thanks!

Issue #3193, opened 2025-08-11
RESUMABLE: feature request - chunk granularity


LPardue:
The authors received an off-list message from a server implementer. They have an existing resumable upload deployment based on a custom protocol and are working to integrate support for draft-ietf-httpbis-resumable-upload.

One of the features that the custom protocol provides is the notion of a chunk granularity. This particular protocol uses an HTTP response header field to advertise a value in bytes. Clients that upload chunks are expected to use an integer multiple of the chunk granularity value (modulo the last chunk, which can be a fragment of it). For example , if a value of 262144 bytes was advertised, a client could send chunks of 262,144, or 524,288, or 2,621,440 bytes.

Other custom reliable upload protocols depend on out-of-band advertisements of this concept. For example, see these requirements listed on a webpage [1]:

> Chunk size must be divisible by 256 KiB (256x1024 bytes). Round your chunk size to the nearest multiple of 256 KiB. Note that the final chunk of an upload that fits within a single chunk is exempt from this requirement.

draft-ietf-httpbis-resumable-upload does not support this notion of chunk granularity. The protocol does support in-band advertisement of limits via the [Upload-Limit](https://httpwg.org/http-extensions/draft-ietf-httpbis-resumable-upload.html#name-limits) field but it defines only the `min-append-size` and `max-append-size` properties. The chunk granularity falls somewhere between these two - the range of chunk sizes can be between a min and a max, but a set of sizes are allowed.

There's a few ways we might be able to support this draft-ietf-httpbis-resumable-upload. However, the key point is that this is a feature request that we need to first consider if we want to support.

[1] https://developers.cloudflare.com/stream/uploading-videos/resumable-uploads/#requirements

LPardue:
Thanks! 

Do you have any sense if the set of values tends to benefit from being an arithmetic progession (i.e a constant delta between values) or if you have a non-lonear range like 266K, 512K, 1M, 2M, 8M

I kinda like the idea of a list to be explicit and allow flexibility in the sequence. The downside might be that a server provides nonsense values in the set and there'd need to be some additional parsing/validation logic defined to be defensive to that. For example, an unsorted list, a list with duplicates, etc. However, any format for expressing a granularity would need _some_ validation against other hints, so a list might not push that much further

Acconut:
[...]

Issue #3215, opened 2025-08-23
RESUMABLE: thoughts on Prior Knowledge


LPardue:
During WGLC review, @MikeBishop raised the following issue on https://lists.w3.org/Archives/Public/ietf-http-wg/2025JulSep/0123.html

> The scenario in Section 3.2 seems dangerous to me. "Prior knowledge" is used to wave away the lack of a discovery step, but the client's prior knowledge might be wrong unless it's fresh. This would result in the client transferring what it knows to be a partial representation and a legacy server/resource attempting to process it as a complete one. How can we guard against this case?

> This reminds me of Expect: 100-continue in that the client really would like to know what the server is going to do before deciding how to proceed. (And I note that the authors made a similar comparison in Section 12.1.) Is there a way the client can proffer a partial upload in such a way that a legacy server will not see it as a completed request? It may be worth instead defining a mechanism to request an upload resource without including any initial content, since an empty payload will be less likely to be usable as the underlying media type. (Perhaps the application/partial-upload media type should be used here as well, to ensure a legacy server will not process it as the underlying media type? But in that case, how does the server eventually wind up with the correct media type of the upload?)

> Section 4.1.4 offers another option -- since the base resource is required to support OPTIONS, perhaps this example should explicitly show the client first making an OPTIONS request to determine whether this resource supports resumable uploads? This requirement on the base resource should probably get a little more emphasis, too. Maybe prior to Section 4 (Upload Resource), there should be a section that talks about differences in interacting with the base resource? See my related comment on 4.1.4.

PR #3312, opened 2025-10-23
resumable: clarify client behavior on 413 response

mertalev:
This PR aims to address https://github.com/httpwg/http-extensions/issues/3193#issuecomment-3394903322
* Makes min-append-size and max-append-size apply to upload creation requests as well
* Recommends retrying with body in this range after receiving a 413
* If limits are unknown, recommends discovering them with an OPTIONS request after a 413

These changes should make clients more resilient to server limits without giving up right away.

This is my first PR here, so sorry in advance if I made a mistake!


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -367,7 +367,9 @@ A server that supports the creation of a resumable upload resource ({{upload-cre
...
+Servers, including intermediaries, can (and often do) apply restrictions on the size of individual request message content. There is no standard mechanism to communicate such existing size restriction. A server that implements one can respond with a 413 Content Too Large status code; see {{Section 15.5.14 of HTTP}}. Appending to an upload resource, as a series of appends, can be used to upload data up to the `max-size` limit without encountering per-message limits. Servers might apply restrictions that are smaller than the append limits, which would also result in a failed request.
+
+When a client receives a response with a 413 Content Too Large status code during upload creation ({{upload-creation}}) or append ({{upload-appending}}), it SHOULD retry the request with a smaller size residing between `min-append-size` and `max-append-size`. If these limits are unknown at the time of receiving such a response, the client SHOULD attempt to discover them through an `OPTIONS` request. Cases where a request uses `min-append-size` yet fails with a 413 Content Too Large response might indicate a deployment mismatch that cannot be recovered from.
```
guoye-zhang: We should probably require 413 to carry an `Upload-Limit` header if it's from a resumable upload server. Upload creation can be attempted without knowledge of server support, and clients can't retry a non-resumable upload.

mertalev: The issue is that a proxy might respond with 413 based on the request's Content-Length, so even if the server is resumable it can't add the header in this case.

mertalev: The OPTIONS request to learn the limits also doubles as a way to discover resumability. Maybe it should be emphasized here more, so as to avoid retrying without knowing about resumability.

guoye-zhang: As an implementor of a generic HTTP client, we can attempt resumable uploads for all uploads since it has little overhead and minimum risk, but we cannot perform an OPTIONS discovery step for all uploads that fail with 413. We need to know that the other side supports resumable uploads before doing something expensive.

mertalev: The OPTIONS request wouldn't be a MUST - it's more that they MUST NOT retry if they don't know the server supports resumability. I suppose the only other way for a generic library to handle this proxy scenario would be to configure it to assume the server is resumable.

guoye-zhang: SHOULD is a recommendation and means that we think it's a good idea. I don't think it's feasible for us since you know there are unupgradable printers that would crash when it receives an OPTIONS request. We have to be as safe as possible when discovering resumable upload support.

Yeah, this has to be an opt-in feature, although a more desirable opt-in behavior probably is to always send OPTIONS to detect Upload-Limit first instead of relying on 413.

mertalev: That makes sense. To summarize:
* If 413 contains Upload-Limit header, the client SHOULD try again within these limits
* If not, the client MAY try OPTIONS to learn limits and try again
* Client MUST NOT retry if it isn't aware of server resumable upload support, either from the above or from prior knowledge

Do those changes sound good?

guoye-zhang: Yes, looks good. I believe the reason we didn't tackle upload creation originally was the complexity of its interaction with discovery. Upload appending can only be performed if the server support is confirmed so its behavior is easier to define.

Minor nit: we stopped using the phrase "prior knowledge" based on the feedback in #3215.

guoye-zhang:
Would prefer the server sending 104 with Location prior to 413. The only thing the client needs to do in that case is to allow 413 retry.

Based on the current spec, the server could send a 5xx after processing the request body to the limit, which would prompt the client to retry and resume. It's a hacky workaround though.

guoye-zhang:
The change looks OK to me. @Acconut what do you think?

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -367,7 +367,9 @@ A server that supports the creation of a resumable upload resource ({{upload-cre
...
+
+A server responding with a 413 Content Too Large status code as a result of its upload limits MUST include the Upload-Limit header in this response, unless it is unable to do so (such as if an intermediary responds in its place). When a client receives a response with a 413 Content Too Large status code during upload creation ({{upload-creation}}) or append ({{upload-appending}}), it SHOULD retry the request with a smaller size residing between `min-append-size` and `max-append-size`. If these limits are unknown at the time of receiving such a response, the client MAY attempt to discover them through an `OPTIONS` request. The client MUST NOT retry the upload request without knowledge of server support of resumable uploads. Cases where a request uses `min-append-size` yet fails with a 413 Content Too Large response might indicate a deployment mismatch that cannot be recovered from.
```
Acconut: This is a good addition overall! Thank you for the PR. The phrasing MUST + "unless unable" seems a bit odd to me. Would it be better to use SHOULD here in the first place?


```suggestion
A server responding with a 413 Content Too Large status code as a result of its upload limits SHOULD include the Upload-Limit header in this response if possible. This might not be possible if intermediaries enforce content size restrictions. When a client receives a response with a 413 Content Too Large status code during upload creation ({{upload-creation}}) or append ({{upload-appending}}), it SHOULD retry the request with a smaller size residing between `min-append-size` and `max-append-size`. If these limits are unknown at the time of receiving such a response, the client MAY attempt to discover them through an `OPTIONS` request. The client MUST NOT retry the upload request without knowledge of server support of resumable uploads. Cases where a request uses `min-append-size` yet fails with a 413 Content Too Large response might indicate a deployment mismatch that cannot be recovered from.
```

mertalev: Thanks! I agree SHOULD makes sense there since the server is already allowed to not add it if it can't.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -369,7 +369,9 @@ When responding to an `OPTIONS` request, if the resource that is the target of t
...
+
+A server responding with a `413 (Content Too Large)` status code as a result of its upload limits SHOULD include the Upload-Limit header in this response if possible. This might not be possible if intermediaries enforce content size restrictions. When a client receives a response with a `413 (Content Too Large)` status code during upload creation ({{upload-creation}}) or append ({{upload-appending}}), it SHOULD retry the request with a smaller size residing between `min-append-size` and `max-append-size`. If these limits are unknown at the time of receiving such a response, the client MAY attempt to discover them through an `OPTIONS` request. The client MUST NOT retry the upload request without knowledge of server support of resumable uploads. Cases where a request uses `min-append-size` yet fails with a `413 (Content Too Large)` response might indicate a deployment mismatch that cannot be recovered from.
```
Acconut:
```suggestion
A server responding with a `413 (Content Too Large)` status code as a result of its upload limits SHOULD include the `Upload-Limit` header field in this response if possible. This might not be possible if intermediaries enforce content size restrictions. When a client receives a response with a `413 (Content Too Large)` status code during upload creation ({{upload-creation}}) or append ({{upload-appending}}), it SHOULD retry the request with a smaller size residing between `min-append-size` and `max-append-size`. If these limits are unknown at the time of receiving such a response, the client can attempt to discover them through an `OPTIONS` request. Retrying with the same request content size will likely yield the same error response. Cases where the request content size matches the `min-append-size` limit yet fails with a `413 (Content Too Large)` response might indicate a deployment mismatch that cannot be recovered from.
```

I think we can reduce the normative language here since it won't help with interoperability much here. It's better to describe the resulting situation, so implementers will know what not to do and why.

After that we can merge this :)

Acconut (approved):
Thank you!

PR #3317, opened 2025-10-26
resumable: clarify greater offset

### LINKED ISSUE #3268: RESUMABLE: What happens when a POST request is made for an already-created incomplete resource?
My use-case is that a given user can only make one resource with a given checksum. This checksum is provided in the Repr-Digest header in the initial request and validated on upload completion. The resource URI itself is not a checksum, but a UUID. Once the client has created the resource with the initial POST request, it should either complete it (possibly in multiple parts) or cancel it. Once cancelled, the user is allowed to create a new resource with a different UUID for the same checksum.

The spec is clear enough that a complete resource should not be modified and provides a response to use for this case. However, what should happen if the client creates this resource without completing it, then tries to create it again?
* Should it be rejected with a 400 code?
* Should it cause the incomplete resource to be invalidated?
* Should it include a Location header to direct it to the existing resource?
* Should it truncate the existing resource and start over?

Given the client needs to know the resource in order to resume or cancel it in this case, I think it makes sense to include the Location header with an error status code to allow them to take further action, possibly also including the Upload-Offset.

mertalev:
Resolves #3268, making it more explicit that the resource offset is allowed to be greater than what the client sent in its current upload. It also consolidates the rejection behavior for lesser and greater offsets.

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -992,6 +992,12 @@ Reference:
 # Changes
 {:removeinrfc="true"}
 
+## Since draft-ietf-httpbis-resumable-upload-11
+{:numbered="false"}
+
+* Clear up different responsibilities of server and upload resource.
+* Relax recommendations on client handling greater offsets.
+
 ## Since draft-ietf-httpbis-resumable-upload-10
 {:numbered="false"}
 


Acconut:
Thank you proposing this text. However, I think that this paragraph is a bit too ambiguous about how clients react to suddenly increasing offsets? Are they expected to follow the offset (i.e. MUST)? Are they recommended to do so (i.e. SHOULD)? Can they decide what to do?

The current text allows clients to abort of offset jumps, but doesn't require this behavior. A client that follows offset jumps would still be compliant with the current text. From that perspective it seems like this PR doesn't change much in the end yet.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -502,7 +502,7 @@ HTTP/1.1 405 Method Not Allowed
...
+The offset can be less than or equal to the total length of the representation data. If the offset is valid and greater than the number of bytes the client has already sent during this upload, which can indicate that other upload sessions have successfully appended to the resource, the client MUST seek to the provided offset of the representation data before continuing the upload. The client is expected to handle backtracking of a reasonable length. If the offset is invalid for this upload, or if the client cannot reproduce the representation data at the received offset, the upload MUST be considered a failure. The client then MUST NOT continue the upload and SHOULD cancel the upload ({{upload-cancellation}}).
```
Acconut: I think we can lose a few more words from this paragraph now that clients can't make any assumption about the relation between the offset and the bytes sent. This should also help with not being too restrictive. 

```suggestion
The offset can be less than or equal to the number of bytes of representation data that the client has already sent. The client is expected to handle backtracking of a reasonable length. On the other hand, the offset can be greater than the amount of sent representation data if the upload resource obtained additional representation data on behalf of the client. If the client is not able to provide the representation data at the given offset, the upload MUST be considered a failure. The client then MUST NOT continue the upload and SHOULD cancel the upload ({{upload-cancellation}}).
```

What do you think, @guoye-zhang?

guoye-zhang: Looks good, yes

mertalev: Looks good to me too!

Acconut (approved):

Issue #3275, opened 2025-10-10
RESUMABLE: Clarify client recovery behavior for 409 Conflict responses


danielresnick:
This issue is a spin-off from the discussion in #3193 regarding chunk granularity, as suggested by @guoye-zhang.

The core issue is that the current specification is ambiguous about whether clients need to handle a 409 Conflict response. This ambiguity can lead to upload failures in recoverable scenarios, undermining the specs goals.

**Problem**

A fully spec complaint client can receive a 409 Conflict response due to factors outside the client's direct control e.g. a network interruption causing the client and server to become desynchronized regarding the current Upload-Offset (client thinks a PATCH request has failed, but it actually succeeded).

The Concurrency section of the spec discusses this, and RECOMMENDS an approach to avoid ever serving clients 409 Conflict responses (latest request wins; previous requests are cancelled), however this is a non-binding recommendation and is acknowledged to be onerous / potentially impractical for servers to implement. A more realistic approach for large-scale servers to meet the requirements of the Concurrency section would be to use a strongly consistent database to maintain session state (current offset, etc) along with optimistic concurrency control to ensure versioning.

Even a proactive client can't avoid this race condition. Consider a scenario where a client's PATCH request times out locally, but is still in-flight to the server. To recover, the client sends a HEAD request to get the current offset. The server might respond to this HEAD request before it has processed the original, in-flight PATCH. By the time the client acts on that HEAD response and sends a new PATCH, the original one may have finally completed, advancing the server's offset and causing the new PATCH to be correctly rejected with a 409 Conflict.

As confirmed by @guoye-zhang, major client implementations like Apple's NSURLSession today treat all 4xx responses as fatal errors and do not attempt to retry / recover.

If a client gives up on a recoverable 409 Conflict, the protocol fails to deliver on its promise of reliable uploads over unreliable connections.

**Proposed Solution**

The spec should be updated to unambiguously state that a client MUST (or at least SHOULD) treat a 409 Conflict response as a recoverable signal.

The recovery mechanism would be via Upload-Offset in the 409 Conflict response and/or the Offset Retrieval mechanism.

Mandating this recovery logic (or at least strongly recommending it) would ensure that it is baked into foundational libraries, providing robust, out-of-the-box reliability for application developers who may not be aware of this specific failure case.

Keen for feedback from others as I’m not sure to what extent this has been discussed previously.

guoye-zhang:
[...]

PR #3319, opened 2025-10-29
resumable: clarify handling of conflicts and retry policies

danielresnick:
See #3275 for discussion.


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -577,10 +579,9 @@ If the client received a final response with the `Upload-Complete: ?0` header fi
 
 - A `2xx (Successful)` status code indicates that representation data was appended but the upload is not complete. The client can continue appending representation data.
 - For a `307 (Temporary Redirect)` or `308 (Permanent Redirect)` status code, the client MAY retry appending to the new URI.
-- For a `4xx (Client Error)` status code, the client SHOULD NOT attempt to retry or resume the upload, unless the semantics of the response allow or recommend the client to retry the request.
-- For a `5xx (Server Error)` status code, the client MAY automatically attempt upload resumption by retrieving the current offset ({{offset-retrieving}}).
-
-If no final response was received at all due to connectivity issues, the client MAY automatically attempt upload resumption by retrieving the current offset ({{offset-retrieving}}).
+- For a `409 (Conflict)` status code, the client SHOULD attempt to resume the upload by using the offset from the `Upload-Offset` response header field.
```
guoye-zhang: I think it's better to keep this flow the same as 5xx: recommending doing an offset retrieval before resuming. Otherwise this creates a new code path, and the text in "Upload Append" about picking the offset is also inconsistent with the recommendation here.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -653,6 +654,56 @@ Content-Type: application/json
 }
 ~~~
 
+C) The following example shows a client attempting to resume an upload with an incorrect offset. The server rejects the request with a `409 (Conflict)` status code and includes the correct offset in the `Upload-Offset` header field. The client then retries the request with the correct offset.
```
guoye-zhang: The previous examples were all happy paths, and my preference is to keep it that way. If we want to include examples for exceptional cases, we should discuss what are the most common failures and why this particular case needs highlighting.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -686,9 +737,15 @@ Resumable uploads, as defined in this document, do not permit uploading represen
...
+Servers SHOULD choose a strategy that best fits their architecture while fulfilling the requirements of this section. Regardless of the chosen strategy, clients MUST be prepared to handle a `409 (Conflict)` response as a recoverable error, as described in {{upload-appending}}.
```
guoye-zhang: All retry behaviors are MAY/SHOULD, do we need a MUST here?

guoye-zhang (commented):
Thanks for opening this PR, the text looks good. Left a few minor comments.

draft-ietf-httpbis-resumable-upload.md
danielresnick: Good catch, fixed to be consistent (SHOULD)

draft-ietf-httpbis-resumable-upload.md
danielresnick: SGTM, removed for now

draft-ietf-httpbis-resumable-upload.md
danielresnick: Updated, how does this look?  It feels wrong to completely omit the option to leverage the Upload-Offset from the response since it's called out as a MUST for the server to return it just below

guoye-zhang (approved):

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -686,9 +687,15 @@ Resumable uploads, as defined in this document, do not permit uploading represen
...
+To meet these requirements, a server can use various strategies. Three common approaches are:
+
+1.  **Preemptive Cancellation**: If the server receives a new request for an upload resource while a previous request for the same resource is in-flight, it abruptly terminates the previous HTTP connection or stream before processing the new request.
+
+2.  **Pessimistic Locking**: The server processes requests for a given upload resource sequentially, effectively creating an exclusive lock on that resource. A new request is only processed after the previous one completes. This can be simpler to implement but may lead to delays if a request hangs from the server's perspective.
+
+3.  **Optimistic Concurrency Control**: By maintaining the upload's state in a strongly consistent datastore, the server can atomically check if the `Upload-Offset` in a request matches the resource's current state. If they do not match, the server can reject the request with a `409 (Conflict)` status code.
```
Acconut: I'm not sure if this is enough to guarantee issue-free uploads. What if the server receives two requests for the same (correct) offset? They both match, but which content is appended to the upload resource? Did you mean to say that the content should be appended in one transaction together with the offset check? But that would make this the same as the pessimistic locking, I think.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -504,18 +504,20 @@ If the client wants to resume the upload after an interruption, it has to know t
...
 - `2xx (Successful)` status code, the client can continue appending representation data to it ({{upload-appending}}) if the upload is not complete yet.
 - `307 (Temporary Redirect)` or `308 (Permanent Redirect)` status code, the client MAY retry retrieving the offset from the new URI.
 - `4xx (Client Error)` status code, the client SHOULD NOT attempt to retry or resume the upload, unless the semantics of the response allow or recommend the client to retry the request.
-- `5xx (Server Error)` status code or no final response at all due to connectivity issues, the client MAY retry retrieving the offset.
+- `5xx (Server Error)` status code or no final response at all due to connectivity issues, the client SHOULD retry retrieving the offset.
 
 ### Server Behavior {#offset-retrieving-server}
 
-A successful response to a `HEAD` request against an upload resource
+Upon receiving a `HEAD` request, a server MAY treat it as a hint to abort a previous, possibly stalled, upload creation or append request for the same resource. See {{concurrency}} for more details on handling concurrency.
```
Acconut: I think this paragraph is a good addition, but should be added a bit later in the section and not be the first.

draft-ietf-httpbis-resumable-upload.md
smatsson: I agree with Acconut here. To me approach 2 and 3 seem very similar in how they would work in practice. The server would receive two requests with the same offset (why though, really?) and one is slightly faster than the other i.e. it will get the exclusive lock and the thus the second request would fail. I think the wording of point 2 hints that the requests are queued and processed after each other which I guess is not the case (the offset would most likely never match for the second request)

draft-ietf-httpbis-resumable-upload.md
danielresnick: Done

draft-ietf-httpbis-resumable-upload.md
danielresnick: I had a go at rewording this to be clearer as it was both too vague and overly specific in different areas - is this any better?  The key distinction between 2 & 3 is that servers don't need to acquire an exclusive lock for the resource which means they're able to achieve higher availability during network partitions

smatsson: Yes much better, thanks!

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -686,9 +687,15 @@ Resumable uploads, as defined in this document, do not permit uploading represen
...
+To meet these requirements, a server can use various strategies. Three common approaches are:
+
+1.  **Preemptive Cancellation**: If the server receives a new request for an upload resource while a previous request for the same resource is in-flight, it abruptly terminates the previous HTTP connection or stream before processing the new request.
+
+2.  **Pessimistic Locking**: The server processes requests for a given upload resource sequentially, effectively creating an exclusive lock on that resource. A new request is only processed after the previous one completes. This can be simpler to implement but may lead to delays if a request hangs from the server's perspective.
+
+3.  **Optimistic Concurrency Control**: The server detects concurrent modifications by atomically checking the resource's state before applying changes. If a conflict is detected, the server rejects the request with a 409 (Conflict) status code, ensuring that only one of multiple parallel requests can succeed.
```
Acconut, line 699: I'm wondering how this would be achieved in practice on the server-side. Does the server read the entire request content first and then tries to apply it in one atomic operation that checks the offset and appends the content? The offset would then be validated after receiving the request content, not when the headers are sent. Not a problem, but a different angle to the task.

Small suggestion:

```suggestion
3.  **Optimistic Concurrency Control**: The server handles concurrent modifications by checking the resource's state and applying changes together in one atomic operation. If a conflict is detected, the server rejects the request with a 409 (Conflict) status code, ensuring that only one of multiple parallel requests can succeed. Upload cancellation is implemented through a similar atomic operation.
```

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -686,9 +687,15 @@ Resumable uploads, as defined in this document, do not permit uploading represen
...
+Servers SHOULD choose a strategy that best fits their architecture while fulfilling the requirements of this section. Regardless of the chosen strategy, clients SHOULD be prepared to handle a `409 (Conflict)` response as a recoverable error, as described in {{upload-appending}}.
```
Acconut, line 701: I think this paragraph isn't necessary as it restates normative language that appeared previously in the section without adding new information.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -686,9 +687,15 @@ Resumable uploads, as defined in this document, do not permit uploading represen
...
+To meet these requirements, a server can use various strategies. Three common approaches are:
+
+1.  **Preemptive Cancellation**: If the server receives a new request for an upload resource while a previous request for the same resource is in-flight, it abruptly terminates the previous HTTP connection or stream before processing the new request.
```
Acconut, line 695: I think we should keep more context from the original wording that explains why this works:

> Since the client is not allowed to perform multiple transfers in parallel, the upload resource can assume that the previous attempt has already failed. Therefore, the server MAY abruptly terminate the previous HTTP connection or stream.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -686,9 +687,15 @@ Resumable uploads, as defined in this document, do not permit uploading represen
...
+
+2.  **Pessimistic Locking**: The server processes requests for a given upload resource sequentially, effectively creating an exclusive lock on that resource. A new request is only processed after the previous one completes. This can be simpler to implement but may lead to delays if a request hangs from the server's perspective.
```
Acconut, line 697: Similarly, the text would benefit from emphasizing the downsides of this (simple) approach:

> However, since network interruptions can occur in ways that cause the request to hang from the server's perspective, it might take the server significant time to realize the interruption and time out the request. During this period, the client will be unable to access the resource and resume the upload, causing friction for the end users.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -578,10 +580,9 @@ If the client received a final response with the `Upload-Complete: ?0` header fi
 
 - A `2xx (Successful)` status code indicates that representation data was appended but the upload is not complete. The client can continue appending representation data.
 - For a `307 (Temporary Redirect)` or `308 (Permanent Redirect)` status code, the client MAY retry appending to the new URI.
-- For a `4xx (Client Error)` status code, the client SHOULD NOT attempt to retry or resume the upload, unless the semantics of the response allow or recommend the client to retry the request.
-- For a `5xx (Server Error)` status code, the client MAY automatically attempt upload resumption by retrieving the current offset ({{offset-retrieving}}).
-
-If no final response was received at all due to connectivity issues, the client MAY automatically attempt upload resumption by retrieving the current offset ({{offset-retrieving}}).
+- For a `409 (Conflict)` status code, the client SHOULD attempt to resume the upload by either retrieving the current offset ({{offset-retrieving}}) or using the value from the response's `Upload-Offset` header field.
```
GrantGryczan, line 585: See my concern here: https://github.com/httpwg/http-extensions/issues/3275#issuecomment-3990724220

But I think that's an underlying design concern rather than a concern with your specific change towards that design, so feel free to resolve this as off-topic.

draft-ietf-httpbis-resumable-upload.md
danielresnick: Yeah good point, removed

Acconut: Did you try to push changes to this PR? I'm not seeing any yet :)

draft-ietf-httpbis-resumable-upload.md
Acconut: I think it's sufficient to tie the retries to the presence of `Upload-Complete: ?0` in this case. It would indicate that the 409 comes from the resumable upload layer.

```suggestion
- For a `409 (Conflict)` status code with the `Upload-Complete: ?0` header field, the client SHOULD attempt to resume the upload by either retrieving the current offset ({{offset-retrieving}}) or using the value from the response's `Upload-Offset` header field.
```

guoye-zhang:
Let's update this to the guidelines discussed during the working group:

    Upload-Complete: ?1 -> do not resume
    409 (Conflict) + Upload-Complete: ?0 -> resume
    413 (Content Too Large) -> resume with smaller chunk
    429 (Too Many Requests) -> resume with delay
    5xx -> resume

GrantGryczan:
Hi @danielresnick, it's been a few months since the last update--are you still interested in working on this? If not, let us know so the draft authors can finish the work, or so that others (possibly myself) can start work on a new PR implementing these changes. I'm very interested in getting these changes implemented! :)

danielresnick:
Hey @GrantGryczan - my sincere apologies, I've been slammed at work and haven't had a chance to progress this, and unfortunately likely won't have bandwidth in the next few months either. If you're able to pick things up that'd be great.

GrantGryczan:
All good, thanks for letting us know!

guoye-zhang:
I opened #3472

draft-ietf-httpbis-resumable-upload.md
GrantGryczan: I personally don't see the point in approach 3 at all. It seems like a bad idea to prioritize an older request over a newer one by rejecting the newer one, considering the client may no longer know about the older one, and this restricts the client from controlling it.

GrantGryczan (changes requested):

Acconut:
Thank you for opening this PR and contributing to the draft, @danielresnick. The work on this topic continues in https://github.com/httpwg/http-extensions/pull/3472, so I'll close this PR.

PR #3320, opened 2025-10-30
resumable: introduce new append-granularity limit

danielresnick:
See discussion on #3193

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -356,6 +356,9 @@ The following key-value pairs are defined:
 `min-append-size`:
 : Specifies a minimum size counted in bytes for the request content in a single upload append request ({{upload-appending}}). The server might reject requests below this limit. A client that is aware of this limit MUST NOT send smaller upload append requests. The value is an Integer. Requests completing the upload by including the `Upload-Complete: ?1` header field are exempt from this limit.
 
+`append-granularity`:
+: Specifies that the size of request content in bytes SHOULD be an integer multiple of this value for an upload append request ({{upload-appending}}), and for an upload creation request ({{upload-creation}}) that contains content. The server might reject requests that do not adhere to this limit (see {{inconsistent-granularity}}). A client that is aware of this limit MUST adhere to this requirement. The value is an Integer. Requests completing the upload by including the `Upload-Complete: ?1` header field are exempt from this limit.
+
 `max-age`:
 : Specifies the remaining lifetime of the upload resource in seconds counted from the generation of the response. After the resource's lifetime is reached, the server might make the upload resource inaccessible and a client SHOULD NOT attempt to access the upload resource as these requests will likely fail. The value is an Integer.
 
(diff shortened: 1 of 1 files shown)


gstrauss:
@danielresnick There should be an exception for the size of the final chunk at the end of the file.

Requiring upload-complete be set for the final chunk would take away the ability of a robust client to send chunks, and then *separately* without any additional body content, to send Upload-Complete.  The ability to send a separate message separates the uploading from marking the upload as complete.

The authors of resumable uploads have made many premature optimizations and I hope that they do not downplay or ignore the robustness of separation of concerns when evaluating this append-granularity feature.

---

Personally, I would prefer that the granularity be a hint, and not a hard error, especially since according to the resumable uploads draft, a server *may* accept part of an upload even if the upload gets disconnected.  If there was a *part* of the granularity chunk committed and the offset is no longer a multiple of the granularity, then a client might wish to send a smaller chunk (smaller than the granularity) even in the middle of an upload in order to re-align with the granularity for further chunks.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -590,6 +593,8 @@ The server might not receive the entire patch document when the upload is interr
 
 If the `Upload-Offset` request header field value does not match the current offset ({{upload-offset}}), the upload resource MUST reject the request with a `409 (Conflict)` status code. The response MUST include the correct offset in the `Upload-Offset` header field. The response can use the problem type {{PROBLEM}} of "https://iana.org/assignments/http-problem-types#mismatching-upload-offset" ({{mismatching-offset}}).
 
+If the request content size violates a server-advertised `append-granularity` limit (see {{upload-limit}}), the server MAY process and accept a portion of the request content that adheres to the granularity, discarding the remainder. In this case, the server SHOULD respond with a `400 (Bad Request)` status code and use the problem type "https://iana.org/assignments/http-problem-types#inconsistent-upload-granularity" ({{inconsistent-granularity}}). The response MUST include an `Upload-Offset` header field indicating the new offset after accepting the partial content, allowing the client to recover and resume the upload.
```
Acconut, line 596: How should a client act when receiving this response? The document currently recommends client to stop the upload once they receive a 400.

danielresnick: Technically the spec says "4xx (Client Error) status code, the client SHOULD NOT attempt to retry or resume the upload, _unless the semantics of the response allow or recommend the client to retry the request_."

I agree that's quite subtle though -- expecting a client dev to read this, and then trawl through all possible response_code+problem_type combinations to decide on resume vs abort behaviour. Particularly because even parsing out the problem type requires doing some json parsing. I'll have a think about the best path forward here.. the most obvious options in my mind are (in no particular order):

1) Switch to a 5xx response code. As discussed on the corresponding issue this is likely to improve client interop, but is a hacky / less pure solution since this isn't actually a server error.

2) Use a different / more specific 4xx response code and improve the wording to call this code out specifically. The problem is I don't see a super obvious pre-existing code in the 4xx space. Maybe 422 Unprocessable Content, but that feels slightly odd vs the suggestion that a server should partially process the provided response.

3) Keep using 400 (Bad Request) and try to improving the wording in the client section to be clearer (e.g. with some combination of improving cross-referencing with problem-types, marking problem-types more explicitly as retryable vs not, rephrasing the 4xx status code handling piece quoted above to lead with a discussion on what's retryable vs not).

Acconut: I think 409 would be an option. If https://github.com/httpwg/http-extensions/pull/3319 lands, clients should resume the upload when they get a 409, which I think it desired in that case.

> 3\. Keep using 400 (Bad Request) and try to improving the wording in the client section to be clearer (e.g. with some combination of improving cross-referencing with problem-types, marking problem-types more explicitly as retryable vs not, rephrasing the 4xx status code handling piece quoted above to lead with a discussion on what's retryable vs not).

I wouldn't want to mandate clients to parse the response body. It would be great if we can keep using status codes as the definitive signal for clients.

danielresnick: Yeah that's a good point. I'd sort of dismissed that option as I was finding it hard to justify granularity mismatch as a state conflict (though I agree that if that PR lands it would solve the problem from a pure behaviour standpoint), but thinking some more I guess an argument could be made that partial acceptance of a chunk (e.g., accepting 256 KiB of a 300 KiB request) creates a direct mismatch between the client's perceived state and the server's actual state. And yes agreed leveraging status codes only is much more robust & likely to get adoption in practice

PR #3327, opened 2025-11-12
RFC6265bis: Address Dnsdir issues

sbingler:
The DNS Directorate review, https://lists.w3.org/Archives/Public/ietf-http-wg/2025JanMar/0140.html, surfaced a number of issues.

Primarily this PR:

- Adds a new name resolution section to clarify the spec's stance on name resolution (ASCII only, DNS default)
- Tightens the canonicalization sub-algorithm to only operate on the expected label types
- Adds a new PSL security consideration section


draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -333,6 +332,16 @@ cookies, such as a web browser API that exposes cookies to scripts.
 The term "top-level navigation" refers to a navigation of a top-level
 traversable.
 
+## Name Resolution System
```
sbingler, line 335: This currently lives in the "Conventions" section, maybe "Overview" would be a better place for it?

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1020,12 +1028,20 @@ A canonicalized host name is the string generated by the following algorithm:
 
 1.  Convert the host name to a sequence of individual domain name labels.
 
-2.  Convert each label that is not a Non-Reserved LDH (NR-LDH) label, to an
-    A-label (see {{Section 2.3.2.1 of RFC5890}} for the former and latter).
+2. If any label is not one of U-label, XN-label, or Non-Reserved LDH (NR-LDH)
+   label (see {{Section 2.3.1 of RFC5890}}) then abort this algorithm and fail
+   to canonicalize the host name.
```
sbingler: The pitfalls this change avoids is if, for example, an input is a reserved label that isn't an xn-label or if it's a non-LDH label (e.x.: "_tcp"). The previous version would attempt to process those into a-labels which isn't possible.

I found the diagram on RFC5890 pg 10 very helpful.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1657,8 +1673,12 @@ user agent MUST process the cookie as follows:
 9.  If the user agent is configured to reject "public suffixes" and the
     domain-attribute is a public suffix:
 
-    1.  If the domain-attribute is identical to the canonicalized
-        request-host:
+    1. Let request-host-canonical be the canonicalized request-host.
+
+    2. If request-host fails to be canonicalized then abort this algorithm and
```
sbingler, line 1679: I suppose if the request-host fails to canonicalize we have bigger issue somewhere, but failing the cookies (repeatedly, for every cookie) seems fine.

draft-ietf-httpbis-rfc6265bis.md
mikewest: I think it's fine here. If you'd prefer it elsewhere, go for it, but this seems perfectly reasonable to me.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -706,11 +715,10 @@ user agent).
 
 The Domain attribute specifies those hosts to which the cookie will be sent.
 For example, if the value of the Domain attribute is "site.example", the user
-agent will include the cookie in the Cookie header field when making HTTP requests to
-site.example, www.site.example, and www.corp.site.example. (Note that a
-leading %x2E ("."), if present, is ignored even though that character is not
-permitted.)  If the server omits the Domain attribute, the user agent
```
mikewest, line 712: Is this not the case anymore? I didn't see a corresponding algorithmic change, but I might have missed it.

draft-ietf-httpbis-rfc6265bis.md
mikewest: Nit: "Not a ... Non-Reserved" is confusing, but I don't have a way of rephrasing this that isn't equally obtuse. :/

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1020,12 +1028,20 @@ A canonicalized host name is the string generated by the following algorithm:
 
 1.  Convert the host name to a sequence of individual domain name labels.
 
-2.  Convert each label that is not a Non-Reserved LDH (NR-LDH) label, to an
-    A-label (see {{Section 2.3.2.1 of RFC5890}} for the former and latter).
+2. If any label is not one of U-label, XN-label, or Non-Reserved LDH (NR-LDH)
+   label (see {{Section 2.3.1 of RFC5890}}) then abort this algorithm and fail
+   to canonicalize the host name.
+
+3.  Convert each U-label to an A-label (see {{Section 2.3.2.1 of RFC5890}}).
```
mikewest, line 1036: Should this include XN-label conversions?

draft-ietf-httpbis-rfc6265bis.md
mikewest: Failing sounds reasonable, but I likewise don't know how that would happen if we've gotten here.

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1671,8 +1691,8 @@ user agent MUST process the cookie as follows:
 
 10. If the domain-attribute is non-empty:
 
-    1.  If the canonicalized request-host does not domain-match the
-        domain-attribute:
+    1.  If request-host-canonical does not domain-match
+        (See {{domain-matching}}) the domain-attribute:
```
mikewest: Nit:

```suggestion
        (see {{domain-matching}}) the domain-attribute:
```

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1720,8 +1740,8 @@ user agent MUST process the cookie as follows:
 
     2.  Their secure-only-flag is true.
 
-    3.  Their domain domain-matches the domain of the newly-created cookie, or
-        vice-versa.
+    3.  Their domain domain-matches (See {{domain-matching}}) the domain of the
```
mikewest: Nit:

```suggestion
    3.  Their domain domain-matches (see {{domain-matching}}) the domain of the
```

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1895,18 +1915,23 @@ is "non-HTTP".
...
-         host of the retrieval's URI is identical to the cookie's domain.
+     *   The cookie's host-only-flag is true and retrieval-host-canonical is
+         identical to the cookie's domain.
 
      Or:
 
-     *   The cookie's host-only-flag is false and the canonicalized
-         host of the retrieval's URI domain-matches the cookie's domain.
+     *   The cookie's host-only-flag is false and retrieval-host-canonical
+         domain-matches (See {{domain-matching}}) the cookie's domain.
```
mikewest: Nit:

```suggestion
         domain-matches (see {{domain-matching}}) the cookie's domain.
```

draft-ietf-httpbis-rfc6265bis.md
sbingler: SGTM

draft-ietf-httpbis-rfc6265bis.md
sbingler: This is still the case, I'm just removing the mention of from section 4 which should be describing the ideal server behavior. It seems strange to mention non-ideal behavior here.

The note still exists in the UA section 5.7 Step 7.1

draft-ietf-httpbis-rfc6265bis.md
sbingler: Agreed. I think it's understandable after reading it once or twice (or thrice).

draft-ietf-httpbis-rfc6265bis.md
sbingler: An xn-label is already in ascii so there's no need to convert it. 

* U-label: café.com
* XN-label: xn--caf-dma.com

draft-ietf-httpbis-rfc6265bis.md
sbingler: Yeah, seemed like a good place for a `NOTREACHED()` but worth calling out nonetheless.

draft-ietf-httpbis-rfc6265bis.md
sbingler: Done

draft-ietf-httpbis-rfc6265bis.md
sbingler: Done

draft-ietf-httpbis-rfc6265bis.md
sbingler: Done

draft-ietf-httpbis-rfc6265bis.md
mikewest: Got it, thanks.

draft-ietf-httpbis-rfc6265bis.md
mikewest: Understood, thank you.

draft-ietf-httpbis-rfc6265bis.md
pspacek: Perhaps easier to digest would be something like:
Only acceptable label types are: U-label, XN-label, or Non-Reserved LDH (NR-LDH) label (see {{Section 2.3.1 of RFC5890}}). Abort this algorithm and fail to canonicalize the host name if any other label type is detected.

draft-ietf-httpbis-rfc6265bis.md
pspacek, line 1058: Perhaps this would be easier to understand?
```suggestion
    *   The last character of the prefix that is not included in the domain
```

pspacek: Sorry, apparently I'm unable to use Github suggestions 🤦‍♂️

draft-ietf-httpbis-rfc6265bis.md
pspacek: FTR it is important for security properties of domain-matching algorithm that NON-LDH labels are not permitted. Doing so would allow crazy domain names like
`something\..example` where `\.` is a literal dot which is _part of_ `something\.` label.

sbingler: I went with a slightly different phrasing

draft-ietf-httpbis-rfc6265bis.md
sbingler: I'm still slightly partial to the existing version.

sbingler:
@mikewest Ping

draft-ietf-httpbis-rfc6265bis.md
```diff
@@ -1020,12 +1029,20 @@ A canonicalized host name is the string generated by the following algorithm:
 
 1.  Convert the host name to a sequence of individual domain name labels.
 
-2.  Convert each label that is not a Non-Reserved LDH (NR-LDH) label, to an
-    A-label (see {{Section 2.3.2.1 of RFC5890}} for the former and latter).
+2. All labels must be one of U-label, XN-label, or Non-Reserved LDH (NR-LDH)
```
pspacek: Together with the step 4, this is effectively 
```suggestion
2. All labels must be one of U-label, A-label, or Non-Reserved LDH (NR-LDH)
```
isn't it?

pspacek: RFC5890 section 2.3.1 says:
> ... For
>   IDNA-aware systems, the valid label types are: A-labels, U-labels,
>   and NR-LDH labels. ...

sbingler: Yep, that seems both correct and a better phrasing.

PR #3330, opened 2025-11-15
[layered-cookies] Add acknowledgements for layered-cookies-01

johannhof:


draft-ietf-httpbis-layered-cookies.md
```diff
@@ -1899,3 +1899,5 @@ Revamped the document to allow for more detailed requirements on browsers in dow
 Many thanks to Adam Barth for laying the groundwork for a modern cookie specification with RFC 6265.
 
 And thanks to John Wilander, Lily Chen, Mike West, Steven Bingler, and Steven Englehardt for improving upon that work in subsequent drafts.
+
+Further thanks goes to Dylan Cutler and Yoav Weiss (in alphabetical order) for their substantial contributions to this document.
```
annevk:
```suggestion
And many thanks to Steven Bingler, Lily Chen, Dylan Cutler, Steven Englehardt, Yoav Weiss, Mike West, and John Wilander for improving upon that work.
```

annevk (commented):
I suggest we try to keep it a bit more concise.

draft-ietf-httpbis-layered-cookies.md
annevk: (I think IETF usually sorts by last name, but I would be happy if we decided to always sort by first name instead.)

johannhof: I don't have a strong opinion on that one, so will take this as-is for now.

PR #3342, opened 2025-12-03
Add note about possible content coding reversal edge cases

### LINKED ISSUE #3339: Reversablilty of encoding wrt unencoded-digest
During WGLC @martinthomson noted:

> There is one thing that occurred to me reading this, which might be worth noting.  This depends on having codings that produce exactly one sequence of bytes when decoded.
>
> RFC 9110 says that a content coding transforms the content "without losing the identity of its underlying media type and without loss of information".  I can think of any number of codings that would not always transform back into the same sequence of bytes.

This is an astute observation. I think we can capture some high-level consiseration(s) to help implementers understand such cases. Even if the likelihood of a reversible yet non-identical-output encoding is niche today.

LPardue:
Fixes #3339

### DIFF
--- a/draft-ietf-httpbis-unencoded-digest.md
+++ b/draft-ietf-httpbis-unencoded-digest.md
@@ -229,7 +229,7 @@ Since the digest is calculated on unencoded representation bytes, validation of
 a message with content coding (as described above) can only succeed where the
 decoded output produces the same byte sequence as the input. While many
 registered content codings behave this way, there is no requirement for them to
-do so and it remains a possibility that a reverse coding could produce a
+do so and it remains a possibility that decoding could produce a
 different byte sequence. In order to avoid unintended validation failures, care
 is advised when selecting content coding for use with `Unencoded-Digest`.
 


draft-ietf-httpbis-unencoded-digest.md
```diff
@@ -219,11 +219,19 @@ Want-Unencoded-Digest: sha-512=3, sha-256=10, unixsum=0
...
-Digests delivered through `Unencoded-Digest` apply to the unencoded representation. If a message is
-received with content coding, a recipient needs to decode the message in order
-to calculate the digest that can subsequently be used for validation. If
-multiple content codings are applied, the recipient needs to decode all
-encodings in order before validation.
+Digests delivered through `Unencoded-Digest` apply to the unencoded
+representation. If a message is received with content coding, a recipient needs
+to decode the message in order to calculate the digest that can subsequently be
+used for validation. If multiple content codings are applied, the recipient
+needs to decode all encodings in order before validation.
```
LPardue, line 226: Changes here are just line-wrapping. No change in content.

draft-ietf-httpbis-unencoded-digest.md
```diff
@@ -219,11 +219,19 @@ Want-Unencoded-Digest: sha-512=3, sha-256=10, unixsum=0
...
+
+Since the digest is calculated on unencoded representation bytes, validation of
+a message with content coding (as described above) can only succeed where the
+decoded output produces the same byte sequence as the input. While many
+registered content codings behave this way, there is no requirement for them to
+do so and it remains a possibility that a reverse coding could produce a
```
martinthomson:
```suggestion
do so and it remains a possibility that decoding could produce a
```

draft-ietf-httpbis-unencoded-digest.md
ioggstream: > If multiple content codings are applied, the recipient
needs to decode all encodings in order before validation.

Isn't this implied by https://datatracker.ietf.org/doc/html/rfc9110#section-8.4-5 ?

I'd remove this one and leave the next paragraph.

LPardue: This text was added to address Roy's feedback that a decoder might overlook doing the right thing. So I'll keep it.

PR #3347, opened 2025-12-03
Add security consideration about encrypted encoding leakage

### LINKED ISSUE #3346: Unencoded digests can be dangerous with encrypted content codings
As part of WGLC review, @ioggstream highlighted that Unencoded-Digest can leak information about an encrypted content coding. This seems like something vital to warn about in the document

LPardue:
This takes some text proposed by @ioggstream and tweaks it a bit

Fixes #3346

### DIFF
--- a/draft-ietf-httpbis-unencoded-digest.md
+++ b/draft-ietf-httpbis-unencoded-digest.md
@@ -329,7 +329,7 @@ presents several risks, {{Sections 6.1, 6.2 and 6.3 of DIGEST-FIELDS}}
 describe relevant considerations and mitigations.
 
 A content coding may provide encryption capabilities, for example "aes128gcm"
-({{?RFC8188}}. Using Unencoded-Digest with such content codings can leak
+({{?RFC8188}}). Using Unencoded-Digest with such content codings can leak
 information about the original data because header fields are visible to anyone
 who can read the HTTP message. This could be used as a side channel.
 


ioggstream (approved):
LGTM

draft-ietf-httpbis-unencoded-digest.md
```diff
@@ -327,6 +328,11 @@ processing. An attacker that can substitute various parts of an HTTP message
 presents several risks, {{Sections 6.1, 6.2 and 6.3 of DIGEST-FIELDS}}
 describe relevant considerations and mitigations.
 
+A content coding may provide encryption capabilities, for example "aes128gcm"
+({{?RFC8188}}. Using Unencoded-Digest with such content codings can leak
```
mikewest: Nit: Either this, or drop the `(`. :)

```suggestion
({{?RFC8188}}). Using Unencoded-Digest with such content codings can leak
```

draft-ietf-httpbis-unencoded-digest.md
```diff
@@ -327,6 +328,11 @@ processing. An attacker that can substitute various parts of an HTTP message
...
+information about the original data because header fields are visible to anyone
+who can read the HTTP message. This could be used as a side channel.
```
mikewest, line 334: Nit: Not sure the last sentence is necessary. 🤷

mikewest (approved):
LGTM % nits. The first should be addressed one way or the other, I'll defer entirely to you on the latter.

draft-ietf-httpbis-unencoded-digest.md
LPardue: oops, thanks!

draft-ietf-httpbis-unencoded-digest.md
LPardue: I was channeling some of the text in https://www.rfc-editor.org/rfc/rfc8188.html#section-4.6

> Because only the payload body is encrypted, information exposed in
   header fields is visible to anyone who can read the HTTP message.
   This could expose side-channel information.
   
  If we think the message is clear enough without the last sentence, happy to drop it, I don't have a strong opinion either way

mikewest: Like I said, I'll defer to you. Parallelism with RFC8188 is good enough justification for me.

LPardue: Im lazy, so for an easy life it's easier to delete it later than add it back, will merge this as is. Thanks!

PR #3348, opened 2025-12-03
Editorial suggestions from Roberto Polli

LPardue:
Most (but not all :smile: ) suggestions from @ioggstream. Some of these triggered memories from RFC editor feedback on 9530 so applying now should ensure some good consistency

- **Add some imports in the conventions and definitions**
- **omit one -> omit the field**
- **gzip example wording**
- **hashing of entire selected representation data**
- **sprinkle some s/coding/codings around**
- **s/body/content**
- **Improve HTTP in examples to avoid a call from the protocol police**

### DIFF
--- a/draft-ietf-httpbis-unencoded-digest.md
+++ b/draft-ietf-httpbis-unencoded-digest.md
@@ -243,10 +243,10 @@ is advised when selecting content coding for use with `Unencoded-Digest`.
 Integrity fields can be used in combination to address different and
 complementary needs, particularly the cases described in {{introduction}}.
 
-In the following examples, the unencoded response data is the string "An
-unexceptional string" following by an LF. For presentation purposes, the
-response content is displayed as a sequence of hex-encoded bytes because it
-contains non-printable characters.
+In the following examples, the selected representation data with no content
+codings applied is: "An unexceptional string" following by an LF. For
+presentation purposes, the response content is displayed as a sequence of
+hex-encoded bytes because it contains non-printable characters.
 
 The first example demonstrates a request that uses content negotiation.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-unencoded-digest.md
```diff
@@ -119,16 +119,21 @@ The definitions "representation", "selected representation", "representation
 data", "representation metadata", and "content" in this document are to be
 interpreted as described in {{!HTTP=RFC9110}}.
 
+This document uses the line folding strategies described in {{!FOLDING=RFC8792}}.
+
+The term "digest" is to be interpreted as described in {{DIGEST-FIELDS}}.
```
mikewest, line 124: I think this is accurate, but I'm not sure it's super helpful. RFC9530 says "[Senders choose a hashing algorithm and calculate a digest from an input related to the HTTP message. The algorithm identifier and digest are transmitted in an HTTP field.](https://www.rfc-editor.org/rfc/rfc9530.html#:~:text=Senders%20choose%20a%20hashing%20algorithm%20and%20calculate%20a%20digest%20from%20an%20input%20related%20to%20the%20HTTP%20message.%20The%20algorithm%20identifier%20and%20digest%20are%20transmitted%20in%20an%20HTTP%20field.)", and then just continues referring to "digest" throughout. If you think this pre-addresses editor feedback, I don't think it's harmful to include, but it doesn't seem like it's doing much work to me.

mikewest (approved):
LGTM. This all seems reasonable to me % one line. As before, I'm fine with it if you think it's helpful. It's just not clear to me what work it's doing.

draft-ietf-httpbis-unencoded-digest.md
LPardue: My read of Roberto's suggestion was to import "digest" as stated in https://www.rfc-editor.org/rfc/rfc9530.html#section-1.4-8.

IIRC we had a bit of discussion during 9530 on these terms, hence landing on that phrasing. Im happy to keep it, even if it seems a bit obvious

mikewest: I'm fine with the text you've proposed. Go ahead and land it like this. :)

ioggstream: Agree with Lucas: let's remember to reuse the definitions when updating 9530 :)

PR #3349, opened 2025-12-08
Julian's ARTART Review

martinthomson:
Two changes: 

* mention that buffering needs to be limited in time, not bytes; and
* add hyperlinks to the two registries that we're adding to.

### DIFF
--- a/draft-ietf-httpbis-incremental.md
+++ b/draft-ietf-httpbis-incremental.md
@@ -217,8 +217,14 @@ accompanied by a connection_limit_reached Proxy-Status response header field
 For performance and efficiency reasons, a small amount of buffering might be
 used by intermediaries, even for incremental messages. Immediate forwarding
 might be exploited to cause an intermediary to waste effort on many small
-packets.  Enabling incremental delivery might instead set limits on the number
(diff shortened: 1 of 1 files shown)


kazuho (commented):
Thank you for the PR.

I really like how the text is changed.

kazuho (approved):

reschke (approved):

draft-ietf-httpbis-incremental.md
```diff
@@ -217,8 +217,14 @@ accompanied by a connection_limit_reached Proxy-Status response header field
 For performance and efficiency reasons, a small amount of buffering might be
 used by intermediaries, even for incremental messages. Immediate forwarding
 might be exploited to cause an intermediary to waste effort on many small
-packets.  Enabling incremental delivery might instead set limits on the number
-bytes that are buffered or the time that buffers are held before forwarding.
+packets.
+
+Enabling incremental delivery might set limits on the number
+bytes that are buffered before forwarding.
```
MikeBishop, line 223:
```suggestion
of bytes that are buffered before forwarding.
```
Error in the existing text, but as long as you're touching the line....

MikeBishop (approved):

PR #3367, opened 2025-12-11
Add some AD suggested wording to content coding decoding discussion

### LINKED ISSUE #3356: Section 5, paragraph 2
_Comment by @MikeBishop_

This is a subtle distinction; thank you for calling it out. A content-coding is defined to operate "without loss of information", but that doesn't necessarily mean byte-for-byte equivalence.

LPardue:
I found this a little difficult to integrate in and hence its quite a mouthful of some sentences. However, I think it captures what was described in the issue.

Fixes #3356

### DIFF
--- a/draft-ietf-httpbis-unencoded-digest.md
+++ b/draft-ietf-httpbis-unencoded-digest.md
@@ -243,7 +243,7 @@ content coding could perform semantically-meaningless
 transformations that nevertheless result in a decoded byte sequence that does
 not exactly match the original unencoded representation. In order to avoid
 unintended validation failures, care is advised when selecting content codings
-for use with `Unencoded-Digest`; many registered content codings do provide
+for use with `Unencoded-Digest`; that said, most registered content codings do provide
 byte-for-byte equivalence and are appropriate.
 
 


draft-ietf-httpbis-unencoded-digest.md
```diff
@@ -232,11 +232,13 @@ needs to decode all encodings in order before validation.
...
-do so and it remains a possibility that decoding could produce a
-different byte sequence. In order to avoid unintended validation failures, care
-is advised when selecting content coding for use with `Unencoded-Digest`.
+decoded output produces the same byte sequence as the input. While {{Section
+8.4.1 of !HTTP=RFC9110}} describes content codings to operate "without loss of
+information", that doesn't necessarily mean a byte-for-byte equivalence. Many
+registered content codings do provide equivalence but there is no requirement
+for it; it remains a possibility that decoding could produce a different byte
+sequence. In order to avoid unintended validation failures, care is advised when
+selecting content coding for use with `Unencoded-Digest`.
```
mikewest: WDYT about this suggestion?

```suggestion
information", that doesn't necessarily mean a byte-for-byte equivalence. It's
entirely possible for content codings to perform semantically-meaningless
transformations that nevertheless result in a decoded byte sequence that does
not exactly match the original unencoded representation.In order to avoid
unintended validation failures, care is advised when selecting content coding
for use with `Unencoded-Digest`.
```

LPardue: I worry that this maybe makes the risk sound too large. I think Julian identified one content coding that behaves that way, and I don't think its widely used that. I hope that implementers of the spec can just assume things like gzip, brotli etc can be used without issues. I'll make a hybrid change that incorporates some of your proposed text.

LPardue: PTAL a look at the latest version after I pushed a commit.

draft-ietf-httpbis-unencoded-digest.md
```diff
@@ -236,11 +236,15 @@ needs to decode all encodings in order before validation.
...
 a message with content codings (as described above) can only succeed where the
-decoded output produces the same byte sequence as the input. While many
-registered content codings behave this way, there is no requirement for them to
-do so and it remains a possibility that decoding could produce a
-different byte sequence. In order to avoid unintended validation failures, care
-is advised when selecting content codings for use with `Unencoded-Digest`.
+decoded output produces the same byte sequence as the input. While {{Section
+8.4.1 of !HTTP=RFC9110}} describes content codings to operate "without loss of
+information", that doesn't necessarily mean a byte-for-byte equivalence. It's
+entirely possible for content codings to perform semantically-meaningless
```
MikeBishop:
```suggestion
information", that doesn't necessarily mean a byte-for-byte equivalence. A
content coding could perform semantically-meaningless
```

draft-ietf-httpbis-unencoded-digest.md
```diff
@@ -236,11 +236,15 @@ needs to decode all encodings in order before validation.
...
+transformations that nevertheless result in a decoded byte sequence that does
+not exactly match the original unencoded representation. In order to avoid
+unintended validation failures, care is advised when selecting content codings
+for use with `Unencoded-Digest`; many registered content codings do provide
+byte-for-byte equivalence and are appropriate.
```
MikeBishop:
```suggestion
for use with `Unencoded-Digest`; that said, most registered content codings do provide
byte-for-byte equivalence and are appropriate.
```

MikeBishop (commented):
Would this help to further scope the size of the risk? It's worth calling out for completeness, but in practice most will be byte-for-byte unless they're media-type-specific.

PR #3373, opened 2025-12-24
Add note about cache busting

### LINKED ISSUE #3276: no-vary: cache busting
This extension will break the "cache busting via query parameters" pattern. Should this be mentioned somewhere?

(https://lists.w3.org/Archives/Public/ietf-http-wg/2025OctDec/0001.html)

nidhijaju:
Fixes #3276.

### DIFF
--- a/draft-ietf-httpbis-no-vary-search.md
+++ b/draft-ietf-httpbis-no-vary-search.md
@@ -35,6 +35,7 @@ author:
     email: jbroman@chromium.org
 
 normative:
+  URI: RFC3986
   HTTP: RFC9110
   HTTP-CACHING: RFC9111
   FETCH:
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-no-vary-search.md
```diff
@@ -131,6 +131,8 @@ And if the resource instead wants to take an allowlist-based approach, where onl
 No-Vary-Search: params, except=("productId")
 ~~~~
 
+Note that common approaches for ensuring caches return non-cached responses by changing query parameters are often ineffective when combined with the No-Vary-Search header field.
```
jeremyroman: formatting: <code>\`No-Vary-Search\`</code>

jeremyroman (commented):
This note seems, to me, too vague to be helpful. If something like this is useful, I think it'd need to explain a bit more why and what can be done about it (e.g., excluding particular query parameters for cache-busting purposes, `Cache-Control: no-cache` from the client; I haven't thought about this in detail lately).

draft-ietf-httpbis-no-vary-search.md
reschke: Wait. They aren't ineffective, they don't work at all. Or am I missing something?

mnot: I think it depends on the specific nvs sent. Perhaps:


```suggestion
Note that "cache busting" by sending unique query parameters to avoid retrieving a cached response can be made ineffective by the `No-Vary-Search` header field.
```

reschke: Yes, but why not be clearer and say "will not work anymore?"

nidhijaju: Thanks for the suggestion!

> Yes, but why not be clearer and say "will not work anymore?"

As Mark was saying, I think it depends on the header. For example, if you use `except=x`, you can exclude the query parameters you use for cache busting, so not all uses of `No-Vary-Search` will cause it to not work anymore.

PR #3374, opened 2025-12-24
Update SF field format for params

### LINKED ISSUE #3287: no-vary: simplify SF field format
Section 3:

> If present, the params entry's value MUST be either a boolean (Section 
3.3.6 of [STRUCTURED-FIELDS]) or an inner list (Section 3.1.1 of 
[STRUCTURED-FIELDS]).

Maybe personal taste, but having varying datatypes for the same param feels wrong. Maybe use empty list instead 
of boolean.

That would change:

~~~
No-Vary-Search: params, except=("productId")
~~~

to

~~~
No-Vary-Search: params=(), except=("productId")
~~~

which would make things both more consistent and readable. Or even drop 
the "no params" syntax.

~~~
No-Vary-Search: except=("productId")
~~~

Also: this should say what format the contents of the inner list can have. (just stating it in the parsing algorithm is not sufficient)

(see https://lists.w3.org/Archives/Public/ietf-http-wg/2025OctDec/0000.html)

nidhijaju:
Fixes #3287.

### DIFF
--- a/draft-ietf-httpbis-no-vary-search.md
+++ b/draft-ietf-httpbis-no-vary-search.md
@@ -1,5 +1,5 @@
 ---
-title: The No-Vary-Search HTTP Response Header Field
+title: The No-Vary-Search HTTP Caching Extension
 abbrev: No-Vary-Search
 category: std
 
@@ -33,8 +33,14 @@ author:
     fullname: Jeremy Roman
     organization: Google LLC
     email: jbroman@chromium.org
+ -
+    fullname: Nidhi Jaju
+    role: editor
+    organization: Google LLC
+    email: nidhijaju@chromium.org
 
 normative:
+  URI: RFC3986
   HTTP: RFC9110
   HTTP-CACHING: RFC9111
   FETCH:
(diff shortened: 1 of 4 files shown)


draft-ietf-httpbis-no-vary-search.md
martinthomson, line 294: This function can return an error, but this algorithm doesn't handle that outcome properly.

draft-ietf-httpbis-no-vary-search.md
```diff
@@ -199,24 +199,17 @@ To _parse a URL search variance_ given _value_:
...
-    1. Otherwise, if _value_\["`params`"] is an array:
-        1. If any item in _value_\["`params`"] is not a string, then return the default URL search variance.
-        1. Set _result_'s no-vary params to the result of applying parse a key ({{parse-a-key}}) to each item in _value_\["`params`"].
-        1. Set _result_'s vary params to __wildcard__.
-    1. Otherwise, return the default URL search variance.
-1. If _value_\["`except`"] exists:
-    1. If _value_\["`params`"] is not true, then return the default URL search variance.
+    1. If _value_\["`params`"] is not an array, then return the default URL search variance.
+    1. If any item in _value_\["`params`"] is not a string, then return the default URL search variance.
+    1. Set _result_'s no-vary params to the result of applying parse a key ({{parse-a-key}}) to each item in _value_\["`params`"].
```
martinthomson, line 218: This should check for error (see below) and return the default variance if there is one.

martinthomson (approved):
I like this version much better.

draft-ietf-httpbis-no-vary-search.md
nidhijaju: Good catch, fixed!

Issue #3075, opened 2025-04-27
Search Variance


mnot:
This terminology is... odd, and at least for me makes it more difficult than necessary to read the spec. Have you considered alternatives? E.g.,

* Query cache key
* Varied cache key
* Varied URL

domenic:
It's not a cache key, for sure, since it's not used as such. The cache keys one might use would be as discussed in https://httpwg.org/http-extensions/draft-ietf-httpbis-no-vary-search.html#section-7, including all three of: the full URL, the URL with query parameters removed, and the URL simplified according to the search variance.

It's also not a URL, so that would be confusing.

If you have other ideas for identifying the struct in https://httpwg.org/http-extensions/draft-ietf-httpbis-no-vary-search.html#name-data-model, let us know...

mnot:
Yeah, this is tough. Will think on it more and come back if anything occurs.

mnot:
Nothing comes to mind quickly.

reschke:
"Query Component Nornalizer"?

PR #3379, opened 2026-01-14
Replace URL search variance with URL variation config

nidhijaju:
Out of all of the options suggested on the issue, this seemed workable so I've uploaded what it would look like. Any thoughts are appreciated!

Fixes https://github.com/httpwg/http-extensions/issues/3075.

### DIFF
--- a/draft-ietf-httpbis-no-vary-search.md
+++ b/draft-ietf-httpbis-no-vary-search.md
@@ -165,7 +165,7 @@ It has the following authoring conformance requirements:
 The dictionary MAY contain entries whose keys are not one of `key-order`, `params`, and `except`, but their meaning is not defined by this specification. Implementations of this specification will ignore such entries (but future documents might assign meaning to such entries).
 
 {:aside}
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-no-vary-search.md
```diff
@@ -165,11 +165,11 @@ It has the following authoring conformance requirements:
 The dictionary MAY contain entries whose keys are not one of `key-order`, `params`, and `except`, but their meaning is not defined by this specification. Implementations of this specification will ignore such entries (but future documents might assign meaning to such entries).
 
 {:aside}
-> As always, the authoring conformance requirements are not binding on implementations. Implementations instead need to implement the processing model given by the obtain a URL search variance algorithm ({{obtain-a-url-search-variance}}).
+> As always, the authoring conformance requirements are not binding on implementations. Implementations instead need to implement the processing model given by the obtain a URL variation config algorithm ({{obtain-a-url-variation-config}}).
```
tfpauly: Should "config" be written out as "configuration" at least the first time, and then introduce the short term?

nidhijaju: Done!

PR #3382, opened 2026-01-15
Update No-Vary-Search authors

nidhijaju:


reschke:
Also: @domenic should be listed as _**Contributor**_ then (separate section).

tfpauly:
@domenic it's totally fine to have you continue as author! The main thing I'm concerned about is availability for AUTH48 approvals (and to a lesser extent other reviews).

During AUTH48, the RFC editor will ask each author for approval. There are processes to handle this, and if you read the following FAQ it says:
https://www.rfc-editor.org/faq/#missingauthor

> One of the authors is no longer available; how do we proceed?
> We recommend one of the following paths:
> 
> 1. The author can be removed as an author and moved to the Acknowledgements section.
> 2. The author can be removed as an author and moved to the Contributors section.
> 3. A stream manager can approve the document in place of the unavailable author. (See the [IESG Statement on AUTH48 State](https://www.ietf.org/iesg/statement/auth48.html).)
> Option 3 is typically used in instances where the missing author made significant contributions to the document, so the other authors are not comfortable removing the individual from the author list

From the emails we received when you asked for others to take on the document, and hearing at least initially that you hadn't responded, the situation had looked like you wouldn't be available or willing to approve, etc. That could lead to being moved later, or needing override approvals. But if you'd like to stay on and just have @nidhijaju join to drive the rest of the process, that's great too.

tfpauly:
Sounds good! I think everyone preferred all along to keep the authors, it was clearly misunderstood about availability.

PR #3385, opened 2026-01-18
no-vary: fix title and intro terminology

reschke:


draft-ietf-httpbis-no-vary-search.md
```diff
@@ -111,16 +111,16 @@ informative:
...
 
-This specification defines a proposed HTTP response header field for changing how URI query parameters impact caching.
+This specification defines an extension to HTTP Caching, changing how URI query parameters impact caching.
 
 --- middle
 
 # Introduction
 
-HTTP caching {{HTTP-CACHING}} is based on reusing resources which match across a number of cache keys. One of the most prominent is the presented target URI ({{Section 7.1 of HTTP}}). However, sometimes multiple URIs can represent the same resource. This leads to caches not always being as helpful as they could be: if the cache contains a response under one URI, but the response is then requested under another, the cached version will be ignored.
+HTTP caching {{HTTP-CACHING}} is based on reusing resources which match across a number of cache keys, with the most important one being the presented target URI ({{Section 7.1 of HTTP}}). However, sometimes multiple URIs can represent the same resource. This leads to caches not always being as helpful as they could be: if the cache contains a response under one URI, but the response is then requested under another, the cached version will be ignored.
```
nidhijaju, line 120: I'm not sure we need to claim it's the most important:

```suggestion
HTTP caching {{HTTP-CACHING}} is based on reusing resources which match across a number of cache keys, with an important one being the presented target URI ({{Section 7.1 of HTTP}}). However, sometimes multiple URIs can represent the same resource. This leads to caches not always being as helpful as they could be: if the cache contains a response under one URI, but the response is then requested under another, the cached version will be ignored.
```

draft-ietf-httpbis-no-vary-search.md
```diff
@@ -111,16 +111,16 @@ informative:
...
 
-The `No-Vary-Search` HTTP header field tackles a specific subset of this general problem, for when different URIs that only differ in
-certain query parameters identify the same resource. It allows resources to declare that some or all parts of the query component do not semantically affect the served resource, and thus can be ignored for cache matching purposes. For example, if the order of the query parameters do not affect which resource is identified, this is indicated using
+The "No-Vary-Search" Caching Extension tackles a specific subset of this general problem, for when different URIs that only differ in
```
nidhijaju: We added a new sentence here saying "This specification defines an extension to HTTP caching, as permitted in {{Section 4 of HTTP-CACHING}}." in https://github.com/httpwg/http-extensions/pull/3383. Could we maybe remove that new sentence and instead bring the reference to Section 4 of HTTP-CACHING down below?

```suggestion
The "No-Vary-Search" response header field defines a caching extension, as described in {{Section 4 of HTTP-CACHING}}, that tackles a specific subset of this general problem, for when different URIs that only differ in
```

draft-ietf-httpbis-no-vary-search.md
reschke: Actually, in simple cases, it's the *only* one relevant. All other pieces of information (accept: etc) only augment the key.

draft-ietf-httpbis-no-vary-search.md
reschke: Yes, but "response header **_field_**".

nidhijaju: Updated the suggestion!

nidhijaju (approved):

PR #3390, opened 2026-02-26
Add params=?0 as an invalid example

nidhijaju:


draft-ietf-httpbis-no-vary-search.md
```diff
@@ -261,6 +261,7 @@ The following inputs are all invalid and will cause the default URL variation co
   * `No-Vary-Search: key-order="not a boolean"`
   * `No-Vary-Search: params="not an inner list"`
   * `No-Vary-Search: params=(not-a-string)`
+  * `No-Vary-Search: params=?0`
```
mnot, line 264: Is that the case though? I think the algorithm only checks for the parameter's presence, not its value...

> If value["params"] exists:[¶](https://httpwg.org/http-extensions/draft-ietf-httpbis-no-vary-search.html#section-5.1-2.6.1)
> If value["params"] is not an array, then return the default URL variation config.[¶](https://httpwg.org/http-extensions/draft-ietf-httpbis-no-vary-search.html#section-5.1-2.6.2.1.1)

nidhijaju: My understanding is that single boolean false is not a valid inner list, and thus 6.1 would say that the default URL variation config is returned, since the result of parsing params does not yield an array of items from the inner list.
Even if that isn't the case, I'd expect 6.2 to return the default URL variation config, since it would not be a string.

I could be missing something though?

mnot: Ah, never mind!

PR #3391, opened 2026-02-27
Advisory is not quite the right word

martinthomson:


draft-ietf-httpbis-incremental.md
```diff
@@ -85,7 +85,7 @@ To help avoid such behavior, this document specifies the "Incremental" HTTP head
 field, which requests that HTTP intermediaries begin forwarding the HTTP message
 downstream before receiving the complete message.
 
-This indication is advisory.
+This indication might not be supported.
```
tfpauly:
```suggestion
This indication might not be supported by intermediaries.
```

This is how you say it later, and I think it is clearer to specify who might not support it. Like, lots of parties won't support sending it, but that doesn't matter.

PR #3397, opened 2026-03-03
RESUMABLE: Fix typo "either ... and"

GrantGryczan:
It should be "either ... or", unless I'm somehow misreading this.


Acconut (approved):
Good catch, thank you!

PR #3400, opened 2026-03-20
Remove redundant requirements

### LINKED ISSUE #3269: Redundant requirements for secondary certs on a non-control stream
The draft states both

> The SERVER_CERTIFICATE frame MUST be sent on the control stream. A SERVER_CERTIFICATE frame received on any other stream MUST not be used for server authentication.

and 

> The SERVER_CERTIFICATE frame applies to the connection, not a specific stream. An endpoint MUST treat a SERVER_CERTIFICATE frame received on any stream other than the control stream as a connection error.

This seems a bit redundant. If you trigger a connection error, then you don't process the frame and so the "MUST not be used for server authentication" can't apply. Consider combining and condensing the requirement text.

ekinnear:
Remove redundant requirements
Fixes #3269

### DIFF
--- a/draft-ietf-httpbis-secondary-server-certs.md
+++ b/draft-ietf-httpbis-secondary-server-certs.md
@@ -269,7 +269,9 @@ the certificate chain and extensions used to create the message.
 A SERVER_CERTIFICATE frame in HTTP/2 (type=0xTBD) carrries a TLS Exported authenticator
 that clients can use to authenticate secondary origins from a sending server.
 
-The SERVER_CERTIFICATE frame MUST be sent on stream 0.
+The SERVER_CERTIFICATE frame applies to the connection, not a specific stream. It
+MUST be sent on stream 0. An endpoint MUST treat a SERVER_CERTIFICATE frame with a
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-secondary-server-certs.md
LPardue: What do you think about combining this paragraph with the line being modified in the PR? For example something aggregated up at line 306

"The SERVER_CERTIFICATE frame applies to the connection, not a specific stream. It MUST be sent on the control stream. Receiving a SERVER_CERTIFICATE frame  on any any stream MUST be treated as a connection error."

The same would apply to the H2 version

tfpauly (approved):

draft-ietf-httpbis-secondary-server-certs.md
ekinnear: Great idea! Moved up and consolidated in [91fd5cb](https://github.com/httpwg/http-extensions/pull/3400/commits/91fd5cb91f0f46d804f0f9f3804f20e36cd0b756).

LPardue (approved):
LGTM thanks

PR #3403, opened 2026-03-24
Fix digest examples

### LINKED ISSUE #3402: Unencoded-Digest: Unable to reproduce example
I tried to reproduce the digest from the example in Section 6 to double-check my understanding of content, representation, and unencoded representation. Here's a quick Node.js script to demonstrate what I did:

```js
const crypto = require('crypto')
const zlib = require('zlib')
const assert = require('assert')

// Representation from figure 2
const repr = Buffer.from([
    0x1f, 0x8b, 0x08, 0x00, 0x79, 0x1f, 0x08, 0x64, 0x00, 0xff,
    0x73, 0xcc, 0x53, 0x28, 0xcd, 0x4b, 0xad, 0x48, 0x4e, 0x2d,
    0x28, 0xc9, 0xcc, 0xcf, 0x4b, 0xcc, 0x51, 0x28, 0x2e, 0x29,
    0xca, 0xcc, 0x4b, 0xe7, 0x02, 0x00, 0x7e, 0xaf, 0x07, 0x44,
    0x18, 0x00, 0x00, 0x00,
])

// Decode the representation
const unencoded = zlib.gunzipSync(repr).toString('utf8')
assert.equal(unencoded, 'An unexceptional string\n') // ok

// Check Unencoded-Digest
const unencodedDigest = crypto.createHash('sha256').update(unencoded).digest('base64')
assert.equal(unencodedDigest, '5Bv3NIx05BPnh0jMph6v1RJ5Q7kl9LKMtQxmvc9+Z7Y=') // ok

// Check Content-Digest of partial response
const partialContent = repr.subarray(0, 10)
const contentDigest = crypto.createHash('sha256').update(partialContent).digest('base64')
assert.equal(contentDigest, 'SotB7Pa5A7iHSBdh9mg1Ev/ktAzrxU4Z8ldcCIUyfI4=') // ok

// Check Repr-Digest
const reprDigest = crypto.createHash('sha256').update(repr).digest('base64')
assert.equal(reprDigest, 'XyjvEuFb1P5rqc2le3vQm7M96DwZhvmOwqHLu2xVpY4=') // failure
```

The unencoded representation, Unencoded-Digest, and Content-Digest are the same as in the draft, but I cannot get the same Repr-Digest. It produces `kwcdt3RBGcsLaj7QSz9AW8MuwJaLjOJqUU/jKixF2oU=` instead of `XyjvEuFb1P5rqc2le3vQm7M96DwZhvmOwqHLu2xVpY4=` from the draft.

This might just be a mistake on my end and I would greatly appreciate if you can help me reproduce the Repr-Digest.

That being said, it might help readers understand the document better if the explicit values that go into the hashing algorithms are shown. For example:

```
Unencoded-Digest = sha256('An unexceptional string\n')
Content-Digest = sha256([0x1f, 0x8b, 0x08, 0x00, 0x79, 0x1f, 0x08, 0x64, 0x00, 0xff])
Repr-Digest = sha256(???)
```

(previously discussed at https://github.com/LPardue/draft-pardue-http-identity-digest/issues/20#issuecomment-4113638906)

LPardue:
@Acconut highlighted in https://github.com/httpwg/http-extensions/issues/3402 that the repr-digest seemed to be wrong. I tried 3 different validation methods and they all agreed. I have no idea how it's been broken since the first draft but now is a good time to fix them :smile: 

While at it, it seemed best to update the earlier examples to also match the later ones. That avoids later confusion if someone asks where those came from.

Closes #3402

### DIFF
--- a/draft-ietf-httpbis-unencoded-digest.md
+++ b/draft-ietf-httpbis-unencoded-digest.md
@@ -155,7 +155,9 @@ where each:
   STRUCTURED-FIELDS}}). This specification does not define any Parameters;
   future extensions may do so. Unknown Parameters MUST be ignored.
 
-For example:
+In the following examples of `Unencoded-Digest` fields, the representation data
+with no content codings applied is: "An unexceptional string" following by an
+LF.
 
 ~~~ http-message
 NOTE: '\' line wrapping per RFC 8792
--- a/draft-ietf-httpbis-unencoded-digest.md
+++ b/draft-ietf-httpbis-unencoded-digest.md
@@ -161,8 +161,8 @@ For example:
 NOTE: '\' line wrapping per RFC 8792
 
 Unencoded-Digest: \
-  sha-512=:YMAam51Jz/jOATT6/zvHrLVgOYTGFy1d6GJiOHTohq4yP+pgk4vf2aCs\
-  yRZOtw8MjkM7iw7yZ/WkppmM44T3qg==:
+  sha-512=:WjyMuMD9EI/v0RoJchcevbo6lF498VyE9564OgXf+98iJptoSvb1Czo9\
+  uVJu2bVU/tOv90huiMG3+YaMX1kipw==:
 ~~~
 
 The `Dictionary` type can be used, for example, to attach multiple digests
@@ -175,9 +175,9 @@ support transitions away from weaker algorithms (see
 NOTE: '\' line wrapping per RFC 8792
 
 Unencoded-Digest: \
-  sha-256=:d435Qo+nKZ+gLcUHn7GQtQ72hiBVAgqoLsZnZPiTGPk=:,\
-  sha-512=:YMAam51Jz/jOATT6/zvHrLVgOYTGFy1d6GJiOHTohq4yP+pgk4vf2aCs\
-  yRZOtw8MjkM7iw7yZ/WkppmM44T3qg==:
+  sha-256=:5Bv3NIx05BPnh0jMph6v1RJ5Q7kl9LKMtQxmvc9+Z7Y=:,\
+  sha-512=:WjyMuMD9EI/v0RoJchcevbo6lF498VyE9564OgXf+98iJptoSvb1Czo9\
+  uVJu2bVU/tOv90huiMG3+YaMX1kipw==:
 ~~~
 
 A recipient MAY ignore any or all digests. Application-specific behavior or
@@ -281,7 +281,7 @@ HTTP/1.1 200 OK
 Content-Type: text/plain
 Content-Encoding: gzip
 Repr-Digest: \
-  sha-256=:XyjvEuFb1P5rqc2le3vQm7M96DwZhvmOwqHLu2xVpY4=:
+  sha-256=:kwcdt3RBGcsLaj7QSz9AW8MuwJaLjOJqUU/jKixF2oU=:
 Unencoded-Digest: \
   sha-256=:5Bv3NIx05BPnh0jMph6v1RJ5Q7kl9LKMtQxmvc9+Z7Y=:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-unencoded-digest.md
```diff
@@ -175,9 +175,9 @@ support transitions away from weaker algorithms (see
 NOTE: '\' line wrapping per RFC 8792
 
 Unencoded-Digest: \
-  sha-256=:d435Qo+nKZ+gLcUHn7GQtQ72hiBVAgqoLsZnZPiTGPk=:,\
-  sha-512=:YMAam51Jz/jOATT6/zvHrLVgOYTGFy1d6GJiOHTohq4yP+pgk4vf2aCs\
-  yRZOtw8MjkM7iw7yZ/WkppmM44T3qg==:
+  sha-256=:5Bv3NIx05BPnh0jMph6v1RJ5Q7kl9LKMtQxmvc9+Z7Y=:,\
+  sha-512=:WjyMuMD9EI/v0RoJchcevbo6lF498VyE9564OgXf+98iJptoSvb1Czo9\
+  uVJu2bVU/tOv90huiMG3+YaMX1kipw==:
```
mikewest, line 182: I don't follow what's being hashed in this example and the one above. Previously, it matched the `Repr-Digest` example in https://www.rfc-editor.org/rfc/rfc9530#section-3, which I thought was intentional given the note above that "When there is no content coding, Unencoded-Digest acts identically to Repr-Digest; for the same hashing algorithm the computed value would be the same."

Perhaps adding some note about the body of the response being hashed could be helpful? Or have I just missed an explanation?

mikewest (commented):
The bottom two examples are clear to me, the top two are less so. :)

draft-ietf-httpbis-unencoded-digest.md
LPardue: In RFC 9530, the examples that accompany the field definition are intended to be minimal and to simply capture how the field in isolation might look. I.e. the field is a dictionary and this is what it looks like if there are two entries. A similar approach has been taken here.

Reusing the example value of `:d435Qo+nKZ+gLcUHn7GQtQ72hiBVAgqoLsZnZPiTGPk=:` from RFC9530 Repr-Digest is actually more confusing IMO because if you drill down into https://www.rfc-editor.org/rfc/rfc9530.html#appendix-B.5, we see the value is based on the brotli encoded representation of `the JSON object {"hello": "world"} followed by an LF.`. For people putting the drafts together, using that value in unencoded digest would appear to be an error (because it is).

Rather than have cross-document dependencies, the first version of this spec made a decision to break any overlap and base the examples in section 6 on `"An unexceptional string" following by an LF.`. 

It's hard to see in the diff but someone looking at the draft could grep the values seen here and find the later examples that includes all the context in the request and response  message exchange. While we could replicate some of that information here, it seems a little bit of a distraction.

Acconut: I think the confusion might stem from the fact that the first example in Section 3 uses sha-512, which is only used in that section. So searching for the sha-512 digest doesn't directly to the examples in Section 6. This could be improve by doing one of these:

- Mention that the examples in Section 3 are based on the representation from Section 6
- Use sha-256 in the first example in Section 3. It's used with the same value in Section 6, so people can understand the value better there
- Include sha-512 in Section 6 as well, so search for this value leads to Section 6.

LPardue: I think this is highlighting the complexity of trying to provide examples. Its really easy to take this too far and overcomplicate things; conceptully or visually. 

Take  a look at https://github.com/httpwg/http-extensions/pull/3403/changes/cb017e0a816d2eb91af3b67451efa61bbd2f4c97 and see if that helps

Acconut: That works for me 👍

Acconut (approved):
Thank you for addressing this, Lucas!

mikewest (approved):
Works for me, thanks!

LPardue:
Cool, thanks for the feedback!

PR #3408, opened 2026-03-29
RESUMABLE: Allow `max-age` limit to decrease as expected

### LINKED ISSUE #3395: RESUMABLE: Impossible restriction on `max-age`
@GrantGryczan found a mistake in the draft. The `max-age` limit is defined as:

> Specifies the remaining lifetime of the upload resource in seconds counted from the generation of the response.

Later on we say:

> In particular, whereas the values for `max-size`, `max-append-size`, and `max-age` can increase without harm, they SHOULD NOT decrease as they can cause ongoing uploads to fail.

If `max-age` cannot be decreased, the upload resource would never expire, which is obviously not the desired outcome. The sentence from the last quote stems from a draft version where `max-age` was an absolute timestamp. The restriction made sense back then, because an upload resource shouldn't expire sooner that previously announced or otherwise the client might run into issues. With the value now being a relative duration, the restriction is problematic and has to be rephrased.

GrantGryczan:
Fixes #3395


guoye-zhang (approved):

smatsson (approved):

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -361,7 +361,7 @@ The following key-value pairs are defined:
...
+Clients usually discover limits through the `Upload-Limit` header field when the upload resource is created ({{upload-creation}}). Throughout the lifetime of the upload resource, these limits SHOULD NOT change in a way that causes failures for clients adhering to the initially discovered limits. In particular, whereas the values for `max-size` and `max-append-size` can increase without harm, they SHOULD NOT decrease as they can cause ongoing uploads to fail. Similarly, the value for `min-append-size` SHOULD NOT increase, and the value for `max-age` SHOULD NOT decrease enough to cause the upload resource's lifetime to end sooner. If the client discovers that it cannot continue the upload while adhering to the limits, it SHOULD stop the current request immediately and cancel the upload ({{upload-cancellation}}).
```
guoye-zhang: Editorial: change to bullets and drop "similarly"

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -361,7 +361,11 @@ The following key-value pairs are defined:
...
+Clients usually discover limits through the `Upload-Limit` header field when the upload resource is created ({{upload-creation}}). Throughout the lifetime of the upload resource, these limits SHOULD NOT change in a way that causes failures for clients adhering to the initially discovered limits. If the client discovers that it cannot continue the upload while adhering to the limits, it SHOULD stop the current request immediately and cancel the upload ({{upload-cancellation}}).
+
+* Although the values for `max-size` and `max-append-size` can increase without harm, they SHOULD NOT decrease as that can cause ongoing uploads to fail.
+* The value for `min-append-size` SHOULD NOT increase.
+* The value for `max-age` SHOULD NOT decrease enough to cause the upload resource's lifetime to end sooner.
```
Acconut: The sentence feels a bit too vague as it allows the value to decrease, but doesn't say exactly what is too much. How about this:

```suggestion
* Between subsequent responses, the end of the upload resource's lifetime as implied by max-age SHOULD NOT decrease by more than the elapsed real time.
```

GrantGryczan: I see your point, though I think there's a mistake in your phrasing here. Did you mean the `max-age` shouldn't decrease by more than the elapsed real time? The end of the upload resource's lifetime shouldn't decrease at all. Something like this:

```suggestion
* The value for `max-age` SHOULD NOT decrease by more than the elapsed real time since the previous response.
```

Acconut: You're correct, thanks! My proposal should read like this:


```suggestion
* Between subsequent responses, the end of the upload resource's lifetime as implied by max-age SHOULD NOT decrease.
```

GrantGryczan: Great! Although I think you missed the tick marks around `max-age`. I'll commit this:

```suggestion
* Between subsequent responses, the end of the upload resource's lifetime as implied by `max-age` SHOULD NOT decrease.
```

Acconut (approved):
Thank you!

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -361,7 +361,11 @@ The following key-value pairs are defined:
...
+Clients usually discover limits through the `Upload-Limit` header field when the upload resource is created ({{upload-creation}}). Throughout the lifetime of the upload resource, these limits SHOULD NOT change in a way that causes failures for clients adhering to the initially discovered limits. If the client discovers that it cannot continue the upload while adhering to the limits, it SHOULD stop the current request immediately and cancel the upload ({{upload-cancellation}}).
+
+* Although the values for `max-size` and `max-append-size` can increase without harm, they SHOULD NOT decrease as that can cause ongoing uploads to fail.
```
LPardue: The bullets feel a bit naked, we're missing a lead in to give some context. How about something like

```suggestion
The following recommendations for limit changes can minimize the risk of causing upload failures:

* `max-size` and `max-append-size` SHOULD NOT decrease.
```

GrantGryczan:
I added a `min-size` recommendation since it seems like it should be a more comprehensive list now, rather than some arbitrary examples like before.

Acconut:
> I added a min-size recommendation since it seems like it should be a more comprehensive list now, rather than some arbitrary examples like before.

min-size and max-size apply at the creation time of the upload resource. Changing its values after the resource has been created shouldn't affect the client's ability to complete the upload. Should we remove the mention of min-size and max-size from this paragraph, which talks about limit changes during the upload resource's timeline?

GrantGryczan:
I believe it could affect the client's ability to complete the upload. If the client says `Upload-Length: 100`, and the server accepts that, but they later change their `max-size` to 50 or their `min-size` to 150, the client is recommended to cancel the upload. Alternatively, if the client knows the upload length will be within a certain range, but hasn't communicated an exact length, it may be unable to complete the upload if the required bounds for the upload length narrow.

Acconut:
My previous comment isn't entirely correct because max-size can also be applied after the upload creation:

> The server might also deactivate the upload resource if the offset ([Section 4.1.1](https://www.ietf.org/archive/id/draft-ietf-httpbis-resumable-upload-11.html#upload-offset)) exceeds the maximum size.

(from https://www.ietf.org/archive/id/draft-ietf-httpbis-resumable-upload-11.html#section-4.1.4-3.2.1)

Then let's keep this and get this PR merged!

Acconut (approved):

GrantGryczan:
Same for `min-size`, to be clear, since the `min-size` could be very large and may only be checked by the server upon the client communicating the length from an upload append request. Am I understanding correctly?

Acconut:
The purpose of min-size is mainly to let the client know that the server won't offer resumable uploads for representations outside of those boundaries. But this doesn't mean the representation will be rejected when uploading via a traditional POST request, for example. A server may decide to only use resumable uploads for larger files as there is a cost of holding state involved for upload resources.

GrantGryczan:
Oh, I didn't know that. That seems very worth clarifying in the draft. But also, I'm not confident in your explanation. It feels possibly partially conflated with `min-append-size`? Unsure. Let me open a new issue for this.

Acconut:
The definitions of max-size and min-size state that the server might not create an upload resource if the indicated representation size falls outside of these boundaries:

> max-size: Specifies a maximum size for the representation data, counted in bytes. The server might not create an upload resource if the length ([Section 4.1.3](https://www.ietf.org/archive/id/draft-ietf-httpbis-resumable-upload-11.html#upload-length)) deduced from the upload creation request is larger than the maximum size. The server might also deactivate the upload resource if the offset ([Section 4.1.1](https://www.ietf.org/archive/id/draft-ietf-httpbis-resumable-upload-11.html#upload-offset)) exceeds the maximum size. The value is an Integer.
> 
> min-size: Specifies a minimum size for the representation data, counted in bytes. The server might not create an upload resource if the length ([Section 4.1.3](https://www.ietf.org/archive/id/draft-ietf-httpbis-resumable-upload-11.html#upload-length)) deduced from the upload creation request is smaller than the minimum size or no length can be deduced at all. The value is an Integer.

It doesn't say that upload creation requests with sizes outside of these limits must be rejected. The server can still process the included representation without involving resumable uploads, which is a feature IMO, but a bug :)

guoye-zhang:
Yeah, this is mainly to support optimistic discovery, the implementation should not reject uploads smaller than min or larger than max since the client might not know the limit ahead of time. If that's not a part of your use case, it's fine to be stricter.

Issue #3424, opened 2026-04-13
[Connect-TCP WGLC] server applicability


yaroslavros:
Statements in first paragraph of 5.1 apply only when comparing templated CONNECT-TCP with CONNECT over cleartext HTTP/1.
When TLS or QUIC is involved virtual servers are distinguished through SNI.

Also, I really don't think URI should be a recommended authorisation technique.

It may also be worth mentioning draft-ietf-intarea-proxy-config as a discovery mechanism.

bemasc:
I don't believe that SNI would be a suitable or correct way to implement this kind of virtual hosting.  HTTP connections can be used for requests to any origin matching the certificate (if DNS agrees) but the SNI only indicates a single hostname.

> It may also be worth mentioning draft-ietf-intarea-proxy-config as a discovery mechanism.

Yes, we can mention that.

bemasc:
@yaroslavros I've posted #3435 to mention PvD.  Please review.

> Also, I really don't think URI should be a recommended authorisation technique.

Please review #3436 for this topic.

bemasc:
#3435 and #3436 are now merged.  The remaining topic for this issue is "Applicability > Server" pg. 1:

> For server operators, template-driven TCP proxies are particularly valuable in situations where virtual-hosting is needed, or where multiple proxies must share an origin. For example, the proxy might benefit from sharing an HTTP gateway that provides DDoS defense, performs request sanitization, or enforces user authorization.

As noted at the top, it is possible to share an IP address among multiple classic CONNECT proxies using TLS SNI.  However, in my experience demultiplexing TLS by SNI is not a recommended arrangement, and is not considered a valid form of HTTP virtual hosting.  As such, I think the current text is correct and sufficient.

This text is not normative.  If anyone would like to see it adjusted, feel free to suggest an improvement.

mnot:
@yaroslavros any further feedback?

yaroslavros:
Looks good to me. Thank you for your exemplary work, @bemasc !

PR #3435, opened 2026-04-20
[connect-tcp] Discuss how to apply PvD configuration

bemasc:
Related to #3424

### DIFF
--- a/draft-ietf-httpbis-connect-tcp.md
+++ b/draft-ietf-httpbis-connect-tcp.md
@@ -272,7 +272,11 @@ The URI template can also be structured to generate high-entropy Capability URLs
 
 ## Clients
 
-Clients that support both classic HTTP CONNECT proxies and template-driven TCP proxies MAY accept both types via a single configuration string.  If the configuration string can be parsed as a URI Template containing the required variables, it is a template-driven TCP proxy.  Otherwise, it is presumed to represent a classic HTTP CONNECT proxy.
+Clients that support both classic HTTP CONNECT proxies and template-driven TCP proxies MAY accept both types via a single configuration string:
+
+* If the configuration string is a valid URI Template containing the required variables, it represents a template-driven TCP proxy.
+* Otherwise the client attempts to parse it as representing a classic HTTP CONNECT proxy.
+* Additionally, both types of proxies can be discovered using proxy provisioning domains {{?I-D.ietf-intarea-proxy-config}} if the input is a URI or represents a proxy of the other type.
 
 In some cases, it is valuable to allow "connect-tcp" clients to reach "connect-tcp"-only proxies when using a legacy configuration method that cannot convey a URI Template.  To support this arrangement, clients SHOULD treat certain errors during classic HTTP CONNECT as indications that the proxy might only support "connect-tcp":
 
--- a/draft-ietf-httpbis-connect-tcp.md
+++ b/draft-ietf-httpbis-connect-tcp.md
@@ -272,27 +272,29 @@ The URI template can also be structured to generate high-entropy Capability URLs
 
 ## Clients
 
-Clients that support both classic HTTP CONNECT proxies and template-driven TCP proxies MAY accept both types via a single configuration string:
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-connect-tcp.md
```diff
@@ -272,7 +272,11 @@ The URI template can also be structured to generate high-entropy Capability URLs
 
 ## Clients
 
-Clients that support both classic HTTP CONNECT proxies and template-driven TCP proxies MAY accept both types via a single configuration string.  If the configuration string can be parsed as a URI Template containing the required variables, it is a template-driven TCP proxy.  Otherwise, it is presumed to represent a classic HTTP CONNECT proxy.
+Clients that support both classic HTTP CONNECT proxies and template-driven TCP proxies MAY accept both types via a single configuration string:
```
yaroslavros: I don't think this is a good suggestion. It may lead to ambiguities - especially given that some other proxy techniques (CONNECT-UDP) also leverage identical URI template.

I would suggest something along the following lines:


```suggestion
Clients that support both classic HTTP CONNECT proxies and template-driven TCP proxies SHOULD have proxy type explicitly configured. Client configured with the HTTP CONNECT proxy hostname MAY use a discovery mechanism described in {{Section 2.1 of ?I-D.ietf-intarea-proxy-config=PVD-PROXY-CONFIG}}.
```

If you insist on encouraging a single configuration string, I would slightly adjust the final bullet as suggested below

draft-ietf-httpbis-connect-tcp.md
```diff
@@ -272,7 +272,11 @@ The URI template can also be structured to generate high-entropy Capability URLs
...
+
+* If the configuration string is a valid URI Template containing the required variables, it represents a template-driven TCP proxy.
+* Otherwise the client attempts to parse it as representing a classic HTTP CONNECT proxy.
+* Additionally, both types of proxies can be discovered using proxy provisioning domains {{?I-D.ietf-intarea-proxy-config}} if the input is a URI or represents a proxy of the other type.
```
yaroslavros:
```suggestion
* Additionally, both types of proxies can be discovered using a mechanism specified in {{Section 2.1 of ?I-D.ietf-intarea-proxy-config=PVD-PROXY-CONFIG}} as long as configuration string contains a hostname.
```

draft-ietf-httpbis-connect-tcp.md
bemasc: I think configuration via a single string is extremely valuable.  There is already a maze of system-level configuration mechanisms and user interfaces that can pass an HTTP proxy string.  These mechanisms are unlikely to ever be updated to incorporate the proposed metadata tag, so we should plan to use them as-is.

draft-ietf-httpbis-connect-tcp.md
bemasc: The intent here is not to limit PvD use to HTTPS records.  Hosting multiple distinct proxies on a hostname and avoiding revealing their presence to unauthorized clients are key motivations for this draft, and hostname-based configuration is incompatible with these goals.

A single string input can support a wide range of configuration starting points that lead to PvD.  Even an IP address could be used, in combination with `.well-known/pvd`.  The most flexible (and in my view best) option would be a full "https://" URL for the PvD JSON.  The purpose of this line is to remind implementers that they can support all of these things without redesigning their configuration input system.

draft-ietf-httpbis-connect-tcp.md
tfpauly: Practically, I'm assuming that if there is a string input field for configuring a proxy, one would generally need to assume it is going to be the legacy HTTP CONNECT. I certainly think we can say "if you enter a string that looks like a CONNECT-TCP template, then a system that knows about CONNECT-TCP needs to use this spec.", but have a warning that entering that on systems that don't know about it may lead to failures to use the proxy.

bemasc:
@tfpauly @yaroslavros I've rewritten this section to incorporate the perspectives you've shared here.  Please review.

yaroslavros (approved):

PR #3453, opened 2026-06-15
RESUMABLE: Clarify "request header field" under server behavior (nit)

GrantGryczan:
While referencing the draft for my implementation, I was confused whether this was referring to the `Upload-Complete` request header, response header, or both. I had to read around for context to figure it out, and I just think it'd be nice if I didn't have to.

The previous paragraph uses the same phrasing.

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -412,7 +412,7 @@ The resource targeted by this initial request is responsible for processing the 
 
 If the `Upload-Complete` request header field is set to true, the client intends to transfer the entire representation data in one request. If the request content was fully processed, no resumable upload is needed and the server proceeds to process the request and generate a response.
 
-If the `Upload-Complete` header field is set to false, the client intends to transfer the representation over multiple requests. If the request content was fully processed, the server MUST include the `Location` response header field pointing to the upload resource and MUST include the `Upload-Limit` header field with the corresponding limits if existing. Servers are RECOMMENDED to use the `201 (Created)` status code.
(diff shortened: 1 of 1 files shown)


LPardue:
Thanks for the PR. However, I think a simpler way to achieve clarity is to change the text to say "If the client sets the `Upload-Complete` header field to true, it intends to ..."

GrantGryczan:
I can change that, but are you sure? This phrasing is used in quite a few places, so I figure we'd want to change it in all those other places too if we changed that here.

I'm also not sure that phrasing is better since the "it" might not be immediately obviously referring to the client, as opposed to the header. Even though interpreting "it" as the header doesn't make as much sense, initially interpreting it that way could be confusing. Especially if we want to write it like that all over the place. (Not that this is a big deal either way; this is kind of a bikeshed. I'm just pointing out why you might not want to do that.)

guoye-zhang (approved):

LPardue:
Yeah, I think there should be an editorial pass to ensure self-consistency and also alignment with the HTTP style guide at https://httpwg.org/admin/editors/style-guide

I expect this to be a task for the editors to do (i.e. me) once some of the technical aspects have resolved.

There are specific places full qualification is needed such as the field definition, but adding redundant qualifying terms everywhere, in general reduces the readability IMO.

GrantGryczan:
Alrighty, up to you! Actually sorry, I won't bother modifying my PR to edit all those other instances, since you probably understand how this should be phrased in other cases better than I do. So if you'd argue for such a pervasive change, feel free to simply close this or make those changes yourself. I did ctrl+F "the client intends" and found a few other similar instances.

Good to know there will be an editorial pass though, makes sense.

LPardue:
Yeah thats reasonable. Lets merge this since it won't materially hurt anything. Just look out for a followup editorial change to do what I proposed.

Thanks again!

PR #3460, opened 2026-06-22
RESUMABLE: Clarify min-size as threshold for resumability

### LINKED ISSUE #3446: RESUMABLE: Clarification on `min-size`
@Acconut wrote the following in https://github.com/httpwg/http-extensions/pull/3408#issuecomment-4439012488:

> The purpose of min-size is mainly to let the client know that the server won't offer resumable uploads for representations outside of those boundaries. But this doesn't mean the representation will be rejected when uploading via a traditional POST request, for example. A server may decide to only use resumable uploads for larger files as there is a cost of holding state involved for upload resources.

And the following in https://github.com/httpwg/http-extensions/pull/3408#issuecomment-4439173526:

> The definitions of max-size and min-size state that the server might not create an upload resource if the indicated representation size falls outside of these boundaries:
> 
> > ...
> 
> It doesn't say that upload creation requests with sizes outside of these limits must be rejected. The server can still process the included representation without involving resumable uploads, which is a feature IMO, but a bug :)

I didn't know this! If this is true, this seems important to clarify in the draft. Right now, the definition of `min-size` _also_ says that it is "a minimum size for the representation data", which is a stronger requirement that I thought disallowed this.

~~I'd argue having an uncommunicated, separate (or a lack of a) limit for non-resumable uploads is pointless, because a server could just use `min-append-size` to ensure the entire upload is done in one request under a certain size. The server could even loosen the `min-append-size` limit after the first request if it only needs that restriction for the purpose of avoiding creating an upload resource.~~ In my opinion, using `min-size` for this ~~is redundant and~~ may make `min-size` less reliable for the purpose of restricting the representation data size.

EDIT: Struck through most of my previous argument because I realize `min-append-size` can't be used to restrict upload resource creation, since it doesn't apply to empty upload creation requests. Is there any common situation where a server would want an actual, final minimum restriction on the total size of an upload, resumable or not? I guess I can't really think of anything.

Lastly, I also wanted to ask for clarification on whether `min-size` can be enforced by servers during upload append requests rather than during upload creation, since I'm not sure whether a previous comment hinted that it might only be enforced on upload creation. The current phrasing as "a minimum size for the representation data", as opposed to a minimum upload length to allow upload resource creation, makes it seem like it would always be enforced.

Acconut:
Fixes https://github.com/httpwg/http-extensions/issues/3446.

This PR adds clarifications around the `min-size` limit. It's not a hard limit as in server will reject representations below this value. Servers might still accept them, although without offering resumable uploads because the benefit of resumability is too small compared to the overhead of an upload resource.


guoye-zhang (approved):

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -387,11 +387,11 @@ A server responding with a `413 (Content Too Large)` status code as a result of
...
+The `Upload-Complete` header field is set to true if the request content includes the entire representation data that the client intends to upload. This is also a requirement for transparently upgrading to resumable uploads from conventional uploads ({{upgrading-uploads}}).
 
-If the client knows the representation data's length, it SHOULD indicate the length in the request to help the server allocate necessary resources for the upload and provide early feedback if the representation violates a limit ({{upload-limit}}). This indication can be done through the `Upload-Length` header field or the combination of the `Content-Length` and `Upload-Complete: ?1` header fields (see {{upload-length}}).
+If the client knows the representation data's length, it SHOULD indicate the length in the request to help the server allocate necessary resources for the upload and provide early feedback if the representation exceeds `max-size` ({{upload-limit}}). This indication can be done through the `Upload-Length` header field or the combination of the `Content-Length` and `Upload-Complete: ?1` header fields (see {{upload-length}}).
```
GrantGryczan: If the server has a `min-size` limit, the server needs to know the length on upload creation so it can reject an incomplete upload under that limit. Even if it is a complete upload, the server still needs to know whether it should create an upload resource for it and send 104 responses accordingly. Therefore, I don't know if changing this to only mention `max-size` is desirable.

Other than this, everything else is great! :D

GrantGryczan (changes requested):

draft-ietf-httpbis-resumable-upload.md
Acconut: That's true, let me revert this.

```suggestion
If the client knows the representation data's length, it SHOULD indicate the length in the request to help the server allocate necessary resources for the upload and provide early feedback if the representation violates a limit ({{upload-limit}}). This indication can be done through the `Upload-Length` header field or the combination of the `Content-Length` and `Upload-Complete: ?1` header fields (see {{upload-length}}).
```

Thanks for the feedback!

PR #3461, opened 2026-06-22
RESUMABLE: Recomment incremental delivery

### LINKED ISSUE #3442: RESUMABLE: Mention `Incremental: ?1`
Resumable uploads work best if request content is delivered incrementally, so the partial content can be processed even if the entire request doesn't complete. We should consider recommending the inclusion of `Incremental: ?1` from [draft-ietf-httpbis-incremental](https://datatracker.ietf.org/doc/draft-ietf-httpbis-incremental/) in content-transferring requests to encourage intermediaries to not buffer.

Acconut:
Closes https://github.com/httpwg/http-extensions/issues/3442.

This mentions the benefit of incremental delivery of the request content for resumable uploads. The `Incremental: ?1` header field is mentioned to express the client's preference.

A few open questions I would welcome feedback on:

1. Should the draft explicitly recommend the client to set `Incremental: ?1`? Are there good reasons to not proactively set it in all cases?
2. Should the `Incremental: ?1` header be included in examples?
3. Should the reference to draft-ietf-httpbis-incremental be normative or informative?

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -54,6 +54,7 @@ normative:
   PROBLEM: RFC9457
   DIGEST-FIELDS: RFC9530
   CONTENT-DISPOSITION: RFC6266
+  INCREMENTAL: I-D.ietf-httpbis-incremental
 
 informative:
   SLOWLORIS:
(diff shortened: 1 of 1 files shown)


guoye-zhang (approved):

LPardue (changes requested):
I would phrase the differently. I'll follow up with an alternative in a little bit

LPardue:
Ok see https://github.com/httpwg/http-extensions/pull/3462, which cuts down on the duplicate text by adding a new section for considerations and forward referencing to it

PR #3462, opened 2026-06-23
RESUMABLE: Alternative incremental considerations

### LINKED ISSUE #3442: RESUMABLE: Mention `Incremental: ?1`

LPardue:
This is an alternative to #3461 that focuses on what HTTP allows and why its good. The
reference to the Incremental draft is left as informative because clients "MAY always"
do whatever they like and they should go read that draft to understand things.

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -66,6 +66,7 @@ informative:
     target:
      "https://web.archive.org/web/20150315054838/http://ha.ckers.org/slowloris/"
   RFC8792:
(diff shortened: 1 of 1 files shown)


guoye-zhang (approved):

Acconut (approved):
That's a much better phrasing. Thank you, Lucas!

PR #3463, opened 2026-06-23
RESUMABLE: Require offset and length to match upon completion

### LINKED ISSUE #3430: RESUMABLE: Unclear how server should handle completing request with insufficient length
If the client sends an upload creation or append request with `Upload-Complete: ?1` whose content isn't large enough to meet the upload length expected by the server, what should the server do? I don't think this edge case is ever explicitly commented on in the draft. If it is, I can't find it.

Some options I can think of:

1. ~~Mark the upload as completed, process the content, and adjust the upload length accordingly. (Can the upload length change? Opened #3431)~~
2. Don't mark the upload as completed, but respond with success, and otherwise process the content as if there's no issue.
3. Don't mark the upload as completed, and respond with an error, but otherwise process the content as if there's no issue.
4. Don't mark the upload as completed, respond with an error, and don't comment on whether the content should be processed.
5. Don't mark the upload as completed, respond with an error, and recommend against processing the content.

I personally would suggest either option 4 or 5.

Acconut:
Closes #3430.

This PR clarifies that the offset and length must match when the upload is about to be completed. The draft had already requirements that the length derived from `Content-Length` is consistent, but there was a loophole when the header is missing:

> The only loophole is that this is only specified for requests with Content-Length. This header isn't always present (e.g. when Transfer-Encoding: chunked is used), so this case isn't prohibited at the moment:
> 
>  -  Empty POST request with Upload-Length: 1000 and Upload-Complete: ?0
>  -  PATCH request with Upload-Complete: ?1, no Content-Length, but 500 bytes of content

(from https://github.com/httpwg/http-extensions/issues/3430#issuecomment-4438288485)

This gap is closed by explicitly saying that the length must be consistent in the end.

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -540,7 +540,7 @@ A successful response to a `HEAD` request against an upload resource
 
 - MUST include the offset in the `Upload-Offset` header field ({{upload-offset}}),
 - MUST include the `Upload-Complete` header field ({{upload-complete}}) indicating whether a final response was produced from processing the uploaded representation,
-- MUST include the representation's length in the `Upload-Length` header field if the client has supplied one through corresponding headers as described in ({{upload-length}}),
+- MUST include the representation's length in the `Upload-Length` header field, unless the client has not supplied one, by providing the corresponding headers as described in ({{upload-length}}),
 - MUST indicate the limits in the `Upload-Limit` header field ({{upload-limit}}), and
 - SHOULD include the `Cache-Control` header field with the value `no-store` to prevent HTTP caching ({{CACHING}}).
 
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -335,7 +335,7 @@ Despite this, a client SHOULD communicate the representation's length to the ser
 
 The `Upload-Length` request and response header field is an Item Structured Header Field ({{STRUCTURED-FIELDS}}). Its value is a non-negative Integer ({{Section 3.3.1 of STRUCTURED-FIELDS}}) and indicates the representation's length as a number of bytes. Other values MUST cause the entire header field to be ignored.
 
-If indicators (1) and (2) are both present in the same request, their indicated lengths MUST match. The representation's length, if known, MUST stay consistent across subsequent requests. A server can use the problem type {{PROBLEM}} of "https://iana.org/assignments/http-problem-types#inconsistent-representation-length" ({{inconsistent-length}}) in responses to indicate inconsistent length indicators.
+If indicators (1) and (2) are both present in the same request, their indicated lengths MUST match. The representation's length, if known, MUST stay consistent across subsequent requests. A server can use the problem type {{PROBLEM}} of "https://iana.org/assignments/http-problem-types#inconsistent-upload-length" ({{inconsistent-length}}) in responses to indicate inconsistent length indicators.
 
 The `Upload-Length` field can be used in response to an offset retrieval; see {{offset-retrieving-server}}.
 
@@ -737,7 +737,7 @@ Upload-Complete: ?0
 
 ## Inconsistent Length
 
-This section defines the "https://iana.org/assignments/http-problem-types#inconsistent-representation-length" problem type {{PROBLEM}}. A server can use this problem type when responding to an upload creation ({{upload-creation}}) or upload append request ({{upload-appending}}) to indicate that the request includes inconsistent length values, as described in {{upload-length}}.
+This section defines the "https://iana.org/assignments/http-problem-types#inconsistent-upload-length" problem type {{PROBLEM}}. A server can use this problem type when responding to an upload creation ({{upload-creation}}) or upload append request ({{upload-appending}}) to indicate that the request includes inconsistent length values, as described in {{upload-length}}.
 
 The following example shows an example response:
 
@@ -750,8 +750,8 @@ Upload-Complete: ?0
 
 {
   "type":"https://iana.org/assignments/http-problem-types#\
-    inconsistent-representation-length",
-  "title": "inconsistent length values for representation"
+    inconsistent-upload-length",
+  "title": "inconsistent length values for upload"
 }
 ~~~
 
@@ -948,10 +948,10 @@ Reference:
 IANA is asked to register the following entry in the "HTTP Problem Types" registry:
 
 Type URI:
-: https://iana.org/assignments/http-problem-types#inconsistent-representation-length
+: https://iana.org/assignments/http-problem-types#inconsistent-upload-length
 
 Title:
-: Inconsistent Representation Length Values
+: Inconsistent Upload Length Values
 
 Recommended HTTP status code:
 : 400
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -335,7 +335,7 @@ Despite this, a client SHOULD communicate the representation's length to the ser
 
 The `Upload-Length` request and response header field is an Item Structured Header Field ({{STRUCTURED-FIELDS}}). Its value is a non-negative Integer ({{Section 3.3.1 of STRUCTURED-FIELDS}}) and indicates the representation's length as a number of bytes. Other values MUST cause the entire header field to be ignored.
 
-If indicators (1) and (2) are both present in the same request, their indicated lengths MUST match. The representation's length MUST stay consistent across subsequent requests. A server can use the problem type {{PROBLEM}} of "https://iana.org/assignments/http-problem-types#inconsistent-representation-length" ({{inconsistent-length}}) in responses to indicate inconsistent length values.
+If indicators (1) and (2) are both present in the same request, their indicated lengths MUST match. The representation's length, if known, MUST stay consistent across subsequent requests. A server can use the problem type {{PROBLEM}} of "https://iana.org/assignments/http-problem-types#inconsistent-representation-length" ({{inconsistent-length}}) in responses to indicate inconsistent length values.
 
 The `Upload-Length` field can be used in response to an offset retrieval; see {{offset-retrieving-server}}.
 
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -328,9 +328,9 @@ When used in an upload creation response ({{upload-creation}}) or an upload appe
 
 The length of the representation data might not be known when starting the transfer, for example, because the representation is taken from a streaming source. The representation's length will, however, be known at the latest when the upload is complete.
 
-Despite this, a client SHOULD communicate the representation's length to the server as soon as it becomes known to aid with resource managament and facilitate early validation. There are two different ways for the client to indicate and for the server to discover the representation's length from requests for creating the upload resource ({{upload-creation}}) or appending to it ({{upload-appending}}):
+Despite this, a client SHOULD communicate the representation's length to the server as soon as it becomes known to aid with resource management and facilitate early validation. There are two different ways for the client to indicate and for the server to discover the representation's length from requests for creating the upload resource ({{upload-creation}}) or appending to it ({{upload-appending}}):
 
-1. If the request includes the `Upload-Complete` field value set to true, the request content is the remaining representation data. The representation's length is then the sum of the current offset ({{upload-offset}}) and request content's length, which might be announced in the `Content-Length` header field.
+1. If the request includes the `Upload-Complete` field value set to true, the request content is the remaining representation data. The representation's length is then the sum of the current offset ({{upload-offset}}) and the request content's length, which might be announced in the `Content-Length` header field.
 2. The request can include the `Upload-Length` header field defined below.
 
 The `Upload-Length` request and response header field is an Item Structured Header Field ({{STRUCTURED-FIELDS}}). Its value is a non-negative Integer ({{Section 3.3.1 of STRUCTURED-FIELDS}}) and indicates the representation's length as a number of bytes. Other values MUST cause the entire header field to be ignored.
@@ -388,7 +388,7 @@ A client can start a resumable upload from any request that can carry content by
 
 The `Upload-Complete` header field is set to true if the request content includes the entire representation data that the client intends to upload. This is also a requirement for transparently upgrading to resumable uploads from conventional uploads ({{upgrading-uploads}}).
 
-If the client knows the representation data's length, it SHOULD indicate the length in the request to help the server allocate necessary resources for the upload and provide early feedback if the representation violates a limit ({{upload-limit}}), as described in {{upload-length}}.
+If the client knows the representation's length, it SHOULD indicate the length in the request to help the server allocate necessary resources for the upload and provide early feedback if the representation violates a limit ({{upload-limit}}), as described in {{upload-length}}.
 
 The client SHOULD respect any limits ({{upload-limit}}) announced in the `Upload-Limit` header field in interim or final responses. In particular, if the allowed maximum size is less than the amount of representation data the client intends to upload, the client SHOULD stop the current request immediately and cancel the upload ({{upload-cancellation}}). If the client knows that the representation data is smaller than `min-size`, it cannot expect resumability to be offered. The client might still attempt to transfer the representation in a single request, either in a request with the `Upload-Complete` header field set to true (see {{upgrading-uploads}}) or via a conventional upload.
 
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -335,7 +335,7 @@ Despite this, a client SHOULD communicate the representation's length to the ser
 
 The `Upload-Length` request and response header field is an Item Structured Header Field ({{STRUCTURED-FIELDS}}). Its value is a non-negative Integer ({{Section 3.3.1 of STRUCTURED-FIELDS}}) and indicates the representation's length as a number of bytes. Other values MUST cause the entire header field to be ignored.
 
-If indicators (1) and (2) are both present in the same request, their indicated lengths MUST match. The representation's length, if known, MUST stay consistent across subsequent requests. A server can use the problem type {{PROBLEM}} of "https://iana.org/assignments/http-problem-types#inconsistent-representation-length" ({{inconsistent-length}}) in responses to indicate inconsistent length values.
+If indicators (1) and (2) are both present in the same request, their indicated lengths MUST match. The representation's length, if known, MUST stay consistent across subsequent requests. A server can use the problem type {{PROBLEM}} of "https://iana.org/assignments/http-problem-types#inconsistent-representation-length" ({{inconsistent-length}}) in responses to indicate inconsistent length indicators.
 
 The `Upload-Length` field can be used in response to an offset retrieval; see {{offset-retrieving-server}}.
 
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -339,7 +339,7 @@ If indicators (1) and (2) are both present in the same request, their indicated 
 
 The `Upload-Length` field can be used in response to an offset retrieval; see {{offset-retrieving-server}}.
 
-The server might not know the length until the upload is complete.
+The server might not know the length until the upload is complete. Once the client attempts to complete the upload, the server MUST, after processing the request content, verify that the offset matches a previously recorded length value. If it does not, the server MUST reject the corresponding request, deactivate the upload resource, and reject further interaction with it.
 
 Note that the length and offset values do not determine whether an upload is complete. Instead, the client uses the `Upload-Complete` ({{upload-complete}}) header field to indicate that a request completes the upload. The offset could match the length, but the upload can still be incomplete.
 
@@ -977,6 +977,7 @@ Reference:
 * Redefine Upload-Complete on the server side.
 * Recommend incremental delivery.
 * Clarify `min-size` limit and its client behavior.
+* Specify that offset and length must match upon completion.
 
 ## Since draft-ietf-httpbis-resumable-upload-10
 {:numbered="false"}
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -326,20 +326,18 @@ When used in an upload creation response ({{upload-creation}}) or an upload appe
 
 ### Length {#upload-length}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -339,7 +339,7 @@ If indicators (1) and (2) are both present in the same request, their indicated
 
 The `Upload-Length` field can be used in response to an offset retrieval; see {{offset-retrieving-server}}.
 
-The server might not know the length until the upload is complete.
+The server might not know the length until the upload is complete. Once the client attempts to complete the upload, the server MUST, after processing the request content, verify that the offset matches a previously recorded length value. If it does not, the server MUST reject the corresponding request, deactivate the upload resource, and reject further interaction with it.
```
guoye-zhang: Does “Upload-Length: 1000 and Upload-Complete: ?0” count as “attempts to complete the upload”?

Could we just say something like “final length must not change once it’s known”?

Acconut: > Does “Upload-Length: 1000 and Upload-Complete: ?0” count as “attempts to complete the upload”?

I wouldn't say so. With "attempting to complete the upload" I referred to sending a request with `Upload-Complete: ?1`

> Could we just say something like “final length must not change once it’s known”?

The spec already includes this. It says that the length indicated in subsequent requests must match. The problem is that the length is currently tied to the presence of either Upload-Length or Content-Length. So if the final request includes neither, no text exists that force the final representation size to match previous length values. A concrete example is this:

- Empty POST request with Upload-Length: 1000 and Upload-Complete: ?0
- PATCH request with Upload-Complete: ?1, no Content-Length, but 500 bytes of content

I think we could either
- add text to say that the length must be acurate if the upload is completed (what this PR tries to do), or
- say that a length is also recorded if the upload completes and it must be consist with previous length values.

Does that make sense?

GrantGryczan:
Why do we deactivate the upload resource instead of simply responding with a 4xx error with `Upload-Complete: ?0`? Couldn't the client recover from this mistake by simply resending the correct final append? This seems particularly problematic if the inconsistent length is merely a transmission error, not some permanent bad client state. Why would such an error result in the entire upload failing?

I'd think we'd want to just make this fall under the existing inconsistent length problem type.

Acconut:
You're right. We can adjust the language to meet the existing requirements on keeping the length consistent.

draft-ietf-httpbis-resumable-upload.md
guoye-zhang: Ah I see. I think we should
1. define the upload length to be the complete full length of the content upfront.
2. Then say it can be derived ahead of completion from the Upload-Length or Content-Length + complete.
3. Once it is known, it must not be changed.
4. Attempting to change it results in a certain error.

guoye-zhang:
I think this should be a hard failure. It's unclear how this could be the result of a transmission error, other than TCP somehow corrupting the bytes which is outside the scope of the protocol to recover. However, I'm also fine with it being retryable since it doesn't complicate the implementation.

GrantGryczan:
Even if it isn't a realistic transmission error, the server implementation may be able to be simpler if it can just reuse the inconsistent upload length 4xx response, rather than special-casing this one means that the upload length can be inconsistent.

draft-ietf-httpbis-resumable-upload.md
Acconut: That's a good approach. Thinking more about this made me realize that the upload length is just the representation's length. We can probably adjust the wording in this section to talk about the representation's length and how the server can discover it through Content-Length and Upload-Length. This would allow us to get rid of the duplication of representation length and the upload resource's recorded length and then we don't have to mention that they have to match in the end. Let me try that out ~~in a PR~~...

Acconut:
I've adjusted the section of length to use _representation's length_ instead of _upload length_. We introduced this term, but it has the same meaning as the representation's length, so I think this makes it easier for the reader to understand what value is communicated. It should also be clearer that the length must stay consistent over the upload. The `Upload-Length` header is left untouched since it's now just a way to communicate the representation's length.

Please let me know what you think.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -326,21 +326,19 @@ When used in an upload creation response ({{upload-creation}}) or an upload appe
 
 ### Length {#upload-length}
 
-The length of an upload is the number of bytes of representation data that the client intends to upload.
+The length of the representation data might not be known when starting the transfer, for example, because the representation is taken from a streaming source. The representation's length will, however, be known at the latest when the upload is complete.
```
GrantGryczan: This last sentence is untrue. The server can mark an upload as completed (I assume you meant "completed", not "complete") before the length is known.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -326,21 +326,19 @@ When used in an upload creation response ({{upload-creation}}) or an upload appe
...
 
-Even the client might not know the total length of the representation data when starting the transfer, for example, because the representation is taken from a streaming source. However, a client SHOULD communicate the length to the server as soon as it becomes known. There are two different ways for the client to indicate and the server to discover the length from requests for creating the upload resource ({{upload-creation}}) or appending to it ({{upload-appending}}):
+Despite this, a client SHOULD communicate the representation's length to the server as soon as it becomes known to aid with resource managament and facilitate early validation. There are two different ways for the client to indicate and for the server to discover the representation's length from requests for creating the upload resource ({{upload-creation}}) or appending to it ({{upload-appending}}):
```
GrantGryczan: Nit: Just a typo.

```suggestion
Despite this, a client SHOULD communicate the representation's length to the server as soon as it becomes known to aid with resource management and facilitate early validation. There are two different ways for the client to indicate and for the server to discover the representation's length from requests for creating the upload resource ({{upload-creation}}) or appending to it ({{upload-appending}}):
```

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -326,21 +326,19 @@ When used in an upload creation response ({{upload-creation}}) or an upload appe
...
 
-1. If the request includes the `Upload-Complete` field value set to true and a valid `Content-Length` header field, the request content is the remaining representation data. The length is then the sum of the current offset ({{upload-offset}}) and the `Content-Length` header field value.
+1. If the request includes the `Upload-Complete` field value set to true, the request content is the remaining representation data. The representation's length is then the sum of the current offset ({{upload-offset}}) and request content's length, which might be announced in the `Content-Length` header field.
```
GrantGryczan: Nit: This sounds better to me with an article here: "the sum of the X and _the_ Y".

```suggestion
1. If the request includes the `Upload-Complete` field value set to true, the request content is the remaining representation data. The representation's length is then the sum of the current offset ({{upload-offset}}) and the request content's length, which might be announced in the `Content-Length` header field.
```

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -326,21 +326,19 @@ When used in an upload creation response ({{upload-creation}}) or an upload appe
...
+The `Upload-Length` request and response header field is an Item Structured Header Field ({{STRUCTURED-FIELDS}}). Its value is a non-negative Integer ({{Section 3.3.1 of STRUCTURED-FIELDS}}) and indicates the representation's length as a number of bytes. Other values MUST cause the entire header field to be ignored.
 
-If indicators (1) and (2) are both present in the same request, their indicated lengths MUST match. If multiple requests include indicators, their indicated values MUST match. A server can use the problem type {{PROBLEM}} of "https://iana.org/assignments/http-problem-types#inconsistent-upload-length" ({{inconsistent-length}}) in responses to indicate inconsistent length values.
+If indicators (1) and (2) are both present in the same request, their indicated lengths MUST match. The representation's length, if known, MUST stay consistent across subsequent requests. A server can use the problem type {{PROBLEM}} of "https://iana.org/assignments/http-problem-types#inconsistent-representation-length" ({{inconsistent-length}}) in responses to indicate inconsistent length values.
```
GrantGryczan: At the end of this line, I don't know if it should say "length values" anymore since you removed the "indicated values" phrasing from earlier sentences. I would think "length indicators" is better, but then it says "to indicate inconsistent length indicators", which is repetitive. This may not matter, so feel free to disregard.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -390,7 +388,7 @@ A client can start a resumable upload from any request that can carry content by
 
 The `Upload-Complete` header field is set to true if the request content includes the entire representation data that the client intends to upload. This is also a requirement for transparently upgrading to resumable uploads from conventional uploads ({{upgrading-uploads}}).
 
-If the client knows the representation data's length, it SHOULD indicate the length in the request to help the server allocate necessary resources for the upload and provide early feedback if the representation violates a limit ({{upload-limit}}). This indication can be done through the `Upload-Length` header field or the combination of the `Content-Length` and `Upload-Complete: ?1` header fields (see {{upload-length}}).
+If the client knows the representation data's length, it SHOULD indicate the length in the request to help the server allocate necessary resources for the upload and provide early feedback if the representation violates a limit ({{upload-limit}}), as described in {{upload-length}}.
```
GrantGryczan: I assume this is (and has been) intentionally redundant with the same recommendation in the upload length section, for ease of reading/searching?

If so, we probably want to at least phrase "representation's length" consistently:

```suggestion
If the client knows the representation's length, it SHOULD indicate the length in the request to help the server allocate necessary resources for the upload and provide early feedback if the representation violates a limit ({{upload-limit}}), as described in {{upload-length}}.
```

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -542,7 +540,7 @@ A successful response to a `HEAD` request against an upload resource
 
 - MUST include the offset in the `Upload-Offset` header field ({{upload-offset}}),
 - MUST include the `Upload-Complete` header field ({{upload-complete}}) indicating whether a final response was produced from processing the uploaded representation,
-- MUST include the length in the `Upload-Length` header field, unless the client has not supplied one, by providing the corresponding headers as described in ({{upload-length}}),
+- MUST include the representation's length in the `Upload-Length` header field if the client has supplied one through corresponding headers as described in ({{upload-length}}),
```
GrantGryczan: Why don't we just remove "through corresponding headers" since that could be misinterpreted as excluding the final request content length itself as a signal? I know you're referring to when `Upload-Complete: ?1` is present, but I don't think that header itself is what _supplies_ the representation's length. This previously referred to either `Upload-Length` or `Upload-Offset`+`Content-Length`, which actually did supply the length.

```suggestion
- MUST include the representation's length in the `Upload-Length` header field if the client has indicated one as described in ({{upload-length}}),
```

GrantGryczan (changes requested):

GrantGryczan:
This may be out of scope, but do we want to rename the `Upload-Length` header field to `Repr-Length` if it's really just an indicator for the representation's length? This seems more broadly useful than just for resumable uploads, and if an HTTP header for that were ever added in the future, it would make `Upload-Length` completely redundant.

Basing this on `Content-Digest` vs. `Repr-Digest`, we have `Content-Length` but no `Repr-Length`.

draft-ietf-httpbis-resumable-upload.md
Acconut: I'm not sure I understand. The length will always be known when the upload is complete (i.e. request with Upload-Complete: ?1 was fully received) because the full representation has been received at that point.

draft-ietf-httpbis-resumable-upload.md
Acconut: There original phrasing has been chosen on purpose based on feedback from the previous Working Group Last Call (https://github.com/httpwg/http-extensions/pull/3204). So let's keep that and just adjust it to use "representation's length".

```suggestion
- MUST include the representation's length in the `Upload-Length` header field, unless the client has not supplied one, by providing the corresponding headers as described in ({{upload-length}}),
```

Acconut:
> This may be out of scope, but do we want to rename the Upload-Length header field to Repr-Length if it's really just an indicator for the representation's length?

There's value in keeping the header naming consistent (Upload-Complete, Upload-Offset, Upload-Limit, Upload-Length), so the prefix serves a good purpose. In addition, renaming it to Repr-Length would raise more questions about its relation to Content-Length and its general use outside of uploads. I'm not sure there's much to gain here.

draft-ietf-httpbis-resumable-upload.md
```suggestion
If indicators (1) and (2) are both present in the same request, their indicated lengths MUST match. The representation's length, if known, MUST stay consistent across subsequent requests. A server can use the problem type {{PROBLEM}} of "https://iana.org/assignments/http-problem-types#inconsistent-representation-length" ({{inconsistent-length}}) in responses to indicate inconsistent length indicators.
```

draft-ietf-httpbis-resumable-upload.md
GrantGryczan: The server can send an early response with `Upload-Complete: ?1`, deeming the upload complete without a request being sent with `Upload-Complete: ?1`.

draft-ietf-httpbis-resumable-upload.md
GrantGryczan: Yeah, but that was before the representation's length was specified as being deduced from things other than headers. I don't think this phrasing is appropriate anymore. The reason that phrasing was chosen is because it clarifies under what circumstances the server must know the length. Citing the aforementioned indicators rather than "headers" should be equally sufficient.

If you disagree, I'd want to hear @LPardue's opinion on this before this is merged. (I don't consider my comment resolved, so if you're able to un-mark it as resolved, that would be nice so I don't have to start a new thread for this review.)

draft-ietf-httpbis-resumable-upload.md
Acconut: In that case the upload is complete and the server learns the length of the transferred representation. No further data can be appended and the length is final.

draft-ietf-httpbis-resumable-upload.md
Acconut: How about this?

> - MUST include the representation's length in the `Upload-Length` header field, unless the server has not discovered the length through the mechanisms described in ({{upload-length}}),

GrantGryczan: This addresses _my_ problem, but to also address the point of the aforementioned PR, shouldn't it mention the client indicating the length rather than the server discovering it?

draft-ietf-httpbis-resumable-upload.md
GrantGryczan: But the client never finished transferring the representation, and it didn't communicate the length through either of the two length indicators this PR documents, so the server cannot know the full representation's length. Setting aside semantics though, the server implementation required to actually compute the length from a request discarded early seems useless and unnecessarily complex when the server no longer needs to be aware of the length. The server still won't necessarily know the length, and unlike for the 2 indicators, it isn't required to.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -948,10 +946,10 @@ Reference:
 IANA is asked to register the following entry in the "HTTP Problem Types" registry:
 
 Type URI:
-: https://iana.org/assignments/http-problem-types#inconsistent-upload-length
+: https://iana.org/assignments/http-problem-types#inconsistent-representation-length
 
 Title:
-: Inconsistent Upload Length Values
+: Inconsistent Representation Length Values
```
guoye-zhang: Should we keep the problem type inconsistent-upload-length since the new name sounds like it might be more widely applicable?

We already defined upload length to be the representation’s length to be uploaded. I suspect most people are still confused about resource vs representation vs content.

draft-ietf-httpbis-resumable-upload.md
guoye-zhang: Can we just rephrase this as “when the client completes the upload”?

guoye-zhang (approved):

draft-ietf-httpbis-resumable-upload.md
Acconut: > Can we just rephrase this as “when the client completes the upload”?

That's possible. Thinking out load, the consequence would be that if the server serves an early response, the upload either

- is completed and does not have a length, or
- is completed, but `offset < length`.

Both seem fine to me since the server stopped the upload early, so it's not unexpected that the upload is halted in such an intermittent state. Using the suggested phrasing would resolve the topic in a rather simple way, wouldn't it?

draft-ietf-httpbis-resumable-upload.md
Acconut: We can also keep the previous name 👍

draft-ietf-httpbis-resumable-upload.md
GrantGryczan: Yes, it would! (I swore I reacted to Guoye's comment with 👍 before, but it disappeared.)

Acconut: Changed this in 77c5b38a

draft-ietf-httpbis-resumable-upload.md
Acconut: Makes sense. I changed it to this in ab6bae66:

> - MUST include the representation's length in the `Upload-Length` header field, unless the client has not supplied the representation's length as described in ({{upload-length}}),

Let me know if you think this isn't accurate.

draft-ietf-httpbis-resumable-upload.md
Acconut: Reverted to `inconsistent-upload-length` in 1d8cc59d.

Acconut:
@guoye-zhang @GrantGryczan Addressed your remaining feedback. Let me know if you think this PR is ready to be merged now.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -415,7 +413,7 @@ If the `Upload-Complete` request header field is set to true, the client intends
 
 If the `Upload-Complete` request header field is set to false, the client intends to transfer the representation over multiple requests. If the request content was fully processed, the server MUST include the `Location` response header field pointing to the upload resource and MUST include the `Upload-Limit` header field with the corresponding limits if existing. Servers are RECOMMENDED to use the `201 (Created)` status code.
 
-The server MUST record the length according to {{upload-length}} if the `Upload-Length` or `Upload-Complete: ?1` header fields are included in the request.
+The server MUST record the representation's length according to {{upload-length}} if the `Upload-Length` or `Upload-Complete: ?1` header fields are included in the request.
```
GrantGryczan, line 416: I know this is an old clause, but what if the client or server fails? Does this account for that sufficiently? For example,

- if a client makes an `Upload-Complete: ?1` request with no `Content-Length` header that the server rejects after receiving only half its content,
- if the client makes a request with `Upload-Length` that has an invalid value for `Upload-Complete`,
- if the client's headers correctly indicate the length and go through to the server, but the body doesn't get sent due to the client losing connection,
- if the client indicates the length, but the server fails with a 5xx response and is unable to record the length.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -326,21 +326,19 @@ When used in an upload creation response ({{upload-creation}}) or an upload appe
...
+The `Upload-Length` request and response header field is an Item Structured Header Field ({{STRUCTURED-FIELDS}}). Its value is a non-negative Integer ({{Section 3.3.1 of STRUCTURED-FIELDS}}) and indicates the representation's length as a number of bytes. Other values MUST cause the entire header field to be ignored.
 
-If indicators (1) and (2) are both present in the same request, their indicated lengths MUST match. If multiple requests include indicators, their indicated values MUST match. A server can use the problem type {{PROBLEM}} of "https://iana.org/assignments/http-problem-types#inconsistent-upload-length" ({{inconsistent-length}}) in responses to indicate inconsistent length values.
+If indicators (1) and (2) are both present in the same request, their indicated lengths MUST match. The representation's length, if known, MUST stay consistent across subsequent requests. A server can use the problem type {{PROBLEM}} of "https://iana.org/assignments/http-problem-types#inconsistent-upload-length" ({{inconsistent-length}}) in responses to indicate inconsistent length indicators.
```
GrantGryczan, line 338: It says indicators (1) and (2) MUST match, but just within indicator (1), what if the content length doesn't match the `Content-Length` header? Perhaps we don't need to comment on that since RFC 9110 already [prohibits this](https://httpwg.org/specs/rfc9110.html#field.content-length):

> a sender MUST NOT forward a message with a Content-Length header field value that is known to be incorrect.

GrantGryczan (commented):
Other than these 2 questions which I'm uncertain of, this seems good.

draft-ietf-httpbis-resumable-upload.md
Acconut: I don't think we should include normative statements in this document about how messages with inaccurate Content-Length values ought to be handled. It's outside of this document's scope and overlaps with the version-specific details of HTTP. If I'm not mistaken, HTTP/2 and HTTP/3 consider messages with an inaccurate Content-Length as malformed.

draft-ietf-httpbis-resumable-upload.md
Acconut: > * if a client makes an `Upload-Complete: ?1` request with no `Content-Length` header that the server rejects after receiving only half its content,
> * if the client makes a request with `Upload-Length` that has an invalid value for `Upload-Complete`,

Then the length is unknown.

> * if the client's headers correctly indicate the length and go through to the server, but the body doesn't get sent due to the client losing connection,
> * if the client indicates the length, but the server fails with a 5xx response and is unable to record the length.

The server can try to record the length based on the request header without the content, but if that process fails the length remains unknown.

In general, I don't think we have to go into every possible failure scenario in the text, as this makes it too verbose, rigid and restrictive. The document shouldn't be an implementation blueprint. It's good to think about these scenarios and it's great that you bring themup, but the document should focus on areas where interoperability matters. The examples you mentioned either reflect transient failures in the message exchange that can be recovered from using the resumption techniques, or are cases of the client emitting invalid messages, where the file upload will fail regardless of what we specify in the document.

draft-ietf-httpbis-resumable-upload.md
GrantGryczan: That's what I figured, but just making sure.

draft-ietf-httpbis-resumable-upload.md
GrantGryczan: I wasn't suggesting explicitly documenting these edge cases; I was only suggesting that the conditions under which the server is required to record the length are perhaps too restrictive. But if you think situations like the examples I gave are already naturally exempt from this requirement, then feel free to mark as resolved.

PR #3466, opened 2026-06-25
RESUMABLE: Include Upload-Limit in response to limit violation

### LINKED ISSUE #3450: RESUMABLE: Recommend `Upload-Limits` in all 4xx responses caused by limit violations
I suggest we recommend servers include the `Upload-Limits` header in all 4xx responses produced because of a client's violation of upload limits. This recommendation already exists for 413 responses, but I ran into a situation where this would be greatly useful for other status codes:

If a server doesn't want to pay the cost of resumable uploads below a certain `min-size` (as per #3446), I'd expect the server to reject any empty upload creation request that has an `Upload-Length` under that `min-size`, since the overhead of an upload resource may be an undesirable. Ideally, an optimistic client should be able to handle that rejection response by resubmitting the request in one non-resumable request rather than as an upload append.

The problem is, the server has no means to communicate that the response failure was due to a `min-size` violation. Therefore, it seems a generalized, optimistic client has no choice but to handle all 4xx upload creation failures by requesting the upload limits and possibly trying again. Most of the time, this results in extra round trips with no benefit. If the client could see the `Upload-Limits` in that response, another round trip wouldn't be needed.

Acconut:
Closes https://github.com/httpwg/http-extensions/issues/3450.


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -374,6 +374,8 @@ The following recommendations for limit changes can minimize the risk of causing
 
 Receivers of `Upload-Limit` parse the Dictionary as described in {{Section 4.2 of STRUCTURED-FIELDS}}. Where the Dictionary is successfully parsed, this document places two additional requirements on Dictionary members. First, a member with an unknown key MUST be ignored. Second, a member with a known key but a value of unexpected type MUST cause the entire `Upload-Limit` header field to be ignored, or alternatively the complete HTTP message MUST be treated as malformed.
 
+When a request is rejected because limits were violated, the response MUST include the `Upload-Limit` header field carrying all limits of the corresponding upload resource.
```
GrantGryczan: As a client implementer, I love the "MUST", but this may be excessively restrictive for servers. What if the part of the server that enforces the limit is unaware of resumable uploads? It could be an intermediary, or some general request size limit configuration from a framework, etc. The server would merely use `Upload-Limit` to communicate these existing limits, at least for `max-append-size`. Perhaps this should be rephrased to exclude cases like those? Either that, or change it to "SHOULD".

Acconut: That's a valid point. We can reuse the phrasing from a related paragraph in the same section:

> A server responding with a `413 (Content Too Large)` status code as a result of its upload limits SHOULD include the `Upload-Limit` header field in this response if possible. This might not be possible if intermediaries enforce content size restrictions.

GrantGryczan: Sure, which reminds me to comment on that excerpt as well.

GrantGryczan:
This remains unmodified in the text:

> A server responding with a `413 (Content Too Large)` status code as a result of its upload limits SHOULD include the Upload-Limit header field in this response if possible. This might not be possible if intermediaries enforce content size restrictions.

Isn't this excerpt redundant now? If there is still value in naming 413 specifically, then perhaps it would be better to use it as an example at the end of the same paragraph as the new, more general recommendation/requirement. You might argue it's better as is for the sake of the surrounding paragraph's flow, but the other stuff related to client handling of 413 responses will (or at least, should) be restructured by #3275 (or a superseding PR) anyway, so that all retry logic is documented in the same place.

draft-ietf-httpbis-resumable-upload.md
Acconut: Changed it to SHOULD in 7b0a1e73.

Acconut:
> Isn't this excerpt redundant now? If there is still value in naming 413 specifically, then perhaps it would be better to use it as an example at the end of the same paragraph as the new, more general recommendation/requirement. You might argue it's better as is for the sake of the surrounding paragraph's flow, but the other stuff related to client handling of 413 responses will (or at least, should) be restructured by #3275 (or a superseding PR) anyway, so that all retry logic is documented in the same place.

I've removed the redundancy in 7b0a1e73. We can consider moving the response code handling somewhere else, but this shouldn't be done as part of this PR.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -378,9 +378,11 @@ When responding to an `OPTIONS` request without the `Upload-Complete` header fie
 
 A client can use an `OPTIONS` request without the `Upload-Complete` header field to discover whether the resource supports resumable uploads and to learn potential limits before creating an upload resource. To reduce the likelihood of failing requests, the limits announced in an `OPTIONS` response SHOULD NOT be less restrictive than the limits applied to an upload once the upload resource has been created, unless the request to create an upload resource included additional information that warrants different limits. For example, a server might announce a general maximum size limit of 1 GB, but reduce it to 100 MB when the media type indicates an image.
 
+When a request is rejected because limits were violated, the response SHOULD include the `Upload-Limit` header field carrying all limits of the corresponding upload resource if possible. This might not be possible if intermediaries enforce restrictions.
```
GrantGryczan: Nit: "if intermediaries enforce restrictions" is rather vague now. What kind of restrictions? This may be better:

```suggestion
When a request is rejected because limits were violated, the response SHOULD include the `Upload-Limit` header field carrying all limits of the corresponding upload resource if possible. This might not be possible if intermediaries enforce some limits.
```

Also, not a nit: "the corresponding upload resource" might not exist at the time of this response if a client makes an upload creation request, and the server responds with a rejection due to a violated limit before an upload resource is created. Not sure how to phrase that.

GrantGryczan (changes requested):

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -548,7 +550,7 @@ A successful response to a `HEAD` or `GET` request against an upload resource
 
 The resource SHOULD NOT generate a response with the `301 (Moved Permanently)` and `302 (Found)` status codes because clients might follow the redirect without preserving the `HEAD` method.
 
-A client does not require response content for an offset retrieval request in order to successfully resume an upload. Therefore, serving response content for a `GET` request is unexpected. Its meaning is not defined by this protocol.
```
guoye-zhang: GitHub is showing a change here but I can't see the difference. Was this changed?

guoye-zhang (approved):

draft-ietf-httpbis-resumable-upload.md
Acconut: That's odd, the paragraph didn't change. Maybe I messed something up when merging main and this is an artifact of that.

draft-ietf-httpbis-resumable-upload.md
Acconut: > Also, not a nit: "the corresponding upload resource" might not exist at the time of this response if a client makes an upload creation request, and the server responds with a rejection due to a violated limit before an upload resource is created. Not sure how to phrase that.

We can refer to just the upload, instead of its resource:

```suggestion
When a request is rejected because limits were violated, the response SHOULD include the `Upload-Limit` header field carrying all limits of the corresponding upload if possible. This might not be possible if intermediaries enforce limits.
```

PR #3470, opened 2026-06-30
RESUMABLE: Clarify when client cannot know limits

### LINKED ISSUE #3321: resumable: Upload-Limit language is ambiguous
Spinning off from https://github.com/httpwg/http-extensions/issues/3193#issuecomment-3469842535

The spec as is (draft 10) is unclear on whether clients MUST respect limits a server sends down via a Upload-Limit header.

"The client SHOULD respect any limits (Section 4.1.4) announced in the Upload-Limit header field in interim or final responses" -> implies optional but strongly recommended

"max-append-size: Specifies a maximum size counted in bytes for the request content in a single upload append request (Section 4.4). The server might reject requests exceeding this limit. A client that is aware of this limit MUST NOT send larger upload append requests. The value is an Integer." -> implies clients are required to respect this limit (since why wouldn't they be "aware" of it? It's included in the spec, and clients are guaranteed to have received it in the Upload Creation response if the server sets it).

Overall I'd be keen to make as clear as possible that it's (at least) strongly recommended that clients which intend to be maximally interoperable with a variety of servers respect all known Upload-Limits. 

I'm happy to take a stab at rewording to try to remove this ambiguity if helpful, but lets first align on the desired end state.

Acconut:
Closes https://github.com/httpwg/http-extensions/issues/3321.

The current draft says:

> The client SHOULD respect any limits ([Section 4.1.4](https://httpwg.org/http-extensions/draft-ietf-httpbis-resumable-upload.html#upload-limit)) announced in the Upload-Limit header field in interim or final responses.

This statement is rather vague. Clients cannot always know the limits before starting an upload and we likely don't want to force clients to discover them via an OPTIONS request first. That's why we say that clients SHOULD respect any limit that they are aware of (either discovered via OPTIONS or in responses), while not forcing them to adhere via MUST.

This PR attempts to add a bit more context around this statement to hopefully clear this up.


guoye-zhang (approved):

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -392,7 +392,7 @@ The `Upload-Complete` header field is set to true if the request content include
...
+Clients are not required to discover limits ({{upload-limit}}) before starting the upload and might therefore not be aware of limits enforced by the server when starting the initial requests. However, the client SHOULD respect any limits ({{upload-limit}}) announced during the upload process in the `Upload-Limit` header field in interim or final responses. In particular, if the allowed maximum size is less than the amount of representation data the client intends to upload, the client SHOULD stop the current request immediately and cancel the upload ({{upload-cancellation}}). If the client knows that the representation data is smaller than `min-size`, it cannot expect resumability to be offered. The client might still attempt to transfer the representation in a single request, either in a request with the `Upload-Complete` header field set to true (see {{upgrading-uploads}}) or via a conventional upload.
```
GrantGryczan: Some nits:

I think this makes the first sentence more concise.

```suggestion
Clients are not required to discover limits ({{upload-limit}}) before starting the upload and might therefore be initially unaware of limits enforced by the server. However, the client SHOULD respect any limits ({{upload-limit}}) announced during the upload process in the `Upload-Limit` header field in interim or final responses. In particular, if the allowed maximum size is less than the amount of representation data the client intends to upload, the client SHOULD stop the current request immediately and cancel the upload ({{upload-cancellation}}). If the client knows that the representation data is smaller than `min-size`, it cannot expect resumability to be offered. The client might still attempt to transfer the representation in a single request, either in a request with the `Upload-Complete` header field set to true (see {{upgrading-uploads}}) or via a conventional upload.
```

In the second sentence, is it necessary to say "during the upload process" when we already say "in interim or final responses"? It's unclear to me why you added that.

```suggestion
Clients are not required to discover limits ({{upload-limit}}) before starting the upload and might therefore be initially unaware of limits enforced by the server. However, the client SHOULD respect any limits ({{upload-limit}}) announced in the `Upload-Limit` header field in interim or final responses. In particular, if the allowed maximum size is less than the amount of representation data the client intends to upload, the client SHOULD stop the current request immediately and cancel the upload ({{upload-cancellation}}). If the client knows that the representation data is smaller than `min-size`, it cannot expect resumability to be offered. The client might still attempt to transfer the representation in a single request, either in a request with the `Upload-Complete` header field set to true (see {{upgrading-uploads}}) or via a conventional upload.
```

GrantGryczan (changes requested):

draft-ietf-httpbis-resumable-upload.md
Acconut: > I think this makes the first sentence more concise.

That's a good suggestion, thanks.

> In the second sentence, is it necessary to say "during the upload process" when we already say "in interim or final responses"? It's unclear to me why you added that.

I want to highlight that limits can be discovered while the upload is ongoing even if the client didn't fetch the limits beforehand and this wording makes this clearer for me.

PR #3471, opened 2026-07-01
RESUMABLE: remove section about Integrity

### LINKED ISSUE #3228: RESUMABLE: Section 10.1 description of how to use Repr-Digest seems wrong

> In Section 10.1, regarding integrity verification, could Repr-Digest also be on the HEAD of the upload resource?

However, looking even further at the text I'm not 100% sure the suggested interactions are correct (and I probably should have caught this earler). 

First, we state this use case which seems fine

> If the client knows the integrity digest of the entire representation data before creating an upload resource, it can include the Repr-Digest header field when creating an upload ([Section 4.2](https://github.com/httpwg/http-extensions/issues/new#upload-creation)). Once the upload is completed, the server can compute the integrity digest of the received representation data and compare it to the provided digest. If the digests don't match, the server SHOULD consider the upload failed, not process the representation further, and signal the failure to the client. This way, the integrity of the entire representation data can be protected.

In other words, the client knew the full data and sent it over for the server to eventually use once it has all the data.

The bit I'm less sure about is 

> the server can compute the digest continuously during the upload and include the Repr-Digest header field in responses to upload creation ([Section 4.2](https://github.com/httpwg/http-extensions/issues/new#upload-creation)) and upload appending requests ([Section 4.4](https://github.com/httpwg/http-extensions/issues/new#upload-appending)) even when the upload is not completed yet.

If the upload is, for example, 1000 bytes of content but the server only received 500 bytes, the hash it would produce over 500 bytes is not the same as over 1000. So is this the representation of the upload resource? Does HTTP allow that to change over time (I think it does). If we agree, the text could probably be clearer on that point. Then, we can answer Mike's original point that Offset Retrieval (HEAD request) returns the state of the upload resource including a Repr-Digest.

LPardue:
While this section was well intentioned, it has not done a great
job of tracking protocol updates in other areas. On closer inspection,
more and more problems reveal themselves. This would take non-trivial
effort to gain consensus and write text that is correct and of equivalent
quality to the other parts of the document.

The integrity aspect is not part of the core resumable upload protocol.
Its possible for people to use the HTTP building blocks like Digest fields
to enhance a deployment and build something that would work. It would be
nice to include text on the topic but without deployment experience we
risk trying to codify something that is not yet well exercised. Delaying
the protocol publication for this topic may hurt to others that do not
intend to use integrity.

As such, this PR proposes to remove the text. By saying nothing on the
topic we do not preclude others from gaining deployment experience and
doing future work to document it (perhaps as an adopted WG item).

Closes #3228

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -765,29 +765,6 @@ Since the codings listed in `Content-Encoding` are a characteristic of the repre
 
 Unlike `Content-Encoding` (see {{Section 8.4.1 of HTTP}}), `Transfer-Encoding` (see {{Section 6.1 of RFC9112}}) is a property of the message, not of the representation. Moreover, transfer codings can be applied in transit (e.g., by proxies). This means that a client does not have to consider the transfer codings to compute the upload offset, while a server is responsible for transfer decoding the message before computing the upload offset. The same applies to the value of `Upload-Length`. Please note that the `Content-Length` header field cannot be used in conjunction with the `Transfer-Encoding` header field.
 
-# Integrity Digests
-
-The integrity of an entire upload or individual upload requests can be verifying using digests from {{DIGEST-FIELDS}}.
-
-## Representation Digests
-
-Representation digests help verify the integrity of the entire representation data that has been uploaded so far, which might stretch across multiple requests.
-
-If the client knows the integrity digest of the entire representation data before creating an upload resource, it can include the `Repr-Digest` header field when creating an upload ({{upload-creation}}). Once the upload is completed, the server can compute the integrity digest of the representation data and compare it to the provided digest. If the digests don't match, the server SHOULD consider the upload failed, not process the representation further, and signal the failure to the client. This way, the integrity of the entire representation data can be protected.
-
-Alternatively, when creating an upload ({{upload-creation}}), the client can ask the server to compute and return the integrity digests using a `Want-Repr-Digest` field conveying the preferred algorithms.
-The response SHOULD include at least one of the requested digests, but might not include it.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -548,7 +548,7 @@ A successful response to a `HEAD` or `GET` request against an upload resource
 
 The resource SHOULD NOT generate a response with the `301 (Moved Permanently)` and `302 (Found)` status codes because clients might follow the redirect without preserving the `HEAD` method.
 
-A client does not require response content for an offset retrieval request in order to successfully resume an upload. Therefore, serving response content for a `GET` request is unexpected. Its meaning is not defined by this protocol.
+A client does not require response content for an offset retrieval request in order to successfully resume an upload. Therefore, serving response content for a `GET` request is unexpected. Its meaning is not defined by this protocol.
```
LPardue, line 551: I can't see what apparently changed here. There was no intentional change.

GrantGryczan: No change, just a GitHub bug it seems.

guoye-zhang (approved):

draft-ietf-httpbis-resumable-upload.md
Acconut: I was an inconsistent line ending that likely your editor and/or git normalized, but GitHub doesn't show this well. I've fixed it on main and updated the PR, so it disappeared.

Acconut (approved):
Thank you, Lucas!

Acconut:
Lucas posted the editor's intent to remove the integrity digests section on the mailing list (https://lists.w3.org/Archives/Public/ietf-http-wg/2026JulSep/0000.html) and so far no objections have been raised. I'm going to merge this for now unless there is consensus to keep it.

PR #3477, opened 2026-07-06
NVS: Address AD review comments

nidhijaju:


MikeBishop (approved):
These all look good. Thank you!

PR #3479, opened 2026-07-06
RESUMABLE: Clarify offset handling

### LINKED ISSUE #3222: RESUMABLE: it might be reasonable for the client to require that the offset be no less than the last Upload-Offset

> In 4.3.1, it might be reasonable for the client to require that the offset be no less than the last Upload-Offset that it received in interim responses during the previous upload. After all, we already said the client can discard that data.

### LINKED ISSUE #3451: RESUMABLE: Client-side upload offset handling seems under-specified
While implementing my generalized resumable upload client, I ran into several edge cases related to `Upload-Offset` as a response header that I seeked for guidance from the draft on, but the draft doesn't say anything about them.

1. _(This is the only purely editorial issue.)_ In a 2xx response to a creation/append request, a client can determine the new upload offset for the `Upload-Offset` response header based on the length of the content it uploaded added to the previous offset. For example, if the client sent `Upload-Offset: 100` and `Content-Length: 50`, the client knows from a 2xx response that the new offset should be 150. This is implied by the [definition of upload offset](https://www.ietf.org/archive/id/draft-ietf-httpbis-resumable-upload-11.html#section-4.1.1-1). It's also specified that the client can [free previous resources](https://www.ietf.org/archive/id/draft-ietf-httpbis-resumable-upload-11.html#section-4.1.1-4) upon receipt of a new `Upload-Offset` value, but the same is not specified for changes to the upload offset in general. I think this suggestion may be better communicated if generalized to the offset implied by responses in general (e.g., as implied by a 2xx response), rather than specific to the `Upload-Offset` header.
2. In a 2xx response to a creation/append request, how should a client treat an `Upload-Offset` value inconsistent with what the client requested? Reusing the previous example, what should the client do if the server replies to that request with `Upload-Offset: 125` or `Upload-Offset: 200` instead of the correct `Upload-Offset: 150`?
3. In an offset retrieval response, how should a client treat an `Upload-Offset` value below its current known offset? In other words, what should the client do if the server tells it that the offset moved backwards?
4. Similarly, should a client handle the case that the offset moved forward further ahead than it knows it ever uploaded the representation, since that implies the server messed up? Or should the client just take that for granted, since that allows a simpler implementation?
5. In an offset retrieval response, how should a client treat a missing or invalid `Upload-Offset`? The [definition of the `Upload-Offset` header](https://www.ietf.org/archive/id/draft-ietf-httpbis-resumable-upload-11.html#section-4.1.1-3) currently says syntactically invalid or negative values "MUST cause the entire header field to be ignored", but ignoring the header on offset retrieval responses is unacceptable, which seems like an oversight in this definition.

I'm guessing that, for some of these, we'd want the client to abort its current request and cancel the upload, similarly to [when the server messes up](https://www.ietf.org/archive/id/draft-ietf-httpbis-resumable-upload-11.html#section-4.2.2-7) the `Location` response header.

### LINKED ISSUE #3454: RESUMABLE: Recommend `Upload-Offset` in 4xx responses with `Upload-Complete: ?0`
For a client handling a 4xx creation/append response with `Upload-Complete: ?0`, it can resume the upload by sending an offset retrieval request, waiting for the response, and then submitting an append request at the retrieved offset. But this isn't necessary if the response includes an up-to-date `Upload-Offset` value already; then, a client can just use that and save a round trip, immediately resuming the upload at the given offset without performing offset retrieval.

This is especially problematic for servers that require that incomplete appends conform to a particular chunk granularity. Since clients have no way to tell what a server's chunk granularity is, a server with a chunk granularity requirement will respond to nearly every incomplete creation/append with a non-completing 4xx response (as per https://github.com/httpwg/http-extensions/issues/3193#issuecomment-3893170872), since most request content lengths won't be a multiple of the server's chunk granularity. Then the client has to perform offset retrieval to send every single chunk but the first and last. This puts unnecessary load on clients and servers wherever chunk granularity is involved.

I'd strongly argue we should recommend servers include the `Upload-Offset` header in 4xx responses with `Upload-Complete: ?0`. We don't have to recommend clients actually use this information, since as Guoye Zhang said in [a comment](https://github.com/httpwg/http-extensions/pull/3319#discussion_r2474114941), that would make client implementations more complex. But clients should at least have the option to opt into that sometimes-very-significant performance optimization, and I expect it to be very simple for servers to include that information.

~~[This paragraph](https://www.ietf.org/archive/id/draft-ietf-httpbis-resumable-upload-11.html#section-4.4.1-2) should also be modified to mention this as one of the sources clients are recommended to take the offset from, though I'm not sure how I'd phrase that.~~ EDIT: I'm not sure why I thought any change was needed there. Looking at it again, it seems like it already covers this.

This could also be extended to ~~3xx-5xx~~ 3xx responses that contain the `Upload-Complete: ?0` header. There is no point in recommending this for 2xx since the offset after a non-completing 2xx response can be inferred from the previous known offset + the requested content length. I also wouldn't recommend extending this to any responses without the `Upload-Complete` header, since those responses are often generated without awareness of resumable uploads.

Acconut:
This PR addresses two comments:

From https://github.com/httpwg/http-extensions/issues/3451:

> In a 2xx response to a creation/append request, a client can determine the new upload offset for the Upload-Offset response header based on the length of the content it uploaded added to the previous offset. For example, if the client sent Upload-Offset: 100 and Content-Length: 50, the client knows from a 2xx response that the new offset should be 150. This is implied by the [definition of upload offset](https://www.ietf.org/archive/id/draft-ietf-httpbis-resumable-upload-11.html#section-4.1.1-1). It's also specified that the client can [free previous resources](https://www.ietf.org/archive/id/draft-ietf-httpbis-resumable-upload-11.html#section-4.1.1-4) upon receipt of a new Upload-Offset value, but the same is not specified for changes to the upload offset in general. I think this suggestion may be better communicated if generalized to the offset implied by responses in general (e.g., as implied by a 2xx response), rather than specific to the Upload-Offset header.

To address this, mention that a client can learn the offset implicitly through responses.

From https://github.com/httpwg/http-extensions/issues/3222:

> In 4.3.1, it might be reasonable for the client to require that the offset be no less than the last Upload-Offset that it received in interim responses during the previous upload. After all, we already said the client can discard that data.

The text has been expanded to that the client must never observe a decreasing offset value, regardless through which means.

Closes https://github.com/httpwg/http-extensions/issues/3451.
Closes https://github.com/httpwg/http-extensions/issues/3222.
Closes https://github.com/httpwg/http-extensions/issues/3454.

### DIFF
--- a/draft-ietf-httpbis-resumable-upload.md
+++ b/draft-ietf-httpbis-resumable-upload.md
@@ -226,7 +226,7 @@ Client                                       Server
 ~~~
 {: #fig-upload-creation-incomplete title="Upload creation with partial representation data"}
 
-2) Next, intermediate parts are appended ({{upload-appending}}) with the `Upload-Complete` field value set to false, indicating that they are not the last part of the representation data. The offset value in the `Upload-Offset` header field is taken from the previous response when creating the upload or appending to it.
+2) Next, intermediate parts are appended ({{upload-appending}}) with the `Upload-Complete` field value set to false, indicating that they are not the last part of the representation data.
 
 ~~~ aasvg
 Client                                       Server
@@ -307,15 +307,19 @@ The state of an upload consists of the following properties that are tracked by 
 
 ### Offset {#upload-offset}
 
-The offset is the number of bytes from the representation data that have been processed, either during the creation of the upload resource ({{upload-creation}}) or by appending to it ({{upload-appending}}). The offset can be retrieved from the upload resource ({{offset-retrieving}}) and is required when appending representation data ({{upload-appending}}) to synchronize the client and resource regarding the amount of transferred representation data.
+The offset is the number of bytes from the representation data that have been processed, either during the creation of the upload resource ({{upload-creation}}) or by appending to it ({{upload-appending}}). The offset is required when appending representation data ({{upload-appending}}) to synchronize the client and resource regarding the amount of transferred representation data.
 
 The offset reflects application-level processing for the upload. Data may have been delivered and acknowledged at the transport layer without yet being reflected in the offset.
 
-Representation data processed by the upload resource cannot be removed again and, therefore, the offset MUST NOT decrease. If the server loses any part of the state, it MUST deactivate the upload resource and reject further interaction with it.
-
 The `Upload-Offset` request and response header field conveys the offset. `Upload-Offset` is an Item Structured Header Field ({{STRUCTURED-FIELDS}}). Its value is a non-negative Integer ({{Section 3.3.1 of STRUCTURED-FIELDS}}) and indicates the current offset as viewed by the message sender. Other values MUST cause the entire header field to be ignored.
 
-The `Upload-Offset` header field in responses serves as an acknowledgement of the processed representation data and as a guarantee that no retransmission of it will be necessary. Clients can use this guarantee to free resources associated to transferred representation data.
+The client obtains the current offset in multiple ways:
+
+- The offset can be explicitly retrieved from the upload resource ({{offset-retrieving}}).
+- Interim and final responses when creating the upload ({{upload-creation}}) or appending ({{upload-appending}}) can include the `Upload-Offset` header field.
+- Final responses for appending representation data ({{upload-appending}}) with a `2XX` status code or the `Upload-Complete: ?1` acknowledge that the entire request content was received and the offset increased by the length of the appended representation data.
+
+Representation data processed by the upload resource cannot be removed again and, therefore, the offset as seen by the client MUST NOT decrease. Clients can use this guarantee to free resources associated to transferred representation data, as no retransmission of it will be necessary. If the server loses any part of the state, it MUST deactivate the upload resource and reject further interaction with it.
 
 ### Completeness {#upload-complete}
 
@@ -577,7 +581,7 @@ Cache-Control: no-store
 
 A client can continue the upload and append representation data by sending a `PATCH` request with the `application/partial-upload` media type ({{media-type-partial-upload}}) to the upload resource. The request content is the representation data to append. The request can benefit from incremental delivery; see {{incremental}}.
 
-The client MUST indicate the offset of the request content inside the representation data by including the `Upload-Offset` header field. To ensure that the upload resource will accept the request, the offset SHOULD be taken from an immediate previous response for retrieving the offset ({{offset-retrieving}}) or appending representation data ({{upload-appending}}).
+The client MUST indicate the offset of the request content inside the representation data by including the `Upload-Offset` header field. To ensure that the upload resource will accept the request, the offset SHOULD be taken from an immediate previous response for retrieving the offset ({{offset-retrieving}}), creating the upload ({{upload-creation}}), or appending representation data ({{upload-appending}}).
 
 The request MUST include the `Upload-Complete` header field. Its value is true in two cases:
 
@@ -694,7 +698,7 @@ HTTP/1.1 204 No Content
 
 Resumable uploads, as defined in this document, do not permit uploading representation data in parallel to the same upload resource. The client MUST NOT perform multiple representation data transfers for the same upload resource in parallel.
 
-Even if the client is well-behaved and doesn't send concurrent requests, network interruptions can occur in such a way that the client considers a request as failed while the server is unaware of the problem and considers the request still ongoing. The client might then try to resume the upload with the best intentions, resulting in concurrent requests from the server's perspective. Therefore, the server MUST take measures to prevent race conditions, data loss and corruption from concurrent requests to append representation data ({{upload-appending}}) and/or cancellation ({{upload-cancellation}}) to the same upload resource. In addition, the server MUST NOT send outdated information in responses when retrieving the offset ({{offset-retrieving}}). This means that the offset sent by the server MUST be accepted in a subsequent request to append representation data if no other request to append representation data or cancel was received in the meantime. In other words, clients have to be able to use received offsets.
+Even if the client is well-behaved and doesn't send concurrent requests, network interruptions can occur in such a way that the client considers a request as failed while the server is unaware of the problem and considers the request still ongoing. The client might then try to resume the upload with the best intentions, resulting in concurrent requests from the server's perspective. Therefore, the server MUST take measures to prevent race conditions, data loss and corruption from concurrent requests to append representation data ({{upload-appending}}) and/or cancellation ({{upload-cancellation}}) to the same upload resource. In addition, the server MUST NOT send outdated information in responses. This means that the offset communicated through responses as defined in {{upload-offset}} MUST be accepted in a subsequent request to append representation data if no other request to append representation data or cancel was received in the meantime. In other words, clients have to be able to use received offsets.
 
 The RECOMMENDED approach is as follows: If a server receives a new request to retrieve the offset ({{offset-retrieving}}), append representation data ({{upload-appending}}), or cancel the upload ({{upload-cancellation}}) while a previous request for creating the upload ({{upload-creation}}) or appending representation data ({{upload-appending}}) to the same upload resource is still ongoing, the server SHOULD prevent race conditions, data loss, and corruption by terminating the previous request before processing the new request. Due to network delay and reordering, the server might still be receiving representation data from an ongoing transfer for the same upload resource, which in the client's perspective has failed. Since the client is not allowed to perform multiple transfers for the same upload resource in parallel, the server can assume that the previous attempt has already failed. Therefore, the server MAY abruptly terminate the previous HTTP connection or stream.
 


draft-ietf-httpbis-resumable-upload.md
```diff
@@ -307,15 +307,19 @@ The state of an upload consists of the following properties that are tracked by
...
 
-The `Upload-Offset` header field in responses serves as an acknowledgment of the processed representation data and as a guarantee that no retransmission of it will be necessary. Clients can use this guarantee to free resources associated to transferred representation data.
+The client obtains the current offset in multiple ways:
+
+- The offset can be explicitly retrieved from the upload resource ({{offset-retrieving}}).
+- Interim and final responses when creating the upload ({{upload-creation}}) or appending ({{upload-appending}}) can include the `Upload-Offset` header field.
+- Final responses for appending representation data ({{upload-appending}}) with a `2XX` status code or the `Upload-Complete: ?1` acknowledge that the entire request content was received and the offset increased by the length of the appended representation data.
```
guoye-zhang, line 320: I think the offset is only relevant for 2xx with `Upload-Complete: ?0`. If `Upload-Complete: ?1`, the offset is no longer relevant since the client won't be resuming.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -575,16 +581,23 @@ Cache-Control: no-store
 
 A client can continue the upload and append representation data by sending a `PATCH` request with the `application/partial-upload` media type ({{media-type-partial-upload}}) to the upload resource. The request content is the representation data to append. The request can benefit from incremental delivery; see {{incremental}}.
 
-The client MUST indicate the offset of the request content inside the representation data by including the `Upload-Offset` header field. To ensure that the upload resource will accept the request, the offset SHOULD be taken from an immediate previous response for retrieving the offset ({{offset-retrieving}}) or appending representation data ({{upload-appending}}).
+The client MUST indicate the offset of the request content inside the representation data by including the `Upload-Offset` header field. To ensure that the upload resource will accept the request, the offset SHOULD be taken from an immediate previous response for retrieving the offset ({{offset-retrieving}}), creating the upload ({{upload-creation}}), or appending representation data ({{upload-appending}}).
```
GrantGryczan, line 584: What about taking it implicitly from 2xx responses?

draft-ietf-httpbis-resumable-upload.md
GrantGryczan, line 701: This "MUST be accepted" requirement should exclude offsets from interim responses. I advise changing "if no other request to append representation data or cancel was received in the meantime" to be more general, including ongoing requests as well as upload creation requests.

GrantGryczan (changes requested):
Excellent! I love the direction of this PR.

GrantGryczan (changes requested):

draft-ietf-httpbis-resumable-upload.md
GrantGryczan, line 320: Are you sure we don't want to comment on the possibility of these sources of information simultaneously contradicting each other (which was the 2nd point in #3451)? I won't argue if not, but I just want to double-check, because it _really_ stands out as odd to me that this real possibility isn't clarified, especially with this new framing.

jjrauch32-lab (approved):

draft-ietf-httpbis-resumable-upload.md
```suggestion
- Final responses for appending representation data ({{upload-appending}}) with a `2XX` status code and the `Upload-Complete: ?0` acknowledge that the entire request content was received and the offset increased by the length of the appended representation data.
```

That seems about right.

Acconut: In those cases, the Upload-Offset value should take the presedence, I think, as the sever can move the offset forward as well. But it's also not a big deal If a client doesn't check that value since it will discover the correct one after a 409.

draft-ietf-httpbis-resumable-upload.md
Acconut: That's also allowed. The line doesn't mention that it SHOULD be taken from the response's header. That omission is intentional and means that taking it implicitly is also fine.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -307,15 +307,19 @@ The state of an upload consists of the following properties that are tracked by
...
 
-The `Upload-Offset` header field in responses serves as an acknowledgment of the processed representation data and as a guarantee that no retransmission of it will be necessary. Clients can use this guarantee to free resources associated to transferred representation data.
+The client obtains the current offset in multiple ways:
+
+- The offset can be explicitly retrieved from the upload resource ({{offset-retrieving}}).
+- Interim and final responses when creating the upload ({{upload-creation}}) or appending ({{upload-appending}}) can include the `Upload-Offset` header field.
+- Final responses for appending representation data ({{upload-appending}}) with a `2XX` status code and the `Upload-Complete: ?0` acknowledge that the entire request content was received and the offset increased by the length of the appended representation data.
```
Acconut:
```suggestion
- Final responses for appending representation data ({{upload-appending}}) with a `2XX` status code and the `Upload-Complete: ?0` acknowledge that the entire request content was received and the offset increased by the length of the appended representation data. However, the value of an `Upload-Offset` header field has presedence, if present.
```

draft-ietf-httpbis-resumable-upload.md
Acconut: > This "MUST be accepted" requirement should exclude offsets from interim responses

We can say that only final responses count.

> I advise changing "if no other request to append representation data or cancel was received in the meantime" to be more general, including ongoing requests as well as upload creation requests.

Can you explain the reason why this would be necessary? Do you have a concrete suggestion for the text?

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -687,28 +698,15 @@ HTTP/1.1 204 No Content
...
+Even if the client is well-behaved and doesn't send concurrent requests, network interruptions can occur in such a way that the client considers a request as failed while the server is unaware of the problem and considers the request still ongoing. The client might then try to resume the upload with the best intentions, resulting in concurrent requests from the server's perspective. Therefore, the server MUST take measures to prevent race conditions, data loss and corruption from concurrent requests to append representation data ({{upload-appending}}) and/or cancellation ({{upload-cancellation}}) to the same upload resource. In addition, the server MUST NOT send outdated information in responses. This means that the offset communicated through responses as defined in {{upload-offset}} MUST be accepted in a subsequent request to append representation data if no other request to append representation data or cancel was received in the meantime. In other words, clients have to be able to use received offsets.
```
Acconut:
```suggestion
Even if the client is well-behaved and doesn't send concurrent requests, network interruptions can occur in such a way that the client considers a request as failed while the server is unaware of the problem and considers the request still ongoing. The client might then try to resume the upload with the best intentions, resulting in concurrent requests from the server's perspective. Therefore, the server MUST take measures to prevent race conditions, data loss and corruption from concurrent requests to append representation data ({{upload-appending}}) and/or cancellation ({{upload-cancellation}}) to the same upload resource. In addition, the server MUST NOT send outdated information in responses. This means that the offset communicated through final responses as defined in {{upload-offset}} MUST be accepted in a subsequent request to append representation data if no other request to append representation data or cancel was received in the meantime. In other words, clients have to be able to use received offsets.
```

Acconut:
@GrantGryczan @guoye-zhang I tried to update the PR following your suggestions. Please review the changes in https://github.com/httpwg/http-extensions/pull/3479/changes/bab2069048e1fcda1c9f3bfcb08774f3ae98550d..93c0425f848e4a0561e4d6a32a43a7e9f110c3ac and let me know if it's still not correct.

draft-ietf-httpbis-resumable-upload.md
GrantGryczan: > We can say that only final responses count.

That works.

> Can you explain the reason why this would be necessary? Do you have a concrete suggestion for the text?

I don't remember what I was trying to say there, and I currently don't see any issue with that part, so disregard.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -307,15 +307,19 @@ The state of an upload consists of the following properties that are tracked by
...
 
-The `Upload-Offset` header field in responses serves as an acknowledgment of the processed representation data and as a guarantee that no retransmission of it will be necessary. Clients can use this guarantee to free resources associated to transferred representation data.
+The client obtains the current offset in multiple ways:
+
+- The offset can be explicitly retrieved from the upload resource ({{offset-retrieving}}).
+- Interim and final responses when creating the upload ({{upload-creation}}) or appending ({{upload-appending}}) can include the `Upload-Offset` header field.
+- Final responses for appending representation data ({{upload-appending}}) with a `2XX` status code and the `Upload-Complete: ?0` acknowledge that the entire request content was received and the offset increased by the length of the appended representation data. However, the value of an `Upload-Offset` header field has presedence, if present.
```
GrantGryczan: `Upload-Offset` can't take precedence over the implicit offset progress from a 2xx response. Otherwise, the `Upload-Offset` header may indicate an earlier offset than the offset implied by the successful append progress, which would be invalid. I'd consider that backwards offset progress, which isn't allowed by the protocol as per the next paragraph. You could fix this by adding that servers must not set this offset to be smaller than the one implied by 2xx.

GrantGryczan (changes requested):

draft-ietf-httpbis-resumable-upload.md
Acconut: How about this then?

```suggestion
- Final responses for appending representation data ({{upload-appending}}) with a `2XX` status code and the `Upload-Complete: ?0` acknowledge that the entire request content was received and the offset increased by the length of the appended representation data. The value of an `Upload-Offset` header field takes precedence, if present. However, its value MUST NOT be smaller than the offset derived by adding the appended representation data's length.
```

GrantGryczan: LGTM, thanks!

GrantGryczan: Although, doesn't this mean clients now go against this recommendation if they don't make `Upload-Offset` take precedence over 2xx?

https://github.com/httpwg/http-extensions/blob/bab2069048e1fcda1c9f3bfcb08774f3ae98550d/draft-ietf-httpbis-resumable-upload.md?plain=1#L584

If so, while I don't personally mind, I think this might go against your original simplicity goal.

Acconut: Clients can choose to not check Upload-Offset in responses, yes. The potential penalty is that they don't notice if the server suddenly advances the offset after a request, which it is allowed to do. But that will reconcile in the next PATCH request resulting in a 409. So I'm not concerned about this.

GrantGryczan: I didn't mean there's an issue with ignoring the `Upload-Offset`. I meant that _because_ there's _no_ issue with ignoring the `Upload-Offset`, it may be undesirable that the cited recommendation could be interpreted as being against ignoring the `Upload-Offset`, since this change defines `Upload-Offset` as taking precedence in the list of ways clients are able to determine the offset.

Acconut: I see. I think we can keep it like that for now and see if this raises questions in later reviews.

GrantGryczan: Given simplicity being a goal, I don't feel terribly comfortable leaving it this way myself. I interpret ignoring `Upload-Offset` to go against the protocol's recommendations when framed this way, and I'd be surprised if new readers of the protocol didn't. Perhaps clarifying that the precedence is optional would work? But really, I'm not sure what precedence even practically entails if it's intended to be equally acceptable for clients to choose either option, so it might suffice to remove the definition of precedence entirely. The bigger problem was the missing constraint against `Upload-Offset` being under the implied 2xx offset, which is now solved by the "MUST NOT".

Acconut: A 2XX response for a PATCH request should mean that the representation data has been appended exactly as supplied, meaning that the offset advanced by the content's length. Any deviation from that should result in a non-2XX response, signaling the client to resync the offset (either via HEAD or the offset in the response). The only valid value for an Upload-Offset in a PATCH's 2XX is `old_offset + content_length`. It then becomes irrelevant whether the client checks Upload-Offset for 2XX and the precedence statement looses its meaning.

GrantGryczan: If that's the case, perfect, then both the precedence statement and the "MUST NOT" should certainly be removed, and it would be much clearer to add "for non-2xx responses" to the `Upload-Offset` point instead. That would certainly help simplify client implementation.

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -307,15 +307,19 @@ The state of an upload consists of the following properties that are tracked by
...
 
-The `Upload-Offset` header field in responses serves as an acknowledgment of the processed representation data and as a guarantee that no retransmission of it will be necessary. Clients can use this guarantee to free resources associated to transferred representation data.
+The client obtains the current offset in multiple ways:
+
+- The offset can be explicitly retrieved from the upload resource ({{offset-retrieving}}).
+- Interim and final responses when creating the upload ({{upload-creation}}) or appending ({{upload-appending}}) can include the `Upload-Offset` header field.
+- Final responses for appending representation data ({{upload-appending}}) with a `2XX` status code and the `Upload-Complete: ?0` acknowledge that the entire request content was received and the offset increased by the length of the appended representation data. The value of an `Upload-Offset` header field takes precedence, if present. However, its value MUST NOT be smaller than the offset derived by adding the appended representation data's length.
```
Acconut:
```suggestion
- Interim and final responses when creating the upload ({{upload-creation}}) or appending ({{upload-appending}}) can include the `Upload-Offset` header field.
- Final responses for appending representation data ({{upload-appending}}) with a `2XX` status code and the `Upload-Complete: ?0` acknowledge that the entire request content was received and the offset increased by the length of the appended representation data. An `Upload-Offset` header field's value MUST match this inferred offset, if present.
```

GrantGryczan:
```suggestion
- Interim and final responses for creating the upload ({{upload-creation}}) or appending ({{upload-appending}}) with a non-`2XX` status code and the `Upload-Complete: ?0` header can include the `Upload-Offset` header field.
- Final responses for appending representation data ({{upload-appending}}) with a `2XX` status code and the `Upload-Complete: ?0` header acknowledge that the entire request content was received and the offset increased by the length of the appended representation data.
```

GrantGryczan (changes requested):

draft-ietf-httpbis-resumable-upload.md
Acconut: Looks good, thanks!

draft-ietf-httpbis-resumable-upload.md
```diff
@@ -307,15 +307,19 @@ The state of an upload consists of the following properties that are tracked by
...
 
-The `Upload-Offset` header field in responses serves as an acknowledgment of the processed representation data and as a guarantee that no retransmission of it will be necessary. Clients can use this guarantee to free resources associated to transferred representation data.
+The client obtains the current offset in multiple ways:
+
+- The offset can be explicitly retrieved from the upload resource ({{offset-retrieving}}).
+- Interim and final responses for creating the upload ({{upload-creation}}) or appending ({{upload-appending}}) with a non-`2XX` status code and the `Upload-Complete: ?0` header can include the `Upload-Offset` header field.
+- Final responses for appending representation data ({{upload-appending}}) with a `2XX` status code and the `Upload-Complete: ?0` header acknowledge that the entire request content was received and the offset increased by the length of the appended representation data.
```
GrantGryczan: Ahh, your revision also had this good sentence in it. Let's include that too. Didn't mean to step on your toes! 😆

```suggestion
- Final responses for appending representation data ({{upload-appending}}) with a `2XX` status code and the `Upload-Complete: ?0` header acknowledge that the entire request content was received and the offset increased by the length of the appended representation data. An `Upload-Offset` header field's value MUST match this inferred offset, if present.
```

GrantGryczan (changes requested):

draft-ietf-httpbis-resumable-upload.md
Acconut: Glad to have this topic finally settled :)

GrantGryczan (approved):

guoye-zhang (approved):

PR #3490, opened 2026-07-15
Avoid using query parameter before it's defined

### LINKED ISSUE #3486: no-vary-search: definition of query param late
(nitpicking)

As reader (very) familiar with URI and HTTP specs, I can't help noticing that "query parameter" is used in the introduction, and people familiar with RFC 3986 will immediately think: that RFC does not define what that is.

A forward reference would be quite lame; but maybe there's another way to handle this?

nidhijaju:
Fixes https://github.com/httpwg/http-extensions/issues/3486.

### DIFF
--- a/draft-ietf-httpbis-no-vary-search.md
+++ b/draft-ietf-httpbis-no-vary-search.md
@@ -112,7 +112,7 @@ informative:
 
 --- abstract
 
-This specification defines an extension to HTTP Caching, changing how URI query component impacts caching. It introduces the `"No-Vary-Search"` response header field, which allows origin servers to signal to caches that certain query parameters do not semantically affect the served response and can be ignored for cache matching purposes.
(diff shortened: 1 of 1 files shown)


reschke (commented):
I don't believe that changing to "key-value pairs" helps, as, from the pov of the URI and HTTP RFCs, the query component is entirely opaque.

I'll try to come up with a proposal, but that will take a few days.

reschke (commented):
I don't believe that changing to "key-value pairs" helps, as, from the pov of the URI and HTTP RFCs, the query component is entirely opaque.

I'll try to come up with a proposal, but that will take a few days.

nidhijaju:
I've attempted to update the text and removed "key-value pairs". Let me know if this looks better.
Happy to incorporate alternate text suggestions as well.

reschke (approved):
Awesome! Thanks!

PR #3495, opened 2026-07-16
RESUMABLE: Generalize completeness response header

GrantGryczan:
@guoye-zhang Sorry, I forgot to mention this in #3493! Now that `Upload-Complete` is only used in creation and append, do we still need this phrasing? You added it [here](https://github.com/httpwg/http-extensions/pull/3404/changes/1332e9c8a9930f16b89dcf86b7c81c966c2bd34b).

If the phrasing is still needed for future-proofing (allowing extended definitions to use the header for other purposes), then you can just close this PR. But I'm guessing it was only added to exclude offset retrieval from this definition.


guoye-zhang:
I actually thought about this, and decided to keep it since otherwise, the only valid value for upload-complete is false in offset retrieval which is somewhat unintuitive, and all existing implementations are non-compliant. Keeping it scoped to creation and appending means that you use it however you want on offset retrieval.

GrantGryczan:
I see, fair enough. But in that case, clients can't rely on the header semantics being consistent for offset retrieval, which was one of the benefits of removing it. Not that it matters to me personally, but worth noting.

Acconut:
I would prefer to keep the current phrasing as-is since it does contribute a bit to clarify the purpose of this header field.

PR #3502, opened 2026-07-28
NVS: Address GENART review comments

nidhijaju:
@mallory: If you could take a look, that would be great. Thank you!


mallory:
> @mallory: If you could take a look, that would be great. Thank you!

Looks great. Thanks!

PR #3504, opened 2026-08-12
Address mnot's last call feedback

nidhijaju:
Addresses https://lists.w3.org/Archives/Public/ietf-http-wg/2026JulSep/0137.html.


mnot:
Looks good to me.

PR #3505, opened 2026-08-13
Add dates for WHATWG references

nidhijaju:
Addresses https://lists.w3.org/Archives/Public/ietf-http-wg/2026JulSep/0145.html.


mnot:
LGTM - RFC Editor will help with this stuff anyway.

PR #3510, opened 2026-08-23
Do not drop a cookie whose value or http-only changed

### LINKED ISSUE #3501: Cookies: New cookie with differing value dropped during duplicate check
Hello,

https://httpwg.org/http-extensions/draft-ietf-httpbis-layered-cookies.html#section-5.4.3-2.18.2.2.1
> If cookie's secure is equal to oldCookie's secure, cookie's same-site is equal to oldCookie's same-site, and cookie's expiry-time is equal to oldCookie's expiry-time, then return null.

This clause is used to check for duplicate cookies, and skip the rest of the Store a Cookie algorithm if the cookie is equal to an existing cookie.

However, it does not check if the value is the same or not.

As such, two cookies having the same attributes, except the value, will result in the cookie with the new value being dropped (and the old value being kept)

(And this would be common for cookies that are removed on browser exit, as they have the same null expiry-time)

I did some quick testing and this does not match major browser behavior - browsers keep the cookie with the new value, even if they have the same same-site, secure, and expiry time (or lack thereof).

I propose revising it to:
> If cookie's secure is equal to oldCookie's secure, cookie's same-site is equal to oldCookie's same-site, cookie's expiry-time is equal to oldCookie's expiry-time, and cookie's value is equal to oldCookie's value, then return null.

annevk:
Store a Cookie compared secure, same-site, and expiry-time against the cookie
being replaced, but not value or http-only, and returned null without storing
when those matched. That dropped the very common case of a cookie being
reissued with a new value, and silently ignored a newly added HttpOnly
attribute.

Compare value and http-only as well, and always store the cookie, returning
null afterwards to signal that it cannot be distinguished from the cookie it
replaced. As a result last-access-time is refreshed again.

Also document what Store a Cookie returns.

Fixes #3501.

Tests: https://github.com/web-platform-tests/wpt/pull/62161

### DIFF
--- a/draft-ietf-httpbis-layered-cookies.md
+++ b/draft-ietf-httpbis-layered-cookies.md
@@ -1225,7 +1225,9 @@ cookie attribute "wins".
 ### Store a Cookie {#store-a-cookie}
 
 To **Store a Cookie** given a cookie _cookie_, boolean _isSecure_, domain or IP address _host_,
-boolean _httpOnlyAllowed_, boolean _allowNonHostOnlyCookieForPublicSuffix_, and boolean _sameSiteStrictOrLaxAllowed_:
+boolean _httpOnlyAllowed_, boolean _allowNonHostOnlyCookieForPublicSuffix_, and boolean
+_sameSiteStrictOrLaxAllowed_, run these steps. They return _cookie_, or null if _cookie_ was not
+stored or cannot be distinguished from the cookie it replaced:
 
 1. Assert: _cookie_'s name's length + _cookie_'s value's length is not 0 or greater than 4096.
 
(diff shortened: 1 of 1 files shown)


johannhof (approved):

annevk:
@johannhof thanks for reviewing, but please don't use merge commits going forward. Rebase and merge is right there.

johannhof:
I actually dislike merge commits as well, but thought (also looking at recent git history) that they were what others do as well in this repo. @mnot any guidance? Happy to go with rebase.

annevk:
Well thanks to us complaining they enabled the other options for this repository, so we can choose and do better.

PR #3511, opened 2026-08-23
Require HttpOnly for the "__Http-" and "__Host-Http-" prefixes

annevk:
Http-prefix compatible required http-only to be false, which is the opposite of
what the prefixes are for. It made "__Http-" and "__Host-Http-" cookies
settable only without an HttpOnly attribute, and rejected the ones set with it,
contradicting the descriptions of both prefixes.

Tests: cookies/prefix/__Http.https.html and
cookies/prefix/__Host-Http.https.html already assert the intended behaviour.


yoavweiss (approved):
Oh my. LGTM. Thanks for catching that!!

johannhof (approved):

PR #3512, opened 2026-08-23
Make a cookie's path a URL path

annevk:
Path-Matches compared serialized paths, which made a cookie path of /foo match a
request path of /foobar, and passed URL paths to the URL path serializer, which
takes a URL. Compare URL path segments instead, using the operations URL now
provides.

The Path attribute was split into a list that kept its leading empty segment,
which is not how a URL path is represented, so such a path matched nothing and no
"__Host-" prefixed cookie could be set. Isomorphic decode the attribute, drop
that segment, and return failure if the result is not an ASCII string, rather
than storing a path no request could ever carry. That check applies to the Path
attribute that wins, so it happens once all attributes have been processed.

Cookie Default Path removed an item from the URL path it was given, or replaced
that path's first item, and it is called with a request URL's path. Operate on a
clone instead.

Also stop passing a path to Store a Cookie, which takes no such argument.

Depends on https://github.com/whatwg/url/pull/929.

Tests: https://github.com/web-platform-tests/wpt/pull/62165

### DIFF
--- a/draft-ietf-httpbis-layered-cookies.md
+++ b/draft-ietf-httpbis-layered-cookies.md
@@ -1025,11 +1025,13 @@ They return a URL path.
 
 1.  Assert: _path_ is a non-empty list.
 
-2.  If _path_'s size is greater than 1, then remove _path_'s last item.
+1.  Let _defaultPath_ be a clone of _path_.
 
-3.  Otherwise, set _path_[0] to the empty string.
+1.  If _defaultPath_'s size is greater than 1, then remove _defaultPath_'s last item.
 
-4.  Return _path_.
+1.  Otherwise, set _defaultPath_[0] to the empty string.
+
+1.  Return _defaultPath_.
 
 
 ### Path Matching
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-layered-cookies.md
```diff
@@ -1025,28 +1025,37 @@ They return a URL path.
...
 
 To determine if a URL path _requestPath_ **Path-Matches** a URL path _cookiePath_, run these steps.
 They return a boolean.
 
-1.  Let _serializedRequestPath_ be the result of URL path serializing _requestPath_.
+1.  If _requestPath_ equals _cookiePath_, then return true.
+
+1.  Let _prefix_ be a clone of _cookiePath_.
+
+1.  If _prefix_ is not empty and _prefix_'s last item is the empty string, then remove _prefix_'s
```
johannhof: Why are we doing this?

johannhof (approved):
This is a great PR, thank you Anne. And I appreciate the notes that helped me understand it 😅 

Not merging immediately in case you want to merge the URL PR first.

PR #3513, opened 2026-08-24
Treat a Domain attribute matching an IP address host as host-only

annevk:
Domain-Match only performs suffix matching when the host is a domain, so a
Domain attribute equal to an IP address host reaches exactly the hosts a
host-only cookie reaches. Storing such a cookie as non-host-only gained nothing
and cost something: host-only is part of what distinguishes one cookie from
another, so it could sit alongside an otherwise identical host-only cookie, with
both being sent.

It also made the "__Host-" prefix reject a cookie whose Domain attribute could
not do any of the subdomain spanning that prefix exists to prevent.


annevk:
This cannot be tested unfortunately, but I'm pretty sure implementations do this already.

johannhof (approved):

